PR #297, opened 2014-04-18
Audio codec refactor

thetheodor:
Remove snd_codec and replace it with CodecFactory and {Wav,Ogg,Opus}Loader
 which use engine/framework/FileSystem.


Senghour11:
Get well

Senghour11:
Interesting

PR #577, opened 2015-02-07
Fix windows build. Define _USE_MATH_DEFINES to get M_PI etc.

gmcode:
On windows at least, math.h/cmath requires _USE_MATH_DEFINES to define M_PI.

g_buildable.cpp etc. requires M_PI and would fail to compile without it.

Longer term solution: don't use M_PI or math.h

### DIFF
--- a/src/engine/qcommon/q_shared.h
+++ b/src/engine/qcommon/q_shared.h
@@ -35,6 +35,14 @@ Maryland 20850 USA.
 #ifndef Q_SHARED_H_
 #define Q_SHARED_H_
 
+// math.h/cmath uses _USE_MATH_DEFINES to decide if to define M_PI etc or not.
+// So define _USE_MATH_DEFINES early before including math.h/cmath
+// and before including any other header in case they bring in math.h/cmath indirectly.
+#ifndef _USE_MATH_DEFINES
+#define _USE_MATH_DEFINES
+#endif
+
+
 // q_shared.h -- included first by ALL program modules.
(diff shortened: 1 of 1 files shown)


Kangz:
Thanks for the fix! It compiled fine with MSVC a few alphas ago so I'm surprised it popped only that recently.

Viech:
Should I still refrain from using `M_PI_2` and the like?

Kangz:
Yes, I don't think it is defined in Windows' headers as when compiling on MSVC, M_PI was working but not M_PI_2.

Amanieu:
Actually M_PI_2 is defined by MSVC.

Here are the math.h definitions in MSVC (only with _USE_MATH_DEFINES):
#define M_E        2.71828182845904523536
#define M_LOG2E    1.44269504088896340736
#define M_LOG10E   0.434294481903251827651
#define M_LN2      0.693147180559945309417
#define M_LN10     2.30258509299404568402
#define M_PI       3.14159265358979323846
#define M_PI_2     1.57079632679489661923
#define M_PI_4     0.785398163397448309616
#define M_1_PI     0.318309886183790671538
#define M_2_PI     0.636619772367581343076
#define M_2_SQRTPI 1.12837916709551257390
#define M_SQRT2    1.41421356237309504880
#define M_SQRT1_2  0.707106781186547524401

Viech:
So, anything I need to dodge? If yes, would it make sense to #undefine to be sure they aren't used?

PR #578, opened 2015-02-07
Remove a stack of unused variables. No intended functional change.

gmcode:


Amanieu:
For the catch clauses just go ahead and remove err entirely instead of commenting it out.

Amanieu:
Looks good to me, thanks

PR #587, opened 2015-02-11
Better handle out of memory conditions when loading ogg files.

gmcode:


Viech:
@gmcode why the close?

Kangz:
It was following a discussion he had on IRC with @Amanieu where they decided these commits where unnecessary given nacl-cgame catches bad_alloc and throws a Drop exception (or fatal?) so what was in this PR wouldn't work.

Amanieu:
It treats out-of-memory errors as fatal, though I guess a Drop could also work in theory.

PR #622, opened 2015-02-22
Stringchanges

gmcode:


gmcode:
Will submit to avoid need for C++14.

PR #623, opened 2015-02-22
Stringchanges

gmcode:


gmcode:
Will try to squash a bit more and resubmit.

gmcode:
Didn't manage to squash so will keep it the same.

src/libs/libRocket/Include/Rocket/Core/StringBase.inl
```diff
@@ -149,6 +150,12 @@ const T* StringBase< T >::CString() const
 }
 
 template< typename T >
+bool StringBase<T>::_IsUsingLocalBuffer() const
```
DolceTriade: libRocket doesn't use the underscore in front of the function anywhere else. Just leave it as Is

src/libs/libRocket/Include/Rocket/Core/StringBase.h
```diff
@@ -185,8 +189,18 @@ class StringBase
 	inline const T& operator[](size_type index) const;
 	inline T& operator[](size_type index);
 
+	// Offer c_str() to make the interface somewhat more consistent with std::string.
+	// It eases code migration to std::string in the future should that be needed.
+	// The name also imparts better expectations about what it does better
+	// than CString() which can be confused with things like MFC's class of the same name.
+	inline const T* c_str() const;
```
DolceTriade: If its inline, define it inside the StringBase.inl file. inl stands for inline so all the functions in that file should be automatically inlined.

src/libs/libRocket/Include/Rocket/Core/StringBase.h
```diff
@@ -185,8 +189,18 @@ class StringBase
...
+
 protected:	
+	inline T* _LocalBuffer();
```
DolceTriade: For this and the function below, libRocket doesn't use this naming convention for protected/private methods. You should remove the underscore.

src/libs/libRocket/Include/Rocket/Core/StringBase.h
DolceTriade: Not necessary as CString() exists.

src/libs/libRocket/Include/Rocket/Core/StringBase.inl
```diff
@@ -165,12 +172,18 @@ void StringBase< T >::Reserve(size_type size)
 	if (value == (T*)local_buffer)
 	{
 		T* new_value = (T*)realloc(NULL, buffer_size);
+		// TODO! Consider throwing here.
```
DolceTriade: librocket doesn't use exceptions.

src/libs/libRocket/Include/Rocket/Core/StringBase.h
gmcode: That's not what I am observing, what about _Assign, _Insert, etc. etc. ?
I'm happy to not use _ but that's why I added it in the first place.

src/libs/libRocket/Include/Rocket/Core/StringBase.h
gmcode: If I have to comply I will, but unless there is more to this I disagree. Using CString in the first place was probably a bad idea IMO as It doesn't convey what c_str() and doesn't help generic code and it doesn't serve a useful purpose to propagate the situation. Plus it's longer and defeats muscle memory for regular string. It's is also an inline function that has no cost to add now so why wait.

src/libs/libRocket/Include/Rocket/Core/StringBase.inl
gmcode: I'll discuss this more on irc and comply based on the outcome of that. But the current argument  makes no sense to me.

Is there anywhere that it says librocket shouldn't use exceptions for any reason? If there is none, perhaps we should add some and the justification for it?

But given the rest of stl does use exceptions it seems exceptions will happen somewhere anyway?

Also given there are only two places to add the bad_alloc support to address the issue properly, it seems deficient not to address the issue unless there is good reason. I mean why force any user of this library to crash with no message or a misleading access violation one if there is no good reason, when we can instead with trivial effort leave it up to the app to decide to crash or not by simply catching the exception or not and without a misleading message. That's what they are for, right?

To do otherwise without good reason seems short sited and illogical unless a real "customer" of the library explicitly needs exceptions disabled and even then it should probably a a compile time enabled thing.

In summary, I think the argument for not using exceptions is insufficiently justified so far to agree with.

src/libs/libRocket/Include/Rocket/Core/StringBase.h
DolceTriade: It would be better. They seem to only use it when the caller has the same function name.

Viech:
@gmcode it's possible to just update an existing pull request with new or changed commits, no need to close and reopen. Just push your branch again, if I remember correctly. :wink:

PR #631, opened 2015-02-25
Correct buffer size for formatted net address with port

f4rnham:
Fixes "Com_sprintf: overflow of 48 bytes buffer" console spam on server listing

Feel free to come with better name than NET_ADDR_W_PORT_STR_MAX_LEN it was derived from NET_AdrToString - NET_AdrToString**wPort**


Kangz:
Great, one less buffer overflow! (did you use a static analyzer to find it?)

f4rnham:
Without static analyzer, Com_sprintf was printing error on every attempt to list servers, so no trouble with reproducing and tracking it down

PR #632, opened 2015-02-25
Respect master volume setting from first millisecond

f4rnham:
Menu background music starts playing before first Audio::Update call which syncs the master volume, so set it also in Audio::Init

### DIFF
--- a/src/engine/audio/Audio.cpp
+++ b/src/engine/audio/Audio.cpp
@@ -138,6 +138,8 @@ namespace Audio {
         InitSounds();
         InitEmitters();
 
+        UpdateListenerGain();
+
         for (int i = 0; i < MAX_GENTITIES; i++) {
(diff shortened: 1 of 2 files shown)


Kangz:
Good catch, thanks for the fix! However in Update, there is more logic related to the state of the window. Maybe it would be better idea to call Update() at the end of Init(). Your call.

Viech:
This sounds like it will greatly please my ears which were often harassed by loud music playing for a second during map load. Thanks! :grinning:

f4rnham:
I don't think running whole Audio::Update is correct there, but I added the minimized / unfocused checks since Audio::Init can be called multiple times during one run (currently only by "snd_restart" command), which I didn't realized before.

And this probably won't fix anything with sound during map loading, only if there is way to skip main menu and load map right after start or "snd_restart" is triggered automatically.

Kangz:
Great :D

PR #647, opened 2015-03-04
Fix "the the" mistakes.

RamchandraApte:
(subject) Fix "the the" mistakes.


DolceTriade:
We only want the changes that are directly in our codebase (ie, in engine/ or gamelogic/). THe changes outside our source tree are very likely to be overwritten, and therefore, useless.

PR #677, opened 2015-03-11
Request for comm(en|i)ts: Decoupling daemon from unvanquished

TimePath:
**Summary**: CMakeLists changes to reduce coupling between daemon and unvanquished.

* Symlinks have been used to test the tree structure and avoid moving and copying the rest of the repo.
    * To merge, replace with the file they point to, and make '/unvanquished' the root.
* Daemon no longer contains unvanquished assets and can be packaged separately.
    * [ ] FIXME: Linux packaging scripts are currently duplicated, probably remove the ones in daemon.
    * [ ] FIXME: So are the travis-ci scripts
* Mods can target daemon in a few lines (with the prerequisite variables set):
```cmake
    ...
    add_subdirectory( daemon daemon_build )
    ...
    include( DaemonGame )
    GAMEMODULE( NAME sgame
      COMPILE_DEF BUILD_VM BUILD_SGAME
      FILES ${SGAMELIST}
    )
```
* Testing/compiling unvanquished remains unchanged.
    * Though in this preview, you will have to `cd unvanquished` first.
* Daemon can easily be used as a git submodule.
    * Recommend splitting into its own repository at some point; duplicate it and remove daemon from one and unvanquished from the other.

**Future improvements**:
* [ ] A CMake defined branding instead of hardcoding "unvanquished" in the engine; DAEMON_BRANDING or similar.
* [ ] Stop depending on the dependency bootstrap provided by dl.unvanquished.net.
* [ ] Build daemon dependencies individually from individual repositories.
* [ ] Expose common engine/game code as a library?  
...


TimePath:
**Gamelogic as the submodule**: With gamelogic as the submodule, it would have to be updated every time it changed. With daemon as the submodule, this is also true, but far less frequent.

TimePath:
Closing, new work [here](https://github.com/DaemonDevelopers/Daemon)

PR #690, opened 2015-03-29
Cleanup and static analysis

TimePath:


TimePath:
Now that everything works, I will begin rewriting the history with your style suggestions.

I tried ninja mostly because I prefer its output for large projects ex: [98/975] vs [10%], and that it defaults to multiple threads unlike make (made debugging travis a bit more pleasant). Make defaults to a single thread, though I haven't tried multiple with it. If make exhibits the same race condition, you lose nothing by using ninja with a single thread.

-Werror is something I feel should be used by CI, but as a compromise I can hide it behind a flag and add 2 more travis builds with -Werror and -Wno-error.

TimePath:
Travis built the wrong commit for some reason: https://travis-ci.org/TimePath/Unvanquished/builds/57127568

PR #707, opened 2015-04-18
Reduce local compile verbosity

TimePath:
I will increase this again when no more warnings are visible


TimePath:
Ready for merge

PR #723, opened 2015-05-14
Changes to make shader debugging easier. Support a -set shaderpath "c…

gmcode:
…:/unvanquished/main" option. See app and git comments.

On Windows use forward slashes instead of backslashes in the option as shown in the above example as the unv app console eats backslashes.
Basically use this command on the command line like so -set shaderpath "c:/unvanquished/main".
In unv app console use the option like so: set shaderpath "c:/unvanquished/main".

So the idea is:
* Issue set shaderpath "c:/unvanquished/main" to tell Unvanquished what folder the glsl folder is in.
* Then make whatever changes you want to make the .glsl files in the glsl folder.
* Then type glsl_restart at the Unvanquished console and unv will load the files directly from disk and compare them to the versions already compiled into Unvanquished and let you know what it thinks has changed.
* The command can be on the console or command line. If on the command line Unv will start by loading the disk versions first.
* If done from the console the built in versions will be used first.
* If Unvanquished can't load the new versions from disk due to syntax errors or whatever, it will try to recover and use the built in version again.
* If that happens, just make whatever changes you need to correct the problems with the disk versions and run glsl_restart again.
* When you're done, look at the console log for a reminder of what changes Unv thinks you've made and commit those to git.
* Then run shaderbuild.sh and commit shaders.cpp to git too.

### DIFF
--- a/src/engine/renderer/gl_shader.cpp
+++ b/src/engine/renderer/gl_shader.cpp
@@ -24,6 +24,63 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "gl_shader.h"
 
 extern std::unordered_map<std::string, const char *> shadermap;
+// Shader Type will be determined later based on command line setting or absence of.
+ShaderType shaderType = ShaderType::Unknown;
+
+namespace // Private types - implementation details.
+{
+	class file_reader
+	{
+	public:
+		file_reader() : data_(nullptr), length_(-1)
+		{
+		}
+		file_reader(Str::StringRef filename) : data_(nullptr), length_(-1)
+		{
+			open(filename.c_str());
+		}
+		int open(Str::StringRef filename)
+		{
+			close();
+			void* new_data = nullptr;
+			int new_length = ri.FS_ReadFile(filename.c_str(), &new_data);
+			if (new_length < 0)
+				return -1;
+			data_ = static_cast<char*>(new_data);
+			length_ = new_length;
+			return length_;
+		}
+		void close()
+		{
+			char* old_data = data_;
+			length_ = -1;
+			data_ = nullptr;
+			if (old_data != nullptr)
+				ri.FS_FreeFile(old_data);
+		}
+		~file_reader()
+		{
+			close();
+		}
+		int length() const
+		{
+			return length_;
+		}
+		char* data()
+		{
+			return data_;
+		}
+
+	private:
+		char* data_;
+		int length_;
+	};
+};
+
+namespace
+{
+	std::unordered_map<std::string, std::string> aliasmap;
+};
 
 // *INDENT-OFF*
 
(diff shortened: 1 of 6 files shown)


src/engine/renderer/gl_shader.cpp
```diff
@@ -22,8 +22,20 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 // gl_shader.cpp -- GLSL shader handling
 
 #include "gl_shader.h"
+#include <type_traits>
+
+// We currently write GLShaderHeader to a file and memcpy all over it.
```
Kangz: nit: typo, copy it all over the place?

src/engine/renderer/gl_shader.cpp
```diff
@@ -22,8 +22,20 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+// Make sure it's a pod, so we don't put a std::string in it or something
+// and try to memcpy over that or binary write an std::string to a file.
+static_assert(std::is_pod<GLShaderHeader>::value, "Value must be a pod while code in this cpp file reads and writes this object to file as binary.");
```
Kangz: Good static assert, I did not know it was supported by gcc 4.6 but as travis is green it is ok. Although you migt want to replace Value in the string by GLShaderHeader

src/engine/renderer/gl_shader.cpp
```diff
@@ -55,27 +67,226 @@ GLShader_motionblur                      *gl_motionblurShader = nullptr;
...
+		size_t sourcePos = 0;
+		size_t keepPos = 0;
+		size_t commentStartPos;
+		size_t commentEndPos;
+		CommentType commentType;
+
+		auto keep = [&](size_t keepLength)
+		{
+			if (keepPos != sourcePos)
+				std::copy(source.begin() + sourcePos, source.begin() + sourcePos + keepLength,
```
Kangz: Ranges for std::copy cannot overlap, you can use memmove which will work.

src/engine/renderer/gl_shader.cpp
```diff
@@ -55,27 +67,226 @@ GLShader_motionblur                      *gl_motionblurShader = nullptr;
...
+					keep(remainingLength);
+				}
+				else
+				{
+					size_t codeLength = commentStartPos - sourcePos;
+					keep(codeLength);
+				}
+				break;
+			}
+			// We found a start and end comment. keep the what's before the comment.
```
Kangz: nit: typo remove "the"

src/engine/renderer/gl_shader.cpp
```diff
@@ -55,27 +67,226 @@ GLShader_motionblur                      *gl_motionblurShader = nullptr;
...
+	// new line character inplace to avoid splicing together lines
+	// and creating new tokens by accident.
+	// We don't use regex as we don't need it. Nor do we want
+	// to force a dependency on it either for such simple needs.
+	// We do not remove #if 0 type comments as buildshader does not do
+	// that. It would be also more complex. Complexity comes from
+	// things like #if X #elif 0 #else X #endif patterns.
+	// The shaders actually do/did that btw.
+	// But we don't need to handle #directives right now nor do we.
+	void StripComments(std::string& source)
```
Kangz: Thanks for fixing the complexity of the algorithm, it was a large change:

src/engine/renderer/gl_shader.cpp
```diff
@@ -55,27 +67,226 @@ GLShader_motionblur                      *gl_motionblurShader = nullptr;
...
+	{
+		assert(commentType != CommentType::None);
+		if (commentType == CommentType::Block)
+		{
+			commentEndPos = source.find("*/", start);
+			if (commentEndPos != std::string::npos)
+				commentEndPos += 2;
+		}
+		else
+			commentEndPos = source.find_first_of("\r\n", start);
```
Kangz: No need to check for \r, it is removed before we get here

src/engine/renderer/gl_shader.cpp
```diff
@@ -55,27 +67,226 @@ GLShader_motionblur                      *gl_motionblurShader = nullptr;
...
+	}
+
+	// Somewhat emulate what buildshader.sh does or doesn't do.
+	// Not wildly efficient but doesn't have to be as it's currently
+	// only used as part of a debugging only feature on small shader files.
+	void NormalizeShaderText( std::string& text )
 	{
-		delete _shaders[ i ];
+		size_t pos = 0;
+		while ((pos = text.find("\r\n", pos)) != std::string::npos)
```
Kangz: Use the Erase-remove if idiom here instead? It would drop the complexity from n² to n

src/engine/renderer/gl_shader.cpp
```diff
@@ -55,27 +67,226 @@ GLShader_motionblur                      *gl_motionblurShader = nullptr;
...
+};
 
-	for ( std::size_t i = 0; i < _deformShaders.size(); i++ )
+std::string GetShaderPath()
+{
+	std::string shaderPath;
+	if (shaderPathEnabledVar.Get())
 	{
-		glDeleteShader( _deformShaders[ i ] );
+		auto shaderPathVar = Cvar_Get("shaderpath", "", CVAR_INIT);
```
Kangz: that's sad, I will take a look to that cvar business.

src/engine/renderer/gl_shader.cpp
gmcode: \r\n is replaced with \n, but any \r that might exist would still be there.

src/engine/renderer/gl_shader.cpp
gmcode: I think you are suggesting using remove_if, but the predicate needs to identify a sequence not a single character and it's only passed a single character. What exact code are you suggesting if it's only a few lines like this implies?

src/engine/renderer/gl_shader.cpp
```diff
@@ -623,23 +695,23 @@ bool GLShaderManager::buildPermutation( GLShader *shader, int macroIndex, int de
...
 		{
-			buildPermutation( shader, i, 0 );
+			buildPermutation( &shader, i, 0 );
 		}
 
 		_shaderBuildQueue.pop();
 	}
 
-	ri.Printf( PRINT_DEVELOPER, "glsl shaders took %d msec to build\n", _totalBuildTime );
+	Log::Debug( "glsl shaders took %d msec to build\n", _totalBuildTime );
```
Kangz: That's going to print it unconditionnzlly which is going to be too verbose for users, use a Logger (Log.h) if you want to use Log

src/engine/renderer/gl_shader.cpp
```diff
@@ -953,6 +1011,8 @@ void GLShaderManager::PrintInfoLog( GLuint object, bool developerOnly ) const
 	int         maxLength = 0;
 	int         i;
 
+	// TODO: Not sure how to upgrade this type of usage to the
```
Kangz: I'd tranform it to GetInfoLog and let the caller decide where to output the string

src/engine/renderer/gl_shader.cpp
```diff
@@ -1039,10 +1099,8 @@ void GLShaderManager::BindAttribLocations( GLuint program ) const
 
 bool GLCompileMacro_USE_VERTEX_SKINNING::HasConflictingMacros( size_t permutation, const std::vector< GLCompileMacro * > &macros ) const
 {
-	for ( size_t i = 0; i < macros.size(); i++ )
+	for (const GLCompileMacro* macro : macros)
```
Kangz: we usualy use auto& in the range based for loops

src/engine/renderer/gl_shader.cpp
Kangz: Ah I see, it is not exactly the same. However I think we can remove all \r characters as they are forbidden in other places in the shader: Feel free to change the script to do that same although I doubt a difference will ever arise.

Kangz:
Thanks, these address most of my concerns; I will look at that cvar thing as soon as I get time. You might want to try Cvar::ROM in the meantime it miht work.

src/engine/renderer/gl_shader.cpp
```diff
@@ -55,27 +63,236 @@ GLShader_motionblur                      *gl_motionblurShader = nullptr;
...
+		}
+		else if (shaderKind == ShaderKind::External)
+		{
+			std::string shaderText;
+			std::string shaderFilename = GetShaderFilename(filename);
+			std::error_code openErr;
+
+			FS::File shaderFile = FS::RawPath::OpenRead(shaderFilename, openErr);
+			if (openErr)
+				ThrowShaderError(Str::Format("Cannot load shader from file: %s\n", shaderFilename));
```
Amanieu: You want to print the actual error message as well, which you can obtain from openErr.message(). The same applies to all other std::error_code uses.

src/engine/renderer/tr_shade.cpp
```diff
@@ -122,6 +122,77 @@ void GLSL_InitGPUShaders()
...
+			Log::Warn( "Switching from external to built-in shaders.");
+		try
+		{
+			GLSL_InitGPUShadersOrError();
+		}
+		catch (const ShaderException&e)
+		{
+			Log::Warn("Built-in shaders failed: %s.", e.what());
+
+			Sys::Error(e.what()); // Fatal.
```
Amanieu: Sys::Error("Built-in shaders failed: %s.", e.what()); and get rid of the warning. Sys::Error will log messages appropriately.

src/engine/renderer/gl_shader.cpp
gmcode: \r characters can get put into the stream by windows hosts. We remove \r\n sequences but any \r that might be there. I have replaced the simple erase loop with something more custom but more efficient. I think we can stop worrying now. This is a debug tool after all. Let's not loose sight of that. I've only bothered to polish it this much as we might want to avoid using buildershader.sh at all eventually or at least in some cases and if we do this code can replace that script, but otherwise the effort isn't justified to worry this much about the performance of a debugging tool that's just used on inherently small files. ;) Even the bug risk isn't worth it. But anyway, done now so everyone should be happy.

src/engine/renderer/gl_shader.cpp
gmcode: I normally do that too. But I didn't like the idea of looping with references to pointers rather than pointers. And auto on it's own risks copying larger things if someone changes the loop.
I'll leave this for now as I think we want to get rid of much of this code anyway. A vector of pointers isn't ideal. Unique and shared ptr  are probably more suitable here but I'll wait until I understand that code to refactor to that and we discuss things in that context.

src/engine/renderer/gl_shader.cpp
gmcode: Thanks will fix.

src/engine/renderer/tr_shade.cpp
gmcode: Thanks. Will fix.

PR #731, opened 2015-05-16
Make CmakeLists.txt more IDE friendly

VadimGit:


Kangz:
It seems that this PR is mostly so that we don't have to add headers to the CMakeLists.txt explicitely to have them show up in IDEs. However introducing a LOT of include dirs like this is really bad in my opinion so I'd rather keep adding headers by hand.

Viech:
What's bad about include dirs? I don't think people who don't use IDEs will remember to add header files given that it compiles without that. I'd like to have a guarantee that IDEs will find all files since otherwise you could be forced to do quite a bit of detective work afer you refactored variable names in only a subset of all project files.

Amanieu:
include_directories() will add the given path to the compiler #include search path

VadimGit:
It's temporarily improvement to clean up source lists and allow IDE to parse CMakeLists.txt. But the best way to add GROUP_HEADER_LIST as well as GROUP_SRC_LIST for each group that  is part of the project (not third party libs).
CMake figures out build dependencies using only source files (SRC_LISTs) and indclude directories will be need only for third party libs. 
But after add HEADER_LISTs to CMake and add them to targets or "void target" they are considered as part of the project and will be shown in any IDE.

DolceTriade:
Can't merge after CMake changes.

TimePath:
Is it still important enough to rebase?

Kangz:
No, the important part of the commit is the addition of tons of include dir, which we don't want to add.

PR #741, opened 2015-06-01
Audio

gmcode:


src/engine/audio/OggCodec.cpp
```diff
@@ -149,20 +149,41 @@ AudioData LoadOggCodec(std::string filename)
 	int sampleRate = oggInfo->rate;
 	int numberOfChannels = oggInfo->channels;
 
-	char buffer[4096];
 	int bytesRead = 0;
 	int bitStream = 0;
 
-	std::vector<char> samples;
+	char* buffer = nullptr;
```
Amanieu: Avoid using raw pointers, they are not exception-safe

src/engine/audio/OggCodec.cpp
```diff
@@ -149,20 +149,41 @@ AudioData LoadOggCodec(std::string filename)
...
+		{
+			if (!bufferCapacity)
+				bufferCapacity = bufferInitialCapacity;
+			else
+				bufferCapacity *= 2;
+			bufferRemaining = bufferCapacity - bufferUsed;
+		}
+		buffer = (char*)realloc(buffer, bufferCapacity);
+		if (!buffer)
+			throw std::bad_alloc();
```
Amanieu: This will leak memory (buffer is not freed)

PR #742, opened 2015-06-01
Audio

gmcode:
Some string to stringref changes. A bug fix to rfind. A performance workaround for Windows debug builds.

Cuts a minute of loading times for debug builds caused by MS debug checks. Setting debug_debug_iterator_level differently would fix this bug that's kind of handy. It's just too much for this situation as audio loads some big files.

### DIFF
--- a/src/engine/audio/OggCodec.cpp
+++ b/src/engine/audio/OggCodec.cpp
@@ -168,9 +168,13 @@ AudioData LoadOggCodec(Str::StringRef filename)
 			else
 				bufferCapacity *= 2;
 			bufferRemaining = bufferCapacity - bufferUsed;
+			char* oldBuffer = buffer;
 			buffer = (char*)realloc(buffer, bufferCapacity);
 			if (!buffer)
+			{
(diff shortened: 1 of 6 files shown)


src/engine/audio/OggCodec.cpp
```diff
@@ -149,20 +149,41 @@ AudioData LoadOggCodec(std::string filename)
...
+		if (bufferRemaining < bufferMinCapacity)
+		{
+			if (!bufferCapacity)
+				bufferCapacity = bufferInitialCapacity;
+			else
+				bufferCapacity *= 2;
+			bufferRemaining = bufferCapacity - bufferUsed;
+			buffer = (char*)realloc(buffer, bufferCapacity);
+			if (!buffer)
+				throw std::bad_alloc();
```
Amanieu: This still leaks memory: if realloc fails the buffer is *not* freed

gmcode: Good observation! Fixed,

PR #743, opened 2015-06-01
Audio

gmcode:
Audio. Try a different track... Let me know what you think.
Replaces unique_ptr for vector.


gmcode:
I believe this patch does build under gcc. The build system seems confused. It works under clang. Please commit if appropriate.

PR #745, opened 2015-06-02
Some suggestions for String.h. Update a few member values and functio…

gmcode:
…ns to const.

There doesn't seem a lot of value in making stringrefs rebindable. It may help performance not to support that (shrug). It may also discourage dubious use.
To support these claims the rebinding feature is only used in one place in Unv and that's kind of confusing....

### DIFF
--- a/src/engine/framework/CommandSystem.cpp
+++ b/src/engine/framework/CommandSystem.cpp
@@ -193,10 +193,10 @@ namespace Cmd {
         commandLog.Debug("Execing command '%s'", command);
 
         std::string parsedString;
-		if (parseCvars)
+        if (parseCvars)
             parsedString = SubstituteCvars(command);
 
-        Args args(parseCvars? parsedString : command);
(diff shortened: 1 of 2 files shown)


src/engine/framework/CommandSystem.cpp
```diff
@@ -193,12 +193,10 @@ namespace Cmd {
...
 
         std::string parsedString;
-        if (parseCvars) {
+        if (parseCvars)
             parsedString = SubstituteCvars(command);
-            command = parsedString;
-        }
 
-        Args args(command);
+        Args args(parseCvars? Str::StringRef(parsedString) : command);
```
DolceTriade: Why is the cast necessary?

Amanieu:
I fixed a few things and merged it manually.

src/engine/framework/CommandSystem.cpp
gmcode: Didn't see this question. The cast is necessary because both operands of the ? operator need to resolve unambiguously.

PR #746, opened 2015-06-02
Change some by val strings to stringrefs to improve performance. NFC

gmcode:


Amanieu:
This doesn't actually improve performance since the IPC code that invokes the handler moves the strings instead of copying them. This change has absolutely no effect on generated code.

PR #747, opened 2015-06-02
Fix potential bug in zip file reading. If there's an read error we ca…

gmcode:
…n still have valid data read. We want to return that and not random chunk to the caller.

I think cases of this issue exist such as CopyTo but have not looked into that.

### DIFF
--- a/src/common/FileSystem.cpp
+++ b/src/common/FileSystem.cpp
@@ -1146,7 +1146,8 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 			if (err)
 				return;
 			depsData.resize(length);
-			zipFile.ReadFile(&depsData[0], length, err);
+			auto read = zipFile.ReadFile(&depsData[0], length, err);
+			depsData.resize(read);
 			if (err)
 				return;
 		}


src/common/FileSystem.cpp
```diff
@@ -1146,7 +1146,8 @@ static void InternalLoadPak(const PakInfo& pak, Util::optional<uint32_t> expecte
 			if (err)
 				return;
 			depsData.resize(length);
-			zipFile.ReadFile(&depsData[0], length, err);
+			auto read = zipFile.ReadFile(&depsData[0], length, err);
+			depsData.resize(read);
```
DolceTriade: Maybe we should just crash harder in this situations instead of limping on.

gmcode:
Hmm actually this still won't be right. It's just better. This edge case needs more consideration as to if what happens for read errors and, partial reads etc and if ParseDeps is called. Perhaps an exception is required if an error not the full length is read regarding the deps file.

src/common/FileSystem.cpp
gmcode: Hard to say, it depends on how low level we see these routines. If we hard crash lf we make that choice for every caller that nobody can recover from failing to load a pak. And we don't know why we failed. What reason can a zip file read fail on for example, out of memory? Do we won't to die because we can't load a large pak file. Hard failure doesn't seem unreasonable to me in this situation though. I think we should be consistent though, see the earlier error checks. We could perhaps hard fail in those situations too. We should generally try to leave strings with lengths that match their valid data though. That makes debugging easier and reduces risk introducing totally random data into places and if it's just to look at.

PR #761, opened 2015-07-01
Make GeoIP look for the databases also in libpath

maek:
When GeoIP is compiled as a static library it looks for the databases in a (absolute) path related to the environment where it was compiled. In order to make it work even in that case, the solution I would propose is to make it look for the databases also in `libpath`.

### DIFF
--- a/src/engine/qcommon/net_ip.cpp
+++ b/src/engine/qcommon/net_ip.cpp
@@ -1985,11 +1985,20 @@ const char *NET_GeoIP_Country( const netadr_t *from )
 
 static GeoIP *NET_GeoIP_LoadData (int db)
 {
-	GeoIP *data = GeoIP_open_type (db, GEOIP_INDEX_CACHE);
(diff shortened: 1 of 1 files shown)


maek:
The Travis CI check fails with
```
error: ‘GEOIP_SILENCE’ was not declared in this scope
```
because you are using `libgeoip-dev_1.4.8` when the current dependency (from `external_deps/build.sh`) is 1.6.4 and GEOIP_SILENCE has been added to GeoIP in the 1.6.3 release.

If you don't want to update the Travis CI recipe but you are still interested in this PR I may remove that flag (I've added it to prevent the engine showing an error on the first path check, thing that always happens when GeoIP is statically compiled).

DolceTriade:
../src/engine/qcommon/net_ip.cpp:1988:58: error: ‘GEOIP_SILENCE’ was not declared in this scope

Perhaps either include a min version in the CMake search function or guard against that flag with a define in the code.

PR #771, opened 2015-07-15
Make PNaCl a first-class CMake toolchain

TimePath:
* Split engine and game build scripts
* Target specific warnings


DolceTriade:
libRocket is already gamelogic only.

PR #773, opened 2015-07-19
allow extension independent soundfile path, fix #642

### LINKED ISSUE #642: extension independent sound file path to be able to update res-packages with new format
Hi, it's not a bug, it's a feature request! :smiley: 

You know, the texture shader path is extension independent. You can ship a `tex-package_1.pk3` containing shaders with `.tga` extension then you can replace it by a `tex-package_2.pk3` with shaders with `.crn` extension without breaking maps that depend on this shaders.

But if you ship a `tex-package_1.pk3` with `.wav` sound files, then you replace it with a  `tex-package_2.pk3` with `.opus` sound files, you must rebuild all the maps that depends on it because the sound file path is extension dependant.

I read somewhere that you want to have sound shaders, but before that and without that, having a extension independant sound file path will be a great improvement :+1:

illwieckz:
Hi, this PR allows the engine to use extension independent sound path, so mappers can just write _elemusic_ to load _elemusic.wav_ or _elemusic.opus_, if one of them is available.

Also, the engine can load the _elemusic.opus_ path if the bsp was compiled with _elemusic.wav_ path. It allows asset managers to convert sound file without having to recompile the bsp.

It's the first time I'm writing some c++ lines in my life, so your comments are welcome, perhaps there is a better way to do it! :wink: 

This will also help out-of-tree map packaging, allowing mappers to compile their map against lossless assets and distribute them with lossy assets.

This PR fixes issue #642.

### DIFF
--- a/daemon/src/engine/audio/SoundCodec.cpp
+++ b/daemon/src/engine/audio/SoundCodec.cpp
@@ -32,26 +32,77 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 namespace Audio {
 
+typedef struct
+{
+	const char *ext;
+	AudioData (*SoundLoader) (std::string);
+} soundExtToLoaderMap_t;
+
+// Note that the ordering indicates the order of preference used
+// when there are multiple sound files of different formats available
+static const soundExtToLoaderMap_t soundLoaders[] =
+{
+	{ ".wav",	LoadWavCodec },
+	{ ".opus",	LoadOpusCodec  },
+	{ ".ogg",	LoadOggCodec  },
+};
+
+static int numSoundLoaders = ARRAY_LEN(soundLoaders);
+
 AudioData LoadSoundCodec(std::string filename)
 {
 
-	size_t position_of_last_dot{filename.find_last_of('.')};
+	std::string ext = FS::Path::Extension(filename);
 
-	if (position_of_last_dot == std::string::npos) {
-		audioLogs.Warn("Could not find the extension in %s", filename);
+	// if filename has extension, try to load it
+	if (ext != "") {
+		// look for the correct loader and use it
+		for (int i = 0; i < numSoundLoaders; i++) {
+			if (ext == soundLoaders[i].ext) {
+				// if file exists, load it
+				if (FS::PakPath::FileExists(filename)) {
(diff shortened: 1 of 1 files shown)


Amanieu:
Use FS::Path::Extension and FS::Path::BaseName instead of doing it manually. Looks good otherwise.

Kangz:
Please consider using the same approach as in tr_image: https://github.com/Unvanquished/Unvanquished/blob/master/daemon/src/engine/renderer/tr_image.cpp#L2030 as it handles paks dependencies too.
I don't like that we silently replace the extension as it is a bit confusing though. Can we only add an extension of none is present?

DolceTriade:
Hm, I didn't read the code close enough. I agree with kangz that prefer extension if given, otherwise, if none is provided or the extension given is not found, then we try all the available extensions and see what sticks.

illwieckz:
> Use FS::Path::Extension and FS::Path::BaseName instead of doing it manually. Looks good otherwise.

Hum, it's ok for `FS::Path::Extension`, it returns `.ogg` if file is `sound/ui/heartbeat2.ogg`.
But `FS::Path::BaseName` is not the good one to use: it returns `heartbeat2.ogg` but I need `sound/ui/heartbeat2`.

Sorry if I fooled you with the “basename” variable, it's the full path without extension. Is there a FS::Path function to achieve that?

Amanieu:
Oh sorry, the correct function should have been FS::Path::StripExtension

illwieckz:
Thanks, and I will use another variable name to not be more explicit. :-)

illwieckz:
So, I modified my patch using `FS::Path::Extension` and `FS::Path::StripExtension` and with a package prioritization when looking for an alternative name if file is missing or use an unknown file format.

The travis-ci's build errored because of `The command "sudo apt-key adv --keyserver x-hkp://pool.sks-keyservers.net --recv-key 06AF4C14" failed and exited with 2 during .`, it's not related to this patch.

Kangz:
That's much better, thanks for the iteration! I'll merge it soon unless @DolceTriade or @Amanieu have other concerns.

illwieckz:
@Amanieu, @Kangz, @DolceTriade, do you need something more (perhaps spacing?) to be able to merge this PR in unv 0.42.0 ?

DolceTriade:
LGTM

illwieckz:
Thanks for the merge.

PR #774, opened 2015-07-25
beacon: display “Here” if player <= 1m from entity

illwieckz:
Hi, this is a little patch that substract 1m to beacon distance since IRL, people don't say something is at 1m if it's less than 1m (so, people truncates, they do not rounds).

Currently is it weird to stand on a telenode and see _1m_, also, the only way to see less than _1m_ is to navigate with `noclip` in the center of the entity. So the current feel not natural at all. In fact, when Unvanquished says something is at 3m, it looks like it's at 2m.

Before:
telenode 1m
egg 1m

After:
telenode 0m
egg 0m

Since I subtract the distance, I added a check to not subtract 0 (only happens with `noclip`, you must fly inside the building to have a 0m distance).

### DIFF
--- a/src/cgame/cg_beacon.cpp
+++ b/src/cgame/cg_beacon.cpp
@@ -613,8 +613,15 @@ static void HandHLBeaconToUI()
 		}
 
 		// distance
-		Com_sprintf( br->distance, sizeof( br->distance ), "%im from here",
-		             (int)round( beacon->dist * QU_TO_METER ) );
+		int distance = (int)floor( beacon->dist * QU_TO_METER );
+		if ( distance > 1 )
+		{
+			Com_sprintf( br->distance, sizeof( br->distance ), "%im from here", distance );
+		}
+		else
+		{
+			Com_sprintf( br->distance, sizeof( br->distance ), "Here" );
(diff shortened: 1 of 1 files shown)


DolceTriade:
I think 0m is wrong. Perhaps a more general word such as "very close" or something similar. @Viech thoughts?

illwieckz:
I agree with the “0m is wrong” and a “very close” is doable and will look good, good idea.

But it's not only that, when you stand less than 3m of a turret, unvanquished displays “3m”, IRL people says 2m if it's less than 3m.

So I can change my patch like that:

if 0:

DolceTriade:
Yes, but I also think "very close" is bad, but another phrase with a similar meaning. Perhaps @kharnov can help with this.

TimePath:
Is "<0m" too obvious?

illwieckz:
Now:

Here

illwieckz:
The code now use `floor()` instead of `round() - 1`

Tomtom-mygit:
“Here” sounds good.

illwieckz:
Thanks.

PR #810, opened 2015-10-12
Colored String Refactoring

mbasaglia:
# Overview
This branch introduces more flexible string-encoded colors, and in doing so
it introduces a significant refactoring of the parts of the code that handle
colors or color-coded strings.

# New Features
* String support hexadecimal notation for colors: <tt>^xf00</tt> and <tt>^#ff0000</tt>.
* The rules that discarded invalid player names are a bit more relaxed.
* The log to file mixed strings with color codes being stripped away and strings which preserved them, now it always strips colors.
* The maximum number of characters in a player name has been increased.
* Some invalid single-character color codes that were previously parsed as colors are no longer supported.

# Bug Fixes
* Inconsistencies in color parsing have been removed, now every part of the game and engine which supports colors supports all of the available color formats.
* Low ammo blinking uses the color specified in the source, previously was interpolated in an odd way.

# Refactors
* All general-purpose color-related functions are declared in the <tt>Color</tt> namespace, declared in common/Color.h
* **BasicColor** class to replace colors represented as C arrays
* All color parsing has been changed to use token iterators
* How the game console handle the log has been completely re-written to accommodate the changes

# Design
* The **Color** class uses **float** to represent the RGBA components. Using bytes would be faster and require less network traffic but it was maintained layout compatibility. Similarly, the parts of the code which used **char** now use **Color32Bit** which maintains the data layout.
* The **ColorAdaptor** class template is used to have an explicit syntax for unsafe casts and allows extension to the color class (like Rocket::Core::Colour).
* The **CString** function uses some ugly and unsafe conventions used by some of the legacy code, this is required for when colors are used with unsafe printf-like functions.
* The tokens of the color parser are exposed because many color-parsing function require special actions.
* The **TokenIterator** class, used to parse color strings also parses utf-8 since all functions that parsed colors also parsed that.

# Factors that affected the design
* Backwards compatibility with layout, string manipulation, and function calls.
* Performance: the new code is never slower than the old one and avoids creating objects where possible.
* Type safety and explicit conversions.
* Ease of use at the call site.


daemon/src/engine/qcommon/qfiles.h
```diff
@@ -575,7 +575,7 @@ typedef struct
 	float  st[ 2 ];
 	float  lightmap[ 2 ];
 	vec3_t normal;
-	byte   color[ 4 ];
+	Color::Color32Bit color;
```
Amanieu: This structure is part of the BSP file format, I think it would be best to leave it as an array of bytes since it is read directly from a file. In particular, byte is a POD type by Color32Bit isn't.

daemon/src/common/Color.h
```diff
@@ -0,0 +1,578 @@
...
+ */
+template<class T>
+	struct ColorComponentTraits
+{
+	// Type used to represent the components
+	using component_type = T;
+	// Maxiumum value for a component
+	static CONSTEXPR component_type component_max = std::numeric_limits<component_type>::max();
+	// Size of a component value in bytes
+	static CONSTEXPR const std::size_t component_size = sizeof(component_type);
```
Amanieu: Don't use both CONSTEXPR and const, since CONSTEXPR expands to "const" if it is not supported, which will result in "const const".

daemon/src/common/Color.h
```diff
@@ -0,0 +1,578 @@
...
+};
+/*
+ * Specialization for normalized floats
+ */
+template<>
+	struct ColorComponentTraits<float>
+{
+	using component_type = float;
+	static CONSTEXPR component_type component_max = 1.0f;
```
Amanieu: Float constants don't work in MSVC.

daemon/src/engine/qcommon/qfiles.h
mbasaglia: Color32Bit is a standard layout type. Even though it isn't a trivial type, its default constructor does nothing but zero-initialize its members

PR #814, opened 2015-10-21
Fix crash in daemonded when calling /systeminfo

kungfooman:
Run "make server", choose a map and call /systeminfo, it should not crash anymore. Info_Print(char *s) will not check any array boundaries, so I just increased the static buffer length, as it is done in ioquake3:

https://github.com/ariya/ioquake3/blob/master/code/qcommon/q_shared.h#L228
https://github.com/ariya/ioquake3/blob/master/code/qcommon/common.c#L521

### DIFF
--- a/daemon/src/engine/qcommon/common.cpp
+++ b/daemon/src/engine/qcommon/common.cpp
@@ -433,8 +433,8 @@ bool Com_AddStartupCommands()
 
 void Info_Print( const char *s )
 {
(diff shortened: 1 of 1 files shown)


Kangz:
Good catch, I'll submit once CI is done.

Thank you for the patch

Kangz:
The Appveyor failure seems unrelated, merging.

PR #815, opened 2015-10-22
Mouse Improvements

mbasaglia:
This branch improves how the mouse is handled:
* When the cvar in_nograb is true, the mouse will never be locked inside the game
* Removed the usused cvar in_uigrab
* Cursor mode: 
  * When cursor mode is active (menu and stuff), mouse movements are sent as coordinates instead of deltas and the in-game cursor is positioned where the cursor of the window manager should be. 
  * When the mode is not active, the old behaviour is used.
  * When switching between the two, the mouse is moved to the center of the window
* The loading screen no longer displays a cursor since it doesn't handle input


daemon/src/engine/null/null_rocket.cpp
```diff
@@ -84,3 +84,4 @@ void Rocket_RegisterProperty( const char*, const char*, bool, bool, const char*
 void Rocket_ShowScoreboard( const char*, bool ) {}
 void Rocket_SetDataSelectIndex( int ) {}
 void Rocket_LoadFont( const char* ) {}
+void Rocket_HideMouse() {}
```
DolceTriade: Actually, I think you can remove this file entirely.

daemon/src/engine/client/cg_api.h
```diff
@@ -113,6 +113,8 @@ typedef enum
 	SORT_FAVOURITES
 } serverSortField_t;
 
+void Rocket_HideMouse();
```
DolceTriade: Should this go in rocket.h?

daemon/src/engine/client/cg_api.h
```diff
@@ -213,6 +215,8 @@ void            trap_Key_SetBinding( int keyNum, int team, const char *cmd );
 void            trap_Key_ClearCmdButtons( void );
 void            trap_Key_ClearStates( void );
 std::vector<int> trap_Key_KeysDown( const std::vector<int>& keys );
+bool            trap_Mouse_GetCursorMode();
```
DolceTriade: Can this be renamed to IsCursorVisible/Active? Mode seems to imply that there are multiple states but this just seems to hide/show the cursor.

daemon/src/engine/sys/sdl_glimp.cpp
```diff
@@ -769,7 +769,7 @@ static int GLimp_SetMode( int mode, bool fullscreen, bool noborder )
 				continue;
 			}
 			SDL_GL_SetSwapInterval( r_swapInterval->integer );
-			SDL_ShowCursor( 0 );
+			//SDL_ShowCursor( 0 );
```
DolceTriade: Any reason to leave it commented? If its not necessary, might be best to just delete it.

src/cgame/cg_local.h
```diff
@@ -1727,8 +1727,8 @@ typedef struct
 	// corpse info
 	clientInfo_t corpseinfo[ MAX_CLIENTS ];
 
-	int          cursorX;
-	int          cursorY;
+	int          cursorX; // unused
```
DolceTriade: If unused, delete?

DolceTriade:
I will make a for-0.45.0 branch. Since this breaks compat, this should be merged into a branch and not master.

daemon/src/engine/client/cg_api.h
mbasaglia: It doesn't show or hide the cursor, it changes how the mouse movement is sent to cgame.

daemon/src/engine/sys/sdl_glimp.cpp
mbasaglia: Ah yeah forgot about that

src/cgame/cg_local.h
mbasaglia: yeah, is that struct safe to be modified?

daemon/src/engine/null/null_rocket.cpp
mbasaglia: Hmm, ok then

daemon/src/engine/client/cg_api.h
mbasaglia: OK

daemon/src/engine/sys/sdl_input.cpp
```diff
@@ -569,94 +567,63 @@ static keyNum_t IN_TranslateSDLToQ3Key( SDL_Keysym *keysym, bool down )
...
 /*
-===============
-IN_DeactivateMouse
-===============
-*/
-void IN_DeactivateMouse( bool showCursor )
+ * Toggles forced cursor, this ensure that the cursor is active and visible
+ * even when not in cursor mode
+ */
+static void ForceCursor(bool force)
```
Kangz: It is not clear how this and SetCursorMode will interact. Maybe it would be possible to evaluate the different conditions every frame and then apply the new state we want to be in?

Kangz:
Overall looks good to me, the logic in sdl_input seems a bit unclear.

Thanks for the cleanup, that has been on our TODO list for ages (and I even tried to do that, twice, without success).

daemon/src/engine/null/null_rocket.cpp
Kangz: Just build all three executables, if the functions are used you'll get link errors.

src/cgame/cg_local.h
Kangz: Yes, it is used only in the VM, struct unsafe to be modified are in tr_types.h, qcommon.h, cg_api.h, cg_msgdefs.h and the same for g_

daemon/src/engine/sys/sdl_input.cpp
mbasaglia: ForceCursor temporarily enables the cursor to avoid it being locked, when called with false it will restore whatever mouse mode was.
And it ensures the cursor is visible.

daemon/src/engine/client/cg_api.h
mbasaglia: Ah also, I ended up not needing the Get function so I can remove it

src/cgame/cg_local.h
DolceTriade: Plus your change breaks compat anyways, so it doesn't matter.

daemon/src/engine/client/cg_api.h
DolceTriade: In that case, true/false is a little opaque. Can you switch to an int with descriptive constant names? ie, MOUSE_MOUSE_DELTA, MOUSE_MODE_ABSOLUTE, since the what "mode" is not clear from the function name or arguments.

daemon/src/engine/sys/sdl_input.cpp
Kangz: My concern was more that SetCursorMode doesn't know if we are in a forced state or not, and might overwrite what ForceCursor did. Is it an issue at all?

Kangz:
Thanks for the fixes, I have one more question but otherwise it looks good to me.

daemon/src/engine/sys/sdl_input.cpp
mbasaglia: I don't think it would be a problem, since the window will most likely not have focus.
TBH I think it would be cleaner to switch cursor mode just once when needed than having all those checks at every frame, but that's how the old code did it and didn't know how to change this.

Kangz: Yeah we ca revisit that later. Thanks for the answer.

Kangz:
Thank you for the patch, merging!

PR #816, opened 2015-10-24
Rewrite how the console history is handled

mbasaglia:
This branch encapsulates the console history in a class.
In doing so, the overall code has been cleaned up a bit, fixing several minor bugs and making it thread-safe.

I'm not sure whether it would be worth merging this class with Console::Field since it's basically only used there.

### DIFF
--- a/daemon/src/engine/framework/ConsoleHistory.h
+++ b/daemon/src/engine/framework/ConsoleHistory.h
@@ -31,10 +31,6 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef FRAMEWORK_CONSOLE_HISTORY_H_
 #define FRAMEWORK_CONSOLE_HISTORY_H_
 
-#include <string>
-#include <vector>
-#include <mutex>
-
 namespace Console {
 
     class History
--- a/daemon/src/engine/framework/ConsoleHistory.h
+++ b/daemon/src/engine/framework/ConsoleHistory.h
@@ -31,18 +31,40 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef FRAMEWORK_CONSOLE_HISTORY_H_
 #define FRAMEWORK_CONSOLE_HISTORY_H_
 
+#include <string>
+#include <vector>
+#include <mutex>
+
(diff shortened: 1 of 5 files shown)


daemon/src/engine/framework/ConsoleHistory.h
```diff
@@ -31,18 +31,40 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef FRAMEWORK_CONSOLE_HISTORY_H_
 #define FRAMEWORK_CONSOLE_HISTORY_H_
 
+#include <string>
+#include <vector>
+#include <mutex>
```
Amanieu: Avoid including system headers all over the place. We include all the system headers in q_shared.h, which is included by Common.h, which is built as a precompiled header and included everywhere.

mbasaglia: So should I include Common.h?
ConsoleHistory.cpp had
```
#include "qcommon/q_shared.h"
#include "qcommon/qcommon.h"
```
Though only stuff from q_shared.h would be needed

Amanieu: Well generally we have it set up so that headers don't need to include other headers. Since Common.h is directly or indirectly included in every .cpp file, just removing the #includes from headers should be fine.

mbasaglia: Should I change the two include mentioned above to Common.h on the cpp file?

Kangz: As I think you've noticed, Common.h is included in every file using a compiler flag (this is need for PCH)

Kangz:
Had a couple comments, I think I'll like that new way history works better. When originally developing it we wanted to have an irssi-style history but it isn't what most people are used to so I have no regret switching to something simpler.

daemon/src/engine/framework/System.cpp
```diff
@@ -277,6 +277,9 @@ static void Shutdown(bool error, Str::StringRef message)
 	// Stop accepting commands from other instances
 	CloseSingletonSocket();
 
+	// Save the console history
+	Console::History::Save();
```
Kangz: This is not needed anymore

Kangz:
Thank you, merging

PR #818, opened 2015-10-25
Fix a typo on sdl_input.cpp

mbasaglia:
By the way, does anyone know if the code enclosed by that #if is really needed?


Kangz:
From what I know of X11 that may very well be the case and since it is hard to test we might as well leave the workaround in place.

PR #826, opened 2015-11-01
Goon Barbs regenerate faster near boosters and creep

caldazar-at:
When an advanced goon stands near a booster, it regenerates barbs 3 times faster.
Standing on creep accelerates barb regeneration by a factor 1.5.


Viech:
I'm fine with the gameplay change as such. The open question is how we visualize it, as it's quite a strong buff and (alien) players should be aware of what's up. Right now it looks like the red healing mist effect of the booster is also used when only aiding in barb generation. I don't think it's a good idea to overload the effect like this. In particular, humans want to see at a glance when an alien is being healed by a booster. (Also an alien that needs no healing doesn't want to give away the booster's position.) In my opinion, a better option would be to have the barb ammo display pulsate with a frequency proportional to the regen frequency. Do you think you could add something like this, @caldazar-at?

caldazar-at:
@Viech I added the config variables and removed the red mist when the alien only regenerates barbs.
I use SS_HEALING_4X and SS_HEALING_8X to determine if the alien is on creep or near a booster, respectively. I'm not too happy about that but the best alternative I found was to iterate over entities which I liked even less. Any suggestions?
To make the barb ammo display pulsate I'll have to ask quite a few stupid questions on IRC. Should be doable, though.

Viech:
I think using the `SS_HEALING_*` flags is fine as network bits are a sparse resource – and coupling barb regeneration to health regeneration makes sense for simplicity, too. Note that the healing flags are set and transmitted even when the alien doesn't need health (and a healing source is available). That's exactly what we need for displaying barb regeneration: If a healing source is available and there is barb missing, the alien HUD can display this accordingly.

Viech:
@DolceTriade can you add annotations to the libRocket related code? (I'd be interested in them, too.)

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,105 @@ class PredictedMineEfficiencyElement : public HudElement
...
+	HudElement ( tag, ELEMENT_ALIENS ),
+	numBarbs( 0 ),
+	regenerationInterval ( 0 ),
+	t0 ( 0 ),
+	offset ( 0 ) {}
+
+	void DoOnUpdate()
+	{
+		// GetAttribute returns empty string when called in constuctor so it's here instead
+		Q_strncpyz ( src, GetAttribute<Rocket::Core::String> ("src", "").CString(), sizeof(src) );
```
DolceTriade: Cache the src in OnAttributeChanged. Store it in a Rocket::Core::String.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,105 @@ class PredictedMineEfficiencyElement : public HudElement
...
+		// draw regenerating barb
+		if ( numBarbs < BG_Weapon( WP_ALEVEL3_UPG )->maxAmmo )
+		{
+			// normalize
+			float opacity = getSin()/3 + 0.5;
+			Com_sprintf( base, sizeof( base ), "<img class='barbs' src='%s' style='opacity: %s; color: grey;' />", src, va("%f", opacity) );
+			Q_strcat( rml, sizeof( rml ), base );
+		}
+
+		SetInnerRML( rml );
```
DolceTriade: I'll clean this part up with respect to not having to call this over and over.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,105 @@ class PredictedMineEfficiencyElement : public HudElement
...
+	}
+
+private:
+
+	float getSin()
```
DolceTriade: GetSin()

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,105 @@ class PredictedMineEfficiencyElement : public HudElement
...
+	{
+		float t = (cg.time-t0) / 1000.0;
+		// frequency in Hz; Interval is in ms
+		float f = 4 * 1000.0/regenerationInterval;
+		return sin( offset + t * 2*M_PI * f );
+	}
+
+	int getInterval()
```
DolceTriade: GetInterval

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,105 @@ class PredictedMineEfficiencyElement : public HudElement
...
+		SetInnerRML( rml );
+	}
+
+private:
+
+	float getSin()
+	{
+		float t = (cg.time-t0) / 1000.0;
+		// frequency in Hz; Interval is in ms
+		float f = 4 * 1000.0/regenerationInterval;
```
DolceTriade: pad operators wtih spaces

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,105 @@ class PredictedMineEfficiencyElement : public HudElement
...
+		return sin( offset + t * 2*M_PI * f );
```
DolceTriade: pad operators with spaces.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,105 @@ class PredictedMineEfficiencyElement : public HudElement
...
+			{
+				Q_strcat( rml, sizeof( rml ), base );
+			}
+		}
+
+		// draw regenerating barb
+		if ( numBarbs < BG_Weapon( WP_ALEVEL3_UPG )->maxAmmo )
+		{
+			// normalize
+			float opacity = getSin()/3 + 0.5;
```
DolceTriade: pad operators with spaces.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,111 @@ class PredictedMineEfficiencyElement : public HudElement
...
+
+		// draw regenerating barb
+		if ( numBarbs < BG_Weapon( WP_ALEVEL3_UPG )->maxAmmo )
+		{
+			// normalize
+			float opacity = GetSin() / 3 + 0.5;
+			Q_strcat( rml, sizeof( rml ), va( "<img class='barbs' src='%s' style='opacity: %s; color: grey;' />", src.CString(), va( "%f", opacity) ) );
+		}
+
+		SetInnerRML( rml );
```
DolceTriade: Currently this will set RML each time, which is quite expensive. Setting RML will remove all children from the parent and then re-add them, which means you will need to re-layout everything. You should only set RML once, and then you should store a reference to each of the img elements, and then call element->SetProperty("opacity", va("%f", opacity)) to modify the look.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,133 @@ class PredictedMineEfficiencyElement : public HudElement
...
+
+	void OnAttributeChange( const Rocket::Core::AttributeNameList& changed_attributes )
+	{
+		HudElement::OnAttributeChange( changed_attributes );
+		if ( changed_attributes.find( "src" ) != changed_attributes.end() )
+		{
+			int m = BG_Weapon( WP_ALEVEL3_UPG )->maxAmmo;
+			if ( m > 0 )
+			{
+				char base[ MAX_STRING_CHARS ];
```
DolceTriade: No need to use char. You can use Rocket::Core::String::Format(<length>, "format", ...);
str += text to append

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,133 @@ class PredictedMineEfficiencyElement : public HudElement
...
+	regenerationInterval ( 0 ),
+	t0 ( 0 ),
+	offset ( 0 ) {}
+
+	void OnAttributeChange( const Rocket::Core::AttributeNameList& changed_attributes )
+	{
+		HudElement::OnAttributeChange( changed_attributes );
+		if ( changed_attributes.find( "src" ) != changed_attributes.end() )
+		{
+			int m = BG_Weapon( WP_ALEVEL3_UPG )->maxAmmo;
```
DolceTriade: nit: avoid single character variable names. People will thank you when they read the code later.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,133 @@ class PredictedMineEfficiencyElement : public HudElement
...
+			t0 = cg.time;
+			offset = asin( GetSin() ) - M_PI_2;
+			regenerationInterval = interval;
+		}
+		numBarbs = barbs;
+	}
+
+	void DoOnRender()
+	{
+		for ( int i=0; i<children.size(); i++)
```
DolceTriade: pad operators with spaces. (=, <)

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,133 @@ class PredictedMineEfficiencyElement : public HudElement
...
+		{
+			// black math magic to avoid sudden jumps in opacity
+			t0 = cg.time;
+			offset = asin( GetSin() ) - M_PI_2;
+			regenerationInterval = interval;
+		}
+		numBarbs = barbs;
+	}
+
+	void DoOnRender()
```
DolceTriade: This should go in DoOnUpdate() since the calls below aren't actual draw calls. They are just altering the state of the elements to be drawn.

caldazar-at: This is in DoOnUpdate().

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,134 @@ class PredictedMineEfficiencyElement : public HudElement
...
+				src = GetAttribute<Rocket::Core::String>( "src", "" );
+				Rocket::Core::String base( va("<img class='barbs' src='%s' />", src.CString() ) );
+				Rocket::Core::String rml;
+
+				for ( int i = 0; i < maxBarbs; i++ )
+				{
+					rml += base;
+				}
+
+				Com_Printf("base = %s", base.CString() );
```
DolceTriade: Delete debug code.

src/cgame/cg_rocket_draw.cpp
Viech: Something's wrong with @DolceTriade. He doesn't sound like himself. :stuck_out_tongue:

Viech:
@DolceTriade did you annotate the commits or the squashed diff ("Files changed")? It seems my annotations (on commits) got deleted with the rebase while yours persist.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,131 @@ class PredictedMineEfficiencyElement : public HudElement
...
+		if ( barbs > numBarbs || ( barbs < numBarbs && numBarbs == maxBarbs ) )
+		{
+			t0 = cg.time;
+			// sin(-pi/2) is minimal
+			offset = -M_PI_2;
+		}
+		// change regeneration speed
+		else if ( interval != regenerationInterval )
+		{
+			// black math magic to avoid sudden jumps in opacity
```
Viech: Doesn't seem to work ingame, I can see opacity jumps when I enter or leave creep.

Observation: Why do you need both `t0` and an `offset`? Wouldn't it be sufficient to calculate only an offset in [0, 2*`M_PI`] such that the sine wave is in the same phase with the new frequency? You could use the same offset to reset the phase when you want to start regeneration at a certain opacity.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,131 @@ class PredictedMineEfficiencyElement : public HudElement
...
+	regenerationInterval ( 0 ),
+	t0 ( 0 ),
+	offset ( 0 ) {}
+
+	void OnAttributeChange( const Rocket::Core::AttributeNameList& changed_attributes )
+	{
+		HudElement::OnAttributeChange( changed_attributes );
+		if ( changed_attributes.find( "src" ) != changed_attributes.end() )
+		{
+			maxBarbs = BG_Weapon( WP_ALEVEL3_UPG )->maxAmmo;
```
Viech: As it stands the data accessed through `BG_Weapon` isn't going to change within a match. As long as new `BarbsHudElement` objects are created on each `vid_restart` or map load (@DolceTriade, is that the case?), then you could also add this to the initialization list above. In direct comparisong it looks like your current approach is more error-resistant but the assumption that the value of `BG_Weapon( WP_ALEVEL3_UPG )->maxAmmo` can change between calls to `OnAttributeChange` but not between calls to `DoOnUpdate` seems a bit arbitrary to me.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,131 @@ class PredictedMineEfficiencyElement : public HudElement
...
+				for ( int i = 0; i < maxBarbs; i++ )
+				{
+					rml += base;
+				}
+
+				// SetInnerRML deletes all child Elements => bring children up to date
+				children.clear();
+				SetInnerRML( rml );
+				for( int i = 0; i < maxBarbs; i++) {
+					children.push_back( GetChild( i ) );
```
Viech: Why do you create a copy of the children? To me it seems like you could just iterate over the superclasses children directly all the time, assuming that `GetChild` accesses them. (Note that I have no idea what these children are in the first place.)

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,131 @@ class PredictedMineEfficiencyElement : public HudElement
...
+
+	float GetSin()
+	{
+		float t = (cg.time - t0) / 1000.0;
+		// frequency in Hz; Interval is in ms
+		float f = 4 * 1000.0 / regenerationInterval;
+		return sin( offset + t * 2 * M_PI * f );
+	}
+
+	int GetInterval()
```
Viech: You could put this into a `BG_*` style function in the shared code to guarantee that this is consistent in both client and server.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,131 @@ class PredictedMineEfficiencyElement : public HudElement
...
+		}
+	}
+
+	void DoOnUpdate()
+	{
+		int barbs = cg.snap->ps.ammo;
+		int interval = GetInterval();
+
+		// start regenerating barb now
+		if ( barbs > numBarbs || ( barbs < numBarbs && numBarbs == maxBarbs ) )
```
Viech: You may want to rename `barbs` to `newNumBarbs` or so for clarity.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,131 @@ class PredictedMineEfficiencyElement : public HudElement
...
+		else if ( interval != regenerationInterval )
+		{
+			// black math magic to avoid sudden jumps in opacity
+			t0 = cg.time;
+			offset = asin( GetSin() ) - M_PI_2;
+			regenerationInterval = interval;
+		}
+		numBarbs = barbs;
+
+		for ( int i = 0; i < children.size(); i++)
```
Viech: Using `i` in small loops is generally fine I guess. Especially for bigger loops I prefer names like `childNum` though.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,131 @@ class PredictedMineEfficiencyElement : public HudElement
...
+		{
+			if (i < numBarbs ) // draw existing barbs
+			{
+				children[i]->SetProperty("color", "white");
+				children[i]->SetProperty("opacity", "1.0");
+			}
+			else if (i == numBarbs ) // draw regenerating barb
+			{
+				float opacity = GetSin() / 3 + 0.5; // normalize
```
Viech: Is the `3` intended? If yes, the comment saying "normalize" is a bit misleading.

src/cgame/cg_rocket_draw.cpp
caldazar-at: @Viech GetChild(i) returns a pointer so I just cache references to the children. I could delete the vector and instead use GetChild() every time, but I don't think there's much of a difference.
The librocket Element to display the barbs has 3 child elements - one to display each barb.

src/cgame/cg_rocket_draw.cpp
caldazar-at: @Viech Yes, 3 is intended. You didn't want the barb to become invisible.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,131 @@ class PredictedMineEfficiencyElement : public HudElement
...
+				children[i]->SetProperty("color", "grey");
```
Viech: I don't think you need the color change. Just experiment with the opacity. That allows for different (colored) barb icons to work well.

src/cgame/cg_rocket_draw.cpp
Viech: You could also derive the weapon held from `cg` if I remember correctly, this would allow the code to work for different alien classes that get some notion of barbs to attack with in the future.

src/cgame/cg_rocket_draw.cpp
Viech: If the number of children will always be consistent with the number of barbs then I would argue to just use `GetChild` and not manage another data structure that does basically the same. (Not mainly for performance reasons but to keep the code small and readable.)

If you can think of us adding non-barb children at some point which would potentially break the code then stick with this but justify it with the variable name, i.e. call it `barbs` or `barbChildren` instead of `children`, because then the point is to have a structure containing *specific* children.

src/cgame/cg_rocket_draw.cpp
Viech: I see. I'd just remove the comment then as normalizing kind of implies bringing something to a scale of [0,1]. Also I do find, even with you currently using grey instead of white, that the opacity gets too high (currently 5/6). I would have imagined something like a range of 0.25 to 0.5, hence `GetSin() / 8.0 + (3.0/8.0)`.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,130 @@ class PredictedMineEfficiencyElement : public HudElement
...
+
+		if ( newNumBarbs < maxBarbs )
+		{
+			// start regenerating barb now
+			if ( newNumBarbs > numBarbs || ( newNumBarbs < numBarbs && numBarbs == maxBarbs ) )
+			{
+				t0 = cg.time;
+				// sin(-pi/2) is minimal
+				offset = -M_PI_2;
+				assert( GetSin() == -1.0 );
```
Viech: Personally I wouldn't spam asserts for stuff that's mainly visual and not really critical. In particular I wouldn't expect exact floats on all platforms.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,130 @@ class PredictedMineEfficiencyElement : public HudElement
...
+			}
+			// change regeneration speed
+			else if ( interval != regenerationInterval )
+			{
+				float sOld = GetSin();
```
Viech: `sinOld` and `cosOld` would be better for readability.

src/sgame/sg_active.cpp
```diff
@@ -1049,7 +1049,8 @@ void ClientTimerActions( gentity_t *ent, int msec )
 	{
 		if ( client->ps.ammo < BG_Weapon( WP_ALEVEL3_UPG )->maxAmmo )
 		{
-			if ( ent->timestamp + LEVEL3_BOUNCEBALL_REGEN < level.time )
+			int bounceballRegen = BG_GetBarbRegenerationInterval(ent->client->ps);
+			if ( ent->timestamp + bounceballRegen < level.time )
```
Viech: The problem with this old approach is that I can shoot while not on creep or near a booster, wait in place for `ballRegenBooster` ms, then enter the booster range and instantly get a barb. This harms the intuition given by the pulse frequency. Alternatively, you could add a `gentity_t` float field where, each frame where you don't have full ammo, you add the frame time divided by the regen interval and when this value gets greater than 1, substract 1 from it (or reset it to 0) and add a barb. (Don't be afraid of adding single-purpose `gentity_t` fields, they will all go into components later.)

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2174,6 +2174,130 @@ class PredictedMineEfficiencyElement : public HudElement
...
+				}
+				regenerationInterval = interval;
+
+				assert( fabs(GetSin() - sOld ) < 0.001);
+				assert( GetCos() * cOld > 0 );
+			}
+		}
+		numBarbs = newNumBarbs;
+
+		for ( int i = 0; i < GetNumChildren(); i++ )
```
Viech: I'll try some things after we merge. I think displaying missing barbs at 25% and existing barbs at 100% opacity and pulsating between 25% and 50% could work. For now this looks good.

Viech:
The things I annotated I'd do myself if you don't want to spend any more time on them. You can also squash the commits or leave it to me. Just tell me when you want me to merge.

caldazar-at:
@Viech Feel free to merge.

Viech:
Merged as 3837d11. Thanks @caldazar-at!

PR #828, opened 2015-11-02
Connectionless commands, rcon and encryption

mbasaglia:
The purpose of this branch is to improve interoperability with remote scripts and to refactor certain parts of the networking code.

### Details
* Improved OOB **getstatus** and **getinfo**
* Added the **ping** OOB command, so scripts can just check if a server is online. (This is more lightweight that getinfo)
* New challenge system, improved, thread-safe, flexible and with nicer challenges. This new system can be set up to have a configurable challenge size, the challenge is stored as binary data and displayed as a hexadecimal string. Challenge representations always end up having the same length unlike the current signed integer representation.
* Higher-level wrappers around nettle for encryption, hashing and similar.
* Secure rcon command (**srcon**), which supports challenges and encryption
* C++ alternative for info string manipulation, used by the connectionless commands, other parts of the code still use the old C version.
* Server-side cvars to configure the requirements for **rcon** and **srcon**
* A command for auto-discovery of rcon settings, so a client only have to know the server password 
* Used MAX_MSGLEN for rcon environment
* Re-written the low-level out of band networking functions
* Improved the **status** server command
* Added server-side **listmaps** command
* Fixed some misc networking bugs
* Cvar server.private replaces server.lanOnly to allow finer control
* Refactored how the engine keeps track of master servers
* Fix localhost name lookup.

### Compatibility
* New clients are still compatible with older servers
* Old clients may have a hard time connecting to new servers as the challenge format has changed

### Testing the new protocols
To test the new OOB commands you can use [this bash script](https://raw.githubusercontent.com/Evil-Ant-Colony/Xonotic-EAC-Scripts/master/server/rcon) or [these python scripts](https://github.com/mbasaglia/Unvanquished-misc-scripts/)

[Review on Reviewable](https://reviewable.io/reviews/unvanquished/unvanquished/828)


daemon/src/common/String.h
```diff
@@ -234,6 +234,18 @@ namespace Str {
             return c;
     }
 
+    /*
+     * Converts a hexadecimal character to the value of the digit it represents.
+     * Pre: cisxdigit(ch)
+     */
+    inline CONSTEXPR_FUNCTION int gethex( char ch )
```
Kangz: Probably said earlier but functions that mimc c functions should still get a uppercase first letter (and you can remove the padding in the parenthesis)

daemon/src/engine/server/CryptoChallege.h
```diff
@@ -0,0 +1,633 @@
...
+
+namespace Encryption {
+
+/*
+ * Adds PKCS#7 padding to the data
+ */
+inline void AddPadding( Data& target, std::size_t block_size = 8 )
+{
+    auto pad = block_size - target.size() % block_size;
+    target.resize( target.size() + pad, pad );
```
Kangz: pad as a second argument? I'd rather have something 0 padded

daemon/src/engine/server/CryptoChallege.h
```diff
@@ -0,0 +1,633 @@
...
+    }
+
+    /*
+     * Removes outdated challenges
+     * PRE: The caller has acquired a lock on mutex
+     */
+    void Cleanup()
+    {
+        auto now = Challenge::Clock::now();
+        challenges.erase(
```
Kangz: fails if the deque is empty, that's a strong caveat of the remove-erase_if idiom.

daemon/src/engine/server/CryptoChallege.h
```diff
@@ -0,0 +1,633 @@
+/*
```
Kangz: Imho this is a lot of code when 50 lines would have been sufficent: we are writing a C++ engine not corporate Java:
 - No need for a Library or a Algorithm interface, I'd rather have a direct bool AES256Encrypt(const Data& in, const Data& key, Data* out)
 - Why isn't Data just a typedef to std::vector?
 - Error isn't used heavily enough to warrant it, just returns bool and take the result by pointer.
 - Base64 and Hex can go to String.h
 - Spare us the javadoc on top of every function. Either the comment brings something useful like a caveat or an example either the name of the method is sufficient.
 - Challenge should be a companion struct of the ChallengeManager without these trivial methods.
 - ChallengeManager doesn't need to be mutexed as it is only accessed from the main thread.

In summary, this is way too complex for what you actually need. Keep it simple, and only write the code that you actually need. We don't care about other algorithms, we don't care about methods that replace trivial one liners. However we do care about a small code size, with as little abstractions as possible when we can avoid them.

daemon/src/engine/server/CryptoChallege.h
mbasaglia: This function follows a standard specification. It's the standard way of padding data for encrypted stuff.

daemon/src/engine/server/CryptoChallege.h
mbasaglia: Are you sure about that?
Both std::remove_if and std::deque::erase are supposed to handle empty ranges correctly.

daemon/src/engine/server/CryptoChallege.h
mbasaglia: Library+Algorithm is to select an algorithm by name at runtime.

Once a secure rcon commands arrives, these steps are needed:
* Base64 decoding
* Key generation
* Decryption
* Hex Decoding
I'm not a fan of exceptions myself but I feel these many checks will be quite error-prone and might result in code duplication.

What trivial methods are you referring to? The functions that return constants are just placeholders until we decide where to get those value from (Cvar, constants or whatever).

mbasaglia: I got an idea, how about a struct which holds Data and an error string, which overloads operator >> or something so you can do something like this:
```C++
Encrypt encrypt(input);
if ( encrypt >> Base64Decode >> Decrypt("Aes256", key) >> HexDecode )
    doSomething();
else
    print( encrypt.error );
```

Kangz: Or you can do:

```cpp
using Data = std::vector<uint8_t>;
bool DecryptBase64(const Data& input, const Data& key, Data* output) {
    Data rawInput;
    if (!Base64Decode(input, &rawInput)) {
        return false;
    }
    Data decryptedHex;
    if (!Decrypt(rawInput, input, &decryptedHex)) {
        return false;
    }
    return HexDecode(decryptedHex, output)
}
```
The code will end up being shorter and easier to understand, it's a double win.

Kangz: Also we don't need to choose an algorithm at runtime as we own both the server, the client and the tools. If some day we need to choose an algorithm at runtime, then we'll introduce the required complexity.

The trivial methods I was speaking about are Data's methods.

mbasaglia: With that you lose the ability of reporting what is going wrong.

mbasaglia: Ah and if we use a fixed algorithm which one shall it be? is AES256 OK?

Kangz: Sure, AES256 is a good standard. As for error reporting, if you really want you can stick a Log::Logger's logger.Notice in there (note that it shouldn't be on by default as we should assume stuff to work)

mbasaglia:
The work is mostly done, so it needs review now.

daemon/src/engine/client/cl_main.cpp
```diff
@@ -1429,7 +1428,67 @@ void CL_Connect_f()
...
+static void CL_RconSend(const Rcon::Message &message)
+{
+	std::string invalid_reason;
+	if ( message.valid(&invalid_reason) )
+	{
+		message.send();
+	}
+	else
+	{
+		Log::Notice("Invalid rcon message: %s\n", invalid_reason.c_str());
```
DolceTriade: Do we have a proper client logger now? If so, maybe we can use it here?

daemon/src/engine/client/cl_main.cpp
```diff
@@ -1429,7 +1428,67 @@ void CL_Connect_f()
...
+	}
+}
+
+struct RconChallengeQueue
```
DolceTriade: Use class if you're going to use public/private.

daemon/src/engine/client/cl_main.cpp
```diff
@@ -2099,16 +2150,12 @@ void CL_CheckForResend()
...
 			Info_SetValueForKey( info, "pubkey", key, false );
 
 			Com_sprintf( data, sizeof(data), "connect %s", Cmd_QuoteString( info ) );
 
-			// EVEN BALANCE - T.RAY
-			pktlen = strlen( data );
-			memcpy( pkt, &data[ 0 ], pktlen );
-
-			NET_OutOfBandData( netsrc_t::NS_CLIENT, clc.serverAddress, ( byte * ) pkt, pktlen );
+			Net::OutOfBandData( netsrc_t::NS_CLIENT, clc.serverAddress, (byte*)data, strlen( data ) );
```
DolceTriade: C++ style cast

daemon/src/engine/client/cl_main.cpp
```diff
@@ -2737,7 +2789,7 @@ void CL_ConnectionlessPacket( netadr_t from, msg_t *msg )
 		return;
 	}
 
-	Log::Debug( "Unknown connectionless packet command." );
+	Log::Debug( "Unknown connectionless packet command.\n" );
```
DolceTriade: No newlines on logs.

daemon/src/engine/framework/Network.cpp
```diff
@@ -0,0 +1,118 @@
...
+#       include <netinet/in.h>
+#endif
+
+// Helper functions defined in net_ip.cpp
+void Sys_SockaddrToString( char *dest, int destlen, struct sockaddr *input );
+void NetadrToSockadr( const netadr_t *a, struct sockaddr *s );
+
+namespace Net {
+
+void OutOfBandData( netsrc_t sock, const netadr_t& adr, byte *format, std::size_t len )
```
DolceTriade: format is an incorrect variable name.

daemon/src/engine/framework/Rcon.cpp
```diff
@@ -0,0 +1,252 @@
...
+#include "common/Common.h"
+#include "engine/qcommon/qcommon.h"
+#include "Rcon.h"
+#include "Crypto.h"
+#include "framework/Network.h"
+#include "engine/server/CryptoChallenge.h"
+
+namespace Rcon {
+
+Cvar::Cvar<std::string> cvar_server_password(
```
DolceTriade: Seems like this should be defined in server.

daemon/src/engine/framework/Rcon.cpp
```diff
@@ -0,0 +1,252 @@
...
+        "0: Allow unencrypted rcon, "
+        "1: Require encryption, "
+        "2: Require encryption and challege check",
+    Cvar::NONE,
+    0,
+    0,
+    2
+);
+
+Cvar::Cvar<std::string> cvar_client_destination(
```
DolceTriade: Likewise for client.

daemon/src/engine/framework/Rcon.cpp
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	return true;
+}
+
+Message Message::decode(const netadr_t& remote, const Cmd::Args& args)
+{
+	if ( args.size() < 3 || (args[0] != "rcon" && args[0] != "srcon") )
+	{
+		return Message("Invalid command");
```
DolceTriade: Sounds kinda ugly to return the same class for both error and success. This might be clearer if you use something like Util::Optional.

daemon/src/engine/qcommon/q_shared.cpp
```diff
@@ -1868,6 +1868,61 @@ char     *QDECL PRINTF_LIKE(1) va( const char *format, ... )
 =====================================================================
 */
 
+std::string InfoMapToString( const InfoMap& map )
+{
+	std::string info_string;
+
+	for ( const auto& pair : map )
+	{
```
DolceTriade: Unnecessary. There is code to iterate infostrings already. Info_NextPair. See cg_rocket_datasource.cpp:CG_Rocket_BuildServerInfo() for example usage.

daemon/src/engine/qcommon/q_shared.cpp
```diff
@@ -1868,6 +1868,61 @@ char     *QDECL PRINTF_LIKE(1) va( const char *format, ... )
...
+{
+	if ( string.empty() )
+	{
+		return InfoMap{};
+	}
+
+	InfoMap map;
+	std::istringstream input( string );
+	input.ignore( 1, INFO_SEPARATOR );
+	while ( input.good() )
```
DolceTriade: Can use Info_NextPair to iterate an info string.

daemon/src/engine/server/CryptoChallenge.h
```diff
@@ -0,0 +1,203 @@
...
+    std::string GenerateString();
+
+    TimePoint    created;
+    Duration     timeout;
+    Crypto::Data challenge;
+    netadr_t     source;
+
+};
+
+class ChallengeManager
```
DolceTriade: Benefit of a singleton vs a namespace + methods? Seem like the latter will work well enough.

daemon/src/engine/server/sv_ccmds.cpp
```diff
@@ -226,96 +226,112 @@ static void SV_MapRestart_f()
...
+			"(begin server status)\n"
+			"hostname: %s\n"
+			"version:  %s\n"
+			"protocol: %d\n"
+			"cpu:      %.0f%%\n"
+			"time:     %s\n"
+			"map:      %s\n"
+			"players:  %d / %d\n"
+			"num score connection address                port   name\n"
+			"--- ----- ---------- ---------------------- ------ ----\n",
```
DolceTriade: The last newline will create an extra newline. Is that intentional.

daemon/src/engine/server/sv_ccmds.cpp
```diff
@@ -369,6 +385,48 @@ static void SV_Systeminfo_f()
 	Info_Print( Cvar_InfoString( CVAR_SYSTEMINFO, false ) );
 }
 
+class ListMapsCmd: public Cmd::StaticCmd
```
DolceTriade: iirc, this is implemented in the gamelogic, and this command will overwrite that. It should be available over rcon as well. If you decide that this is better, please delete the gamelogic function from sg_cmds.cpp:Cmd_ListMaps_f()

daemon/src/engine/server/sv_ccmds.cpp
```diff
@@ -369,6 +385,48 @@ static void SV_Systeminfo_f()
...
+		FS::RefreshPaks();
+		std::error_code ignore;
+		for ( const auto& pak : FS::GetAvailablePaks() )
+		{
+			FS::PakPath::LoadPakPrefix(pak, "maps", ignore);
+		}
+
+		for ( const auto& pak : FS::PakPath::ListFiles("maps", ignore) )
+		{
+			if ( Str::IsSuffix(".bsp", pak) )
```
DolceTriade: You also need to verify the bsp filename == pakfile name sans "map-". We will not load mismatching maps so listing them here is pointless.

daemon/src/engine/server/sv_ccmds.cpp
```diff
@@ -369,6 +385,48 @@ static void SV_Systeminfo_f()
...
+		std::vector<std::string> maps;
+
+		FS::RefreshPaks();
+		std::error_code ignore;
+		for ( const auto& pak : FS::GetAvailablePaks() )
+		{
+			FS::PakPath::LoadPakPrefix(pak, "maps", ignore);
+		}
+
+		for ( const auto& pak : FS::PakPath::ListFiles("maps", ignore) )
```
DolceTriade: file might be a more clear variable name?

daemon/src/engine/server/sv_ccmds.cpp
DolceTriade: Make ".bsp" a constant variable.

daemon/src/engine/server/sv_ccmds.cpp
```diff
@@ -369,6 +385,48 @@ static void SV_Systeminfo_f()
...
+		{
+			if ( Str::IsSuffix(".bsp", pak) )
+			{
+				maps.push_back( pak.substr(0, pak.size() - 4) );
```
DolceTriade: Call the ".bsp" variable.size() here instead of a magic number.

daemon/src/engine/server/sv_ccmds.cpp
```diff
@@ -369,6 +385,48 @@ static void SV_Systeminfo_f()
...
+		Log::Notice("Listing %d maps:", maps.size());
+
+		for ( const auto& map: maps )
+		{
+			Log::Notice("%s", map.c_str());
+		}
+	}
+};
+
+#if BUILD_SERVER
```
DolceTriade: Why the ifdef?

daemon/src/engine/framework/Network.cpp
mbasaglia: Can  you elaborate?

daemon/src/engine/framework/Rcon.cpp
mbasaglia: It's used by the client as well (when sending rcon commands).
I'm not sure if it's better to split the cvars into server/client or just keep a single set for simplicity.

daemon/src/engine/framework/Rcon.cpp
mbasaglia: How would you handle error messages in that case? (they'd need to go both on the socket reply and to stdout)

daemon/src/engine/server/sv_ccmds.cpp
mbasaglia: There's an #if that only enables it on a dedicated server.
The other definition is on cgame.
They have  slightly different goals, the cgame one tries to have a pretty format for humans (formatting looks a bit broken on that btw. While this new one is mainly to be machine readable.

daemon/src/engine/server/sv_ccmds.cpp
mbasaglia: I think we need a single place that does something like that (ie: return a list of all map names).
Currently there are similar loops in 3 or 4 places.

Shall I do that as part of this branch?

daemon/src/engine/framework/Network.cpp
DolceTriade: I mean the variable name doesn't match what the variable contains. "data" might be more appropriate.

daemon/src/engine/framework/Rcon.cpp
DolceTriade: Ah, fair point.

daemon/src/engine/server/sv_ccmds.cpp
DolceTriade: Ah, you are correct. Apparently listmaps doesn't exist on the dedi.

daemon/src/engine/server/sv_ccmds.cpp
DolceTriade: Well, given that you added it, might as well.

daemon/src/common/String.h
```diff
@@ -241,6 +242,18 @@ namespace Str {
             return c;
     }
 
+    /*
+     * Converts a hexadecimal character to the value of the digit it represents.
+     * Pre: cisxdigit(ch)
+     */
+    inline CONSTEXPR_FUNCTION int GetHex(char ch)
```
Kangz: What benefit would come from this being constexpr? If there are no benefits, I'd rather have this inlined at LTO.

daemon/src/engine/client/cl_main.cpp
Kangz: Also why do we need locking here? At this point of the read I assume this will always be called from the main thread

daemon/src/engine/framework/Crypto.h
```diff
@@ -0,0 +1,115 @@
...
+#define COMMON_CRYPTO_H_
+
+
+namespace Crypto {
+
+	using Data = std::vector<uint8_t>;
+
+	Data RandomData( std::size_t bytes );
+
+	inline Data String(const std::string& string)
```
Kangz: FromString?

daemon/src/engine/framework/Crypto.h
```diff
@@ -0,0 +1,115 @@
...
+	{
+		return Data(string.begin(), string.end());
+	}
+
+	inline std::string String(const Data string)
```
Kangz: ToString?

daemon/src/engine/framework/Crypto.h
```diff
@@ -0,0 +1,115 @@
...
+		Data Md5( const Data& input );
+	} // namespace Hash
+
+	/*
+	 * Adds PKCS#7 padding to the data
+	 */
+	void AddPadding( Data& target, std::size_t block_size = 8 );
+
+	/*
+	 * Encrypts using the AES256 algorthim
```
Kangz: typo algorithm

daemon/src/engine/framework/Crypto.h
```diff
@@ -0,0 +1,115 @@
...
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+===========================================================================
+*/
+#ifndef COMMON_CRYPTO_H_
+#define COMMON_CRYPTO_H_
+
+
+namespace Crypto {
```
Kangz: No need to add nested namespaces, not sure want need PRE and POST

daemon/src/engine/framework/Crypto.cpp
```diff
@@ -0,0 +1,289 @@
...
+}
+
+namespace Encoding {
+
+/*
+ * Translates binary data into a hexadecimal string
+ */
+Data HexEncode( const Data& input )
+{
+    std::ostringstream stream;
```
Kangz: The C way seems faster here xD

daemon/src/engine/framework/Crypto.cpp
```diff
@@ -0,0 +1,289 @@
...
+ * Encrypts using the AES256 algorthim
+ * PRE: Key is 256-bit (32 octects) long,
+ *      cypher_text.size() is an integer multiple of the AES block size
+ * POST: output contains the decrypted data
+ * Note: If the decoding fails, output is unchanged
+ * Returns true on success
+ */
+bool Aes256Decrypt( Data cypher_text, const Data& key, Data& output )
+{
+    if ( key.size() != AES256_KEY_SIZE || cypher_text.size() % AES_BLOCK_SIZE )
```
Kangz: Assert?

daemon/src/engine/framework/Crypto.h
mbasaglia: Nested namespaces are to group related functions.

I have pre/post comments whenever I cannot guarantee the function does the right thing if the precondition is not met.

daemon/src/engine/framework/Crypto.cpp
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    aes256_ctx ctx;
+    nettle_aes256_set_decrypt_key( &ctx, key.data() );
+    output.resize( cypher_text.size(), 0 );
+    nettle_aes256_decrypt( &ctx, cypher_text.size(),
+        output.data(), cypher_text.data() );
+
+    // Strip the PKCS#7 padding
+    if ( !output.empty() )
```
Kangz: I'm not 100% familiar with PKCS7 but how do we detect what the padding length was?

daemon/src/engine/framework/CvarSystem.cpp
```diff
@@ -479,6 +479,17 @@ namespace Cvar {
         return info;
     }
 
+    void PopulateInfoMap(int flag, InfoMap& map) {
```
Kangz: GetCvqrsWithFlag?

daemon/src/engine/framework/Network.h
```diff
@@ -0,0 +1,66 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+===========================================================================
+*/
+#ifndef COMMON_NETWORK_H
+#define COMMON_NETWORK_H
+
+namespace Net {
+
+inline const std::string& OOBHeader()
```
Kangz: extern cont char* ?

daemon/src/engine/framework/Crypto.cpp
mbasaglia: Like it's done below.

According to PKCS#7 you always add at least one byte, and all of the padding bytes have as value the number of padding bytes.

daemon/src/engine/framework/Network.h
mbasaglia: That doesn't have size()

daemon/src/engine/framework/Network.cpp
```diff
@@ -0,0 +1,118 @@
...
+            if ( NET_IS_IPv6( address.type ) )
+            {
+                result = '[' + result + ']';
+            }
+                result += ':' + std::to_string(ntohs(ExtractPort(address)));
+        }
+        return result;
+    }
+
+    return "";
```
Kangz: ASSERT here?

daemon/src/engine/framework/Network.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+unsigned short ExtractPort(const netadr_t& address)
```
Kangz: GetPort?

daemon/src/engine/framework/Rcon.cpp
Kangz: Imho it should be split.

daemon/src/engine/client/cl_main.cpp
```diff
@@ -1441,46 +1500,30 @@ CL_Rcon_f
...
 	else
 	{
-		if ( !strlen( rconAddress->string ) )
+		if ( Rcon::cvar_client_destination.Get().empty() )
 		{
-			Log::Notice( "%s", "Connect to a server "
-			            "or set the 'rconAddress' cvar "
-			            "to issue rcon commands\n");
+			Log::Notice( "Connect to a server or set the '%s' cvar to issue rcon commands\n",
+				"rcon.client.destination"
```
Kangz: Nice, you could introduce CvarProxy::GetName() and use it here.

daemon/src/engine/framework/Rcon.cpp
```diff
@@ -0,0 +1,252 @@
...
+    Cvar::NONE,
+    ""
+);
+
+Cvar::Range<Cvar::Cvar<int>> cvar_server_secure(
+    "rcon.server.secure",
+    "How secure the Rcon protocol should be: "
+        "0: Allow unencrypted rcon, "
+        "1: Require encryption, "
+        "2: Require encryption and challege check",
```
Kangz: Typo: challenge

Kangz: nit: server.rcon.something instead of rcon.server.something?

daemon/src/engine/framework/Rcon.cpp
```diff
@@ -0,0 +1,252 @@
...
+	  remote_(remote)
+{}
+
+Message::Message( std::string error_message )
+	: error_(std::move(error_message))
+{}
+
+const std::string& Message::command() const
+{
+	return command_;
```
Kangz: In general method names are CamelCase, we use GetCommand and don't postfix private members with _

daemon/src/engine/framework/Rcon.cpp
mbasaglia: You have rcon.server and rcon.client cvars, I don't think they should be namespaced so differently.

daemon/src/engine/framework/Network.cpp
mbasaglia: There are other address types, I'm not sure whether this can end up being called on them or not.

daemon/src/engine/framework/Rcon.h
```diff
@@ -0,0 +1,99 @@
...
+	Unencrypted,        // Allow unencrypted rcon
+	EncryptedPlain,     // Require encryption
+	EncryptedChallenge, // Require encryption and challege check
+	Invalid             // Invalid protocol
+};
+
+/*
+ * RCon message sent from a client to a server
+ */
+class Message
```
Kangz: I'm not a fan of your approach here: it seems to me that Message does to much and there is no clear separation between what is received, errors, etc. From what I understand you did this to have all the Rcon code at the same place, right? Message should be a method-less structure and the various methods should be changed to functions operating on that structure.

For example Acceptable would become bool Rcon::IsAcceptable(const Message& message, std::string* error);
The code using that function might have to do a bit more work to create the answer from the error string, but it would still be cleaner.

daemon/src/engine/qcommon/q_shared.cpp
```diff
@@ -1868,6 +1868,61 @@ char     *QDECL PRINTF_LIKE(1) va( const char *format, ... )
...
+std::string InfoMapToString( const InfoMap& map )
+{
+	std::string info_string;
+
+	for ( const auto& pair : map )
+	{
+		if ( pair.first.find( INFO_SEPARATOR ) != std::string::npos ||
+			pair.second.find( INFO_SEPARATOR ) != std::string::npos )
+		{
+			Log::Notice( "Can't use keys or values with a \\\n" );
```
Kangz: Log::Warn and a more explicit message? I think the Info_ code has stuff to add escaping so you might wnat to do the same.

daemon/src/engine/server/CryptoChallenge.h
```diff
@@ -0,0 +1,203 @@
...
+        return std::unique_lock<std::mutex>{mutex};
+    }
+
+    /*
+     * Removes outdated challenges
+     * PRE: The caller has acquired a lock on mutex
+     */
+    void Cleanup();
+
+    std::mutex mutex;
```
Kangz: Likewise, this is going to be used from the main thread only so no need for locking.

daemon/src/engine/server/CryptoChallenge.cpp
```diff
@@ -0,0 +1,180 @@
...
+	Sys::GenRandomBytes(data.data(), data.size());
+	std::ostringstream stream;
+	stream.setf(std::ios::hex, std::ios::basefield);
+	stream.fill('0');
+	for ( auto ch : data )
+	{
+		stream.width(2);
+		stream << int( ch );
+	}
+	return stream.str();
```
Kangz: This could reuse code to get a hex string.

daemon/src/engine/framework/Rcon.h
mbasaglia: What's the advantage of having a struct and functions that act on it rather than a proper class?
None of the attributes need to be changed once constructed. This cannot be ensured with a struct.

daemon/src/engine/server/server.h
```diff
@@ -486,4 +460,27 @@ void     SV_Netchan_FreeQueue( client_t *client );
...
+	NoAdvertise, // Do not advertise but reply to all out of band messages
+	NoStatus,    // Do not advertise nor reply to status out of band messages but allow all connections
+	LanOnly,     // Block everything except for LAN connections
+};
+
+/*
+ * Returns whether the server has a private level equal to or greater than
+ * the one provided
+ */
+bool SV_Private(ServerPrivate level);
```
Kangz: nit: better name?

daemon/src/engine/server/sv_ccmds.cpp
```diff
@@ -226,96 +226,112 @@ static void SV_MapRestart_f()
...
+		auto seconds = svs.time / 1000 % 60;
+		auto minutes = svs.time / 1000 / 60 % 60;
+		if ( auto hours = svs.time / 1000 / 60 / 60 / 60 )
 		{
-			Log::Notice( " " );
+			time_string = Str::Format("%02d:", hours);
+		}
+		time_string += Str::Format("%02d:%02d", minutes, seconds);
+
+		Log::Notice(
```
Kangz: In commands you want to use Print so that it gets into the Rcon print handler.

daemon/src/engine/server/sv_client.cpp
```diff
@@ -35,73 +35,20 @@ Maryland 20850 USA.
...
+		return;
 	}
 
-	challenge->pingTime = svs.time;
-
-	NET_OutOfBandPrint( netsrc_t::NS_SERVER, from, "challengeResponse %i", challenge->challenge );
-
-	return;
+	auto challenge = ChallengeManager::Get().GenerateChallenge( from );
+	Net::OutOfBandPrint( netsrc_t::NS_SERVER, from, "challengeResponse %s\n", challenge );
```
Kangz: Is the \n necessary here?

daemon/src/common/String.h
mbasaglia: I made it into a constexpr because it used to be a macro in qcommon.
I guess it can become just a normal function, it's only being used in a couple places anyway.

daemon/src/engine/framework/Crypto.cpp
mbasaglia: Maybe streams are a bit of an overkill.
It can still be done in a C++ way though.

daemon/src/engine/framework/Crypto.h
Kangz: Namespace: grouping in the header file should be enough, you only have a dozen symbols in there anyway.
PRE / POST: This isn't in the style of the "new" code, we usually have verbose comments instead.

daemon/src/engine/framework/Crypto.cpp
Kangz: I see, the code below makes sense now

daemon/src/engine/framework/Rcon.h
Kangz: Here message is used both as a structure that contains rcon related data, and as an error message. This makes code using it very unclear. You can pass the structure around with const. Imho classes + methods should be used for dynamic dispatch and when you need to mutate internal state in a controlled way. This is not the case here.

daemon/src/engine/client/cl_main.cpp
mbasaglia: Looks like all calls will stem from Com_EventLoop. If a client is running a local game, will it run from a single thread?

daemon/src/engine/qcommon/q_shared.cpp
mbasaglia: I'm trying to deprecate those old Info functions, they are a bit ugly and potentially unsafe.
Info_NextPair for instance has no bound checking and makes a few assumptions on the pointers it receives without documenting them.

daemon/src/engine/qcommon/q_shared.cpp
mbasaglia: That message was taken straight from Info_SetValueForKey, which fails for \ (and a few other characters, for some mysterious reason)

mbasaglia:
*[daemon/src/engine/client/cl_main.cpp, line 1440 \[r4\]](https://reviewable.io:443/reviews/unvanquished/unvanquished/828#-KBdgvZ5LT-f4j95HnLC-r4-1440:-KDEqCoi9iumD7q8kfrl:-1569062231) ([raw file](https://github.com/unvanquished/unvanquished/blob/82648841c48f19eefd3faf24a604be3f3b7322bd/daemon/src/engine/client/cl_main.cpp#L1440)):*
Do you know which logger should be used?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/unvanquished/unvanquished/828)*

illwieckz:
For information, the _“[Reviewable](https://reviewable.io/reviews/unvanquished/unvanquished/828)”_ service used in that PR added some `refs/reviewable/pr828/r[1-8]` git references in origin repository [I have pruned](https://github.com/illwieckz/split_unvanquished/commit/b86ebbed437e65ab55386be9632547b696667f72?diff=unified) in the [Daemon](https://github.com/DaemonDevelopers/Daemon) repository before pushing it. These references are still hosted there in [Unvanquished](https://github.com/Unvanquished/Unvanquished) repository. They look to be kind of duplicates of `refs/pull/828/[head|merge]` (GitHub's _“Pull Request”_ stuff), but the _Reviewable_ way.

PR #829, opened 2015-11-07
Run CBSE toolchain from CMake

DolceTriade:


daemon/cmake/DaemonCBSE.cmake
```diff
@@ -0,0 +1,78 @@
...
+            find_package(Git REQUIRED)
+            if (GIT_FOUND)
+                execute_process(
+                    COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
+                    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
+            endif()
+        endif()
+    endif()
+    # Check if python has all the dependencies
+    # TODO: Execute pip directly here and install them
```
Kangz: I'm not sure this is a good idea, I wouldn't want a build script to automatically install global programs. (I'm not 100% sure how pip works so I might be wrong)

Kangz:
LGTM! Thanks a lot for tackling this and actually making it work on AppVeyor

daemon/cmake/DaemonCBSE.cmake
DolceTriade: Yeah, I believe pip installs things to /usr/local by default. pip is not installed by default so, maybe i need to change this method to specify which deps are required.

DolceTriade:
It doesn't actually work on AppVeyor. For some reason, AppVeyor stopped testing this branch :(

daemon/cmake/DaemonCBSE.cmake
mbasaglia: It could use virtualenv to install the required python packages without polluting the host system.

DolceTriade: Perhaps. I think I should leave installing the deps up to the user. We don't want to add virtualenv as another dep.

Viech: I would also just add the relevant python libraries to the build dependencies. I wouldn't want to add more downloads to cmake even though we have the NaCl one and even if we can make pip download into the Unvanquished source tree. Unlike the NaCl dependency, python libraries are available on all platforms.

DolceTriade: I'll just remove the TODO.

DolceTriade:
AppVeyor is broken on master. Merging anyways.

Viech:
The new dependencies need to be added to the `Buildtime` section of [README.md](https://github.com/Unvanquished/Unvanquished/blob/master/README.md).

I identified the following:
- python ≥ 2
- python-yaml
- python-jinja

Is that correct? Personally I didn't test if compilation works with python2. It does with python3 and the latest libraries.

Viech:
Also, there is a section named `CBSE Toolchain` in the readme that may not to get updated.

DolceTriade:
I tested with python2. It works fine.

PR #831, opened 2015-11-09
Binary messages

Kangz:
I'd like to merge the binary message work little by little to avoid putting up a HUGE PR.

In here you'll find:
 - Some cleanup
 - The addition of the number of entities in the snapshot so that we can preallocate the entity vector
 - A rewrite (and potential bugfix) of the client side entity packet parsing in order to put entities in a vector owned by the snapshot (will make it easier to move that part of the netcode to the gamelogic)

### DIFF
--- a/daemon/src/engine/client/cl_cgame.cpp
+++ b/daemon/src/engine/client/cl_cgame.cpp
@@ -295,24 +295,13 @@ bool CL_GetSnapshot( int snapshotNumber, snapshot_t *snapshot )
 		return false;
 	}
 
-	// if the entities in the frame have fallen out of their
-	// circular buffer, we can't return it
-	if ( cl.parseEntitiesNum - clSnap->parseEntitiesNum >= MAX_PARSE_ENTITIES )
-	{
-		return false;
-	}
-
 	// write the snapshot
 	snapshot->snapFlags = clSnap->snapFlags;
 	snapshot->ping = clSnap->ping;
 	snapshot->serverTime = clSnap->serverTime;
 	memcpy( snapshot->areamask, clSnap->areamask, sizeof( snapshot->areamask ) );
 	snapshot->ps = clSnap->ps;
-
-	snapshot->entities.reserve(clSnap->numEntities);
-	for (unsigned i = 0; i < clSnap->numEntities; i++) {
-		snapshot->entities.push_back(cl.parseEntities[( clSnap->parseEntitiesNum + i ) & ( MAX_PARSE_ENTITIES - 1 ) ]);
-	}
+	snapshot->entities = clSnap->entities;
 
 	CL_FillServerCommands(snapshot->serverCommands, clc.lastExecutedServerCommand + 1, clSnap->serverCommandNum);
 	clc.lastExecutedServerCommand = clSnap->serverCommandNum;
--- a/daemon/src/engine/client/cl_cgame.cpp
+++ b/daemon/src/engine/client/cl_cgame.cpp
@@ -1239,8 +1239,8 @@ void CGameVM::QVMSyscall(int index, Util::Reader& reader, IPC::Channel& channel)
 
 		case CG_GETCURRENTSNAPSHOTNUMBER:
 			IPC::HandleMsg<GetCurrentSnapshotNumberMsg>(channel, std::move(reader), [this] (int& number, int& serverTime) {
-	            number = cl.snap.messageNum;
-	            serverTime = cl.snap.serverTime;
+				number = cl.snap.messageNum;
+				serverTime = cl.snap.serverTime;
 			});
 			break;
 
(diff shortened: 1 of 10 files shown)


daemon/src/engine/client/cl_parse.cpp
```diff
@@ -68,38 +66,16 @@ MESSAGE PARSING
...
-
-	if ( unchanged )
-	{
-		*state = *old;
-	}
-	else
-	{
-		MSG_ReadDeltaEntity( msg, old, state, newnum );
-	}
+    entityState_t entity;
```
DolceTriade: Indentation seems fishy here.

Kangz: Yeah it is a mix of old and new (ew), fixed by reindenting all of it.

daemon/src/engine/client/cl_cgame.cpp
```diff
@@ -1327,7 +1264,9 @@ void CGameVM::QVMSyscall(int index, Util::Reader& reader, IPC::Channel& channel)
 
 		case CG_SETUSERCMDVALUE:
 			IPC::HandleMsg<SetUserCmdValueMsg>(channel, std::move(reader), [this] (int stateValue, int flags, float scale) {
-				CL_SetUserCmdValue(stateValue, flags, scale);
+                cl.cgameUserCmdValue = stateValue;
```
DolceTriade: Indentation here is wrong as well.

daemon/src/engine/client/client.h
```diff
@@ -74,11 +74,10 @@ typedef struct
 	int           cmdNum; // the next cmdNum the server is expecting
 	playerState_t ps; // complete information about the current player at this time
 
-	unsigned      numEntities; // all of the entities that need to be presented
-	int           parseEntitiesNum; // at the time of this snapshot
-
 	int           serverCommandNum; // execute all commands up to this before
 	// making the snapshot current
+
+    std::vector<entityState_t> entities;
```
DolceTriade: Inconsistent indentation.

DolceTriade:
I am style police now.

daemon/src/engine/client/cl_parse.cpp
```diff
@@ -108,148 +84,83 @@ CL_ParsePacketEntities
...
+        oldEntityNum = oldSnapshot->entities[0].number;
+    }
+
+    // Likewise when we don't have an old snapshot, oldEntities just has to be an empty
+    // vector so that we skip step (4)
+    std::vector<entityState_t> dummyEntities;
+    auto& oldEntities = oldSnapshot? oldSnapshot->entities : dummyEntities;
+    auto& newEntities = newSnapshot->entities;
+
+    unsigned int numEntities = MSG_ReadShort(msg);
```
mbasaglia: MSG_ReadShort returns int, is the cast safe? can it ever return negative values?

daemon/src/engine/client/cl_parse.cpp
```diff
@@ -108,148 +84,83 @@ CL_ParsePacketEntities
...
-		{
-			// delta from baseline
-			if ( cl_shownet->integer == 3 )
-			{
-				Com_Printf( "%3i:  baseline: %i\n", msg->readcount, newnum );
-			}
+        // (1) all entities that weren't specified between the previous newEntityNum and
+        // the current one are unchanged and just copied over.
+        while (oldEntityNum < newEntityNum) {
+            newEntities.push_back(oldEntities[oldIndex]);
```
mbasaglia: Can oldEntities ever be empty? You always access oldEntities[0] with no checks.

daemon/src/engine/client/cl_parse.cpp
Kangz: Good point, it returns an unsigned integer, unless we read past the end of the message in which case it returns -1. Will fix.

daemon/src/engine/client/cl_parse.cpp
Kangz: It is very unlikely to have a completely empty snapshot (given that the client itself is an entity) but it is better to not assume that. Will fix.

daemon/src/engine/client/client.h
Kangz: :( does it mean that inside a single code block indentation has to be consistent?

daemon/src/engine/client/cl_parse.cpp
Kangz: Actually, this is correct since we later check that we didn't go past the end of the message.

daemon/src/engine/client/cl_parse.cpp
Kangz: Likewise, this is correct because oldEntityNum is set to MAX_GENTITIES earlier in the function if the old snapshot doesn't have entities.

daemon/src/engine/client/cl_parse.cpp
DolceTriade: Should add comment to note this.

Kangz: I moved the code to just after it, do you still want the comment?

Kangz:
Still waiting for a formal LGTM

daemon/src/engine/client/cl_parse.cpp
DolceTriade: Nah.

DolceTriade:
LGTM

Kangz:
Thank you for the reviews, merging.

Issue #832, opened 2015-11-09
Incorrect usages of std::move


mbasaglia:
I've came across some incorrect usages of std::move, for now I'm fixing CvarSystem, but it might be the case that it's used incorrectly throughout the codebase.

Some instances are probably innocuous but useless, others can have a slight negative impact on performance (Forcing a premature conversion from an prvalue to a lvalue).
But the biggest issue is that some places read from a moved-from variable, which results at best in an unspecified value, at worst undefined behaviour.

necessarily-equal:
I noted some suspect ones in Dæmon, I haven't had a look at Unvanquished as of now.

* daemon/src/shared/CommonProxies.cpp line 88 and line 112
* daemon/src/engine/server/sv_sgame.cpp, function QVMSyscall
* daemon/src/engine/audio/Sample.cpp?
* daemon/src/common/FileSystem.cpp, function HandleFileSystemSyscall
* daemon/src/engine/client/cl_cgame.cpp, function QVMSyscall, HandleCommandBufferSyscall
* daemon/src/engine/framework/CommandBufferHost.cpp, CommandBufferHost::Syscall
* daemon/src/engine/framework/CommonVMServices.cpp, HandleMiscSyscall

PR #833, opened 2015-11-10
Fix some issues with cvars

mbasaglia:
This branch ensures the cvar description gets properly registered and fixes some of the issues related to #832.

### DIFF
--- a/daemon/src/common/Command.h
+++ b/daemon/src/common/Command.h
@@ -206,7 +206,7 @@ namespace Cmd {
 
     // Engine calls available everywhere
 
-    void AddCommand(std::string name, const CmdBase& cmd, std::string description);
+    void AddCommand(const std::string& name, const CmdBase& cmd, std::string description);
     void RemoveCommand(const std::string& name);
     Environment* GetEnv();
 
--- a/daemon/src/common/Cvar.h
+++ b/daemon/src/common/Cvar.h
@@ -232,7 +232,7 @@ namespace Cvar {
 
     bool Register(CvarProxy* proxy, const std::string& name, std::string description, int flags, const std::string& defaultValue);
     std::string GetValue(const std::string& cvarName);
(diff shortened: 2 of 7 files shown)


Kangz:
LGTM, though we should strive to use Str;;StringRef instead of const std::string& (pinging @Amanieu about this, to be sure)

mbasaglia:
Those strings end up being assigned to std::string objects

Kangz:
Yes but Str::StringRef is made to avoid the allocation that happens when passing a const char* to std::string&

mbasaglia:
But it would fail horribly if passed a std::string prvalue

Kangz:
Not sure, @Amanieu?

mbasaglia:
Amanieu said it was OK to merge.

if more refactoring is needed it should be done some other time, this was just a small bug fix

Exaeta:
Um, this is a misunderstanding of how std::move works.

mbasaglia:
No, variables were accessed after being moved from, which leads to undefined behaviour

mbasaglia:
And all the moves resulted in multiple copies anyway

Exaeta:
@mbasaglia, moves do not cause copies... That's the entire point of a move.

mbasaglia:
if the value is passed by value multiple times without moving it before it reaches its destination, there is no point in moving for the last assignment

Exaeta:
@mbasaglia Actually that is wrong. Pass by value can use move semantics.

mbasaglia:
Not if you are assigning from an lvalue.

Exaeta:
@mbasaglia I think you misunderstand how move works?
I gave you an example of a copy-free situation. IF you are assigning from an lvalue you need a copy will be made anyway, moves don't significantly slow it down.

mbasaglia:
Except our code doesn't look like that

PR #838, opened 2015-11-15
[RFC] Assert gamelogic will work with a given engine version

DolceTriade:
Note that I should probably make the AssertEngineVersion in some common location. Will do that next, but I wanted feedback on the essence of the idea.

### DIFF
--- a/daemon/src/common/String.cpp
+++ b/daemon/src/common/String.cpp
@@ -127,6 +127,16 @@ namespace Str {
         return LongestPrefixSize(ToLower(text1), ToLower(text2));
     }
 
+    std::vector< std::string > Expand( Str::StringRef string, char delimiter ) {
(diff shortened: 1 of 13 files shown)


TimePath:
What do you do about new engine w/o some deprecated feature + old gamelogic?

daemon/src/common/String.cpp
```diff
@@ -127,6 +127,16 @@ namespace Str {
         return LongestPrefixSize(ToLower(text1), ToLower(text2));
     }
 
+    std::vector< std::string > Expand( Str::StringRef string, char delimiter ) {
```
TimePath: Name suggestion: `Str::Split`, as in other languages

DolceTriade:
functionality exists. gg.

PR #846, opened 2015-11-22
Refactor medkit healing

caldazar-at:
Activating a medkit and subsequent healing is now handled in a single function.
This function is called from ClientThink_real now, instead of ClientTimerActions.
Due to a different timing mechanism, medkits now heal slightly faster (~200 ms).

### DIFF
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -723,7 +723,7 @@ bool ClientInactivityTimer( gentity_t *ent, bool active )
 	return true;
 }
 
-// TODO: Move to MedikitComponent.
+// TODO: Move to MedkitComponent.
 static void G_ReplenishHumanHealth( gentity_t *self )
 {
 	gclient_t *client;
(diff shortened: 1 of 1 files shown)


Viech:
I'm going to test this once I fix my development setup, currently I'm unable to launch Unvanquished at all due to some audio backend issue. So far there's two things I'd point out:

First, if you start on refactoring things in the server side gamelogic, you would ideally try to move them to the CBSE subsystem. The problem about the `MedkitComponent` would be that there isn't any other component for human equipment yet and we'd need to design a wrapper for the equipment handling code (or even refactor the latter) first. I didn't work on CBSE stuff for a while and can't quite tell how to do this best right now but I'll look into it next. (Right now I'd imagine messages for adding and removing of eqipment that the `MedkitComponent` would react on. However, I faintly remember conversations about actually adding and removing whole components, as opposed to just enabling/disabling them.)

Secondly, the reason the old medkit code looks like it is, in particular the `MEDKIT_STARTUP_TIME` bit, is that previously we could only do integer damage and healing. A nicer approach would be to make use of floating point health and just heal by some amount every couple of server frames. The initial amount would then be lower and smoothly go up over time. To the user (and the network) this wouldn't look much different but the code would be easier to read and more importantly it would be framerate-independent and we could change the calculation interval to whatever we want without altering the healing logic (given that you take the actual time since the last execution into account, see the `ThinkingComponent` and how it's used by other components).

Kangz:
Yeah, I can work on something to add support for dynamically added components, it will have a runtime cost though (only for entities that use runtime togglable components).

Viech:
This PR is now contained in #973 and will be automatically closed when the latter gets merged.

PR #847, opened 2015-11-27
WIP: Breakpad integration / crash dump generation

slipher:
This code is for generating minidump files when the program crashes, using the Breakpad library, which can be analyzed to get stack traces. Not intended to be merged immediately, but I'd the work so far to be reviewed before doing the final steps. So far it generates dumps for Windows and Linux engine (also works for gamelogic DLLs), and cgame NaCl. A few specific points I'd like feedback on:

- How should the code be organized better?
- Should I rate limit somehow crash dumps from NaCl? This could be a problem e.g. if the main menu gets stuck in a loop reloading cgame, or a VM sent bogus requests to write a dump.
- Does anyone care if it is implemented for native-exe VMs?

Still TODO:
- Dumps from NaCl sgame (a matter of copy/pasting cgame code)
- Dumps from OS X engine
- (Linux) Detect if the crash server fails to start, since if it doesn't, the program will hang after crashing. (maybe use smth else instead of fork/exec)
- Change the action of the USE_BREAKPAD flag (?) For users compiling the engine themselves, an engine crash dump is probably not very useful, so it makes sense for it to default to off. But NaCl crash dumps (which don't actually require Breakpad) would still be useful in this case.
- Figure out how to get symbol info when compiling in Release mode. This is needed for the release builds of the engine.
- Remove some files from the breakpad codebase that we don't use.
- Name NaCl dump files something other than a truncated time stamp.


daemon/CMakeLists.txt
```diff
@@ -413,7 +414,8 @@ if (WIN32)
     set(LIBS_BASE ${LIBS_BASE} winmm ws2_32)
 elseif (NACL)
     find_library(NACL_EXCEPTION nacl_exception)
-    set(LIBS_BASE ${LIBS_BASE} ${NACL_EXCEPTION})
+    find_library(NACL_MINIDUMP minidump_generator)
```
mbasaglia: Should this be guarded by USE_BREAKPAD?

Kangz:
Looks promising so far but reviewing has been made difficult by Github failing to show such a big diff.

Could you make breakpad a submodule (it is unlikely to change often) as well as lss? Since we are already using submodules I think this is ok. Once you have done that, you can squash most of the commits (you can keep linux separate if you want) so that we don't introduce the breakpad source code in our repo then remove it.

I'll do a more detailed review when this is done, (or when I get to a command line git client)

slipher:
@Kangz Ishq and I are thinking not to do a submodule, since the Breakpad repo is huge and contains numerous programs that we don't need for the game.

Viech:
I assume this is now #865. (For cross-reference.)

slipher:
@Viech correct.

Kangz:
How about making a repo of our own with source drops of the useful parts of breakpad?

DolceTriade:
If we want...I don't think the breakpad source is ever really going to change.

PR #853, opened 2015-12-08
Console command spam

Kangz:
Please take a look. I couldn't store information in cgs because it gets erased and recreated in CG_Init


DolceTriade:
LGTM

PR #856, opened 2015-12-12
fix wrong command

Keziolio:


Kangz:
LGTM, thanks for the fix!

PR #857, opened 2015-12-14
Create SpectatorEntity

DolceTriade:


src.cmake
```diff
@@ -179,6 +179,7 @@ set(SGAMELIST
     ${GAMELOGIC_DIR}/sgame/components/ThinkingComponent.cpp
     ${GAMELOGIC_DIR}/sgame/components/TrapperComponent.cpp
     ${GAMELOGIC_DIR}/sgame/components/TurretComponent.cpp
+    ${GAMELOGIC_DIR}/sgame/components/SpectatorClassComponent.cpp
```
Viech: `SpectatorComponent` for consistency please.

src/sgame/components/SpectatorClassComponent.cpp
```diff
@@ -0,0 +1,35 @@
+#include "SpectatorClassComponent.h"
+
+SpectatorClassComponent::SpectatorClassComponent(Entity& entity, ClientComponent& r_ClientComponent)
+	: SpectatorClassComponentBase(entity, r_ClientComponent)
+{}
+
+void SpectatorClassComponent::HandlePrepareNetCode() {
+	gclient_t* cl = entity.oldEnt->client;
+
+	if (cl && cl->sess.spectatorState == SPECTATOR_FOLLOW) {
```
Viech: Instead of nesting the three `if` statements, maybe just return early if the respective condition is not met?

src/sgame/components/SpectatorClassComponent.cpp
```diff
@@ -0,0 +1,35 @@
...
+			{
+				// Save
+				int score = cl->ps.persistant[ PERS_SCORE ];
+				int ping = cl->ps.ping;
+
+				// Copy
+				cl->ps = other->ps;
+
+				// Restore
+				cl->ps.persistant[ PERS_SCORE ] = score;
```
Viech: Can you explain why these two lines are needed, in particular the score one? Why is it the only field from `ps.persistant` that is also persistant accross different spectator states?

src/sgame/sg_main.cpp
```diff
@@ -2991,9 +2991,20 @@ void G_RunFrame( int levelTime )
 void G_PrepareEntityNetCode() {
```
Viech: I don't like special casing the PrepareNetCode message for spectators *too much*. Can we not live with having spectators lag up to a single server frame behind? A lot of the server side gamelogic (in particular thinkers) assumes that such a lag doesn't matters.

src/sgame/sg_main.cpp
```diff
@@ -2991,9 +2991,20 @@ void G_RunFrame( int levelTime )
...
+			if (oldEnt->entity->Get<SpectatorClassComponent>()) {
+				specEntities.push_back(oldEnt->entity);
+				continue;
+			}
 			oldEnt->entity->PrepareNetCode();
 		}
 	}
+
+	// Prepare netcode for specs
+	for (Entity* e : specEntities) {
```
Viech: Just use the `ForEntities` templated iterator here. The reason it's not done above is, as the TODO implies, you cannot currently iterate over *all* entities with it, just over a subset of all entities that has a given nonempty subset of components.

The syntax is like this:

    ForEntities<SpectatorComponent>([&](Entity& entitiy, SpectatorComponent& spectatorComponent){
        entity.PrepareNetcode();
    });

src/sgame/sg_main.cpp
Viech: Given that spectating related code is hard to test and your approach is at least as correct as my proposal, I won't insist on putting too much effort into this. Consider adding a comment like this, though: `TODO: Find out whether grouping netcode preparation by spectator state is necessary.`.

src/sgame/components/SpectatorClassComponent.cpp
Kangz: I think that otherwise that the spectator would show with the spectatee's score in the scoreboard.

DolceTriade: Correct. If you do /serverstatus or click on "view server info", it would show the player being spectated's score.

src/sgame/sg_main.cpp
DolceTriade: It's not so much that the information is out of date, but rather that the information can be completely wrong. HandlePrepareNetcode can do anything. Currently, it mainly just clamps the float health to an int, but that may not be what happens in the future. That's why we must call call PrepareNetcode after all the state has been written.

Viech: Note that `HandlePrepareNetcode` will never be allowed to do anything but exporting the component's state to the entity's network record. No member of the component may be changed. We just don't mark it `const` because it has an autogenerated declaration, as far as I know.

That being said, I think the only type of "wrong" the information can be is "outdated by one frame". And for clients, we're already talking about client frames (with asynchronous rate) that the server mimics, as far as I understand it. I do see that we may not want additional lag for some reason. I'm just hesitant to allow special cases in CBSE related code by default, because reducing the number of special cases to an elegant minimum is the point of CBSE. :stuck_out_tongue:

DolceTriade: My point is that the exported state pre-HandlePrepareNetcode can be completely different for whatever reason.

src/sgame/components/SpectatorClassComponent.cpp
DolceTriade: Done.

src.cmake
DolceTriade: SpectatorClassComponent was being consistent with AlienClass/HumanClass. Do you think spectatorcomponent is better?

Viech: Yeah, `AlienClass` is merely to distinguish from `AlienBuildable`.

DolceTriade: Done.

Viech:
LGTM

PR #860, opened 2015-12-20
add a floating point fudge factor to the floating complain to silence…

### LINKED ISSUE #850: Log spam
Our server has a 110MB daemon.log of this error:
`^3Warn: A resource storage stores an invalid fraction of all build points: 1.0`

It seems to be caused when one or both teams has no drill or leech.

DolceTriade:
… log spam

Fixes #850

### DIFF
--- a/daemon/src/engine/qcommon/q_math.cpp
+++ b/daemon/src/engine/qcommon/q_math.cpp
@@ -524,7 +524,7 @@ Traces a ray from inside a rectangle and returns the point of
 intersection with the rectangle
 ==================
 */
-float ProjectPointOntoRectangleOutwards( vec2_t out, const vec2_t point, const vec2_t dir, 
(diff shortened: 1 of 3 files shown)


DolceTriade:
Commit message is wrong. I'll merge in the fixed commit message once Viech says this looks good.

DolceTriade:
For some reason, the FLT_EPSILON fudge factor doesn't seem to help.I picked that value because its defined as LINE_EPSILON in q_math.cpp (though not defined publicly).

DolceTriade:
This is an optimization error, since it doesn't happen in Debug mode. 1e-5 is the right fudge factor. Should I declare it as a variable or is it fine as it is?

Kangz:
We might want to keep this in mind when doing floating point asserts too. A constant would be best.

Viech:
Either way, LGTM.

DolceTriade:
Added constant. Will merge now.

PR #861, opened 2015-12-22
Call fsync when writing to stdout

DolceTriade:
This is required on OSX; otherwise, you will be spammed with
"Resource temporarily unavailable" and will eventually segfault.


DolceTriade:
@Amanieu Please take a look. I wonder why this is required.

Kangz:
Could you summarize the IRC discussion here? I seem to remember Amanieu had a better idea on how to fix it.

DolceTriade:
I won't merge this yet. I haven't been able to reproduce this again (so this fix might have not fixed it at all). Next time I can reproduce this, I'll experiment further.

mbasaglia:
std::cout?

illwieckz:
@DolceTriade here is a way to reproduce this bug on GNU/Linux: #893

illwieckz:
I just did that:

```diff
diff --git a/daemon/src/engine/sys/con_tty.cpp b/daemon/src/engine/sys/con_tty.cpp
index 6466042..212da1c 100644
--- a/daemon/src/engine/sys/con_tty.cpp
+++ b/daemon/src/engine/sys/con_tty.cpp
@@ -67,7 +67,7 @@ static Console::Field TTY_field(INT_MAX);
 
 void WriteToStdout(const char* text) {
        if (write( STDOUT_FILENO, text, strlen(text)) < 0) {
-        Log::Warn("Error writing to the terminal: %s", strerror(errno));
+        // Log::Warn("Error writing to the terminal: %s", strerror(errno));
     }
 }
```

And it fixed my problem, like pertubed` said on IRC (which is obvious):

> the fix to this is known, remove the warning message, when there's an error writing to the terminal, don't report it by writing to the terminal

Amanieu:
A better solution: if `write` returns `EAGAIN`, use `select` to wait for stdout to be ready.
```c
fd_set fds;
FD_ZERO(&fds);
FD_SET(&fds, STDOUT_FILENO);
select(2, NULL, &fds, NULL, NULL);
```
and then try writing one more time (give up if it still doesn't work).

Issue #314, opened 2014-06-11
Running a /devmap disables cheat protection for the next game you join


Veyrdite:
EDIT: Severely edited since first report

The game keeps track of whether or not you have permission to cheat and edit cheating variables.  This is determined by whether or not you start a /map or a /devmap
## Current behaviour

If the first match you play is non-cheating, then all successive matches (even with /devmap) will not allow you to properly cheat.
If the first match you play is cheating, then all successive matches (even online) can be cheated in.

This bug only seems to apply to cheat-protected variables (like r_showskeleton) and not commands (noclip, god)
## Reproduction

Forcing cheat perms on:
- start the game
- /devmap parpax
- /map parpax
- note that you can change cheat-protected variables
- join an online match
- note that you can change cheat-protected variables

Forcing cheat perms off:
- start the game
- /map parpax
- /devmap parpax
- note that you can NOT change cheat-protected vars
- repeat, etc

![utcs](https://cloud.githubusercontent.com/assets/6168874/3240034/81a6ae8c-f11b-11e3-9622-7ba257d37649.jpeg)

Viech:
@Kangz `sv_pure` defaults to `0` and as far as I know the pure check has been removed in the new filesystem. So if clients are loading the wrong pk3s that should (technically) be a different problem. Are you still on this by the way?

Amanieu:
I cannot reproduce the pure issue, closing.

Viech:
@caldazar-at wrote on IRC:
> I noticed that sometimes debug cvars (cg_thirdpersion, cg_drawbbox) are still active after joining a server. and then you can't reset them because of cheat protection. is that a known bug?
I think this is basically the same issue: Cvars that can only be enabled in devmap mode stay set on the next non-devmap match and cannot even be disabled.

caldazar-at:
I found a way to reproduce that bug:
1) start unvanquished (duh!)
2) /devmap plat23
3) /cg_thirdperson 1
4) /cg_drawbbox 1
5) /quit (/disconnect is not enough)
6) start unvanquished againdaa6cd8
7) /map plat23 (or connect to a server): now you can see bboxes and use 3rd persion without cheatmode
8) /map plat23 this solves the problem.

All that happened on current master branch (commit daa6cd8) with pak from alpha 37.
![Uploading debug_cvars.jpg . . .]()

Calinou:
I was able to join a server in third person mode. Using Alpha 38, on 64-bit Debian testing.

[...]

PR #867, opened 2015-12-24
Always reset Cvar::CHEAT cvars to default value.

DolceTriade:


DolceTriade:
This prevents cases where you set a cheat cvar (say r_showTris 1), quit, and then start the game, r_showTris will still be one. If you join a server, it will continue to be 1. Maybe we should have a better method of reseting cvar cheat values.

Kangz:
Good catch, looks good to me, cheat cvars were disabled in in Cvar::SetCheatsEnabled but I forgot to take into account user-defined cvars.

Viech:
Maybe it would be cleaner to add it to `inline bool IsArchived() const`, such that `CVAR_CHEAT` implies `CVAR_TEMPORARY`?

Also, #314 may want to be closed.

PR #872, opened 2015-12-27
new default bindings proposal

Keziolio:


main/default.cfg
```diff
@@ -7,45 +7,33 @@ unbindall
...
 //
+
 bind  MOUSE1        +attack
 bind  MOUSE2        +attack2
 
-bind  [             weapprev
-bind  ]             weapnext
-bind  mwheelup      weapprev
-bind  mwheeldown    weapnext
+bind  mwheelup      "weapprev; +useitem"
```
DolceTriade: Doesn't quite work for us, as grenade is a usable right now. Calling +useitem on it will throw it upon switching.

TimePath:
+1

TimePath:
Suggestion: use C for medkit

TimePath:
Suggestion: use 1 for blaster instead of Z

TimePath:
Suggestion: move taunt to X or C

TimePath:
Possibly avoid putting medkit/nades next to each other

TimePath:
T can be freed up by moving it to enter if needed

main/default.cfg
```diff
@@ -7,45 +7,33 @@ unbindall
...
 
-bind  c             +movedown
+bind  CTRL          +movedown
 bind  SPACE         +moveup
 
-bind  UPARROW       +forward
-bind  DOWNARROW     +back
-bind  LEFTARROW     +left
-bind  RIGHTARROW    +right
+
```
mbasaglia: I'm not sure it's a good idea to remove default quake movement/aiming bindings as people might be used to them.

Viech:
- `+speed` is useless for the average player (it slows you down to walking speed), maybe put duck on `ALT` in addition to `CTRL`. This is nice for spectating (flying) as you can descend and ascend with your thumb on `ALT` and `SPACE`.
- `+useitem` is deprecated as no item in the game except the grenade and the medkit needs to be activated and I plan to keep it like this (for example, the new jetpack overrides the jump mechanics instead of being activated). You could put the grenade on the middle mouse button in addition to `g`.
- `y` and `z` should be used with care as long as some keyboard layouts have them swapped while all other letter keys equal the position on a standard english keyboard (QWERTZ vs. QWERTY). Ideally the same command is executed by both. `f` is a good key for weapon toggling, taunt is not that important.
- Does the vsay menu binded to `o` still exist?

main/default.cfg
Viech: People who have played quake should be able to change their bindings accordingly. We do not want *new* players to use the arrow keys as it would be a great disadvantage (as there are very few keys nearby).

DolceTriade:
In its current iteration, +speed is useful for stamina control. Also in competitive settings to silence footsteps. We need to have a proper bind there. Vsay menu doesn't exist anymore.

kmsyzi:
- does modcase still work? perhaps put sprint on shift and have alt as a modifier so alt+shift activiates +speed.

- same for chat, you could use enter for all chat and alt or shift + enter for teamchat, lots of games do it that way.

- this frees up t, maybe use that for the (rather unused) waypoint menu, or vsay menu if it gets reimplemented. 

- ctrl for crouch? or keep it on c? most modern shooters use ctrl i believe. if we move it to ctrl then c is another key freed for something else, maybe use that for medkit, voip (if its ever reimplemented) can then go on v. 

- number keys (1-9) are  unused for default bindings. if we get the vsay menu back we could have 1-5 open each respective menu and use the numbers for that by default. 

- x is free since sprint would be on shift, maybe use that for taunt? 

- not sure what to do in regards to the swapped location of z and y on qwerty/z keyboards, is there any way to have default configs generated based on users keyboard layout? i have no idea if theres a way to go about checking that, otherwise i say just keep itemtoggle blaster on z.

main/default.cfg
mbasaglia: btw, those keys are supposed to replace the mouse, not wasd

mbasaglia:
What is the status on this? Some new changes have been committed a while ago, are any of the old comments still valid?

@Keziolio Change the label to PR-Needs-Review if you think is ready for more review.

Keziolio:
I addressed almost all comments in the last commit, changing what was wrong before. The majority of the comments are now invalid.

Viech:
It looks like an improvment over the current state, even though I feel another redesign will be necessary once we sorted out some remainers of old mechanics (like weapon scrolling). I'd say we can merge this for now. I'll leave @DolceTriade the honor to have another look and push the button.

main/default.cfg
```diff
@@ -7,65 +7,46 @@ unbindall
...
 
-bind  ,             +moveleft
-bind  .             +moveright
 
 //
 // MISC
 //
 
-bind  t             message_public
+bind  ENTER         message_public
```
DolceTriade: Let's keep these keys next to each other. If we use enter to initiate chat, we should have a way to change visibility of the chat from inside the chat prompt.

DolceTriade:
One nit, and otherwise I'm fine with this change. We can either wait for that feature (probably never) or revert that bit of the change.

Viech:
@Keziolio any more thoughts? Otherwise I'll fix it up locally and merge in the next days. (No need to have a revert on master.)

DolceTriade:
By revert, I didn't mean "git revert", but the verb "revert" where we just change it locally and commit.

Keziolio:
No more thoughts on this, go ahead  😋

PR #873, opened 2015-12-28
Replace all uses of <ctype.h> functions with locale-independent versions

slipher:
I was prompted to do this since I was getting some errors from `std::`-prefixed uses of these functions.

In one place (player name validation), I thought it would be better to use a Unicode-aware function to check for spaces, instead of `cisspace`. Is there any existing function for that?

### DIFF
--- a/daemon/src/common/Color.cpp
+++ b/daemon/src/common/Color.cpp
@@ -259,11 +259,11 @@ TokenIterator::value_type TokenIterator::NextToken(const char* input)
         {
             return value_type( input, input+2, parent->DefaultColor() );
         }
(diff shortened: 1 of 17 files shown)


mbasaglia:
What kind of errors were you getting?

DolceTriade:
With VS2015, if you don't include <cctype> instead of<ctype.h> you can't use the C++ versions. This probably makes part of my vs2015 commits redundant.

slipher:
If you include `cctype` the functions are defined in the `std` namespace. If you include `ctype.h` they are only guaranteed to be in the global namespace.

DolceTriade:
LGTM

PR #874, opened 2015-12-29
Hack to stop undefined symbol error when using SDL2 + VS2015

slipher:


DolceTriade:
LGTM

mbasaglia:
Will it cause problems if someone builds sdl from source?

I'm not sure if we can detect the sdl abi from cmake but might be worth investigating.

DolceTriade:
It will not cause problems. If they build SDL from source this symbol is extraneous and unused.

slipher:
It's no longer possible to compile anything with the older MSVC anyway.

PR #875, opened 2015-12-30
Mark VM loading options as Cvar::CHEAT

DolceTriade:
This forces you to use the VM in the pk3. This ensures
you never have a mismatch with the server you are
connecting to.

### DIFF
--- a/daemon/src/engine/framework/VirtualMachine.h
+++ b/daemon/src/engine/framework/VirtualMachine.h
@@ -86,12 +86,14 @@ enum vmType_t {
 	TYPE_END
 };
 
+extern std::unordered_map<std::string, int> vmCvarFlagsMap;
+
 
 struct VMParams {
 	VMParams(std::string name)
 		: logSyscalls("vm." + name + ".logSyscalls", "dump all the syscalls in the " + name + ".syscallLog file", Cvar::NONE, false),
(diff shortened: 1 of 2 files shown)


DolceTriade:
Any further comments?

mbasaglia:
What if the vm in the pk3 is different?

daemon/src/engine/framework/VirtualMachine.h
```diff
@@ -86,12 +86,17 @@ enum vmType_t {
 	TYPE_END
 };
 
+static std::unordered_map<std::string, int> vmCvarFlagsMap = {
```
slipher: Doesn't this need to be in a .cpp file?

DolceTriade: Hm, seems to compile fine.

DolceTriade:
What do you mean? The VM in the pk3 should be what the server is using or else the server is configured incorrectly. In that case, you're screwed either way.

daemon/src/engine/framework/VirtualMachine.h
Kangz: Here each file including VirtualMachine.h will have a new instance of that map, putting it in a cpp would be better.

DolceTriade:
Moved map to cpp.

PR #878, opened 2016-01-03
remove references to sourceforge

cu-kai:


mbasaglia:
How about linking to https://github.com/Unvanquished/Unvanquished/releases/latest ?

PR #880, opened 2016-01-04
Kangz/breakpad fixup compatbreakw

Kangz:
PTAL, an additional fixup to the breakpad code that breaks VM compat. Thanks!


DolceTriade:
LGTM

Kangz:
Just a rebase, there's only one added commit (the last one)

slipher:
There might be something wrong with `Make trap_CrashDump a shared VM syscall` ... it gives me `Error during initialization: IPC: Unexpected end of message` on startup. I double-checked that I built client and nacl-vms and -set vm.cgame.type 1.

PR #881, opened 2016-01-04
Kangz/misc

Kangz:
PTAL, stuff found while working on the asserts and logs branch. Thanks!

### DIFF
--- a/daemon/src/common/Color.cpp
+++ b/daemon/src/common/Color.cpp
@@ -1,7 +1,7 @@
 /*
 ===========================================================================
 Daemon BSD Source Code
-Copyright (c) 2015, Daemon Developers
+Copyright (c) 2013-2016, Daemon Developers
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/daemon/src/common/Platform.h
+++ b/daemon/src/common/Platform.h
@@ -1,7 +1,7 @@
 /*
 ===========================================================================
 Daemon BSD Source Code
-Copyright (c) 2013-2014, Daemon Developers
+Copyright (c) 2013-2016, Daemon Developers
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/daemon/src/common/Platform.h
+++ b/daemon/src/common/Platform.h
@@ -32,7 +32,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define COMMON_PLATFORM_H_
 
 // Platform-specific configuration
-#ifdef _WIN32
+#if defined(_WIN32)
 #define DLL_EXT ".dll"
 #define EXE_EXT ".exe"
 #define PATH_SEP '\\'
--- a/src/sgame/sg_api.cpp
+++ b/src/sgame/sg_api.cpp
@@ -1,22 +1,30 @@
 /*
 ===========================================================================
+Daemon BSD Source Code
+Copyright (c) 2013-2016, Daemon Developers
+All rights reserved.
 
-Daemon GPL Source Code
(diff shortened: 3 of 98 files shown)


daemon/src/common/Color.cpp
```diff
@@ -1,7 +1,7 @@
 /*
 ===========================================================================
 Daemon BSD Source Code
-Copyright (c) 2015, Daemon Developers
+Copyright (c) 2013-2016, Daemon Developers
 All rights reserved.
```
mbasaglia: Is this attribution correct? This code got introduced in 2015

daemon/src/common/Platform.h
```diff
@@ -32,7 +32,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define COMMON_PLATFORM_H_
 
 // Platform-specific configuration
-#ifdef _WIN32
+#if defined(_WIN32)
```
mbasaglia: Any particular reason for this style change?

src/sgame/sg_api.cpp
```diff
@@ -1,22 +1,30 @@
...
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL DAEMON DEVELOPERS BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
mbasaglia: hmm... will this cause licensing issues?

daemon/src/common/Color.cpp
Kangz: I think it is valid to say that the "Daemon BSD source code" (which this file is part of) started in 2013 and it makes it easier to update all the copyrights in the future.

daemon/src/common/Platform.h
Kangz: I like it better than ifdef but it is just a change I made by reflex, can be removed if you want.

src/sgame/sg_api.cpp
Kangz: No, we wrote this file from scratch (function declarations may have been copied but it is unavoidable to make the functions declarations look the same)

Kangz: This file was written a long time ago when we started porting sgame to NaCl and is a reimplementation of parts of the old QVM handling code in g_main.c and g_api.c to use NaCl instead.

daemon/src/common/Platform.h
DolceTriade: I think it's fine. The rest of the code (except the include guard) uses #if defined

DolceTriade:
LGTM

Kangz:
Thanks, merging.

PR #882, opened 2016-01-05
Miscellaneous build process improvements.

Kangz:
Some changes I found while sorting old branches. PTAL

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,13 @@
 cmake_minimum_required(VERSION 2.8)
 project(Unvanquished C CXX)
 
+# When we build in the source dir, the daemon executable cannot be create
+# as there is already a daemon directory, so we disallow building in the
+# source dir.
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
+    message(FATAL_ERROR "Building in the source directory is not supported, try building in a 'build' directory. You may have to delete the CMakeCache.txt file and CMakeFiles directory that are next to the CMakeLists.txt.")
+endif()
+
 set(Daemon_OUT ${CMAKE_CURRENT_BINARY_DIR})
 add_subdirectory(daemon daemon_build)
 include(DaemonGame)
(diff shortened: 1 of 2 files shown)


libs/libRocket/Source/Core/precompiled.cpp
```diff
@@ -27,3 +27,6 @@
 
 #include "precompiled.h"
 
+// XCode's ranlib complain on empty object, adding any function prevents this
+void RocketDummyFunction() {
```
DolceTriade: Please make this commit in the libRocket repo. I will make this a submodule very soon.

Kangz: Great :D
Let's keep this PR open until this happens.

mbasaglia: If it isn't being used elsewhere might be better to declare it as static.

Kangz: Yeah, static doesn't work.

DolceTriade: Either way, the commit needs to happen in the librocket repo so it doesn't get lost. then we update the librocket here by copying what's in the librocket repo.

mbasaglia: Out of curiosity, do we keep librocket up to date manually or do we use git subtrees?

DolceTriade: Manually. It will be a submodule soon.

CMakeLists.txt
```diff
@@ -27,6 +27,13 @@
 cmake_minimum_required(VERSION 2.8)
 project(Unvanquished C CXX)
 
+# When we build in the source dir, the daemon executable cannot be create
+# as there is already a daemon directory, so we disallow building in the
+# source dir.
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
```
mbasaglia: Do you think it's worth having a smarter comparison which accounts for symlinks and whatnot?

Kangz: I don't think so, as it would be odd for someone to have build point to its parent directory.

mbasaglia:
Now we have the librocket submodule, so this can continue.

Viech:
@Kangz, you having an eye on this?

Kangz:
Now I do, added to my TODO list

PR #883, opened 2016-01-06
Kangz/nuke

Kangz:
PTAL, I felt like nuking code today and found the RoQ decoder :3


daemon/src/engine/qcommon/q_shared.cpp
```diff
@@ -1347,35 +1347,12 @@ Safe strncpy that ensures a trailing zero
 =============
 */
 
-#ifndef NDEBUG
```
mbasaglia: Id' be careful altering NDEBUG-guarded code, as it's set by cmake

daemon/src/engine/qcommon/q_shared.h
```diff
@@ -503,8 +460,6 @@ signed char  ClampChar( int i );
 int          DirToByte( vec3_t dir );
 void         ByteToDir( int b, vec3_t dir );
 
-#if 1
```
mbasaglia: Love these :heart:

daemon/src/engine/qcommon/q_shared.cpp
Kangz: Yes, in this case we really have no use for the Hunk Debug, since we are trying to get rid of it

DolceTriade: What's wrong with this code? I don't think this is specific to Hunk Debug, but rather to any Q_strncpyz across the engine.

Kangz: I thought it was another chunk. I thought that noone uses that feature of Q_strncpyz so I might as well remove it but I'm not attached to that.

DolceTriade: You can probably remove it since moving to std::string has more or less deprecated this function. There used to be a define that would #define Q_strncpyz( src, dst, size ) to Q_strncpyzDebug( src, dst, size, __FILE__, __LINE__ ) or something and it would be used engine wide and help track down null src/dst bugs.

Kangz:
Any changes I should do before landing this?

DolceTriade:
LGTM

Kangz:
Thanks, merging.

PR #884, opened 2016-01-08
Lower the required CMake version

Kangz:
PTAL, required for some stable linux distros


mbasaglia:
Has it been tested with that version?

Kangz:
It has been tested with that version on Linux and I checked all the commands used in the Windows path (and introduced in the same patch the required version was bumped) are defined in 2.8.12. Perturbed told me he might have been looking at the 2.8 doc and not the 2.8.12 doc when he checked that the command didn't exist.

PR #885, opened 2016-01-13
Kangz/assert and logs

Kangz:
PTAL, this adds asserts to Daemon, which is a good thing.
I'm planning an UNREACHABLE and UNIMPLEMENTED macro too but that can be done later, as can be BREAKPOINT on PNaCl.


daemon/src/common/Debugger.cpp
```diff
@@ -0,0 +1,127 @@
...
+DISCLAIMED. IN NO EVENT SHALL DAEMON DEVELOPERS BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+===========================================================================
+*/
+
```
DolceTriade: I thought we don't indent in #ifdefs

daemon/src/common/Assert.h
```diff
@@ -0,0 +1,159 @@
...
+
+#if !defined(DAEMON_SKIP_ASSERT_SHORTHANDS)
+    #define ASSERT DAEMON_ASSERT
+    #define ASSERT_EQ DAEMON_ASSERT_EQ
+    #define ASSERT_NQ DAEMON_ASSERT_NQ
+    #define ASSERT_LT DAEMON_ASSERT_LT
+    #define ASSERT_LE DAEMON_ASSERT_LE
+    #define ASSERT_GT DAEMON_ASSERT_GT
+    #define ASSERT_GE DAEMON_ASSERT_GE
+#endif
```
mbasaglia: Why the ASSERT_EQ and similar? Wouldn't ASSERT(a == b) be more readable?

daemon/src/common/Assert.h
```diff
@@ -0,0 +1,159 @@
...
+        } while(DAEMON_ASSERT_LOOP_CONDITION)
+#else
+    #if defined(_MSC_VER)
+        #define DAEMON_ASSERT_CALLSITE_HELPER(file, func, line, directExpression, _1, _2, _3) \
+            __assume(directExpression)
+    #elif defined(__GNUC__)
+        #define DAEMON_ASSERT_CALLSITE_HELPER(file, func, line, directExpression, _1, _2, _3) \
+            if (!(directExpression)) { \
+                __builtin_unreachable(); \
+            }
```
mbasaglia: So, will ASSERT() generate code also on release builds?

daemon/src/common/Assert.h
Kangz: ASSERT only tells you that the values are different, ASSERT_EQ also tells you what the values are.

daemon/src/common/Assert.h
Kangz: Yes, it will generate code equivalent to compiler hints for optimization. Looking at it now I realize on GCC and clang it might call function calls. I'll need to find a better builtin if one exists.

mbasaglia: __builtin_expect?

Kangz: Nope that's branch prediction overriding

Kangz: Clang has __builtin_assume http://clang.llvm.org/docs/LanguageExtensions.html#builtin-functions

daemon/src/common/Debugger.cpp
Kangz: I'm in favor of indenting them, except for the ifdef at the start of headers

daemon/libs/recast/RecastAssert.h
```diff
@@ -22,12 +22,15 @@
...
-#	include <assert.h> 
-#	define rcAssert assert
-#endif
+//#ifdef NDEBUG
+//// From http://cnicholson.net/2009/02/stupid-c-tricks-adventures-in-assert/
+//#	define rcAssert(x) do { (void)sizeof(x); } while((void)(__LINE__==-1),false)
+//#else
+//#	include <assert.h>
+//#	define rcAssert assert
+//#endif
```
mbasaglia: Can this be removed completely?

daemon/src/common/Debugger.cpp
```diff
@@ -0,0 +1,127 @@
...
+#if defined(_WIN32)
```
mbasaglia: I wonder if it's better to handle the conditionals in CMake and have different files for different implementations

daemon/libs/recast/RecastAssert.h
Kangz: It can, will do.

daemon/src/common/Debugger.cpp
Kangz: For something as simple as this I'd rather have a single file, that's just a personal preference.

daemon/src/engine/renderer/tr_main.cpp
```diff
@@ -235,12 +235,12 @@ void R_TBNtoQtangents( const vec3_t tangent, const vec3_t binormal,
...
-	// assert( fabs(DotProduct( normal2, normal2 ) - 1.0f) < 0.01f );
-	// assert( fabs(DotProduct( tangent2, binormal2 ) ) < 0.01f );
-	// assert( fabs(DotProduct( tangent2, normal2 ) ) < 0.01f );
-	// assert( fabs(DotProduct( binormal2, normal2 ) ) < 0.01f );
+	// ASSERT_LT(fabs(DotProduct(tangent2, tangent2) - 1.0f), 0.01f);
+	// ASSERT_LT(fabs(DotProduct(binormal2, binormal2) - 1.0f), 0.01f);
+	// ASSERT_LT(fabs(DotProduct(normal2, normal2) - 1.0f), 0.01f);
+	// ASSERT_LT(fabs(DotProduct(tangent2, binormal2)), 0.01f);
+	// ASSERT_LT(fabs(DotProduct(tangent2, normal2)), 0.01f);
+	// ASSERT_LT(fabs(DotProduct(binormal2, normal2)), 0.01f);
```
mbasaglia: Maybe not needed?

src/sgame/sg_buildable.cpp
```diff
@@ -4153,7 +4153,7 @@ static void BuildableSpawnCBSE(gentity_t *ent, buildable_t buildable) {
 		}
 
 		default:
-			assert(false);
+			ASSERT(false);
```
mbasaglia: Is this needed?

daemon/src/common/Assert.h
Kangz: Done

daemon/libs/recast/RecastAssert.h
Kangz: Done

daemon/src/engine/renderer/tr_main.cpp
Kangz: Removed

src/sgame/sg_buildable.cpp
Kangz: As long as we don't use an enum class here, I think it is good to keep it though few other places in the code do that.

Kangz:
I think this is ready to merge. Any further comments?

mbasaglia:
It's failing all CI buildsf

Kangz:
That's why it isn't merged yet

Kangz:
Passes CI builds (previous failures were flakes it seems?). Merging. Thank you all for the reviews.

Issue #862, opened 2015-12-22
Unvanquished locks mouse even if out of game


DolceTriade:
If you load a map and switch away from unvanquished, when Unvanquished loads, it will look the mouse to the center of the screen. You have to go to Unvanquished and open console or open a menu to unlock the mouse.

mbasaglia:
This should get fixed once #821 is implemented.

DolceTriade:
Could we try to get this in for this release?

mbasaglia:
Looks like the fact that the loading screen doesn't handle any form of input might cause some issues (ie: you can't detect that you lost focus during loading)

Keeping focus detection at every input frame would allow the cursor to be unlocked right away if at the map start the window doesn't have focus but this will happen after the locking occurs. So the mouse will be free but jumps at the center of the game window.

PR #889, opened 2016-01-17
More mouse improvements

### LINKED ISSUE #821: Improved cursor mode
There could be three cursor modes:
* CURSOR_SYSTEM - Sends coordinates, shows the system cursor
* CURSOR_CUSTOM - Sends coordinates, hides the system cursor
* DELTAS - Sends deltas, hides the cursor

This will make ForceCursor redundant, its calls should be replaced and moved away from the frame loop and the currently present checks should become events.

mbasaglia:
### Three mouse modes
* System cursor: The mouse is handled by the system (movement and cursor)
* Custom cursor: The mouse is free to move and rendered by the game
* Deltas: The mouse is hidden and can't move, input is sent as movement deltas

### Mouse focus engine event
A new input event is sent from the engine to the client to signify that the game area has lost or gained focus. This can happen when the window has lost focus, the console is opened and similar situations.

### All mouse modes are handled by cgame
With the new event and mouse modes, is up to the cgame vm to decide what to do with the cursor, the engine never forces the cursor to be in any particular state

## Bug Fixes
Since the new behaviour is more flexible and cgame has all the information it needs, the mouse is no longer locked on map load.

This merge request closes #821 and #862.

## Remaining issues
<del>The loading screen locks all input event so the focus status has to be evaluated at each frame (instead of reacting from SDL events that change it).
Once the loading screen is over, the mouse will jump to the center of the Unvanquished window (without locking), this is caused by the fact that it's impossible to know what happened to the mouse when the loading screen was running.</del>


DolceTriade:
In general, the code looks ok to me. However, Mouse still gets stuck in middle of screen when I switch away during map load.

daemon/src/engine/null/null_input.cpp
```diff
@@ -28,6 +28,7 @@ If you have questions concerning this license or the applicable additional terms
 
 #include "qcommon/q_shared.h"
 #include "qcommon/qcommon.h"
+#include "client/client.h"
```
DolceTriade: Why is this include file required?

daemon/src/engine/sys/sdl_input.cpp
```diff
@@ -561,31 +561,56 @@ static keyNum_t IN_TranslateSDLToQ3Key( SDL_Keysym *keysym, bool down )
...
+    return mouse_mode;
+}
 
 /*
- * Activates the cursor, the system cursor will still be disabled and
- * the game is responsible of drawing the cursor image
+ * Enables or disables the cursor
  */
-void IN_SetCursorActive( bool active )
+void IN_SetMouseMode(MouseMode new_mode)
```
DolceTriade: Perfer camelcase for variable names for consistency with the other variables in the file.

DolceTriade:
Ok, turns out the mouse getting stuck is an edge case. If you start a map when the game doesn't have focus (via your tty terminal console, or even through the unvanquished IPC socket, it will get stuck.

daemon/src/engine/null/null_input.cpp
mbasaglia: For the MouseMode enum

mbasaglia:
Now everything seems to be working fine.

DolceTriade:
Awesome. LGTM.

Kangz:
LGTM too

mbasaglia:
Merged in for-0.48.0

PR #890, opened 2016-01-21
Breakpad fixes

slipher:
Fix hang after crashing on Linux and unnecessary console popping up on Windows.

### DIFF
--- a/daemon/CMakeLists.txt
+++ b/daemon/CMakeLists.txt
@@ -567,7 +567,8 @@ if (USE_BREAKPAD)
             endif()
         endforeach(breaklib)
 
-        add_executable(crash_server src/engine/crash_server/crash_server_windows.cpp)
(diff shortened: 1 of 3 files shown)


Kangz:
LGTM

daemon/CMakeLists.txt
```diff
@@ -571,6 +571,11 @@ if (USE_BREAKPAD)
         target_link_libraries(crash_server
             srclibs-breakpad-crash_generation_server srclibs-breakpad-common)
         target_compile_definitions(crash_server PRIVATE UNICODE)
+        if (MINGW)
```
Amanieu: A better way of doing this is to add `WIN32` to `add_executable`: https://cmake.org/cmake/help/v3.0/command/add_executable.html

slipher: Done

DolceTriade:
LGTM.

PR #891, opened 2016-01-30
[WIP] Profiler

Keziolio:
## Things to do
- decide in which format to export the profile data:
  - chrome json (needs reversing, testing, and probably changes to everything (and in the future chrome can change its format without telling to anyone))
  - custom format and custom tool (currently in a "draft" state, but working)
- test for nacl code profiling (for now i tested it only on the renderer, attemps for cgame failed )
  - pass an argument to the macro to distinguish threads? (once this is working)
- move the currently random location of PROFILE() to useful functions
- maybe a compile option to remove everything and ship the game without the (tiny) overhead

Warning: logs tend to be very heavy, use profiler.frame to capture only one frame.


daemon/src/common/Profiler.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+namespace Profiler{
+
+    /*
+         * Function execution times are saved in the samples vector, every element
+         * contains the name of the function, the type of sample (function start,
+         * function end, and if there's a new frame), and a timestamp.
+         */
+    std::vector <Profiler::Point>                   samples;
+    std::chrono::high_resolution_clock::time_point  startTime;
```
mbasaglia: It might be better to use steady_clock, as it's ensured that time points are sequential

daemon/src/common/Profiler.h
```diff
@@ -0,0 +1,34 @@
...
+        public:
+            std::string label;
+            Profile(std::string);
+
+            ~Profile();
+        };
+
+}
+
+#define PROFILE() Profiler::Profile Profile_Function(__PRETTY_FUNCTION__);
```
mbasaglia: AFAIK __PRETTY_FUNCTION__ is non-standard

daemon/src/engine/audio/Audio.cpp
```diff
@@ -182,6 +183,7 @@ namespace Audio {
     }
 
     void Update() {
+        PROFILE()
```
mbasaglia: I'm a bit suspicious of intrusive profiling

daemon/src/common/Profiler.cpp
```diff
@@ -0,0 +1,186 @@
...
+        time = std::chrono::duration_cast < std::chrono::microseconds > (timeElapsed).count();
+    }
+
+
+    /// At the Beginning and End of every function
+
+    Profile::Profile( std::string label ): label( label ){
+        if(!enabled)
+            return;
+        samples.push_back( Point( START, label ) );
```
mbasaglia: These global variables are always accessed in a non-thread-safe way...

daemon/src/common/Profiler.cpp
Amanieu: Use `Sys::SteadyClock` instead. `std::chrono::steady_clock` is really bad on win32.

daemon/src/engine/audio/Audio.cpp
Keziolio: What do you mean?

daemon/src/common/Profiler.cpp
Keziolio: I'm not really sure on how to fix that

mbasaglia: Currently it'd need locking.
Locking would have a significant performance overhead but I doubt you can alter a vector lock-free

daemon/src/engine/audio/Audio.cpp
mbasaglia: Currently you have to alter the implementation of the functions you are profiling.

I think you should take a look at libunwind and libbacktrace, afaik they allow call stack introspection but I've never used them myself so I can't give you more details.

Kangz:
First of all apologies for being so slow to answer, I have started this many times and ran into some design problems and real life was hectic. The current PR is a good work in progress and shows the feasibility of such a profilers but it is lacking on several points. I really like the idea of having a "profile next frame" command, this will be super handy. Also the RAII recording of events is really good though we need to make sure the C++ standard guarantees the destructor will be run only at the function's end.

The problems with the current approach can be grouped as follows:
 - Performance: the PROFILE macro should have the smallest footpring possible
 - Multithreading: this covers both multiple threads in the engine and VM - engine parallelism
 - Usage (though this is a somewhat lesser point)

If you are willing to iterate on the PR, I'll commit to reviewing your follow-up changes in 48hours and you wil hopefully get to get a better point of view of te architecture of our engine. Now on to comments:

Performance
-----------
We want to have PROFILE() consume as little time as possible so as to not interfere with what it is trying to measure.

The first performance problem is that PROFILE() does allocation so can take from almost no time, to a lot if you have to get pages from the system or if you need to realloc a large allocation. This performance variability, and allocations being expensive on average means they should be avoided if possible. There are currently three allocation made per sample, two in Point::label and one in Profile::label, these could be avoided by passing const char* around. If we later need to pass more than a string, we should try to just store the relevant values in a buffer and have processing happen later. Likewise the point vector push_back function will reallocate its internal buffer when it is full, causing a big and unpredictable spike in performance. I'd suggest using a preallocated fixed-size buffer to store sample data, this buffer could get allocated at profile start time and overflowing data would just be dropped (with a warning).

Without allocations the slowest part of PROFILE() would probably become Sys::SteadyClock::now(). This is a good general purpose clock but we don't know exactly what the implementation of it is and it might be orders of magnitude slower than using and RDSTC-based method. (RDSTC is an x86 CPU register containing the cycles elapsed since boot). We'd probably want to use QueryPerformanceCounter on Windows an CLOCK_MONOTONIC on Posix, or maybe an rdstc compiler builtin.

Multithreading
--------------
Our engine is many-threaded, with two threads used by the engine process itself (main and renderer) and up to two other parallel executions in the virtual machines. There's also special case threads for signal, and command pipes but these are mostly idling so we don't really care. The current approach uses mutable global variable with no clear semantics, which will lead to breakage when we add PROFILE() in renderer code.

The two global variables used are the samples vector and profile enabled variables. The profile vector should not be shared by threads (even under a mutex) as it would lead to a lot of cache invalidations which would deteriorate performance. This means every thread should have its own samples buffer in a TLS variable (thread_local in C++11), it would check at the beginning of each "frame" if it should allocate and at the end of the frame it would hand out the buffer to the profiling subsystem (under a lock). This would reduce locking and sharing to a minimum. Enabled could either be copied at frame start or be in a global variable (with std::atomic, but it seems to be less clean).

Then we have VMs, they would work the same way the regular threads work except that "enabled" would have to be pushed to them, and that instead of just handing a pointer to the profiling subsystem, they would have to copy everything though IPC. This will require the splitting of profiling code between common/ shared/ and framework/ but we can look at that in a later step.

Usage
-----
I think we should have good tools to view profiling info and that's why I suggested to use the Chrome tracing format. Apparently this is not that easy, so going with our own format is a good option. Changing formats later shouldn't be too difficult anyway. I'lll ping the relevant Chromium developers to get an up to date specification. The format you currently have works and will get us started.

While the issue was originally about profiling I think we might want to extend the scope of it to general tracing. One thing in particular I have struggled with in the past is understanding how VM calls get nested, and hacked some tracing locally to produce flame graphs like you are doing. Having a way to toggle tracing of VM calls for a frame would be insanely handy. I'm sure there are other things we would want to trace too, so having different "tracers", one of which is profiling would be good.

Finally when adding profiling to the code we want to choose which function we add PROFILE() to so that it provides the right granularity: too much and it slows down everything, too little it is useless. I can help with this at a later stage. I don't think using libunwind is good, as it would make PROFILE() more expensive.

PR #894, opened 2016-01-31
Console bugs

slipher:
For the 2nd one, there was a Color which was supposed to be statically initialized to white, but apparently ran before White's constructor in an MSVC build, as all the color text showed up as black.

### DIFF
--- a/daemon/src/common/Color.h
+++ b/daemon/src/common/Color.h
@@ -352,18 +352,18 @@ enum {
 }; // enum
 } // namespace Constants
 
-extern Color Black;
-extern Color Red;
-extern Color Green;
-extern Color Blue;
-extern Color Yellow;
-extern Color Orange;
-extern Color Magenta;
-extern Color Cyan;
-extern Color White;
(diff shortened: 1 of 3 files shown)


daemon/src/common/Color.h
```diff
@@ -352,18 +352,18 @@ enum {
...
-extern Color Blue;
-extern Color Yellow;
-extern Color Orange;
-extern Color Magenta;
-extern Color Cyan;
-extern Color White;
-extern Color LtGrey;
-extern Color MdGrey;
-extern Color LtOrange;
+const Color Black    = { 0.00, 0.00, 0.00, 1.00 };
```
mbasaglia: They could be constexpr

DolceTriade:
Could you make it constexpr and we can merge this?

DolceTriade:
constexpr seems to complicate stuff with MSVC. We'll leave as is for now.

mbasaglia:
Even with our macros?

PR #895, opened 2016-02-06
Rework animation interpolation to avoid skipping frames

slipher:
The old algorithm caused animation frames to be skipped frequently, even with a high enough FPS that several rendering frames would occur per the length of one animation frame (but it was jerky with low FPS as well).

Also remove duplication in a nearly identical copy of the same function.


Kangz:
LGTM, would there be other functions that are obvious code duplication?

DolceTriade:
LGTM

DolceTriade:
@Kangz Most of the animation loading and CG_{Buildable|Weapon|Player} all are generally very similar. Further all the CG_{Buildable|Weapon|Player}Animation functions are very similar.

Kangz:
FACTOR ALL!

PR #914, opened 2016-03-06
Cache constant strlen outside of cycles

dimhotepus:
Cache `strlen` value outside of cycles to improve performance.

Sometimes still need to use `int` instead of `std::size_t` because of old
code relying on passing int.

### DIFF
--- a/daemon/src/engine/qcommon/q_shared.cpp
+++ b/daemon/src/engine/qcommon/q_shared.cpp
@@ -180,7 +180,9 @@ Com_CharIsOneOfCharset
 */
 static bool Com_CharIsOneOfCharset( char c, const char *set )
 {
-	for (unsigned i = 0; i < strlen( set ); i++ )
+	std::size_t len = strlen( set );
+
+	for ( std::size_t i = 0; i < len; i++ )
 	{
 		if ( set[ i ] == c )
 		{
@@ -1226,8 +1228,9 @@ int Com_HexStrToInt( const char *str )
 	if ( str[ 0 ] == '0' && str[ 1 ] == 'x' )
 	{
 		int n = 0;
+		std::size_t len = strlen( str );
 
-		for (unsigned i = 2; i < strlen( str ); i++ )
+		for ( std::size_t i = 2; i < len; i++ )
 		{
 			char digit;
 
--- a/src/cgame/cg_gameinfo.cpp
+++ b/src/cgame/cg_gameinfo.cpp
@@ -41,11 +41,11 @@ CG_ParseInfos
 int CG_ParseInfos( const char *buf, int max, char *infos[] )
 {
 	char *token;
-	int  count;
+	int  count = 0;
 	char key[ MAX_TOKEN_CHARS ];
 	char info[ MAX_INFO_STRING ];
 
-	count = 0;
+	auto infoPostfixLen = strlen( "\\num\\" ) + strlen( va( "%d", MAX_ARENAS ) );
 
 	while ( 1 )
 	{
@@ -98,7 +98,7 @@ int CG_ParseInfos( const char *buf, int max, char *infos[] )
 		}
 
 		//NOTE: extra space for arena number
-		infos[ count ] = (char*) BG_Alloc( strlen( info ) + strlen( "\\num\\" ) + strlen( va( "%d", MAX_ARENAS ) ) + 1 );
+		infos[ count ] = (char*) BG_Alloc( strlen( info ) + infoPostfixLen + 1 );
 
 		if ( infos[ count ] )
 		{
--- a/src/cgame/cg_rocket_datasource.cpp
+++ b/src/cgame/cg_rocket_datasource.cpp
@@ -888,7 +888,7 @@ void CG_Rocket_BuildDemoList( const char* )
 	char  demolist[ 4096 ];
 	char demoExt[ 32 ];
 	char  *demoname;
-	int   i, len;
+	int   i;
 
 	Com_sprintf( demoExt, sizeof( demoExt ), "dm_%d", ( int ) trap_Cvar_VariableIntegerValue( "protocol" ) );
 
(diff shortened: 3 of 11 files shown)


src/cgame/cg_rocket_datasource.cpp
```diff
@@ -32,6 +32,7 @@ Maryland 20850 USA.
 ===========================================================================
 */
 
+#include <cstddef>     // for std::size_t
```
mbasaglia: Maybe you can use **auto** and avoid this inclusion

src/cgame/cg_gameinfo.cpp
```diff
@@ -41,11 +41,11 @@ CG_ParseInfos
 int CG_ParseInfos( const char *buf, int max, char *infos[] )
 {
 	char *token;
-	int  count;
+	int  count = 0;
 	char key[ MAX_TOKEN_CHARS ];
 	char info[ MAX_INFO_STRING ];
 
-	count = 0;
+	int infoPostfixLen = strlen( "\\num\\" ) + strlen( va( "%d", MAX_ARENAS ) );
```
mbasaglia: Maybe you can use size_t (or auto) here.

src/cgame/cg_rocket_datasource.cpp
dimhotepus: Looks like a good point. 

Out of curiosity, is it for reduce compile time (less headers, even system => up compile performance), or just for 'do not touch original quake? sources', or smth else?

src/cgame/cg_gameinfo.cpp
dimhotepus: `infoPostfixLen` is used in:
`infos[ count ] = (char*) BG_Alloc( strlen( info ) + infoPostfixLen + 1 );`

BG_Allock take an `int` parameter, looks like it is better to change it
to std::size_t, since `malloc` / `memset` accept `size_t`, and using signed 
type looks like at least strange.

`void *BG_Alloc( int size )
{
	void *ptr = malloc( size );

	if ( ptr )
	{
		memset( ptr, 0, size );
	}

	return ptr;
}`

May be you have some thoughts about?

daemon/src/engine/qcommon/q_shared.cpp
```diff
@@ -34,6 +34,9 @@ Maryland 20850 USA.
 
 // q_shared.c -- stateless support routines that are included in each code dll
 #include "q_shared.h"
+
+#include <cstddef>     // for std::size_t
```
Kangz: that shouldn't be necessary

src/cgame/cg_rocket_datasource.cpp
Kangz: This include is not necessary

src/cgame/cg_gameinfo.cpp
mbasaglia: Since strlen returns a size_t, it's very likely that size_t + int would return a size_t anyway.
But yeah, I guess you can change BG_Alloc to accept a size_t.

src/cgame/cg_rocket_datasource.cpp
mbasaglia: Mostly for simplicity.

daemon/src/engine/qcommon/q_shared.cpp
dimhotepus: Haven't noticed `q_shared.h` already includes `stddef.h`.

Does transitive include is clear there and `q_shared` will always has 
'std::size_t' or 'size_t' types imported?

Kangz: Yes we are basically guaranteed that q_shared includes most of the C++ and C stdlib

dimhotepus: Ok, i got a point. Thank you.

src/cgame/cg_gameinfo.cpp
dimhotepus: Nullifying memory during allocation in `BG_Alloc` is making its usage easier.

But isn't it a performance loss to do it for strings or some structures, as some 
usages are shown? 
`BG_strdup` is clear example of inefficient usage.

May be that is better to change `BG_Alloc` contract and do null memory on
allocation only if needed?

mbasaglia: In that case you could just use malloc.
And there might be parts of the code that use whatever is being returned from BG_Alloc, either directly or indirectly, that expect the whole buffer to be zero-initialized.
So I'd just leave it as it is, it'd be very easy to introduce silent bugs that are extremely hard to spot.

dimhotepus: Ok, replaced some BG_Alloc calls to malloc in low-level functions https://github.com/dimhotepus/Unvanquished/commit/15ec4d5aa0a779563b508a3e2a92709a1beb2555.

daemon/src/engine/qcommon/q_shared.cpp
```diff
@@ -180,7 +180,7 @@ Com_CharIsOneOfCharset
 */
 static bool Com_CharIsOneOfCharset( char c, const char *set )
 {
-	for (unsigned i = 0; i < strlen( set ); i++ )
+	for (std::size_t i = 0, len = strlen( set ); i < len; i++ )
```
Kangz: nit: I'd rather have len initialized before the loop.

daemon/src/engine/qcommon/q_shared.cpp
```diff
@@ -1227,7 +1227,7 @@ int Com_HexStrToInt( const char *str )
 	{
 		int n = 0;
 
-		for (unsigned i = 2; i < strlen( str ); i++ )
+		for (std::size_t i = 2, len = strlen( str ); i < len; i++ )
```
Kangz: nit: Same here.

Kangz:
LGTM with just a couple more nits

daemon/src/engine/qcommon/q_shared.cpp
dimhotepus: Fixed. 
I prefer reduce scope of vars to least possible (less probability of unneeded intersection with out side vars, reduce chance of incorrect usage).

daemon/src/engine/qcommon/q_shared.cpp
dimhotepus: Fixed.

dimhotepus:
Squash commits as required.

Kangz:
Thanks for the improvements, merging.

dimhotepus:
Thank you, too.

Issue #903, opened 2016-02-15
'missing' models in debug build


TimePath:
The following functions should check for file existence first:

* `CG_ParseWeaponFile`
* `CG_InitBuildables`
* `CG_RegisterClientModelname`

mbasaglia:
There's a merged pull request referencing this issue can it be closed?

DolceTriade:
I reverted that PR.

jaytersen:
It was only a partial revert, i believe the issue is fixed.

PR #921, opened 2016-04-02
Fix 'missing' model warnings in debug build

jaytersen:
This is a proposed fix for issue #903

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -630,7 +630,6 @@ void CG_InitBuildables()
 		//Prefer md5 models over md3
 
 		if ( cg_highPolyBuildableModels.integer &&
-		     CG_FileExists( va( "models/buildables/%s/%s.iqm", buildableName, buildableName ) ) &&
 		     ( bi->models[ 0 ] = trap_R_RegisterModel( va( "models/buildables/%s/%s.iqm",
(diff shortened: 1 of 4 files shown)


src/cgame/cg_buildable.cpp
```diff
@@ -630,6 +630,7 @@ void CG_InitBuildables()
 		//Prefer md5 models over md3
 
 		if ( cg_highPolyBuildableModels.integer &&
+		     CG_FileExists( va( "models/buildables/%s/%s.iqm", buildableName, buildableName ) ) &&
```
mbasaglia: Maybe now it's a good idea to store the result of va(...) in a variable, we don't want to risk divergence between the condition and the actual value.

src/cgame/cg_buildable.cpp
```diff
@@ -630,6 +630,7 @@ void CG_InitBuildables()
...
 		     ( bi->models[ 0 ] = trap_R_RegisterModel( va( "models/buildables/%s/%s.iqm",
```
mbasaglia: By the way, looking at RE_RegisterModel(), it seems the value 0 is used for errors, so maybe we can add a check to the trap function (or the engine-side function) and return 0 if the file doesn't exist. This will avoid having to check each time.

Kangz:
LGTM
@DolceTriade, PTAL

DolceTriade:
LGTM

PR #922, opened 2016-04-02
Add a profiler.

Keziolio:
Made another pull request because the old one was messy, sorry

Here's the useful message from kangz https://gist.github.com/Keziolio/9c483257c958f7c898096952f40c5be6

The multi-threading part is still missing and I don't know how to deal with it unfortunately

http://keziolio.github.io/profiler/ here's a buggy implementation of a tool to see the output of the profiler, it's not yet usable but i can update it if there's interest.

Other famous formats like the chromium one and flamegraph were tried but they are not suitable for this case.

### DIFF
--- a/daemon/src/common/Profiler.cpp
+++ b/daemon/src/common/Profiler.cpp
@@ -84,13 +84,13 @@ namespace Profiler{
 
     /// At the Beginning and End of every function
 
-    Profile::Profile( const char * label ): label( label ){
+    ProfilerGuard::ProfilerGuard( const char * label ): label( label ){
         if(!enabled)
             return;
         samples.push_back( Point( START, label ) );
     }
 
-    Profile::~Profile(){
+    ProfilerGuard::~ProfilerGuard(){
         if(!enabled)
             return;
         samples.push_back( Point( END, label ) );
--- a/daemon/src/common/Profiler.cpp
+++ b/daemon/src/common/Profiler.cpp
@@ -132,7 +132,7 @@ namespace Profiler{
 
     class ProfilerStartCmd: public Cmd::StaticCmd {
     public:
-        ProfilerStartCmd(): Cmd::StaticCmd("profiler.start", Cmd::RENDERER, "starts performance profiling") {
+        ProfilerStartCmd(): Cmd::StaticCmd("profiler.start", Cmd::BASE, "starts performance profiling") {
         }
 
         void Run(const Cmd::Args& args) const OVERRIDE {
@@ -153,7 +153,7 @@ namespace Profiler{
 
     class ProfilerStopCmd: public Cmd::StaticCmd {
     public:
-        ProfilerStopCmd(): Cmd::StaticCmd("profiler.stop", Cmd::RENDERER, "stop performance profiling") {
+        ProfilerStopCmd(): Cmd::StaticCmd("profiler.stop", Cmd::BASE, "stop performance profiling") {
         }
 
         void Run(const Cmd::Args& args) const OVERRIDE {
@@ -168,7 +168,7 @@ namespace Profiler{
 
     class ProfilerFrameCmd: public Cmd::StaticCmd {
     public:
-        ProfilerFrameCmd(): Cmd::StaticCmd("profiler.frame", Cmd::RENDERER, "profile the next rendered frame") {
+        ProfilerFrameCmd(): Cmd::StaticCmd("profiler.frame", Cmd::BASE, "profile the next rendered frame") {
         }
 
         void Run(const Cmd::Args& args) const OVERRIDE {
--- a/daemon/src/common/Profiler.cpp
+++ b/daemon/src/common/Profiler.cpp
@@ -1,0 +1,184 @@
+/*
+===========================================================================
+Daemon BSD Source Code
+Copyright (c) 2013-2016, Daemon Developers
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of the Daemon developers nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL DAEMON DEVELOPERS BE LIABLE FOR ANY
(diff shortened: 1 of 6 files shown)


daemon/src/engine/client/cl_main.cpp
```diff
@@ -50,6 +50,8 @@ cvar_t *cl_wavefilerecord;
 #include "framework/Crypto.h"
 #include "framework/Network.h"
 
+#include "../common/Profiler.h"
```
Kangz: I think common/Profiler.h should work. Also headers in the common/ folder should be put in common/Common.h so that precompiled headers pick them up.

daemon/src/common/Profiler.cpp
```diff
@@ -0,0 +1,184 @@
...
+        if(!enabled)
+            return;
+        samples.push_back( Point( END, label ) );
+    }
+
+
+    void Stop(){
+        enabled=false;
+
+        auto file = FS::HomePath::OpenWrite("profile.log"); ///TODO add timestamp to file
```
Kangz: Writing to the file should be done in streaming and not only at Stop, otherwise your sample vector will grow indefinitely if you just call start.

daemon/src/common/Profiler.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Profiler{
+
+        enum Type{ START, END, FRAME };
+
+        class Point{
+
+        public:
+            Type type;
+            std::string label;
```
Kangz: This should be const char* probably.

daemon/src/common/Profiler.h
```diff
@@ -0,0 +1,35 @@
...
+            std::chrono::microseconds::rep time;
+
+
+            Point(Type,const char *);
+        };
+
+        void Update();
+        void Stop();
+
+        class Profile
```
Kangz: Profile is a bit generic for a class name, maybe ProfilerGuard?

daemon/src/common/Profiler.cpp
```diff
@@ -0,0 +1,184 @@
...
+            Stop();
+        }
+    };
+
+    static ProfilerStopCmd ProfilerStopCmdRegistration;
+
+
+    class ProfilerFrameCmd: public Cmd::StaticCmd {
+    public:
+        ProfilerFrameCmd(): Cmd::StaticCmd("profiler.frame", Cmd::RENDERER, "profile the next rendered frame") {
```
Kangz: The commands are base commands, not renderer commands.

Kangz:
If I remember correctly, you profiled the profiler and found it to be very fast, I'm still concerned about perf but I guess we can revisit it later while keeping the same interface. Likewise having multiple profilers (called tracers at that point) would be awesome but we would be able to keep mostly the same interface when changing that so it could be adressed later.

Multithreading is really important from the start though, or we'll have weird crashes as soon as we start adding PROFILE to renderer code. Using std::vectors in TLS would be great. I think the message earlier explains how it could be done but I can elaborate if you want.

daemon/src/engine/client/cl_main.cpp
mbasaglia: Do you think the profiles should be in Common.h? It doesn't seem something that must be visible everywhere, from what I understand it's going to be used only on a selected number of functions.

daemon/src/common/Profiler.cpp
mbasaglia: Wouldn't file IO at every function call make this very slow?

mbasaglia:
> Other famous formats like the chromium one and flamegraph were tried but they are not suitable for this case.

May I ask why? The format for flamegraph seems very close to the one you are already using...

daemon/src/common/Profiler.cpp
Keziolio: I was concerned about speed too, writing to the file at every function call is out of discussion, it may be feasible at every frame but it would still have a performance impact...

Keziolio:
For the multithreading part, I'm going to make some experiments, the original message is clear enough, I'll take that direction.


For the flamegraph format, i tried to make a script that converts from my format to theirs, and stopped at half after figuring out that it's not suitable, the format itself is different, this profiler tracks the time at which each function starts and ends, their format keeps track of all the tree of the functions with "samples", i think it's not necessary to do this, and this works with seconds, not samples.
And, their svg visualizer is not that much better than mine, not sure if there are more advanced softwares compatible with their format but... A converter can be made at any time.

daemon/src/common/Profiler.cpp
```diff
@@ -0,0 +1,201 @@
...
+            ss << std::to_string(i.time) << ";" << type << ";" << i.label << ";" << i.tid << "$\n";
+           // std::cout << ss.str();
+            file.Write(ss.str().c_str(), ss.str().length());
+        }
+
+        file.Close();
+        samples.clear();
+    }
+
+    void Sync(){ //TODO: optimize?
```
Kangz: There should be a call to this in the renderer too.

daemon/src/common/Profiler.cpp
```diff
@@ -0,0 +1,201 @@
...
+        std::lock_guard<std::mutex> lck (mtx);
+
+        samples.insert(samples.end(), localSamples.begin(), localSamples.end());
```
Kangz: Would be nice to:
 - avoid allocations most frame
 - have the thread id so the visualizer can show multiple flame graphs
 - avoid copies when giving the samples

Here's a quick idea:
```cpp
size_t nextCapacity = local
Samples.capacity - std::max(localSamples.capacity() / 256, 1);
threadSamples.push_back(std::make_pair(threadID, std::move(localSamples)));
// At that point localSamples is empty
localSamples.reserve(nextCapacity);```

File: daemon/src/common/Profiler.cpp [hunk 8]
```diff
@@ -0,0 +1,201 @@
...
+
+
+namespace Profiler{
+
+    /*
+         * Function execution times are saved in the samples vector, every element
+         * contains the name of the function, the type of sample (function start,
+         * function end, and if there's a new frame), and a timestamp.
+         */
+    std::vector <Profiler::Point>                   samples;
```
Kangz: Can you put all global variables in an anonymous namespace so the symbols don't get exposed to other compilation units?

daemon/src/common/Profiler.h
```diff
@@ -0,0 +1,45 @@
...
+
+        enum Type{ START, END, FRAME };
+
+        class Point{
+
+        public:
+            Type type;
+            const char* label;
+            std::chrono::microseconds::rep time;
+            std::thread::id tid;
```
Kangz: the thread id doesn't have to be in each sample

daemon/src/common/Profiler.cpp
```diff
@@ -0,0 +1,201 @@
...
+        localSamples.push_back( Point( END, label ) );
+    }
+
+
+    void Stop(){
+        enabled=false;
+
+        auto file = FS::HomePath::OpenWrite("profile.log"); ///TODO add timestamp to file
+
+        std::sort(samples.begin(), samples.end());
```
Kangz: You'll need to lock this access too, to make the lock as tiny as possible, you can std:::move samples into a local variable.

Viech:
What's up with this one, @mbasaglia and @Kangz?

mbasaglia:
@Viech I was skeptical of the design of this profiler before and I still am.
This profiler looks slow, unsafe, non-extensible and too intrusive.

Viech:
In this case I'll close this for now. Feel free to ask for a re-open when there were significant changes to address these issues.

Issue #910, opened 2016-02-21
Remove the static char[] hack where possible.


mbasaglia:
There are several functions that return a char* and use static buffer to do so. Some rotate through an array of buffers to be called several times in a row.

This is not ideal in most situations and could be refactored to a std::string.
Some instances might want to keep the current behaviour to avoid allocation overhead but I have to believe that in most places is just an ugly hack.

Kangz:
These ugly hacks were relevant in 1998 but not anymore, let's get rid of them.

TimePath:
`#define va(...) Str::Format(__VA_ARGS__).c_str()`?

PR #924, opened 2016-04-03
Make the colors convertible to a C++ string instead of a char*

mbasaglia:
See #910, I wanted to be more thorough but many instances of that hack are a bit of a can of worms so I've removed it just for the color stuff.

### DIFF
--- a/daemon/src/common/Color.cpp
+++ b/daemon/src/common/Color.cpp
@@ -173,35 +173,29 @@ CONSTEXPR_FUNCTION bool Has8Bits( uint8_t v )
  * Uses multiple character arrays to allow multiple calls
  * within the same sequence point
  */
-const char* CString ( const Color32Bit& color )
+std::string ToString ( const Color32Bit& color )
 {
-	static const int length = 9;
-	static const int nstrings = 4;
-	static char text[nstrings][length];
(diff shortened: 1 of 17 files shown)


daemon/src/common/Color.cpp
```diff
@@ -173,35 +173,29 @@ CONSTEXPR_FUNCTION bool Has8Bits( uint8_t v )
...
 	{
-		sprintf(text[i], "^#%02x%02x%02x",
-			(int)intcolor.Red(),
-			(int)intcolor.Green(),
-			(int)intcolor.Blue()
-		);
+		text += '#';
+		text += Str::HexDigit(color.Red() & 0xf);
+		text += Str::HexDigit((color.Red() >> 4) & 0xf);
+		text += Str::HexDigit(color.Green() & 0xf);
```
Kangz: nit: We could use an ostringstream, or a buffer on the stack here for better performance.

Kangz:
LGTM

daemon/src/common/Color.cpp
mbasaglia: Since strings usually grow by a factor of 4 I doubt it'd make a difference.

PR #925, opened 2016-04-04
Use Str::ctolower instead of std::tolower

dimhotepus:
A little buildfix

C:\projects\unvanquished\daemon\src\common\Color.cpp(231): error C2039: 'tolower': is not a member of 'std'
c:\projects\unvanquished\daemon\src\common\Optional.h(412): note: see declaration of 'std'


mbasaglia:
Our implementation of ctolower (and friends) looks extremely naive...
But I guess it's good enough.

dimhotepus:
Thanks

PR #927, opened 2016-04-10
Fix infinite stack growth under heavy load on stdout.

jaytersen:
Under heavy I/O on stdout, write() returns -1 and sets
errno to EAGAIN. The return value check causes a
warning to the log which again causes a write to
stdout and on we go. This commit ensures we do not
log to stdout when there is a problem with stdout.
It also uses select to wait until stdout is ready
for writing in case EAGAIN is returned.

### DIFF
--- a/daemon/src/engine/sys/con_tty.cpp
+++ b/daemon/src/engine/sys/con_tty.cpp
@@ -39,6 +39,7 @@ Maryland 20850 USA.
 #include <termios.h>
 #include <fcntl.h>
 #include <sys/time.h>
+#include <sys/select.h>
 
 #include "framework/ConsoleField.h"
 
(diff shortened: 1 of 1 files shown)


jaytersen:
I should mention that i used the _shaderlist_ command in-game to provoke a crash

daemon/src/engine/sys/con_tty.cpp
```diff
@@ -66,9 +67,27 @@ static struct termios TTY_tc;
...
+	int ret;
+
+	ret = write( STDOUT_FILENO, text, strlen(text) );
+
+	if ( ret < 0 && errno == EAGAIN )
+	{
+		fd_set set;
+		FD_ZERO( &set );
+		FD_SET( STDOUT_FILENO, &set );
+		select( STDOUT_FILENO + 1, NULL, &set, NULL, NULL);
```
DolceTriade: Doesn't select block? Will that block the main game loop?

jaytersen: Yes. The alternative is to drop the output.

jaytersen: Maybe a separate thread with some intermediary buffering would do the trick.

Kangz: Either blocking or dropping is fine, with a slight preference for blocking because that only happens when there is a lot of console spam, which is rare.

Viech:
So, how do we go on from here, @jaytersen, @Kangz and @DolceTriade?

DolceTriade:
LGTM

Viech:
I'll let @Kangz decide on it, then.

Kangz:
LGTM, happy that it uses blocking :)

PR #929, opened 2016-04-11
Config header

TimePath:
Move a few configurable constants into a common header which could later be generated by cmake (for other gamelogics)


update-version-number.sh
```diff
@@ -19,10 +19,10 @@ SOURCE_PATH="`dirname "${BASH_SOURCE[0]}"`"
 VERSION="${1}"
 VERSION_SHORT="$(echo "$VERSION" | cut -d. -f1,2)"
 
-# Update q_shared.h
+# Update Defs.h
 TMP_FILE="`mktemp /tmp/config.XXXXXXXXXX`"
-sed "s|\(#define \+PRODUCT_VERSION *\"\)[^\"]*\"|\1${VERSION}\"|" "${SOURCE_PATH}/daemon/src/engine/qcommon/q_shared.h" > "${TMP_FILE}"
-mv "${TMP_FILE}" "${SOURCE_PATH}/daemon/src/engine/qcommon/q_shared.h"
+sed "s|\(#define \+PRODUCT_VERSION *\"\)[^\"]*\"|\1${VERSION}\"|" "${SOURCE_PATH}/daemon/src/common/Defs.h" > "${TMP_FILE}"
+mv "${TMP_FILE}" "${SOURCE_PATH}/daemon/src/common/Defs.h"
```
mbasaglia: Why don't we use CMake to generate this file?

daemon/src/engine/qcommon/common.cpp
```diff
@@ -1318,7 +1320,7 @@ void Com_WriteConfigToFile( const char *filename, void (*writeConfig)( fileHandl
 		return;
 	}
 
-	FS_Printf( f, "// generated by Unvanquished, do not modify\n" );
+	FS_Printf( f, "// generated by " PRODUCT_NAME ", do not modify\n" );
```
mbasaglia: Maybe this should just become Daemon

tarball.sh
```diff
@@ -4,8 +4,8 @@ set -e
 
 cd "$(dirname "$0")"
 
-PACKAGE_NAME="$(sed -e '/^#define PRODUCT_NAME_LOWER/! d; s/[^\"]*\"//; s/\".*$//; s/[^0-9a-z.]/-/g' daemon/src/engine/qcommon/q_shared.h)"
-PACKAGE_VERSION="$(sed -e '/^#define PRODUCT_VERSION/! d; s/[^\"]*\"//; s/\".*$//; s/[^0-9a-z.]/-/g' daemon/src/engine/qcommon/q_shared.h)"
+PACKAGE_NAME="$(sed -e '/^#define PRODUCT_NAME_LOWER/! d; s/[^\"]*\"//; s/\".*$//; s/[^0-9a-z.]/-/g' daemon/src/common/Defs.h)"
+PACKAGE_VERSION="$(sed -e '/^#define PRODUCT_VERSION/! d; s/[^\"]*\"//; s/\".*$//; s/[^0-9a-z.]/-/g' daemon/src/common/Defs.h)"
```
mbasaglia: tarball.sh could also be generated by CMake

Kangz:
LGTM

Eventually all this should be replaced to not be defines.

daemon/src/engine/qcommon/common.cpp
DolceTriade: Would make more sense to have this be the game name rather than the engine. (People were already wtf is daemon when we released Unvanquished.)

PR #931, opened 2016-04-11
Share syscall definitions

TimePath:
Provide gamelogic->engine syscall definitions for use in alternate gamelogics without requiring them to duplicate between updates


src/cgame/cg_api.cpp
```diff
@@ -39,7 +39,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 int VM::VM_API_VERSION = CGAME_API_VERSION;
 
-static IPC::CommandBufferClient cmdBuffer("cgame");
+extern IPC::CommandBufferClient cmdBuffer;
```
Kangz: Not a fan of having an extern declared in a cpp. We could have a .h for that.

src/sgame/sg_api.cpp
```diff
@@ -34,7 +34,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "shared/VMMain.h"
 #include "shared/CommonProxies.h"
 
-static IPC::SharedMemory shmRegion;
+extern IPC::SharedMemory shmRegion;
```
Kangz: ditto

Kangz:
LGTM, except that I think the VM::HandleSyscall functions should be moved too.

src/cgame/cg_api.cpp
TimePath: We have a .h for that

PR #936, opened 2016-04-26
Split lagometer and ping to separate elements.

jaytersen:
This also gives the lagometer element a width and
height so it shows up on the screen.

Screenshot: https://imgur.com/CRSRgUt


DolceTriade:
LGTM

PR #938, opened 2016-05-21
Implement binary messaging

TimePath:


daemon/src/engine/client/cg_msgdef.h
```diff
@@ -719,6 +721,9 @@ using CGameTextInptEvent = IPC::SyncMessage<
 using CGameFocusEventMsg = IPC::SyncMessage<
 	IPC::Message<IPC::Id<VM::QVM, CG_FOCUS_EVENT>, bool>
 >;
+using CGameRecvMessageMsg = IPC::SyncMessage<
+	IPC::Message<IPC::Id<VM::QVM, CG_RECV_MESSAGE>, IPC::SharedMemory, size_t, int>
```
TimePath: So long as this is a `SyncMessage`, the lifetime of `SharedMemory` doesn't matter

daemon/src/engine/client/cl_cgame.cpp
```diff
@@ -1034,6 +1034,13 @@ void CL_SetCGameTime()
 	}
 }
 
+void CL_CGameBinaryMessageReceived(byte *buf, size_t size, int serverTime)
+{
+	auto shm = IPC::SharedMemory::Create(size);
+	memcpy(shm.GetBase(), buf, size);
+	cgvm.SendMsg<CGameRecvMessageMsg>(shm, size, serverTime);
```
TimePath: Need to forward `size`, as `SharedMemory::Create` rounds up

daemon/src/engine/server/sv_client.cpp
```diff
@@ -1526,25 +1514,22 @@ void SV_ExecuteClientMessage( client_t *cl, msg_t *msg )
...
+	} else if (c == clc_moveNoDelta) {
+		SV_UserMove(cl, msg, false);
+	} else if (c != clc_EOF) {
+		Log::Warn("bad command byte for client %i\n", (int) (cl - svs.clients));
 	}
-	else if ( c == clc_moveNoDelta )
-	{
-		SV_UserMove( cl, msg, false );
-	}
-	else if ( c != clc_EOF )
```
TimePath: Bug here: `clc_EOF` always follows no matter what

daemon/src/engine/server/sg_msgdef.h
```diff
@@ -103,7 +103,7 @@ using GetEntityTokenMsg = IPC::SyncMessage<
     IPC::Reply<bool, std::string>
 >;
 using SendGameStatMsg = IPC::Message<IPC::Id<VM::QVM, G_SEND_GAME_STAT>, std::string>;
-using SendMessageMsg = IPC::Message<IPC::Id<VM::QVM, G_SEND_MESSAGE>, int, int, std::vector<char>>;
+using SendMessageMsg = IPC::Message<IPC::Id<VM::QVM, G_SEND_MESSAGE>, int, size_t, std::vector<uint8_t>>;
```
TimePath: Could be sent as `IPC::SharedMemory`, but would have to be a `IPC::SyncMessage`

daemon/src/engine/client/cg_api.h
```diff
@@ -221,6 +221,8 @@ void            trap_SetMouseMode( MouseMode mode );
 void            trap_S_StopBackgroundTrack();
 void            trap_R_RemapShader( const char *oldShader, const char *newShader, const char *timeOffset );
 bool        trap_GetEntityToken( char *buffer, int bufferSize );
+void 			trap_SendMessage(uint8_t *buf, size_t buflen);
```
Kangz: std::vector<uint8_t> ?

daemon/src/engine/client/cg_msgdef.h
Kangz: Yes, however creating a SharedMemory involves a SyncMessage so it shouldn't be done every frame. Instead consider putting this message in the command buffer

Kangz: Ah it is engine -> VM. How about creating a commandbuffer for that direction too instead of making a shmem each time?

daemon/src/engine/client/cl_cgame.cpp
```diff
@@ -1591,6 +1632,18 @@ void CGameVM::QVMSyscall(int index, Util::Reader& reader, IPC::Channel& channel)
...
 
+		case CG_SEND_MESSAGE:
+			IPC::HandleMsg<SendMessageMsg>(channel, std::move(reader), [this](size_t len, std::vector<uint8_t> message) {
+				CL_SendBinaryMessage(message.data(), len);
+			});
+			break;
+
+		case CG_MESSAGE_STATUS:
+			IPC::HandleMsg<MessageStatusMsg>(channel, std::move(reader), [this](int& status) {
+				status = Util::ordinal(CL_BinaryMessageStatus());
```
Kangz: You should be able to send enums through IPC without casting them to int.

daemon/src/engine/client/cg_msgdef.h
TimePath: Does a command buffer involve copying memory around? Ideally shared memory would be created once and reused

daemon/src/engine/client/cg_api.h
TimePath: No reason vector.data() can't be used by the gamecode if it wants to. Keep it general?

daemon/src/engine/client/cl_cgame.cpp
TimePath: Will try

daemon/src/engine/client/cg_msgdef.h
TimePath: Looked into it, there's no copying. However, the extra features don't seem useful

daemon/src/engine/client/cg_api.h
TimePath: An API needs to be built on top of this anyway, calls to SendMessage clobber previous messages from the same frame

TimePath: `Util:: Writer` can be used for this purpose

daemon/src/engine/client/cg_msgdef.h
Kangz: It is just that I think we'll need a command buffer in that direction eventually so we could introduce it now.

daemon/src/shared/client/cg_api.cpp
```diff
@@ -104,6 +104,20 @@ bool trap_GetEntityToken( char *buffer, int bufferSize )
 	return res;
 }
 
+void trap_SendMessage(uint8_t *buf, size_t buflen)
```
Kangz: The point of passing a vector by value here is that you could move it in the message. Whereas with the current API a copy is forced.

daemon/src/engine/server/sv_client.cpp
```diff
@@ -1502,15 +1495,10 @@ void SV_ExecuteClientMessage( client_t *cl, msg_t *msg )
 	}
 
 	// read optional clientCommand strings
-	do
+	for (;;)
```
Kangz: or while(true) :P

TimePath: rustlike: `#define loop for (;;)`

daemon/src/engine/client/cg_msgdef.h
TimePath: I'm not sure how command buffers are supposed to be used, but shared memory is now allocated just once

TimePath:
@kangz: changes have been made

daemon/src/engine/client/cl_cgame.cpp
```diff
@@ -1065,9 +1065,9 @@ static messageStatus_t CL_BinaryMessageStatus(void)
 	return messageStatus_t::MESSAGE_WAITING;
 }
 
-void CL_CGameBinaryMessageReceived(const byte *buf, size_t size, int serverTime)
+void CL_CGameBinaryMessageReceived(const uint8_t *buf, size_t size, int serverTime)
 {
-	auto shm = IPC::SharedMemory::Create(size);
+	static auto shm = IPC::SharedMemory::Create(MAX_BINARY_MESSAGE);
```
Kangz: Or just pass the message by value, if we find it later to be a bottleneck we can change it. This will make it possible to make CGameRecvMessageMsg an async message

Kangz: It was already, but it is flawed because you can't know the shmem will have been consumed already.

TimePath: I know in xonotic passing by value would be a bit of a bottleneck. What else needs to be done for this to be merged?

Kangz: What's the bitrate that xonotic uses? We can send a LOT of data through async IPC.

Kangz:
I'm not a huge fan of the way this is integrated with the netcode (would have liked a proper clc_ number and something less hacky). Still I tink this is ok to merge because having a proper integration in the netcode is HARD (I tried multiple times, there's a ton of things to untangle).

daemon/src/engine/client/cl_cgame.cpp
TimePath: Hm, not *that* much really, but we do most of our networking with binary messages in the gamecode and not the engine

PR #939, opened 2016-05-29
Support loading pak files which don't follow naming convention

TimePath:
Defines a flag for allowing the use of legacy pak naming, the primary use case being in xonotic

### DIFF
--- a/daemon/src/common/Defs.h
+++ b/daemon/src/common/Defs.h
@@ -42,6 +42,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 /** Default base package */
 #define DEFAULT_BASE_PAK    PRODUCT_NAME_LOWER
 
+#define DAEMON_FS_LEGACY_PAKS false
+
 /** URI scheme and server browser filter */
 #define GAMENAME_STRING     "unv"
 #define GAMENAME_FOR_MASTER PRODUCT_NAME_UPPER
--- a/daemon/src/common/Defs.h
+++ b/daemon/src/common/Defs.h
@@ -42,8 +42,6 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 /** Default base package */
 #define DEFAULT_BASE_PAK    PRODUCT_NAME_LOWER
 
-#define DAEMON_FS_LEGACY_PAKS false
-
 /** URI scheme and server browser filter */
 #define GAMENAME_STRING     "unv"
 #define GAMENAME_FOR_MASTER PRODUCT_NAME_UPPER
--- a/daemon/src/common/FileSystem.cpp
+++ b/daemon/src/common/FileSystem.cpp
@@ -2155,7 +2155,12 @@ static void AddPak(pakType_t type, Str::StringRef filename, Str::StringRef baseP
 	Util::optional<uint32_t> checksum;
 	if (!ParsePakName(filename.begin(), filename.end() - suffixLen, name, version, checksum) || (type == pakType_t::PAK_DIR && checksum)) {
 		fsLogs.Warn("Invalid pak name: %s", fullPath);
-		return;
+		if (!DAEMON_FS_LEGACY_PAKS) {
+			return;
+		} else {
+			name = filename.substr(0, filename.size() - suffixLen);
+			version = "1";
+		}
 	}
 
 	availablePaks.push_back({std::move(name), std::move(version), checksum, type, std::move(fullPath)});
@@ -2346,7 +2351,7 @@ bool ParsePakName(const char* begin, const char* end, std::string& name, std::st
 
 	// Get the name of the package
 	const char* underscore1 = std::find(nameStart, end, '_');
-	if (underscore1 == end)
+	if (underscore1 == end || (DAEMON_FS_LEGACY_PAKS && underscore1 == begin))
 		return false;
 	name.assign(begin, underscore1);
 
@@ -2392,7 +2397,7 @@ std::vector<PakInfo> GetAvailableMapPaks()
 	std::vector<PakInfo> infos;
 	for ( const auto& pak : FS::GetAvailablePaks() )
 	{
-		if ( Str::IsPrefix("map-", pak.name) )
+		if (DAEMON_FS_LEGACY_PAKS || Str::IsPrefix("map-", pak.name))
 		{
 			infos.push_back(pak);
 		}
--- a/daemon/src/common/FileSystem.cpp
+++ b/daemon/src/common/FileSystem.cpp
@@ -2161,10 +2161,11 @@ static void AddPak(pakType_t type, Str::StringRef filename, Str::StringRef baseP
 	std::string name, version;
 	Util::optional<uint32_t> checksum;
 	if (!ParsePakName(filename.begin(), filename.end() - suffixLen, name, version, checksum) || (type == pakType_t::PAK_DIR && checksum)) {
-		fsLogs.Warn("Invalid pak name: %s", fullPath);
 		if (!UseLegacyPaks()) {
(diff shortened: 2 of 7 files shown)


daemon/src/common/Defs.h
```diff
@@ -42,6 +42,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 /** Default base package */
 #define DEFAULT_BASE_PAK    PRODUCT_NAME_LOWER
 
+#define DAEMON_FS_LEGACY_PAKS false
```
TimePath: Eventually all of these flags will be overridable in cmake, making this easier to specify

daemon/src/common/FileSystem.cpp
```diff
@@ -2346,7 +2351,7 @@ bool ParsePakName(const char* begin, const char* end, std::string& name, std::st
 
 	// Get the name of the package
 	const char* underscore1 = std::find(nameStart, end, '_');
-	if (underscore1 == end)
+	if (underscore1 == end || (DAEMON_FS_LEGACY_PAKS && underscore1 == begin))
```
TimePath: Apparently we have some maps that start with underscores, this fixes loading them with an empty name

daemon/src/engine/server/server.h
```diff
@@ -379,7 +379,7 @@ void SV_GetPlayerPubkey( int clientNum, char *pubkey, int size );
 void SV_CreateBaseline();
 
 void SV_ChangeMaxClients();
-void SV_SpawnServer( const char *server );
+void SV_SpawnServer(const std::string pakname, const std::string server);
```
TimePath: We need the pakname later, it is no longer `"map-" + name"`
Must be an owned string (at some point, so why not here) as the filesystem is cleared in this function invalidating any prior references

daemon/src/common/FileSystem.cpp
```diff
@@ -2155,7 +2155,12 @@ static void AddPak(pakType_t type, Str::StringRef filename, Str::StringRef baseP
 	Util::optional<uint32_t> checksum;
 	if (!ParsePakName(filename.begin(), filename.end() - suffixLen, name, version, checksum) || (type == pakType_t::PAK_DIR && checksum)) {
 		fsLogs.Warn("Invalid pak name: %s", fullPath);
```
TimePath: Left this here intentionally, we do want to migrate over time

daemon/src/common/Defs.h
mbasaglia: Can we use a cvar instead of a compile-time constant? It might be useful to be able to toggle it without recompilation.

daemon/src/common/FileSystem.cpp
```diff
@@ -2392,7 +2397,7 @@ std::vector<PakInfo> GetAvailableMapPaks()
 	std::vector<PakInfo> infos;
 	for ( const auto& pak : FS::GetAvailablePaks() )
 	{
-		if ( Str::IsPrefix("map-", pak.name) )
+		if (DAEMON_FS_LEGACY_PAKS || Str::IsPrefix("map-", pak.name))
```
mbasaglia: Why don't we use more high-level functions?
Like IsMapPak or something. Then it might be easier to change/extend the naming system

DolceTriade: +1

Viech:
Not sure why @DolceTriade was assigned, isn't this something for either @Kangz or @Amanieu?

DolceTriade:
prolly cuz I was most active when this was made

illwieckz:
I'm just reading that topic. Why not, for games that have a very large heritage they can't throw away, having the ability to get some pakdir with the new vfs, and some other pakdir using the legacy naming.

For example, xonotic can use a `pkg/` dir for all official stuff (with proper name, version, prefix and dependencies), and a legacy `data/` dir for legacy contrib stuff. If people wants to put legacy contrib stuff n `data/`, they can, at they own risk.

The idea would not be to discriminate the legacy data dir and the modern pkg dir with their names (because we must allow people to add plenty of arbitrary pakpaths), but with a kind of directory metadata, for example an empty file at this directory root.

For example, if there is a `.legacy` file inside a pak dir, use legacy naming, if not, use advanced naming.

This way, the Xonotic project just have to ship a `pkg/` dir (or any name they like) with their modern new packages, plus a `data/` dir (or any name they like) with a file named `data/.legacy` inside, where people can put legacy stuff if they want. Same for the `.xonotic/ user dir`, the game just have to create these dirs:

```
.xonotic/
.xonotic/pkg/
.xonotic/data/
```

plus this file:

```
.xonotic/data/.legacy
```

So user can put ported maps in `pkg/` an unported maps in `data/`, and they still can create any arbitrary pakpaths, having the ability to create modern ones or legacy ones.

I think this per-directory option is better than a global option like a cvar or a compile-time option. By the way, it could be good to disable the legacy support for games without heritage, to enforce good practices from the start.

Kangz:
+1

illwieckz:
Related to my comment above, if there is a mechanism implemented to detect legacy pak directories and normal pak directories, the client engine must know if an autodownloaded pak comes from a legacy pak dir (for example testing if the parent dir on the server has a `.legacy` file) and must store the file localy on the default pak dir for legacy stuff (`data` in xonotic's example).

TimePath:
As much as I like the `.legacy` idea, it would require updating all map pk3 files
Will go with the cvar approach for now

illwieckz:
My `.legacy` idea was to flag directories (pakpaths) to put _unmodifed_ paks in it. :wink:

TimePath:
Controlled by a cvar, we can re-visit a hybrid solution at a later time

daemon/src/common/FileSystem.cpp
```diff
@@ -122,6 +122,13 @@ template<> struct SerializeTraits<FS::LoadedPakInfo> {
 
 namespace FS {
 
+static Cvar::Cvar<bool> fs_legacypaks("fs_legacypaks", "Don't use versioned pk3 files", Cvar::NONE, false);
```
DolceTriade: New Cvar scheme? Let's not keep adding new cvars in the old style.

daemon/src/common/FileSystem.cpp
DolceTriade: Perhaps change the log message too to indicate legacy pak usage.

DolceTriade:
LGTM except for the few nits

daemon/src/engine/server/sv_init.cpp
```diff
@@ -447,7 +447,7 @@ clients along with it.
 This is NOT called for map_restart
 ================
 */
-void SV_SpawnServer( const char *server )
+void SV_SpawnServer(const std::string pakname, const std::string server)
```
mbasaglia: Any reason to pass by const value instead of const reference?

mbasaglia: Also, shouldn't "server" be called "map" or something like that?

mbasaglia:
There are a couple things that could be separate functions (https://github.com/Unvanquished/Unvanquished/pull/939/files#diff-c043ab13bf5eceb630bdcec72692275aR2407 and https://github.com/Unvanquished/Unvanquished/pull/939/files#diff-54a2522a75205ecd7f520f169199786aR63) but I guess it can be done later.

daemon/src/engine/server/sv_init.cpp
TimePath: Must be an owned string (at some point, so why not here) as the filesystem is cleared in this function invalidating any prior references

daemon/src/common/FileSystem.cpp
TimePath: New one being `fs.legacypaks`?

DolceTriade: On second thought, we can leave this as is. All the other filesystem cvars are stil fs_* so changing this one a lone would be confusing.

PR #942, opened 2016-06-19
Sound.cpp slight optimization + cleanup.

nathanbarnesduncan:
Used short circuiting to get slightly better performance and cleaned up the loop in the GetSource function in Sound.cpp

### DIFF
--- a/daemon/src/engine/audio/Sound.cpp
+++ b/daemon/src/engine/audio/Sound.cpp
@@ -137,13 +137,7 @@ namespace Audio {
                 return &source;
             }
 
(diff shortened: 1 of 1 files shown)


Kangz:
I don't think this results in an optimization because the previous iteration of the code had a continue statement. Merging the two identical ifs is a good think though: LGTM

nathanbarnesduncan:
Sorry if that wasn't clear: switching the order of the two if statements is what gave the performance increase in my testing (with optimization flags) as there were quite a few unnecessary (if best < 0) calls.

PR #946, opened 2016-07-05
Fix errors in the quaternion functions.

### LINKED ISSUE #945: Quaternion library is inconsistent.
Vec3 a = {1.0f,2.0f,3.0f}, b;
    quat_t aq;
    AnglesToQuat(a.Data(), aq);
    QuatToAngles(aq, b.Data());
    turretLogger.Debug("%s %s", Utility::Print(a), Utility::Print(b));

→ `Debug: (1.00, 2.00, 3.00) (1.00, 178.00, 177.00)`

Viech:
This is supposed to fix #945.

### DIFF
--- a/daemon/src/engine/qcommon/q_math.cpp
+++ b/daemon/src/engine/qcommon/q_math.cpp
@@ -3098,50 +3098,29 @@ void QuatToAngles( const quat_t q, vec3_t angles )
 	angles[ ROLL ] = RAD2DEG( atan2( 2.0f * ( q[ 3 ] * q[ 0 ] + q[ 2 ] * q[ 1 ] ), ( -q2[ 0 ] + q2[ 1 ] - q2[ 2 ] + q2[ 3 ] ) ) );
 }
 
-void QuatMultiply0( quat_t qa, const quat_t qb )
+void QuatMultiply( const quat_t qa, const quat_t qb, quat_t qc )
+{
+	/*
+	*	   from matrix and quaternion faq
+	*	   x = w1x2 + x1w2 + y1z2 - z1y2
+	*	   y = w1y2 + y1w2 + z1x2 - x1z2
+	*	   z = w1z2 + z1w2 + x1y2 - y1x2
+	*
+	*	   w = w1w2 - x1x2 - y1y2 - z1z2
+	*/
+
+	qc[0] = qa[3] * qb[0] + qa[0] * qb[3] + qa[1] * qb[2] - qa[2] * qb[1];
+	qc[1] = qa[3] * qb[1] + qa[1] * qb[3] + qa[2] * qb[0] - qa[0] * qb[2];
+	qc[2] = qa[3] * qb[2] + qa[2] * qb[3] + qa[0] * qb[1] - qa[1] * qb[0];
+	qc[3] = qa[3] * qb[3] - qa[0] * qb[0] - qa[1] * qb[1] - qa[2] * qb[2];
+}
+
+void QuatMultiply2( quat_t qa, const quat_t qb)
 {
 	quat_t tmp;
 
-	QuatCopy( qa, tmp );
-	QuatMultiply1( tmp, qb, qa );
-}
-
-void QuatMultiply1( const quat_t qa, const quat_t qb, quat_t qc )
-{
-	/*
-	 *	   from matrix and quaternion faq
-	 *	   x = w1x2 + x1w2 + y1z2 - z1y2
-	 *	   y = w1y2 + y1w2 + z1x2 - x1z2
-	 *	   z = w1z2 + z1w2 + x1y2 - y1x2
-	 *
-	 *	   w = w1w2 - x1x2 - y1y2 - z1z2
-	 */
-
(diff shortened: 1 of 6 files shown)


Viech:
This does not yet seem to make quats and angles fully functional. See #945 for details.

Viech:
@cmf028 can you explain what angle representation is assumed by your new code?

If I assume the order **pitch, yaw, roll**, as most of the game seems to do, then the test gives me:
```
t = (   0.00,   90.00,    0.00), h = (  45.00,    0.00,    0.00) → a = (   0.00,   90.00,   45.00).
```
I would assume that when I rotate by a yaw of 90° first, then by a pitch of 45°, it should give me a yaw of 90° and a pitch of 45° total. This however does not equal the result of the test.

gimhael:
The result is correct. The angles are applied in the order roll, pitch, yaw. The names and order of the angles are derived from airplane physics. Assuming a plane traveling in +x direction with +y right- and +z up-vectors, the roll is a rotation around the x-axis, pitch is rotation around the y-axis and yaw is around the z-axis. If you apply the 90 degrees yaw rotation first, the +x vector is rotated to the +y vector and so the 45 degrees "pitch" doesn't turn the nose up or down, but rotates the plane around its new direction vector.
You get the same result by using a 45 degree roll and then the 90 degree yaw, as your test case shows.

Viech:
I though the order of operations for airplanes was yaw-pitch-roll? Can you state more precisely: what's the order of yaw, pitch and roll within the vector and what is the order of their execution?

(Also if the order is roll-pitch-yaw as you state, then in your argument wouldn't the result be a 45° roll and a 90° yaw but a 45° roll followed by a 90° pitch?)

Viech:
Also, can one explain why not to use @DolceTriade's first approach where he simply transformed the quat to a matrix and the matrix to angles, analogoues to how `QuatFromAngles` does it? It seems to guarantee a small amount of consistency, something that we really need (at this point more so than performance I would argue).

cmf028:
The following are equivalent interpretations of the set of angles:

Given a set of angles (pitch, yaw, roll)

Intrinsic rotation:
Rotate the object around its local z axis by yaw,
then rotate the object around its local y axis by pitch,
then rotate the object around its local x axis by roll.

Extrinsic rotation:
Rotate object around the (global) x axis by roll, then
rotate the object around the (global) y axis by pitch, then
rotate the object around the (global) z axis by yaw.

Note that in intrinsic rotation the rotation axes are changing with each elemental rotation, but in extrinsic rotation the rotation axes are constant.

The format of the angles argument to QuatToAngles is the same as the rest of the engine.

cmf028:
I simultaneously updated the commented out code in QuatFromAngles in the latest commit so that it works now.

I would rather use the other version of QuatFromAngles instead of converting QuatToAngles to use MatrixToAngles.

Viech:
Thanks for clarification. I still seem to miss a detail on the test case though.

The test I wrote is supposed to mimic my turret code. On the client side, the turret model has a set of bones that depend on each other: The barrel bone moves together with the headpiece bone and the headpiece bone moves together with the torso bone. More precisely, first the torso bone's roll controls the turret's yaw, then the head bone's pitch controls the turrets pitch (relative to the current torso orientation) and last the barrel bone's roll actually controls the barrel's roll (relative to the turret's head, so that the barrel spins around its own local axis). So the intrinsic order of roations that the turret needs to aim at stuff is yaw-pitch-roll. Note that on the server side, the turret barrel's roll is never considered – the client only modifies it autonomously whenever the firing flag is set.

What I need the `AbsoluteAngles` function defined in #945 to do is the following: Given a torso orientation (relative to the world) `t`, in my last example it defines a yaw of 90° (`(0,90,0)`), and a head rotation `h` (relative to the torso), in the example a pitch of 45° (`(45,0,0)`), I want the absolute orientation `a` to be angles in a yaw-pitch-roll intrinsic order obtained by first applying the torso orientation and then rotating this by the head rotation. However, in the example I would want the resolut `a` to be a pitch of 45°, a yaw of 90° and no roll, that is `(45,90,0)` instead of `(0,90,45)`. How to achieve this and where's my error in reasoning?

For reference the example output:
> t = (   0.00,   90.00,    0.00), h = (  45.00,    0.00,    0.00) → a = (   0.00,   90.00,   45.00).

gimhael:
The error in your reasoning is that (45,0,0) is not a rotation relative to the turret base, but it is always relative to the world y-axis.

If you have a rotation l defined in a local reference system and you have a rotation r that defines the orientation of this local reference system, then the rotation l in world space is (r * l * r^-1), which basically means you rotate the local axes to the world axes by r^-1, then apply the l rotation and then return back to the local space.

So in your example the turret head rotation in the world space is h' = (t * h * t^-1). If you then combine this with the base t transformation you get the transformation (t * h * t^-1) * t = t * h.

So turning the base by yaw t and then the head relative to the base by pitch h is the same as first turning the head by h (in world coordinates) and then the base by t.

Viech:
Thank you! With this in mind I can make it work and even explain why it works. For me the pull request looks good. Before a merge I'd give @cmf028 a chance to make QuatFromAngles use his fixed code instead of matrices, if he wants to do so.

Viech:
Squashed and merged.

PR #947, opened 2016-07-05
Revert commit that reworked animation frame seeking.

DolceTriade:
This animation does not work as intended for buildables. There are instances where the attack animation only partially plays or is skipped entirely.

### DIFF
--- a/src/cgame/cg_animation.cpp
+++ b/src/cgame/cg_animation.cpp
@@ -27,11 +27,15 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 ===============
 CG_RunLerpFrame
 
+Sets cg.snap, cg.oldFrame, and cg.backlerp
 cg.time should be between oldFrameTime and frameTime after exit
 ===============
 */
 void CG_RunLerpFrame( lerpFrame_t *lf, float scale )
 {
+	int         f, numFrames;
+	animation_t *anim;
+
 	// debugging tool to get no animations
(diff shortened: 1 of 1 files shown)


Viech:
If I understood it correctly, then if 9ac589f was merged, some animations would not play smoothly. I'd rather not introduce a regression if we potentially can fix both issues by spending some more time on it.

DolceTriade:
The reason this commit was made was because animations like rifle reload seemed slightly jerky to me, but I was the only person who saw that. This will return the animation system to the way it was before.

Viech:
I'm sure @slipher can confirm jerkiness given that he made the original fix?

DolceTriade:
Probably. That said, I'm not sure why this happens at all. I found one issue with the original commit, but am uncertain why my commit doesn't fix it. Therefore, it's likely we'll see more unforeseen issues from this change.

Viech:
I'll assign @slipher then. He can choose to:
- Rework the commit, creating an alternative pull request. (Prefered! :smiley:)
- Give up on it, in which case *this* pull request would get merged and the jerkiness gets an issue to be investigated further.

DolceTriade:
Honestly, the real fix for this is to refactor our animation system to

 - Not abuse legs/torsoAnim for things that are not the legs/torso.
 - All of the animation code is more or less the same with very minor changes. It needs to be consolidated.
 - Undo the weird double anim thing going on with buildables (which is what I suspect is why all the other animations are generally fine but buildables are not.)

Viech:
Superseded by #951.

PR #950, opened 2016-07-10
Update Detour library

cmf028:
We currently use a local copy of the library from 2013. Hopefully this will fix a few issues with bots.


Kangz:
LGTM

PR #951, opened 2016-07-10
Fix premature termination of buildable animations

slipher:


Viech:
Rebased 761ddf0 as 57a86bb on current master.

PR #953, opened 2016-07-19
couple of mem leaks fixes

devnexen:


Kangz:
LGTM, will merge after it passes Travis.

Two things: this memleak happened because of our move away of the Q3 memory zones, in regular Quake3 the tokens would have been freed when one of the zone was cleared. Also, how did you find the leak? It should happen only when the parsing fail, was it failing to parse one of the files we ship?

devnexen:
I found via static code analysis.

Kangz:
Nice, thanks for your contribution!

Issue #877, opened 2016-01-03
CMake requires curses to be installed even when curses are disabled.


mbasaglia:
[daemon/CMakeLists.txt](https://github.com/Unvanquished/Unvanquished/blob/master/daemon/CMakeLists.txt) defines the CMake option USE_CURSES which is mean to disable curses.
It's being forwarded to the compiler and is being checked in the code but it isn't being checked before [find_package(CursesW REQUIRED)](https://github.com/Unvanquished/Unvanquished/blob/master/daemon/CMakeLists.txt#L468).

PR #962, opened 2016-09-25
Honor the USE_CURSES option in various places

jaytersen:
This is a fix for issue #877

### DIFF
--- a/daemon/src.cmake
+++ b/daemon/src.cmake
@@ -301,7 +301,7 @@ set(QCOMMONLIST
     ${ENGINE_DIR}/sys/con_common.cpp
 )
 
-if (NOT APPLE AND USE_CURSES)
+if (NOT APPLE AND (USE_CURSES OR WIN32))
     set(ENGINELIST ${ENGINELIST}
         ${ENGINE_DIR}/sys/con_curses.cpp
     )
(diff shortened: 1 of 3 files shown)


daemon/src.cmake
```diff
@@ -301,7 +301,7 @@ set(QCOMMONLIST
     ${ENGINE_DIR}/sys/con_common.cpp
 )
 
-if (NOT APPLE)
+if (NOT APPLE AND (USE_CURSES OR WIN32))
```
mbasaglia: Instead of so many checks, we could change the default value of USE_CURSES based on the OS and then just use that. (this would affect the other NOT APPLE above).

jaytersen: Yeah it is a bit much. 
It seems like windows must use curses enabled (effectively ignores USE_CURSES), Linux can use curses and apple must not use curses. So it seems to me that apple and windows shouldn't even have the option while linux is the only one that should be able to enable/disable curses.

mbasaglia: Still, if the user wants to override it, I don't see why not. It would also make these checks easier and more consistent.

jaytersen: But it wont compile/link if the user overrides it. We could hardcode USE_CURSES on win/apple to ON/OFF while having it configurable on linux. That would also make the checks easier.

mbasaglia: What I wonder is if someone compiles curses from source or something to enable support.
Though I don't use either apple or win so I don't know if it's possible.

jaytersen: Yeah there is that of course. So then curses wise there should be no difference between linux and mac in the cmake files, except for the default setting,

PR #963, opened 2016-10-09
using non blocking call if possible to generate randomness

devnexen:


daemon/src/common/System.cpp
```diff
@@ -346,13 +354,23 @@ void GenRandomBytes(void* dest, size_t size)
 	if (nacl_secure_random(dest, size, &bytes_written) != 0 || bytes_written != size)
 		Sys::Error("nacl_secure_random failed");
 #else
+#ifdef __linux__
```
mbasaglia: Maybe it's more readable to have
```
///...
#elif defined(__linux__) && defined(HAS_GETRANDOM_SYSCALL)
//...
#elif defined(__linux__) 
//...
#endif
```

DolceTriade:
Seems like the error here is somewhere else? This looks ok to me.

Viech:
@DolceTriade what error are you talking about? Also, OK to merge now (also @mbasaglia)?

mbasaglia:
TBH I'd just use <random> instead of all this custom code, but the changes themselves look ok to me.

illwieckz:
Hi @devnexen! The _Dæmon Engine_ is now extracted from this source tree, you can find its dedicated repository there:

- https://github.com/DaemonEngine/Daemon

Can you replay your pull request on this repository? Merci et bienvenue ! :wink:

devnexen:
Sure :-)

Issue #799, opened 2015-08-22
some pending renderer regressions introduced in alpha 38 and alpha 40


illwieckz:
Hi, these regressions are not spotted by current unvanquished assets, but this foreign assets show things that worked before and are now broken, so unvanquished things can break in the future.
- How looks the railgun map from Wolf:ET on unvanquished 0.37.0:
  ![railgun](http://dl.illwieckz.net/u/thomas-debesse/bazar/unvanquished/bugs/etmain-regression/unvanquished_0.40.0_2015-08-22_005244_000.jpg)
  Only the snow particle is missing (perhaps it never worked).
- How looks the railgun map from Wolf:ET on unvanquished 0.38.0:
  ![railgun](http://dl.illwieckz.net/u/thomas-debesse/bazar/unvanquished/bugs/etmain-regression/unvanquished_0.38.0_2015-08-22_024024_000.jpg)
  Many surface shaders are broken, trees are broken.
- How looks the railgun map from Wolf:ET on unvanquished 0.40.0:
  ![railgun](http://dl.illwieckz.net/u/thomas-debesse/bazar/unvanquished/bugs/etmain-regression/unvanquished_0.37.0_2015-08-22_030137_000.jpg)
  The fog is broken and the skybox too (the white background is the void).

illwieckz:
For information, `daemon` complains about that:

```
WARNING: unknown shader stage parameter 'detail' in shader 'textures/snow_sd/icelake2_opaque' 
WARNING: unknown shader stage parameter 'detail' in shader 'textures/mp_railgun/lmterrain_6' 
WARNING: unknown shader stage parameter 'detail' in shader 'textures/mp_railgun/lmterrain_5' 
WARNING: unknown shader stage parameter 'detail' in shader 'textures/mp_railgun/lmterrain_4' 
WARNING: unknown shader stage parameter 'detail' in shader 'textures/mp_railgun/lmterrain_3to4' 
WARNING: unknown shader stage parameter 'detail' in shader 'textures/mp_railgun/lmterrain_3' 
WARNING: unknown shader stage parameter 'detail' in shader 'textures/mp_railgun/lmterrain_2to5' 
WARNING: unknown shader stage parameter 'detail' in shader 'textures/mp_railgun/lmterrain_2to4' 
WARNING: unknown shader stage parameter 'detail' in shader 'textures/mp_railgun/lmterrain_2to3' 
[...]

PR #954, opened 2016-07-24
Placed daemon-engine in submodule

Elzair:
Since the daemon engine is now being used for multiple projects (such as Xonotic), I think it would be best to place it in its own repository. I used `git filter-branch` to create a repository containing only the commits pertinent to the _daemon_ subdirectory, removed the _daemon_ subdirectory, and added the [new repository](https://github.com/Elzair/daemon-engine) as a submodule under the subdirectory _daemon_. (**NOTE:** Since the _.gitignore_ ignores any files named "daemon", I changed the name to "daemon-engine"). I also removed the submodules _daemon/libs/breakpad_ and _daemon/libs/recastnavigation_ from the repository and added them to the engine repository.

I tested the new system on Arch Linux x86_64. It built okay except for the following part.

`[ 99%] Linking CXX executable cgame-native-exe
CMakeFiles/cgame-native-exe.dir/src/cgame/rocket/rocket.cpp.o: In function`Rocket::Core::Lua::LuaTypeRocket::Core::Lua::Cvar::thunk(lua_State_)':
rocket.cpp:(.text._ZN6Rocket4Core3Lua7LuaTypeINS1_4CvarEE5thunkEP9lua_State[_ZN6Rocket4Core3Lua7LuaTypeINS1_4CvarEE5thunkEP9lua_State]+0x29): undefined reference to `lua_rotate'
rocket.cpp:(.text._ZN6Rocket4Core3Lua7LuaTypeINS1_4CvarEE5thunkEP9lua_State[_ZN6Rocket4Core3Lua7LuaTypeINS1_4CvarEE5thunkEP9lua_State]+0x6e): undefined reference to`lua_rotate'
CMakeFiles/cgame-native-exe.dir/src/cgame/rocket/rocket.cpp.o: In function `Rocket::Core::Lua::LuaTypeRocket::Core::Lua::Cmd::thunk(lua_State_)':
rocket.cpp:(.text._ZN6Rocket4Core3Lua7LuaTypeINS1_3CmdEE5thunkEP9lua_State[_ZN6Rocket4Core3Lua7LuaTypeINS1_3CmdEE5thunkEP9lua_State]+0x29): undefined reference to `lua_rotate'
rocket.cpp:(.text._ZN6Rocket4Core3Lua7LuaTypeINS1_3CmdEE5thunkEP9lua_State[_ZN6Rocket4Core3Lua7LuaTypeINS1_3CmdEE5thunkEP9lua_State]+0x6e): undefined reference to`lua_rotate'
[...]

PR #966, opened 2016-10-12
do not abort shader parsing on unknown keyword warning, ref #799

illwieckz:
When an unknown keyword is found in a shader, the parser displays a warning, drops the current line but returns false, invalidating the whole shader, which is not a warning behavior. This change makes the shader parsing continue after the warning. If the shader must not continue, the parser must not raises a warning but an error. It was the behavior before the refactoring of this part. See #799.

### DIFF
--- a/daemon/src/engine/renderer/tr_shader.cpp
+++ b/daemon/src/engine/renderer/tr_shader.cpp
@@ -2424,7 +2424,7 @@ static bool ParseStage( shaderStage_t *stage, const char **text )
 		{
 			Log::Warn("unknown shader stage parameter '%s' in shader '%s'", token, shader.name );
 			SkipRestOfLine( text );
-			return false;
+			continue;
 		}
 	}
 


illwieckz:
Before:
etmain railgun daemon warning return
After:
etmain railgun daemon warning continue

illwieckz:
Perhaps we can implement a cvar to make warning acting like an error (like `-Werror` in C compiler) for mappers who create brand new stuff and want to be cautious, but it' would be another change.

DolceTriade:
Yes, could you add a cvar like r_strictShader or something? Otherwise, I'm fine with this CL.

illwieckz:
It would be a bigger change since if we implement this cvar, we have to look at all shader warnings to put the `if strict: return false; else: continue` stuff, or implement a factored version, which is outside my scope currently.

illwieckz:
If I do another `continue` instead of a `return` in another part of code, the renderer becomes able to load xonotic shader while ignoring the darkplace's specific shaders, printing a warning.

Before:
solarium daemon invalid bad shader
After:
solarium daemon permissive bad shader

I haven't pushed this another change yet. In fact, there is absolutely no urge to merge that current PR, and it can be done after the engine split. But I think it's good to talk about that change. I think the shaders must be permissive. In fact they are already permissive. The first commit I put in this PR is just reenabling a permissiveness that were there before (I checked older revisions, before the code refactoring).

For the other commit I haven't pushed, I don't know if the permissiveness was there before. In fact there is many places in the code where the shader parser prints a warning and continue, and some places were it prints a warning and return. Perhaps one of these warnings must be an error instead and must return, but I guess most of them are simple warnings and can continue. It's why there is really no urge to merge this PR, especially since there is a code split pending.

But I think it could be very good to improve the permissiveness of the shader parser to help people porting games on Dæmon, while allowing brand new game to enable a strict mode (and that can be a default for Unvanquished!). That's probably something I can work on, I'm not a c++ master but heh, it's just adding a cvar, and writing some `continues` and `return` at some place.

In fact the hard point of that job is to test the permissiveness, because it needs to test faulty shaders. That's why I wrote the two commits, the one already visible, and the other one I done for this xonotic test. I was able to test how the shader parser reacted.

mbasaglia:
If we add the cvar, we could refactor the code to look something like this:
```
if ( shader_warning("Message") == Fail )
    return false;
else
    continue;
```

Where shader_warning could be checking the cvar or other things if needed.
We could also have a severity argument to allow certain warnings and fail on others.

illwieckz:
Something I'm not sure is when `SkipRestOfLine( text );` is needed and when is not.

illwieckz:
Just to say I will have a very busy time the next 4 weeks, so I'm only requesting for comments at this time, it would be an opportunity to do more than the current commit does. Do not block nothing because of that (especially, do not block #954 because of that).

PR #972, opened 2016-10-29
Remove a leftover from the VOIP nuke

maek:


DolceTriade:
These are compat breaking changes. I think we should leave them in, or use a reserved keyword so we don't change all the enum values.

mbasaglia:
Would it make sense to start using explicit values for enums, so we avoid adding compat breakages every time we add or remove values?

maek:
To be fair there's also Speex that should be removed from:

- daemon/cmake/FindSpeex.cmake
- daemon/external_deps/build.sh
- .travis.yml

Viech:
We're currently in a state where we prefer not to break compat. Can you rework the PR as outlined, @maek?

daemon/src/engine/qcommon/qcommon.h
```diff
@@ -308,7 +308,6 @@ enum svc_ops_e
   svc_serverCommand, // [string] to be executed by client game module
   svc_download, // [short] size [size bytes]
   svc_snapshot,
-  svc_voip, // not wrapped in USE_VOIP, so this value is reserved.
```
wtfbbqhax: Instead of removing it, rename it to `svc_unused` and comment appropriately.

illwieckz:
Hi @maek, since the engine was extracted, can you replay this PR (reworked as outlined) on https://github.com/DaemonEngine/Daemon/ ? :smiley:

illwieckz:
Note that α51 will break compatibility with older clients due to heavy VFS change, and protocol number will be bumped, so if there is some compat-breaking changes to do, it's the very right moment to do them, before α51 or never.

illwieckz:
@maek in case of compat-breaking changes, target the v0.51.0 Dæmon's branch for your PR.

DolceTriade:
These files have moved into daemon. I cherry picked your changes and put them in the for-0.51.0 branch.

PR #974, opened 2016-12-14
cg_players: Refactor handling of delta animations.

DolceTriade:
Create the concept of a SkeletonModifier that can perform arbitrary
modifications of a skeleton. Use this to allow handling hand deltas
for human models.

### DIFF
--- a/src/cgame/cg_animdelta.cpp
+++ b/src/cgame/cg_animdelta.cpp
@@ -1,0 +1,119 @@
+/*
+===========================================================================
+Copyright (C) 1999-2005 Id Software, Inc.
+Copyright (C) 2000-2009 Darklegion Development
+
+This file is part of Daemon.
+
+Daemon is free software; you can redistribute it
+and/or modify it under the terms of the GNU General Public License as
+published by the Free Software Foundation; either version 2 of the License,
+or (at your option) any later version.
+
+Daemon is distributed in the hope that it will be
+useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with Daemon; if not, write to the Free Software
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+===========================================================================
+*/
+
+#include "cg_animdelta.h"
+
+namespace
+{
+
+int LoadDeltaAnimation( weapon_t weapon, const char* modelName, bool iqm )
+{
+	int handle = 0;
+	if ( iqm )
+	{
+		handle = trap_R_RegisterAnimation( va( "models/players/%s/%s.iqm:%s_delta", modelName, modelName, BG_Weapon( weapon )->name ) );
+	}
+	else
+	{
+		handle = trap_R_RegisterAnimation( va( "models/players/%s/%s_delta.md5anim", modelName, BG_Weapon( weapon )->name ) );
+	}
+	return handle;
+}
+
+}  // namespace
+
+bool AnimDelta::ParseConfiguration(clientInfo_t* ci, const char* token2, const char** data_p)
+{
+	if ( Q_stricmp( token2, "handBones" ) ) return false;
+	char* token = COM_Parse2( data_p );
+	if ( !token || token[0] != '{' )
+	{
+		Log::Notice( "^1ERROR^7: Expected '{' but found '%s' in %s's character.cfg", token, ci->modelName );
+		return false;
+	}
+	while ( 1 )
+	{
+		token = COM_Parse( data_p );
+		if ( !token || token[0] == '}' ) break;
+		int index = trap_R_BoneIndex( ci->bodyModel, token );
+		if ( index < 0 )
+		{
+			Log::Warn("AnimDelta: Error finding bone '%s' in %s", token, ci->modelName );
+		}
+		boneIndicies_.push_back( index );
+	}
+	return true;
+}
+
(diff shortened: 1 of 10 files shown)


src/cgame/cg_animdelta.cpp
```diff
@@ -0,0 +1,115 @@
...
+		{
+			VectorSubtract( delta.bones[ boneIndicies_[ j ] ].t.trans, base.bones[ boneIndicies_[ j ] ].t.trans, weaponDeltas[ j ].delta );
+			QuatInverse( base.bones[ boneIndicies_[ j ] ].t.rot );
+			QuatMultiply( base.bones[ boneIndicies_[ j ] ].t.rot, delta.bones[ boneIndicies_[ j ] ].t.rot, weaponDeltas[ j ].rot );
+		}
+	}
+	return true;
+}
+
+bool AnimDelta::Apply(const entityState_t* es, refSkeleton_t* skeleton)
```
cmf028: This returns "true" in all possible paths. This makes the return value essentially useless

src/cgame/cg_animdelta.cpp
```diff
@@ -0,0 +1,115 @@
...
+		Q_strncpyz( newModelName, "human_nobsuit_common", sizeof( newModelName ) );
+	}
+	else
+	{
+		Q_strncpyz( newModelName, ci->modelName, sizeof( newModelName ) );
+	}
+
+	refSkeleton_t base;
+	refSkeleton_t delta;
+	for ( int i = WP_NONE; i < WP_NUM_WEAPONS; ++i )
```
cmf028: Do we really need to load an animation for WP_NONE?

src/cgame/cg_animdelta.cpp
DolceTriade: True. I thought I might need to error out, but turns out I didn't. Made void.

src/cgame/cg_animdelta.cpp
```diff
@@ -0,0 +1,115 @@
...
+	{
+		int handle = LoadDeltaAnimation( static_cast<weapon_t>( i ), newModelName, ci->iqm );
+		if ( !handle ) continue;
+		Log::Notice("Loaded delta for %s %s", newModelName, BG_Weapon( i )->humanName);
+		trap_R_BuildSkeleton( &delta, handle, 1, 1, 0, false );
+		// Derive the delta from the base stand animation.
+		trap_R_BuildSkeleton( &base, ci->animations[ TORSO_STAND ].handle, 1, 1, 0, false );
+		auto& weaponDeltas = deltas_[ i ];
```
cmf028: prefer .insert with std::move()

DolceTriade: You mean deltas_.insert(std::make_pair(i, std::move(std::vector<delta_t>);?

src/cgame/cg_animdelta.cpp
```diff
@@ -0,0 +1,115 @@
...
+		weaponDeltas.resize( boneIndicies_.size() );
```
cmf028: If the above is changed to .insert, this can be changed to just use the count parameter of the std::vector constructor to avoid an allocation.

src/cgame/cg_skeleton_modifier.h
```diff
@@ -0,0 +1,36 @@
...
+
+===========================================================================
+*/
+#ifndef CG_SKELETON_MODIFIER_H
+#define CG_SKELETON_MODIFIER_H
+
+// Forward declare due to circular dependency.
+struct clientInfo_t;
+
+class SkeletonModifier
```
DolceTriade: The motivation behind this abstraction is that there are several types of modifiers we apply to the skeleton. Some legit and some hacks. I'd like to turn all of these into modifiers so that we don't have ugly if statements littered around in our main code path. For instance, I want to make a modifier for when we combine the leg skeleton and the torso skeleton and some modifiers for our general hacks like how we prevent the taunt animation from clipping through the body and how we limit the pitch angle of the player to prevent the player arms from clipping through the model.

src/cgame/cg_local.h
```diff
@@ -909,16 +913,14 @@ typedef struct
...
-	int         handBones[ MAX_BONES ];
-	int         numHandBones;
-
 	sfxHandle_t customFootsteps[ 4 ];
 	sfxHandle_t customMetalFootsteps[ 4 ];
 
 	char        voice[ MAX_VOICE_NAME_LEN ];
 	int         voiceTime;
-} clientInfo_t;
+	std::vector<std::shared_ptr<SkeletonModifier>> modifiers;
```
DolceTriade: We need to store pointers here so that we can share one copy of a modifier among all the different clientInfos. There is one per player model and they are generally shared.

src/cgame/cg_animdelta.cpp
DolceTriade: Ah, clever. Done.

src/cgame/cg_animdelta.cpp
```diff
@@ -0,0 +1,115 @@
...
+	if ( !token || token[0] != '{' )
+	{
+		Log::Notice( "^1ERROR^7: Expected '{' but found '%s' in %s's character.cfg", token, ci->modelName );
+		return false;
+	}
+	while ( 1 )
+	{
+		token = COM_Parse( data_p );
+		if ( !token || token[0] == '}' ) break;
+		boneIndicies_.push_back( trap_R_BoneIndex( ci->bodyModel, token ) );
```
DolceTriade: Done.

DolceTriade:
Any more comments?

DolceTriade:
I'll take that as a no. If you have any more comments, feel free to ping me on IRC (or leave them here)

PR #975, opened 2016-12-23
Fix static initialization order bug affecting curses console

slipher:


daemon/src/engine/sys/con_curses.cpp
```diff
@@ -66,7 +66,9 @@ static bool     forceRedraw = false;
 static int      stderr_fd;
 #endif
 
-static const int LOG_LINES = ( LINES - 3 );
+static int NumLogLines() {
```
DolceTriade: inline?

DolceTriade:
Could you also expand on the bug a little?

daemon/src/engine/sys/con_curses.cpp
slipher: I'm sure it will be inlined anyway. Everyone says the inline keyword is an anachronism.

slipher:
None of the text is displayed in the curses console.

DolceTriade:
I meant in the commit message and the reason behind the initialization order bug. Like, "NUM_LINES isn't initialized in time because so and so"

slipher:
All right, I added some more descriptions.

DolceTriade:
Sweet thanks!

PR #976, opened 2016-12-27
[Review] Skeletal animation refactoring

slipher:


src/cgame/cg_segmented_skeleton.cpp
```diff
@@ -0,0 +1,121 @@
...
+	// rotate torso
+	if ( torsoControlBone >= 0 && torsoControlBone < skeleton->numBones )
+	{
+		// HACK: convert angles to bone system
+		quat_t rotation;
+		QuatFromAngles( rotation, ctx.torsoYawAngle, 0, 0 );
+		QuatMultiply2( skeleton->bones[ torsoControlBone ].t.rot, rotation );
+	}
+
+	// HACK: limit angle (avoids worst of the gun clipping through the body)
```
DolceTriade: I meant to put all of these hacks into their own "HumanBaseHacks" as they don't work for the battlesuit. (So basically everything below this comment is a hack and really only works for the base human.

PR #983, opened 2017-02-10
Make Daemon a submodule [the real one] fix #954 #677 ref #704 #771

illwieckz:
It looks like to work this time.
fix #954 #677 ref #704 #771

Side note: I removed `daemon` references in `.gitignore` from the Unvanquished repository. I guess it was a rest of some previous build system. Since CMake builds in `build/` repository there is no need to ignore files named `daemon` since they are never produced in repository root anymore. Keeping this rest can produces issues since the `daemon` engine directory shares this name. I also removed the related `daemon-tty` in `.gitignore` because it's a rest too.

PR #984, opened 2017-02-11
Bot stuff

slipher:
Add some functionality to the bot AI to avoid getting stuck so often. Plus a couple other small bugs.

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -575,6 +575,12 @@ void Cmd_Give_f( gentity_t *ent )
 		G_AddMomentumGeneric( (team_t) ent->client->pers.team, amount );
 	}
 
+	if ( Q_strnicmp( name, "bp", strlen("bp") ) == 0 )
+	{
+		float bp = trap_Argc() < 3 ? 300.0f : atof( name + strlen("bp") );
+		level.team[ent->client->pers.team].totalBudget += bp;
+	}
+
 	if ( G_Dead( ent ) || ent->client->sess.spectatorState != SPECTATOR_NOT )
 	{
 		return;
--- a/main/bots/camper.bt
+++ b/main/bots/camper.bt
@@ -1,5 +1,7 @@
 selector
-{	
+{
+	behavior unstick
+	
 	condition alertedToEnemy
 	{
 		action fight
--- a/main/bots/default.bt
+++ b/main/bots/default.bt
@@ -1,5 +1,7 @@
 selector
 {
+	behavior unstick
+	
 	sequence
 	{
 		condition alertedToEnemy
--- a/main/bots/reckless.bt
+++ b/main/bots/reckless.bt
@@ -1,5 +1,7 @@
 selector
 {
+	behavior unstick
+	
 	condition alertedToEnemy
 	{
 		action fight
--- a/main/bots/unstick.bt
+++ b/main/bots/unstick.bt
@@ -1,0 +1,37 @@
+// Subroutine to try to prevent bots from getting stuck
+
+condition(stuckTime > 10000)
+{
+	selector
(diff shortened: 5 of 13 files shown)


illwieckz:
From what I see they all managed to escape from there:

bot trap

but, it looks like they fall in another issue:

bot trap
bot trap

illwieckz:
interesting bot stuck case:

bot trap

illwieckz:
Perhaps it's related to another bug (we already seen similar things in the past), but it looks like these bots are trying to _not walk on navmeshes_.

bot trap

illwieckz:
Something that can be good to keep in mind is blocks can have navmeshes in them:
bot trap
bot trap

I don't know if it can lead to some bugs, it looks like these navmeshes are too fare from outside to “grab” a bot, by the way.

illwieckz:
Well, I'm not able to reproduce on other maps like plat23 for example, looks to be an another and unrelated bug.

illwieckz:
That other bug is related to the fact the eggs are placed in disconnected navmeshes. Definitely another bug. So everything looks OK to me in this PR.

illwieckz:
Example in some maps:

bot escape

Well this one was very unlucky, but found a way.
bot escape

This one can't escape with current code (it has too much place to move), these are the 10%:
bot escape
By the way he felt while evolving to dragoon.

illwieckz:
As noted on the IRC channel, the current "give bp" implementation can overflow, it must be fixed before merging.

And it's time to ask a question, it would be the time to give to "build points" a proper name, since they are now mined as a resource, it's more than a "point", it's part of the unvanquished universe.

I remember the other "give" cheat available: ammo, fuel, funds, health, momentum, poison, stamina. So, how can be named build ressources?

slipher:
Sometimes the problems are caused by navmesh defects. Here is one where you can a polygon going through a wall:
image

slipher:
Regarding `give bp`, I see using cheats as similar to "voiding your warranty" for the duration of the game — it's the user's fault if they break it with overflow. Besides, none of the other code that adds BP checks for overflow, so it seems incongruous to check it in only this one place.

illwieckz:
> Besides, none of the other code that adds BP checks for overflow.

Ok, so it means it's another issue elsewhere and is probably not preventing the merge, see #987 for the dedicated thread about that specific bug which is not related to stuck bots.

illwieckz:
For some reason, I prefer to merge the engine split first even if this PR does not modify any engine files, because this way the last commit before the split commit on this repository is also the last commit before the split commit in the other repository. It will help future historiographer work.

As wrote there:
https://github.com/Unvanquished/Unvanquished/pull/983#issuecomment-279776063
I will merge the split in two days. I guess we will be able to merge the present PR right after if no one reviewer as something to say against merging that bot stuff.

src/sgame/sg_cmds.cpp
```diff
@@ -575,6 +575,12 @@ void Cmd_Give_f( gentity_t *ent )
 		G_AddMomentumGeneric( (team_t) ent->client->pers.team, amount );
 	}
 
+	if ( Q_strnicmp( name, "bp", strlen("bp") ) == 0 )
+	{
+		float bp = trap_Argc() < 3 ? 300.0f : atof( name + strlen("bp") );
```
DolceTriade: Why not just reuse amount?

DolceTriade:
Looks good. i don't feel strongly about the amount vs bp change.

src/sgame/sg_cmds.cpp
slipher: As I understand, more narrowly scoped variables are usually preferable. `amount` is only there at the top because of C89.

illwieckz:
so, is this can be merged?

PR #990, opened 2017-03-11
Avoid overwriting submodules during builds

mbasaglia:
Note that there is similar code in Daemon, fixed in DaemonEngine/Daemon/pull/8

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,20 +34,6 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
     message(FATAL_ERROR "Building in the source directory is not supported, try building in a 'build' directory. You may have to delete the CMakeCache.txt file and CMakeFiles directory that are next to the CMakeLists.txt.")
 endif()
 
-# Check if we need to initialize submodules
-file(GLOB DENG_RESULT ${CMAKE_CURRENT_SOURCE_DIR}/daemon)
-list(LENGTH ${DENG_RESULT} DENG_RES_LEN)
-if(DENG_RES_LEN EQUAL 0)
-  if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git")
-    find_package(Git REQUIRED)
-    if(GIT_FOUND)
-      execute_process(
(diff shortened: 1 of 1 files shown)


DolceTriade:
I think we need this here to initialize the repo (daemon won't be checked out yet...). Instead, we should figure out why the check is failing. According to the code, it should only initialize the repo if the daemon dir is empty, otherwise, it should do nothing...

mbasaglia:
To be honest I don't think it should be part of the build system at all, it should be done when you check it out and that's it

DolceTriade:
I think that's fair, however, I think it's much easier for developers to get started if stuff magically works for them. We've taken effort to have CMake initialize people's build environment accordingly (ie, with the build_deps tarballs). I think we should maintain this automatic initialization as long as it only happens once and never again.

Viech:
Could also add "Clone instructions" between "Dependencies" and "Build Instructions" in `README.md` (and update the other sections, if necessary). But I generally agree with @DolceTriade that such magic is nice if it's error safe.

Viech:
On this branch, my `daemon` folder is still subject to a forced checkout. See DaemonEngine/Daemon/pull/7. Am I missing something here?

illwieckz:
As @gimhael noticed last month on dev channel, doing:

```
[submodule "daemon"]
	path = daemon
	url = https://github.com/DaemonEngine/Daemon.git
	update = none
```

works too.

I don't know if it's a feature or an ugly workaround by the way.

mbasaglia:
Looks like an ugly workaround to me

DolceTriade:
I am fine with removing submodule init from CMake, or fixing submodule init in CMake. THis leaves everyone else as a tie breaker.

slipher:
This is only for the case where the user just downloaded the repo and forgot to clone --recursive or submodule init, right? If it annoys some people why don't we print a warning telling what command to do instead.

DolceTriade:
I haven't seen this issue. If no one else reports seeing it, I'ma close this PR.

Viech:
Reproduction steps were checking out another commit inside `daemon` and re-running cmake? If so, then I also cannot reproduce.

illwieckz:
Yes that was the reproduction steps.

PR #991, opened 2017-03-11
cmake: Fix the Glob command so that it actually globs the directory

DolceTriade:
Since the syntax was incorrect, the glob wasn't returning the contents
of the directory. Therefore, git submodule update --init --recursive
was being run each time, resetting the submodules.


Viech:
Should the same be done to daemon, instead of the related commit that sneaked into DaemonEngine/Daemon#8?

DolceTriade:
I don't think so. I think the decision should be left to the parent project. Having it in one location is sufficient.

DolceTriade:
Can I merge?

Viech:
It does not appear to work for me. `cmake` still checks out `heads/master` in the `daemon` submodule:

    ~/src/Unvanquished/build (git)-[fix_submodule_checkout] % LANG=C cmake ..        
    -- git reported v0.50.0-167-gdc5d0ed63+dirty
    -- client
    -- server
    Submodule path 'daemon': checked out 'b6e41f7baae75adbd7121e70d7b73b7e19ec4b6a'
    -- Configuring done
    -- Generating done
    -- Build files have been written to: […]/src/Unvanquished/build

mbasaglia:
Is this "feature" really worth the effort?

Viech:
I'm also not sure if this feature should exist at all. I'd be just as fine with a short readme section on "Getting the source"/"Cloning the repository" that explains the submodule related git commands.

DolceTriade:
Fair enough. It does work for me, but I'd have to debug further on your setup.. We can instead update the readme and tell people to init submodules before building.

DolceTriade:
I'm going to go ahead and merge this since we currently do have this feature anyways. We can always decide to remove it later.

illwieckz:
hu, it does not work at all on my end, and since the "empty dir" detection fails, the submodule is not fetched:

```sh
CMake Error at CMakeLists.txt:52 (add_subdirectory):
  The source directory

    /home/illwieckz/dev/Unvanquished/daemon

  does not contain a CMakeLists.txt file.
```

PR #994, opened 2017-03-28
add g_neverEnd cvar to never end games on missing spawn

illwieckz:
- allows to load maps that miss spawn entities,
  this can be helpful to:
  - load unported maps from third-party games to examine them
  - edit layout in-game without needing a default one
  - load test maps that are not meant to be played
- needs cheats to be enabled, so it only works on devmap games

### DIFF
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -190,6 +190,8 @@ vmCvar_t           g_instantBuilding;
 
 vmCvar_t           g_emptyTeamsSkipMapTime;
 
+vmCvar_t           g_neverEnd;
+
 // <bot stuff>
 
 // bot buy cvars
@@ -340,6 +342,7 @@ static cvarTable_t gameCvarTable[] =
 	{ &g_debugFire,                   "g_debugFire",                   "0",                                0,                                               0, false    , nullptr       },
 
 	// gameplay: basic
+	{ &g_neverEnd,                    "g_neverEnd",                    "0",                                CVAR_SERVERINFO,                                 0, true     , nullptr       },
 	{ &g_timelimit,                   "timelimit",                     "45",                               CVAR_SERVERINFO,                                 0, true     , nullptr       },
 	{ &g_friendlyFire,                "g_friendlyFire",                "1",                                CVAR_SERVERINFO,                                 0, true     , nullptr       },
 	{ &g_friendlyBuildableFire,       "g_friendlyBuildableFire",       "1",                                CVAR_SERVERINFO,                                 0, true     , nullptr       },
@@ -2273,6 +2276,10 @@ can see the last frag.
 */
 void CheckExitRules()
 {
+	if ( g_cheats.integer && g_neverEnd.integer ) {
+		return;
+	}
+
 	// if at the intermission, wait for all non-bots to
 	// signal ready, then go to next level
 	if ( level.intermissiontime )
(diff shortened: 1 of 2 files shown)


illwieckz:
Example to load the `railgun` map from WolfET:

```
mkdir ~/.unvanquished/pkg/map-railgun_2.60b.pk3dir
./daemon -pakpath ~/.etlegacy/etmain -set fs_legacypaks on +set g_neverEnd 1 +devmap railgun
```

illwieckz:
note: to edit layout in-game while there is no one missing layout, more development is needed to spawn without spawn entities, perhaps add a cheat command named `spawn` that just makes you spawning where you are spectating?

src/sgame/sg_main.cpp
```diff
@@ -2273,6 +2276,10 @@ can see the last frag.
 */
 void CheckExitRules()
 {
```
Kangz: Use Cvar::Cvar<bool> g_neverEnd(<stuff>); so you don't have to put it in the array nor declare it in sg_extern.h

mbasaglia: Would that work? I thought the same thing bug then I couldn't find any use of the Cvar class in sgame. If they do work, maybe it's worth stating so in the comment above the definition of vmCvar_t because it isn't very clear.

Kangz: It should work, see https://github.com/DaemonEngine/Daemon/blob/master/src/shared/CommonProxies.cpp that actually implements vmCvar_t on top of Cvar::Cvar

Kangz: In particular this comment: https://github.com/DaemonEngine/Daemon/blob/master/src/shared/CommonProxies.cpp#L370

illwieckz: In fact it very hard to figure out how it works because everything looks implicit on that part, even that array is undocumented, if there is a better way to do, why not, I just need to know how to do it.

illwieckz: Just a side note, is it possible to comment under the line instead of above it? That's what GitHub expects and it always display previous line and drop following lines in comment thread so the related code is not displayed when a comment is set above the line.

illwieckz:
@Kangz get a look at my new commit that uses `Cvar::Cvar`. :wink: 

It looks like to be the very first time this mechanism is used in game code, all other occurrences I found where on engine side. Doing `grep Cvar::Cvar src/` was printing nothing.

src/sgame/sg_main.cpp
```diff
@@ -190,6 +190,8 @@ vmCvar_t           g_instantBuilding;
 
 vmCvar_t           g_emptyTeamsSkipMapTime;
 
+Cvar::Cvar<bool>   g_neverEnd("g_neverEnd", "cheat to never end a game, helpful to load a map without spawn for testing purpose", Cvar::NONE, false);
+
```
Kangz: FYI: You could also place the declaration just above the function where it is used. No need to change it though

Kangz:
I think it is used in some of daemon's shared gamelogic code but not in the non-Daemon gamelogic code. Thanks for the iteration, LGTM

illwieckz:
Do I have to squash the commits? By the way it's a good example to show how to convert an old cvar to the new format.

Kangz:
I like a squashed history for fixups, but it isn't a strong opinion and I don't think we settled on a specific option yet.

illwieckz:
I was clearly in need for such kind of commits to understand how to implement it. Sometime it's useful to see how people do step by step.

PR #1001, opened 2017-04-23
[cmake] make DAEMON_DIR an option, fix #1000

### LINKED ISSUE #1000: build against a Dæmon directory that is outside of source tree
Having to deal with Dæmon as a submodule is annoying, what could be far more convenient would be to just clone Unvanquished and Daemon this way:

```sh
cd ~/src
git clone https://github.com/Unvanquished/Unvanquished.git
git clone https://github.com/DaemonEngine/Daemon.git
```

then build Dæmon and Unvanquished this way:

```sh
cd Daemon
mkdir build; cd build
cmake ..
make
```

```sh
cd Unvanquished
mkdir build; cd build
cmake -DDAEMON_PATH=~/src/Daemon ..
make
```

We are already able to run the game that way:

```sh
~/src/Daemon/build/daemon -libpath ~/src/Unvanquished/build
```

So it would make sense to have the convenience to build them that way, and nobody would have to complain about that annoying submodule thing. By the way the `daemon` submodule can be kept for people who just want to build the game without having to hack the code, fetching the submodule if `DAEMON_PATH` cmake option is not set.

Feel free to suggest a better cmake option name if you think having a better idea for it of if there is some best practice to follow. :wink:

illwieckz:
As explained in #1000, this change makes user able to specify an arbitrary path for Dæmon directory.

On a game code point of view, the Dæmon source is just a kind of library. If this directory is not set, the `daemon/` submodule is used.

On an engine code point of view, the engine is standalone and being constrained by a game code tree is painful. Being constrained to make engine development in game vm source tree is painful.

The `daemon` submodule is handy for game vm developers who never touch the engine or for users who just want to build the game, but that submodule makes engine development very painful. So this option is proposed to make engine developers life easier, to allow developers to not caring about gamecode submodule everytime they hack on engine.

See [how I build](https://github.com/illwieckz/UnvBuildEnv/blob/master/Makefile#L36-L52) all unvanquished stuff (engine, gamevm, assets) with a single `make`, and [how I run](https://github.com/illwieckz/UnvBuildEnv/blob/master/Makefile#L60-L70) the game.

Basically I'm just building like that:

```
[user@box:src/Daemon]
$ cmake -H. -Bbuild && cmake --build build
[user@box:src/Unvanquished]
$ cmake -H. -Bbuild -DDAEMON_DIR=$(realpath ../Daemon) && cmake --build build
[user@box:src/UnvanquishedAssets]
$ make build
```

And I run the game like that:

```
[user@box:src]
$ Daemon/build/daemon -libpath Unvanquished/build -pakpath UnvanquishedAssets/build/test
```

That's very handy, especially when dealing with engine stuff everyday.

It would also make life easier to get hands on multiple Dæmon based games at the same time, but it was thought first for people hacking engine stuff on daily schedule, allowing engine developers to get rid of the _please develop engine stuff in game vm source tree_ requirement and annoyance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,22 +34,36 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
     message(FATAL_ERROR "Building in the source directory is not supported, try building in a 'build' directory. You may have to delete the CMakeCache.txt file and CMakeFiles directory that are next to the CMakeLists.txt.")
 endif()
 
-# Check if we need to initialize submodules
-file(GLOB DENG_RESULT ${CMAKE_CURRENT_SOURCE_DIR}/daemon)
-list(LENGTH ${DENG_RESULT} DENG_RES_LEN)
-if(DENG_RES_LEN EQUAL 0)
-  if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git")
-    find_package(Git REQUIRED)
-    if(GIT_FOUND)
-      execute_process(
-        COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
-        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
-    endif()
-  endif()
+set(SUBMODULE_LIST
+  "libs/libRocket"
+  "src/utils/cbse")
+
+set(DAEMON_DIR_STRING "Path to the Daemon Engine source.")
+if(NOT DAEMON_DIR)
+  set(DAEMON_DIR "${CMAKE_CURRENT_SOURCE_DIR}/daemon" CACHE STRING "${DAEMON_DIR_STRING}")
+  LIST(APPEND SUBMODULE_LIST "daemon")
+else()
+  set(DAEMON_DIR "${DAEMON_DIR}" CACHE STRING "${DAEMON_DIR_STRING}")
 endif()
 
+foreach(SUBMODULE_PATH ${SUBMODULE_LIST})
+  # Check if we need to initialize submodule
+  file(GLOB DENG_RESULT "${CMAKE_CURRENT_SOURCE_DIR}/${SUBMODULE_PATH}")
+  list(LENGTH ${DENG_RESULT} DENG_RES_LEN)
+  if(DENG_RES_LEN EQUAL 0)
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git")
+      find_package(Git REQUIRED)
+      if(GIT_FOUND)
+        execute_process(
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -34,22 +34,25 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
...
-file(GLOB DENG_RESULT ${CMAKE_CURRENT_SOURCE_DIR}/daemon)
-list(LENGTH ${DENG_RESULT} DENG_RES_LEN)
-if(DENG_RES_LEN EQUAL 0)
-  if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git")
-    find_package(Git REQUIRED)
-    if(GIT_FOUND)
-      execute_process(
-        COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
-        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+if (NOT DAEMON_DIR)
```
DolceTriade: Undo the if statement and maybe try:
set(DAEMON_DIR "${CMAKE_CURRENT_SOURCE_DIR}/daemon" CACHE STRING "Path to the Daemon Engine source.")

illwieckz: oh, nice.

but how can I do the magic submodule update when it fallbacks on default?

DolceTriade: Should just work

CMakeLists.txt
```diff
@@ -34,22 +34,25 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
...
+  set (DAEMON_DIR "daemon")
+  # Check if we need to initialize submodules
+  file(GLOB DENG_RESULT ${CMAKE_CURRENT_SOURCE_DIR}/daemon)
```
DolceTriade: CHange this to ${DAEMON_DIR}

illwieckz:
I made some change, what do you think about them?

I also noticed that non-daemon submodules (cbse and librocket) were dependent on daemon one being empty to be pulled, it means a non-empty daemon submodule with empty cbse submodule or empty librocket submodule was making cmake fail. I fixed that: daemon submodule is pulled if empty and DAEMON_DIR unset, and cbse and librocket submodules are pulled if empty, whatever the state of other ones. So this PR not only add daemon build convenience but also fixes existing issues for other submodules.

illwieckz:
I've simplified the code, each submodule is in a submodule list and if `DAEMON_DIR` is not set, `daemon` is added in submodule list, and a simple `foreach` fetches all submodules in this list.

Kangz:
The foreach approach LGTM.

DolceTriade:
Me as well. Fell free to merge.

illwieckz:
OK, so let's push that button !

PR #935, opened 2016-04-16
Fix a bunch of PVS-Studio static analyser warnings

dimhotepus:
A list of fixes from PVS-Studio static analyser tool.

PR #1013, opened 2017-07-11
Fix a bunch of PVS-Studio static analyser warnings №2

illwieckz:
This a PR replaying #935 by @dimhotepus on Unvanquished source tree after the engine split.

It obsoletes #935.

This is all @dimhotepus' work.


Kangz:
LGTM

PR #965, opened 2016-10-10
Fixes from clang code inspector

dimhotepus:

PR #1014, opened 2017-07-23
Fixes from clang code inspector №2

illwieckz:
This a PR replaying #965 by @dimhotepus on current master.

It obsoletes #965.

This PR is smaller because some of the fixes were already done in #1013, also some code were just deleted in our master branch, and obviously, deleted code does not have to be fixed.

This is all @dimhotepus' work.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -1184,7 +1184,7 @@ static bool admin_match( void *admin, const void *match )
 	}
 
 	G_SanitiseString( ( ( g_admin_admin_t * ) admin )->name, n1, sizeof( n1 ) );
-	return !!strstr( n1, n2 );
+	return strstr( n1, n2 ) != nullptr;
 }
 
 static int admin_out( void *admin, char *str )
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -1184,7 +1184,7 @@ static bool admin_match( void *admin, const void *match )
 	}
 
 	G_SanitiseString( ( ( g_admin_admin_t * ) admin )->name, n1, sizeof( n1 ) );
-	return strstr( n1, n2 ) ? true : false;
+	return !!strstr( n1, n2 );
 }
 
 static int admin_out( void *admin, char *str )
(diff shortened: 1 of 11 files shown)


src/sgame/sg_admin.cpp
```diff
@@ -3245,9 +3245,10 @@ bool G_admin_warn( gentity_t *ent )
 	if( ( found = G_ClientNumbersFromString( name, pids, MAX_CLIENTS ) ) != 1 )
 	{
 		count = G_MatchOnePlayer( pids, found, err, sizeof( err ) );
-		ADMP( va( "%s %s %s %s %s", QQ( "^3$1$: ^7$2t$\n $3$$4$" ), "warn", ( count > 0 ) ?  Quote( Substring( err, 0, count ) ) : Quote( err ),
-		                                      ( count > 0 ) ? Quote( Substring( err, count + 1, strlen( err ) ) ) : "",
-				                              ( count > 0 ) ? "" : "" ) );
+		ADMP( va( "%s %s %s",
+				  QQ( "^3$1$: ^7$2t$" ),
+				  "warn",
+				  ( count > 0 ) ? Quote( Substring( err, 0, count ) ) : Quote( err ) ) );
```
illwieckz: reminder from the [old thread](https://github.com/Unvanquished/Unvanquished/pull/965/files#r84266908):

@mbasaglia on 20 Oct 2016 Member
> I'm not familiar with Substring(), is the new code really equivalent? I'd expect the old one to skip err[count]

@Kangz 13 days ago Member
> Yeah this one looks fishy, however this line is full of undefined behavior because Substring returns a pointer to an internal static buffer. I'd still be fine with this change even if we don't know if it is semantically equivalent, because it fixes an UB in rarely used code.

src/sgame/sg_admin.cpp
```diff
@@ -1184,7 +1184,7 @@ static bool admin_match( void *admin, const void *match )
 	}
 
 	G_SanitiseString( ( ( g_admin_admin_t * ) admin )->name, n1, sizeof( n1 ) );
-	return strstr( n1, n2 ) ? true : false;
+	return !!strstr( n1, n2 );
```
illwieckz: reminder from the [old thread](https://github.com/Unvanquished/Unvanquished/pull/965/files#r84266517):

@mbasaglia on 20 Oct 2016 Member
> TBH I have some mixed feelings about bang bang

@Kangz 13 days ago Member
> Same here.

illwieckz: The fact is, in original code:

```c
return strstr( n1, n2 ) ? true : false;
```

`strstr( n1, n2 )` is not a boolean, so it's an issue for the `?` test even if it's not an issue for the `return`

I guess the best way to fix it would be to do:

```c
return strstr( n1, n2 ) != NULL;
```

What do you think about it?

DolceTriade: SGTM

illwieckz: see newer commit

Kangz:
LGTM

PR #964, opened 2016-10-10
Correct listmap cmd filtering & output

dimhotepus:
- Fix map filtering for filter containing 'map-' chars

When `listmaps` filter contains one of "map-" chars, filter returns all
map names. The reason is we take into account invisible "map-" prefix
for map when filter. Fixed by removing "map-" from map name before filter.
- Fix output maps names in columns & asc map names sorting. 

Before fix, maps names were unordered and not outputted in columns, as
it were before. Fixed by ensuring ADMBP( "" ) send newline instead of
do nothing. Was broken when forced end buffer with new line implicitly.

PR #1019, opened 2017-09-10
introduce MAX_MAPLIST_COLS

illwieckz:
After #964, I thought it would be useful to also set a `MAX_MAPLIST_COLS` constant to make code easier to read and modify. Also, one day we can imagine to substitute this constant and the ROWS one by variables relative to the screen size or things like that.

So currently it's just another constant in code so we know what is what. By the way, I have a question: what's the purpose of the other hardcoded value set to `2`? Is it `MAX_MAPLIST_COLS - 1`? Is it always `2`? And Why?

Of course if that `2` is relative to `MAX_MAPLIST_COLS` please not merge this change before engraving this relation in code!

_Edit: it is! It's effectively `MAX_MAPLIST_COLS - 1`, updated the commit accordingly._

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -3788,6 +3788,8 @@ List all maps on the server
 
 #define MAX_MAPLIST_MAPS 256
 #define MAX_MAPLIST_ROWS 9
+#define MAX_MAPLIST_COLS 3
+
 void Cmd_ListMaps_f( gentity_t *ent )
 {
 	char search[ 16 ] = { "" };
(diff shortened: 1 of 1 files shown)


DolceTriade:
LGTM. It's not clear what the extra +2 is for. If I had to guess, it seems it's for formatting purposes, as if they were printing two additional maps, but they aren't printing anything else. Does deleting it make stuff look funky?

illwieckz:
Yeah, it's `MAX_MAPLIST_COLS - 1`. It's to be sure to fill an extra row when the number of map is not a multiple of `MAX_MAPLIST_COLS` and `MAX_MAPLIST_ROWS`, and `1` being the minimal number of map to trigger the filling (because `1 extra map + MAX_MAPLIST_COLS - 1 = MAX_MAPLIST_COLS` so `N extra map + MAX_MAPLIST_COLS - 1 > MAX_MAPLIST_COLS`, the `max( 1, ( rows + MAX_MAPLIST_ROWS - 1 ) / MAX_MAPLIST_ROWS )` code is doing that `>` test implicitely.

Without that, any line with less columns than `MAX_MAPLIST_COLS` will no be counted and displayed.

With 33 maps with 2 rows per page and 4 columns per row, removing that "extra +2" shows this:

```
]/listmaps 4
 test                 transit              uncreation           vega                
 thunder              tremor               usstremor            western             
 
listmaps: listing 8 of 33 maps 
listmaps: page 4 of 4 
```

4 page means 32 maps, so the 5th page for the remaining map (here, yocto) is missing and can't be asked for display.

With 33 maps with 3 rows per page and 4 columns per row, using `MAX_MAPLIST_COLS - 1` instead of `2` shows this:

```
]/listmaps 4
 test                 transit              uncreation           vega                
 thunder              tremor               usstremor            western             
 
listmaps: listing 8 of 33 maps 
listmaps: page 4 of 5; use 'listmaps 5' to see more 
]/listmaps 5
 yocto               
 
listmaps: listing 1 of 33 maps 
listmaps: page 5 of 5 
```

illwieckz:
Updated PR.

DolceTriade:
Cool, nice explanation.

src/sgame/sg_cmds.cpp
```diff
@@ -3853,19 +3854,19 @@ void Cmd_ListMaps_f( gentity_t *ent )
 			   }
 	);
 
-	int rows = ( mapNamesCount + 2 ) / 3;
+	int rows = ( mapNamesCount + MAX_MAPLIST_COLS - 1 ) / MAX_MAPLIST_COLS;
```
DolceTriade: Please add a comment here about why the MAX_MAPLIST_COLS - 1 is necessary.

illwieckz: done

PR #1025, opened 2017-10-03
Fix -Wimplicit-fallthrough warnings

DolceTriade:


DolceTriade:
Actually, this depends on the requsite daemon PR.

src/cgame/cg_weapons.cpp
```diff
@@ -757,10 +757,12 @@ static bool CG_ParseWeaponFile( const char *filename, int weapon, weaponInfo_t *
 				switch( weapon )
 				{
 					case WP_MACHINEGUN:
+						DAEMON_FALLTHROUGH;
```
Kangz: Same comments, multiple cases together shouldn't need FALLTHROUGH

PR #1026, opened 2017-10-08
Don't call every think component as many times as there are entities.

slipher:


src/sgame/sg_main.cpp
```diff
@@ -2858,6 +2858,11 @@ void G_RunFrame( int levelTime )
 		}
 	}
 
+	// ThinkingComponent should have been called already but who knows maybe we forgot some.
```
DolceTriade: Instead of this, do you think we should include a  frame counter and make sure the think counter matches the frame counter? Throw an assert instead or a log telling people that they should report this. I did this one time, and I got a bunch of bugreport. Otherwise, should prolly be ok since we shouldn't be thinking if enough time hasn't passed right?

Kangz: I think the intention with the ForEntities above was to call all the new-style thinkers before G_RunThink, so around line 2814.

slipher: That would make sense. I was afraid of breaking something by changing the order of execution, but I guess I'll go ahead and try it.

slipher: Never mind, actually I don't think it makes sense. Look at the stuff about "free before thinking" and "free after thinking" around line 2600.

slipher:
Added logging if the think component isn't called the expected number of times.

src/sgame/components/SpawnerComponent.cpp
```diff
@@ -46,6 +49,14 @@ void SpawnerComponent::Think(int timeDelta) {
 	Entity* blocker = GetBlocker();
 
 	if (blocker) {
+		if (!blocker->oldEnt) {
```
DolceTriade: We should make macros for this.

LOG_EVERY_INTERVAL()

and have another variant

LOG_EVERY_N().

Having this type of thing is super ugly for logs every time we want to do it.

DolceTriade: we can defer this change till later.

illwieckz: @slipher does this mean we can merge that?

Issue #1017, opened 2017-09-07
mad unpowered turret


illwieckz:
Unpowered turret display is broken on current master, see [this video](http://dl.illwieckz.net/b/unvanquished/bugs/mad-unpowered-turrets/unvanquished_2017-09-06_225009_001.webm).

The bug was introduced by commit 10120d2.

I noticed some other similar bugs with some other models, perhaps it's the same bug.

illwieckz:
The _antares_ map is the perfect place to reproduce the bug. If you don't see any turret going mad (you must see at least one of them going mad due to current layout letting one unpowered), just deco the rc and enjoy.

illwieckz:
ping @Viech, you're probably the man who understands the best this part of the code. :wink:

illwieckz:
The very _wtf_ stuff about this is that if you power the turret (like building a drill near of it) at the exact frame when the turret is dislocated, it remains dislocated on the floor, being unable to orient itself against aliens and shooting through the floor!

So useless defenseless turret:

[![mad unpowered turret](http://dl.illwieckz.net/b/unvanquished/bugs/mad-unpowered-turrets/unvanquished_2017-09-06_230045_000.jpg)](http://dl.illwieckz.net/b/unvanquished/bugs/mad-unpowered-turrets/unvanquished_2017-09-06_230045_000.jpg)

Weird dislocated turret:

[![mad unpowered turret](http://dl.illwieckz.net/b/unvanquished/bugs/mad-unpowered-turrets/unvanquished_2017-09-06_230232_000.jpg)](http://dl.illwieckz.net/b/unvanquished/bugs/mad-unpowered-turrets/unvanquished_2017-09-06_230232_000.jpg)

``¯\_(ツ)_/¯``

illwieckz:
There is a chance that #985 being the same bug.

illwieckz:
How to repro: make a turret unpowered

illwieckz:
fixed in 831eaef and in 831eaef

PR #1029, opened 2017-10-11
Fix animation issues related to power.

DolceTriade:


src/sgame/sg_buildable.cpp
```diff
@@ -775,10 +775,13 @@ static bool CompareBuildablesForPowerSaving(Entity* a, Entity* b)
 void G_UpdateBuildablePowerStates()
 {
 	std::list<Entity*> buildables;
+	std::list<Entity*> unpoweredBuildables;
```
slipher: Why are we using doubly linked lists? I don't see anything where a vector wouldn't suffice.

Also these declarations could be moved inside the loop.

src/sgame/sg_buildable.cpp
```diff
@@ -792,39 +795,57 @@ void G_UpdateBuildablePowerStates()
...
+				if (entity->Get<HealthComponent>()->Alive()) {
+					deficit -= BG_Buildable(entity->oldEnt->s.modelindex)->buildPoints;
+				}
 
-			if (deficit <= 0) break;
+				if (deficit <= 0) break;
+			}
+		} else if (deficit < 0) {
+			// Make our deficit positive for ease of calculation.
+			deficit *= -1;
```
slipher: `int surplus = -deficit` ?

src/sgame/sg_buildable.cpp
```diff
@@ -792,39 +795,57 @@ void G_UpdateBuildablePowerStates()
...
+		int deficit = level.team[team].spentBudget - (int)level.team[team].totalBudget - unpoweredBuildableTotal;
 
-		// Do not power down more buildables if there is no deficit.
-		if (deficit <= 0) continue;
+		// Exactly at our limit. Nothing else to do.
+		if (deficit == 0) continue;
 
-		buildables.sort(CompareBuildablesForPowerSaving);
+		// We have surplus bp, but nothing else to power on, so we're done here.
+		if (deficit < 0 && unpoweredBuildables.empty()) continue;
```
slipher: I think you could remove this `if` block and the one before it.

src/sgame/sg_buildable.cpp
```diff
@@ -775,10 +775,13 @@ static bool CompareBuildablesForPowerSaving(Entity* a, Entity* b)
 void G_UpdateBuildablePowerStates()
 {
 	std::list<Entity*> buildables;
```
slipher: Rename to `poweredBuildables`?

DolceTriade: Done

src/sgame/sg_buildable.cpp
DolceTriade: idk, maybe there is no random access and everything is accessed linearly so you never take a resize hit? Moved the declarations inside the loop.

src/sgame/sg_buildable.cpp
DolceTriade: nice lol

src/sgame/sg_buildable.cpp
DolceTriade: What do you mean? I think that the explicit cases are much clearer about the code intention.

src/sgame/sg_buildable.cpp
slipher: When everything is accessed linearly that just makes the vector more preferable. Vectors are the default choice (for a small, trivially copyable object at least) since the higher amount of allocations for a linked list is more costly than copying a contiguous block of memory a few times.

src/sgame/sg_buildable.cpp
slipher: So `deficit < 0` and you want to power up as many of the N unpowered buildables as possible. I don't see anything qualitatively different about what happens when N ==0 from N != 0, so for me it is harder to read if there is a special case for N == 0.

DolceTriade: While its negligible, you'll also be unnecessarily going through a list of buildables. I think I'll leave this one as is.

src/sgame/sg_buildable.cpp
DolceTriade: Done.

illwieckz:
Note that I intensively tested the code on my server playing with real people multiple games. Even the "addressed comment" changes.

illwieckz:
@DolceTriade can you sum up what this fix was doing in a way it makes sense to user? (to help me to write the release note)

DolceTriade:
@Viech: I'm not sure. I haven't tested.

@illwieckz: This stops the turret from doing weird stuff when its unpowered (we had that weird bug where the turrets would fall on their side. (ie, fixes this issue: https://github.com/Unvanquished/Unvanquished/issues/1017)

illwieckz:
oh thanks for the explanation, I was completely forgotting that weird bug. :grin: 
well, if I'm right it looks like it was a bug introduced after 0.50.0 release so I don't have to talk about it

PR #1030, opened 2017-10-12
Licensing fixes

illwieckz:
The very important changes are the ones in COPYING.txt:

- remove non-free shaderlab exception
  there is no shaderlab's file anymore
- remove non-GPL code from Wolf:ET source tree exception
  it was basically unzip, zlib, md4, jpeg, curl, ft2, and png
  they are now external.
- add sgame/sg_api.cpp as BSD licensed exception
- do not talk redirect to CC file, it's not there anymore
  keep the url

It fixes also some headers refering to Unvanquished or Daemon inconsistently, for example:

`src/sgame/sg_spawn_mover.cpp`
> **Daemon** GPL Source Code
> Copyright (C) 2012 **Unvanquished** Developers
> 
> This file is part of the **Daemon** GPL Source Code (**Daemon** Source Code).
> 
> **Daemon** Source Code is free software: you can redistribute it and/or modify

`src/cgame/Filter.h`
> Copyright 2015 **Unvanquished** Developers
> 
> This file is part of **Daemon**.
> 
> **Unvanquished** is free software: you can redistribute it and/or modify

Btw, all Daemon references in headers were replaced by Unvanquished ones.

It also unify some header spacing and line blanking.

It adds `Unvanquished GPL Source Code' mention to GPL headers when missing, so a simple recursive grep can find the related files.

There is still some files to be fixed because they don't have any header at all, you can find them this way:

```sh
find src -type f | sort > /tmp/srcfiles
grep -r 'Unvanquished .* Source Code' src | cut -f1 -d: | sort -u > /tmp/headfiles
comm -2 -3 /tmp/srcfiles /tmp/headfiles 
```

For example almost all files in `src/sgame/components` lack proper header, what to do with them?

Some look autogenerated, it would be nice to make the generator includes a license header too, and also check if all autogenerated files all contains a statement saying it's autogenerated (some have, I'm not sure all have).

Edit: See DaemonEngine/Daemon#48 for a similar PR on Daemon side.

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -1,23 +1,25 @@
 /*
 ===========================================================================
+
 Copyright (C) 1999-2005 Id Software, Inc.
 Copyright (C) 2000-2009 Darklegion Development
 
-This file is part of Daemon.
+This file is part of the Unvanquished GPL Source Code (Unvanquished Source Code).
 
-Daemon is free software; you can redistribute it
+Unvanquished is free software; you can redistribute it
 and/or modify it under the terms of the GNU General Public License as
 published by the Free Software Foundation; either version 2 of the License,
 or (at your option) any later version.
 
-Daemon is distributed in the hope that it will be
+Unvanquished is distributed in the hope that it will be
 useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License
-along with Daemon; if not, write to the Free Software
+along with Unvanquished; if not, write to the Free Software
 Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
 ===========================================================================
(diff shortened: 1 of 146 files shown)


src/cgame/cg_buildable.cpp
```diff
@@ -2088,6 +2094,7 @@ static int CG_SortDistance( const void *a, const void *b )
 CG_PlayerIsBuilder
 ==================
 */
+
```
DolceTriade: Need to fix regex? We shouldn't be adding extra lines everywhere

illwieckz: that was just making comments consistent: in this kind of file there is random blank lines (or not) after each comment, it was half multiline comment plus space and half multiline comment without space… not very costly to unify that (i.e. I wrote the regexp to do that).

DolceTriade: We should do the opposite...remove the extra newline between comments and the start of the function.

illwieckz: well, it's not related to licensing so it can be submitted as a separate PR, I removed that commit from this PR history.

DolceTriade:
LGTM

PR #1033, opened 2017-10-16
do not hardcode bot names in ui rml, allow ui to reuse user defined botnames, fix #765

### LINKED ISSUE #765: bot add by GUI does not use defined bot name
Hi, if we add bot with _Add an alien bot_ or _Add an human bot_ menu entries, the bot created uses the default names (_Lavishment_ and _Pulsifer_), not the names defined by `bot names` commands.

But when using the `bot add` commands, the bot uses one of the user-defined bot names.

illwieckz:
- do not hardcode bot names in ui rml
- allow ui to reuse user defined botnames
- ~~set default bot names in G_BotInit()~~

### DIFF
--- a/main/ui/menu_ingame.rml
+++ b/main/ui/menu_ingame.rml
@@ -67,8 +67,8 @@
 
 							<h2> Bots </h2>
 							<indent>
-								<blocklink onClick='Events.pushevent("exec bot add Lavishment aliens", event)'> Add an alien bot </blocklink>
-								<blocklink onClick='Events.pushevent("exec bot add Pulsifer humans", event)'> Add a human bot </blocklink>
+								<blocklink onClick='Events.pushevent("exec bot add * aliens", event)'> Add an alien bot </blocklink>
+								<blocklink onClick='Events.pushevent("exec bot add * humans", event)'> Add a human bot </blocklink>
 								<blocklink onClick='Events.pushevent("exec bot del all", event)'> Remove all bots </blocklink>
 							</indent>
 
--- a/src/sgame/sg_client.cpp
+++ b/src/sgame/sg_client.cpp
@@ -707,7 +707,7 @@ static const char *G_UnnamedClientName( gclient_t *client )
 
 	client->pers.namelog->unnamedNumber = number;
 	Com_sprintf( name, sizeof( name ), "%.*s%d", (int)sizeof( name ) - 11,
-	             g_unnamedNamePrefix.string[ 0 ] ? g_unnamedNamePrefix.string : UNNAMED_PLAYER,
+	             g_unnamedNamePrefix.string[ 0 ] ? g_unnamedNamePrefix.string : UNNAMED_PLAYER"#",
 	             number );
 
 	return name;


src/sgame/sg_bot.cpp
```diff
@@ -514,6 +514,22 @@ void G_BotInit()
 	{
 		InitTreeList( &treeList );
 	}
+
+	// default bot names
```
DolceTriade: I don't think we should hard code names here.

main/ui/menu_ingame.rml
```diff
@@ -67,8 +67,8 @@
 
 							<h2> Bots </h2>
 							<indent>
-								<blocklink onClick='Events.pushevent("exec bot add Lavishment aliens", event)'> Add an alien bot </blocklink>
-								<blocklink onClick='Events.pushevent("exec bot add Pulsifer humans", event)'> Add a human bot </blocklink>
+								<blocklink onClick='Events.pushevent("exec bot add * aliens", event)'> Add an alien bot </blocklink>
```
DolceTriade: This might be overkill, but what about using lua to generate random bot names? Otherwise, just this change is sufficient to give the bots unique names.

src/sgame/sg_bot.cpp
illwieckz: Where is the best place to do that? Or is there another way (like a default config file shipped with the game)?

main/ui/menu_ingame.rml
illwieckz: We can also not define default bot names at all, server owners will do it by the way.

illwieckz:
I added a commit to fix a very rare bug when `g_unnamedNamePrefix` is empty, so the prefix used for unnamed players is the same when `g_unnamedNamePrefix` contains the default value or when it's empty (which must use the default value).

main/ui/menu_ingame.rml
illwieckz: We can also name unnamed bots as `Bot#1` etc.

illwieckz:
Well perhaps we just have to put that missing `#` where `UNNAMED_PLAYER` is defined in daemon instead of adding it to prefix anytime we need the prefix…

main/ui/menu_ingame.rml
DolceTriade: I think this is acceptable honestly. We could also make a default set.

illwieckz: btw, to answer the question, I think it would be overkill, or at least out of focus of this PR (same if we want to add the ability to set default bot prefix different than default player prefix).

src/sgame/sg_bot.cpp
DolceTriade: I don't think we have one. We don't have a script that runs by default. We could modify the code here

https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_bot.cpp#L44

to have a default list of names, and then switch between the hardcoded list and the custom one when someone calls G_AddBotNames() or something. I'm happy with just Bot#1, etc.

illwieckz:
I removed the code that was hardcoding default name. We can do that in another PR (adding a default list, generating a random one, setting a per-bot default prefix… or any other solution). Current code already works: bots use names from user-defined name lists or fallback on default prefix and padding.

src/sgame/sg_bot.cpp
illwieckz: Doing `Bot#1` means duplicating a lot of code for prefix handling and padding computation. So it's a good idea but I don't want to include this within this PR. It can be done later.

Issue #986, opened 2017-02-12
scoreboard is not removed on game start


illwieckz:
When the next map is another map, the whole engine restarts so nobody can notice the bug. But when there is no map rotation (and probably when the next map in rotation is the same), the scoreboard is not removed once the game restart. Since there is no nextmap to load, the game restart without engine restart.

slipher:
There are more cases where UI elements can persist after they have outlived their usefulness, e.g.
- Position in spawn queue after switching to spectator
- Build menu after switching teams, selling ckit or evolving
- Evolve menu after swiching teams

illwieckz:
Other persisting UI elements:

- Buildable/Weapons/Evolve forms that were locked when the menu was opened, but then unlocked while the menu was still opened (has to close menu and reopen it)
- Switch team UI persists even if team is selected on console

bmorel:
There's also #1102

bmorel:
>  scoreboard is not removed on game start

When was this fixed? I got it few days ago on sweet's server I think, and it runs vanialla IIRC

PR #1034, opened 2017-10-16
call OnMapRestart on MapRestart, ref #986

illwieckz:
It looks like OnMapRestart is expected to be called on MapRestart.


illwieckz:
Note that `CG_OnMapRestart()` was called explicitly on `restart` console command, but not when the map was restarted because there was no map rotation list or because the next map in rotation list is the same (which occurs often enough when people callvote for a nextmap wich is already the next map in rotation list).

illwieckz:
For reference, this was the fix for the issue of scoreboard not being wiped from screen on map restart.

DolceTriade:
Oh nice find!

PR #1037, opened 2017-10-24
sgame: Limit the number of visible characters in a name to 32.

DolceTriade:
Note, that the since each character can be up to 4 bytes, we make
sure the underlying buffer can hold 128 chars. This ensures that
people can't have super long names in english, but shorter names in
other languages.

### DIFF
--- a/src/sgame/sg_client.cpp
+++ b/src/sgame/sg_client.cpp
@@ -928,6 +928,13 @@ const char *ClientUserinfoChanged( int clientNum, bool forceName )
 			trap_SendServerCommand( ent - g_entities, va( "print_tr %s %s %s", QQ( "$1t$ $2$" ), Quote( err ), Quote( newname ) ) );
 			revertName = true;
 		}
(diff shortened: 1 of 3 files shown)


src/sgame/sg_client.cpp
```diff
@@ -928,6 +928,13 @@ const char *ClientUserinfoChanged( int clientNum, bool forceName )
 			trap_SendServerCommand( ent - g_entities, va( "print_tr %s %s %s", QQ( "$1t$ $2$" ), Quote( err ), Quote( newname ) ) );
 			revertName = true;
 		}
+		else if ( Q_UTF8_Strlen( newname ) > MAX_NAME_CHARACTERS )
```
Kangz: Does this take colors into account?

DolceTriade: Colors do not count against you, however, emoticons do.

Kangz: Sounds good, thanks

DolceTriade:
Merging. Travis failure seems unrelated.

PR #1038, opened 2017-10-31
remove some leftover (break compat)

illwieckz:
remove some leftover (break compat)


illwieckz:
I had these ones in my mind ;-)

PR #1048, opened 2017-11-06
Remove cg_highPolicy*Models options.

slipher:
Remove these 3 cvars which allowed some old Tremulous models to be used. Also nuke code for md3 buildables since there are none remaining.

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -633,16 +633,14 @@ void CG_InitBuildables()
 		//Load models
 		//Prefer md5 models over md3
 
-		if ( cg_highPolyBuildableModels.integer &&
-		     CG_FileExists( va( "models/buildables/%s/%s.iqm", buildableName, buildableName ) ) &&
+		if ( CG_FileExists( va( "models/buildables/%s/%s.iqm", buildableName, buildableName ) ) &&
 		     ( bi->models[ 0 ] = trap_R_RegisterModel( va( "models/buildables/%s/%s.iqm",
 		                                                   buildableName, buildableName ) ) ) )
 		{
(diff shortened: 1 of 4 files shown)


DolceTriade:
removing support for md3 animations is a little more hairy.. You also need to remove the animation.cfg code. I think we should leave in md3 animation support so we can test stuff more easily (md3 models are more readily available than iqm models). Otherwise, looks good.

slipher:
All right, I un-nuked md3.

Viech:
@DolceTriade as soon as the last MD3 model is removed from the game I want to immediately get rid of this format that lets us have roughly twice the necessary amount of model related code all over cgame. MD5 and IQM need to suffice.

DolceTriade:
Well, md3 is less of a problem than our shitty grafting of skeletal animation on the existing code. A more clever programmer than I was years ago would certainly have factored out the common logic into more functions instead of hastily copying and pasting random stuff around at 3 AM :-/

Kangz:
Which MD3 models are useful for testing? If they're only static models we should re-export them to IQM and get rid of MD3 code.

DolceTriade:
In general, if we want to add any new buildables or weapons, we'll likely not want to wait for iqm models. We can download any number of ready made q3 models and use those.

Viech:
If model format support is nicely seperated at some point (with strict interfaces of what models need to be able to do), I don't care how many formats we support. But as long as we waste a lot of time figuring out why, for instance, our missile model or impact models do not display, I'd rather have just one or two supported formats that work really well.

And I can't think of an MD3 model that we would want to include freshly in the game. For testing we can reuse any existing model until we get a proper one, as we have already done so far.

PR #1051, opened 2017-11-07
Buff and simplify the Spiker.

Viech:
It will now always shoot but retain selected spikes to diminish the risk of
hitting friends. Sense range is increased slightly.

Note that without a proper trajectory trace, it still can and will occasionally hit friendly buildables.

### DIFF
--- a/src/sgame/components/BuildableComponent.h
+++ b/src/sgame/components/BuildableComponent.h
@@ -87,7 +87,7 @@ class BuildableComponent: public BuildableComponentBase {
 		 * Protect current animation from being disrupted by selected, less important animations.
 		 * A value of 0 protects for the rest of the frame only.
 		 */
-		void ProtectAnimation(int time) { protectAnimationUntil = level.time + time; }
+		void ProtectAnimation(int duration) { protectAnimationUntil = level.time + duration; }
 		void UnprotectAnimation() { protectAnimationUntil = -1; }
 
 		/**
--- a/src/sgame/components/BuildableComponent.h
+++ b/src/sgame/components/BuildableComponent.h
@@ -83,6 +83,18 @@ class BuildableComponent: public BuildableComponentBase {
 		void SetAimAngles(const Vec3 aimAngles) { this->aimAngles = aimAngles; }
 		Vec3 AimAngles() const { return this->aimAngles; }
 
+		/**
+		 * Protect current animation from being disrupted by selected, less important animations.
+		 * A value of 0 protects for the rest of the frame only.
+		 */
+		void ProtectAnimation(int time) { protectAnimationUntil = level.time + time; }
+		void UnprotectAnimation() { protectAnimationUntil = -1; }
+
+		/**
+		 * While the current animation is protected, less important animations should not be played.
+		 */
+		bool AnimationProtected() const { return (protectAnimationUntil > level.time); }
+
 	private:
 		lifecycle_t state;
 
@@ -92,6 +104,8 @@ class BuildableComponent: public BuildableComponentBase {
 		int  markTime;
 
 		Vec3 aimAngles; /**< Aim angles relative to world. */
+
+		int protectAnimationUntil;
 
 		// TODO: Move gentity_t.deconstruct and gentity_t.powered here.
 		//bool powered;
--- a/src/sgame/components/SpikerComponent.cpp
+++ b/src/sgame/components/SpikerComponent.cpp
@@ -70,7 +70,10 @@ void SpikerComponent::Think(int timeDelta) {
 
 		expectedDamage += damage;
 
-		if (distance < SPIKER_SENSE_RANGE) {
+		// Start sensing (frequent search for best moment to shoot) as soon as an enemy that can be
+		// damaged is close enough. Note that the Spiker will shoot eventually after it started
+		// sensing, and then returns to a less alert state.
+		if (distance < SPIKER_SENSE_RANGE && !sensing) {
 			sensing = true;
 
 			if (!lastSensing) {
@@ -82,8 +85,6 @@ void SpikerComponent::Think(int timeDelta) {
 		}
 	});
 
-	// Shoot if a viable target leaves sense range even if no damage is expected anymore. This
-	// guarantees that the spiker always shoots eventually when an enemy gets close enough to it.
 	bool senseLost = lastSensing && !sensing;
 
 	if (sensing || senseLost) {
@@ -101,6 +102,9 @@ void SpikerComponent::Think(int timeDelta) {
 				entity.oldEnt->s.number, lastExpectedDamage);
 		}
 
+		// Shoot when the expected damage has decreased, so that it has experienced a local maximum
+		// since its last evaluation. Shoot also whenever all viable targets leave the sense range.
+		// The latter guarantees that the spiker always shoots eventually after sensing.
 		if ((sensing && lessDamage) || senseLost) {
 			Fire();
 		}
@@ -113,13 +117,13 @@ void SpikerComponent::Think(int timeDelta) {
 // TODO: Use a proper trajectory trace, once available, to ensure friendly buildables are never hit.
 bool SpikerComponent::SafeToShoot(Vec3 direction) {
 	const missileAttributes_t* ma = BG_Missile(MIS_SPIKER);
+	float missileSize = (float)ma->size;
 	trace_t trace;
 	vec3_t mins, maxs;
 	Vec3 end = Vec3::Load(entity.oldEnt->s.origin) + (SPIKER_SPIKE_RANGE * direction);
 
 	// Test once with normal and once with inflated missile bounding box.
-	for (int i = 0; i <= 1; i++) {
-		float traceSize = (float)ma->size * ((1.0f-(float)i) + (float)i*SAFETY_TRACE_FUDGE_FACTOR);
+	for (float traceSize : {missileSize, missileSize * SAFETY_TRACE_FUDGE_FACTOR}) {
 		mins[0] = mins[1] = mins[2] = -traceSize;
 		maxs[0] = maxs[1] = maxs[2] =  traceSize;
 		trap_Trace(&trace, entity.oldEnt->s.origin, mins, maxs, end.Data(), entity.oldEnt->s.number,
@@ -138,7 +142,7 @@ bool SpikerComponent::Fire() {
 	gentity_t *self = entity.oldEnt;
 	// Check if still resting.
 	if (restUntil > level.time) {
-		return false;
+		logger.Verbose("Spiker #%i wanted to fire but wasn't ready.", entity.oldEnt->s.number);
 	}
 
 	// Play shooting animation.
--- a/src/sgame/components/SpikerComponent.cpp
+++ b/src/sgame/components/SpikerComponent.cpp
@@ -17,6 +17,8 @@ constexpr float ROWOFFSET = 0.5f;
 constexpr float SPIKE_RANGE = 500.0f;
 /** Bounding box scaling factor Used to improve friendly fire prevention. */
 constexpr float SAFETY_TRACE_INFLATION = 3.0f;
+/** Maximum angle added to or substracted rom the 180° radius of damage, in positive degrees. */
+constexpr float GRAVITY_COMPENSATION_ANGLE = 25.0f;
 
 SpikerComponent::SpikerComponent(
 	Entity &entity, AlienBuildableComponent &r_AlienBuildableComponent)
@@ -24,7 +26,9 @@ SpikerComponent::SpikerComponent(
 	, activeThinker(AT_NONE)
 	, restUntil(0)
 	, lastExpectedDamage(0.0f)
-	, lastSensing(false) {
+	, lastSensing(false)
+	, gravityCompensation(0) {
+	SetGravityCompensation();
 	RegisterSlowThinker();
 }
 
@@ -41,6 +45,14 @@ void SpikerComponent::HandleDamage(float amount, gentity_t *source, Util::option
 
 		Fire();
 	}
+}
+
+void SpikerComponent::SetGravityCompensation() {
+	Vec3 dorsal = Math::Normalize(Vec3::Load(entity.oldEnt->s.origin2));
+	Vec3 upwards = Vec3({0.0f, 0.0f, 1.0f});
+	float uprightness = Math::Dot(dorsal, upwards);
+	// A negative value means the Spiker's radius of damage increases.
+	gravityCompensation = -uprightness * std::sin(DEG2RAD(0.5f * GRAVITY_COMPENSATION_ANGLE));
 }
 
 void SpikerComponent::Think(int timeDelta) {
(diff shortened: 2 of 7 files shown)


src/sgame/components/SpikerComponent.cpp
```diff
@@ -87,44 +65,75 @@ void SpikerComponent::Think(int timeDelta) {
...
+// TODO: Use a proper trajectory trace, once available, to ensure friendly buildables are never hit.
+bool SpikerComponent::SafeToShoot(Vec3 direction) {
+	const missileAttributes_t* ma = BG_Missile(MIS_SPIKER);
+	trace_t trace;
+	vec3_t mins, maxs;
+	Vec3 end = Vec3::Load(entity.oldEnt->s.origin) + (SPIKER_SPIKE_RANGE * direction);
+
+	// Test once with normal and once with inflated missile bounding box.
+	for (int i = 0; i <= 1; i++) {
+		float traceSize = (float)ma->size * ((1.0f-(float)i) + (float)i*SAFETY_TRACE_FUDGE_FACTOR);
```
slipher: How about `for (float traceSize : {float(ma->size), ma->size * SAFETY_TRACE_FUDGE_FACTOR})`

src/sgame/components/SpikerComponent.cpp
```diff
@@ -87,44 +65,75 @@ void SpikerComponent::Think(int timeDelta) {
...
-				}
+		float damage = (targetArea / effectArea) * (float)SPIKER_MISSILES *
+			(float)BG_Missile(MIS_SPIKER)->damage;
+
+		expectedDamage += damage;
+
+		if (distance < SPIKER_SENSE_RANGE) {
+			sensing = true;
+
+			if (!lastSensing) {
```
slipher: I assume you wouldn't want this to run more than once per frame, so perhaps the condition should be like `!sensing && !lastSensing`.

illwieckz:
> Note that without a proper trajectory trace, it still can and will occasionally hit friendly buildables.

That's ok, rockets can also hit friendly people or structure when following a target after being shot. :-)

src/sgame/components/SpikerComponent.cpp
Viech: Thanks, the check was there before but below `sensing = true`, so I removed it assuming it was entirely pointless.

src/sgame/components/SpikerComponent.cpp
Viech: Thanks, will do!t

Viech:
@illwieckz The rockets move and have a mind of their own, which makes things more complicated. Still they would only hit friendly structures if they miss a moving enemy that was hiding that structure from damage at some point before, which is the best I could do with reasonable effort.

For the Spiker I'd like the most safety as well.

illwieckz:
Yeah I just wanted to say that it looks ok to me because this behavior can clearly be part of the game and it's even realistic because of the spiker shape, it would be hard to explain how the spiker can aim precisely _each_ spike without having eyes everywhere. So there is probably no need for a better aiming and neither more complex implementation. The fact spiker can hit nearby team mates can be part of the lore. So I'm curious to see how this change behave in game but perhaps this part does not have to be improved more. :-)

Viech:
I had liked to also improve the spiker visually by adding spike impact models, but cgame does net yet properly support models as particles, so I'll postpone that until @DolceTriade has looked into that.

Feel free to test this branch in the next devgame, @illwieckz. If you like it, it's ready to be merged.

src/sgame/components/SpikerComponent.cpp
```diff
@@ -19,175 +39,215 @@ void SpikerComponent::HandleDamage(float amount, gentity_t *source, Util::option
...
-		int spikes = (int)round(((float)SPIKER_MISSILES * perimeter) / totalPerimeter);
+		int spikes = (int)round(((float)MISSILES * perimeter) / totalPerimeter);
 
 		for (int spike = 0; spike < spikes; spike++) {
 			float azimuth = 2.0f * M_PI * (((float)spike + 0.5f * crandom()) / (float)spikes);
-			float altitudeVariance = 0.5f * crandom() * M_PI_2 / (float)SPIKER_MISSILEROWS;
+			float altitudeVariance = 0.5f * crandom() * M_PI_2 / (float)MISSILEROWS;
 
 			RotatePointAroundVector(dir, zenith, rowBase, RAD2DEG(azimuth));
 			RotatePointAroundVector(dir, rotAxis, dir, RAD2DEG(altitudeVariance));
```
slipher: This line doesn't look right. To change (only) the altitude, the rotation axis can't just be any vector perpendicular to `zenith`; it should be the cross product of `zenith` and `dir`. 
Maybe it would be easier to use a spherical coordinate system.

src/sgame/components/SpikerComponent.cpp
```diff
@@ -2,13 +2,33 @@
...
+/** Delay between shots. */
+constexpr int   COOLDOWN = 5000;
+/** Shoot prematurely whenever this much damage is expected. */
+constexpr float DAMAGE_THRESHOLD = 60.0f;
+/** Total number of missiles to shoot. Some might be skipped if they endanger allies.
+ *  Actual values is +/- SPIKER_MISSILEROWS. */
+constexpr float MISSILES = 26;
+/** Number of rows from which to launch spikes. */
+constexpr int   MISSILEROWS = 4;
+/** 0.0: Spikes are shot upwards, 1.0: Spikes are shot sideways. */
```
slipher: I believe the comment on `ROWOFFSET` is not accurate.

src/sgame/components/SpikerComponent.cpp
slipher: I think it would be good to explain somewhere that a "row" is a group of missile launch vectors with a common angle from the spiker's vertical axis (but all missiles are fired from the same origin). It took a while for me to understand this.

src/sgame/components/BuildableComponent.h
```diff
@@ -83,6 +83,18 @@ class BuildableComponent: public BuildableComponentBase {
 		void SetAimAngles(const Vec3 aimAngles) { this->aimAngles = aimAngles; }
 		Vec3 AimAngles() const { return this->aimAngles; }
 
+		/**
+		 * Protect current animation from being disrupted by selected, less important animations.
+		 * A value of 0 protects for the rest of the frame only.
+		 */
+		void ProtectAnimation(int time) { protectAnimationUntil = level.time + time; }
```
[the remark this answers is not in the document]
Viech: Done.

src/sgame/components/SpikerComponent.cpp
Viech: What's wrong with it? Is it the other way round?

src/sgame/components/SpikerComponent.cpp
```diff
@@ -19,175 +39,215 @@ void SpikerComponent::HandleDamage(float amount, gentity_t *source, Util::option
...
-	      friendlyDamage = 0.0f;
-	bool  sensing        = false;
+	float expectedDamage = 0.0f;
+	bool  sensing = false;
 
-	// Calculate a "scoring" of the situation to decide on the best moment to shoot.
+	// Calculate expected damage to decide on the best moment to shoot.
 	ForEntities<HealthComponent>([&](Entity& other, HealthComponent& healthComponent) {
 		// TODO: Check if "entity == other" does the job.
 		if (entity.oldEnt == other.oldEnt)                                return;
```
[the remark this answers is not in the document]
Viech: True, even though I'd still like to know whether `entity == other` would work!

src/sgame/components/SpikerComponent.cpp
```diff
@@ -19,175 +39,215 @@ void SpikerComponent::HandleDamage(float amount, gentity_t *source, Util::option
...
-			if (distance < SPIKER_SENSE_RANGE) {
-				sensing = true;
-				if (sensing && !lastSensing) {
-					RegisterFastThinker();
-				}
+		const missileAttributes_t* ma = BG_Missile(MIS_SPIKER);
+		float spikeDamage  = ma->damage;
+		float distance     = Math::Length(toTarget);
+		float bboxDiameter = Math::Length(otherMins) + Math::Length(otherMaxs);
+		float bboxEdge     = 0.57735026918962584f * bboxDiameter; // diameter = sqrt(3) * edge
```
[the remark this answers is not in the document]
Viech: Thanks, will use it. (Why is it `M_SQRT2` but `M_ROOT3`?!)

src/sgame/components/SpikerComponent.cpp
```diff
@@ -19,175 +39,215 @@ void SpikerComponent::HandleDamage(float amount, gentity_t *source, Util::option
...
-		float altitude = (((float)row + SPIKER_ROWOFFSET) * M_PI_2) / (float)SPIKER_MISSILEROWS;
+	for (int row = 0; row < MISSILEROWS; row++) {
+		float altitude = (((float)row + ROWOFFSET) * M_PI_2) / (float)MISSILEROWS;
 		float perimeter = 2.0f * M_PI * cos(altitude);
 
 		RotatePointAroundVector(rowBase, rotAxis, zenith, RAD2DEG(M_PI_2 - altitude));
 
 		// Attempt to distribute spikes with equal distance on all rows.
-		int spikes = (int)round(((float)SPIKER_MISSILES * perimeter) / totalPerimeter);
+		int spikes = (int)round(((float)MISSILES * perimeter) / totalPerimeter);
```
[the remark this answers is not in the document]
Viech: Indeed, but it should be an int!

src/sgame/components/SpikerComponent.cpp
Viech: Hmm, it should work if the two rotations are swapped, right?

What do you mean exactly by using a spherical coordinate system here? Isn't this already what I'm doing as I rotate the unit vector zenith first by the angle altitude and then by the angle perimeter?

Viech: I'll add such comment.

Frankly I'd rather just use some probability density function for generating the launch directions to make the code less complex, but I want the vectors to be somewhat "evenly" distributed (with large angles between them), as otherwise the damage variance goes through the roof. I don't know a better way than what I did to achieve this efficiently.

src/sgame/components/SpikerComponent.cpp
Viech: I'll actually just remove it, because only `0.5f` makes sense with the rest of the code.

src/sgame/components/SpikerComponent.cpp
slipher: How about [this](https://stackoverflow.com/a/26127012) algorithm? I think you could change y to be in (0, 1) instead of (-1, 1) and use it for a half sphere.

slipher: Yeah, I guess the only real difference would be to do the entire altitude rotation at once, which you have already done now.

Viech: Looks simple enough. I'll bookmark it and maybe try it out when I find the time, but I wouldn't consider the current approach with the fixes to be a PR blocker as it works reasonably well in practice.

Viech:
Travis error is due to `The command "if [ "$TRAVIS_OS_NAME" == "linux" ]; then sudo apt-get -qq update; fi" failed and exited with 100 during .` and thus not related to the code.

@illwieckz should this go on `master`, `for-0.51.0` or be tested in a dev game first?

illwieckz:
The code is running on my server and was tested during some online games yesterday. No obvious wrong stuff were spotted so we can merge it into `for-0.51.0` directly. Nice job !

Viech:
Rebased on `for-0.51.0`, merging when the tests complete.

Viech:
Actually @illwieckz why did I just rebase this? It's not breaking compat and does not require new or updated assets. Hence this should go on `master`, right?

illwieckz:
Hmm, yes, merging it to master was not problematic at all. I just suggested to merge to `for-0.51.0` directly for the sole reason people just have to fetch that branch to test everything expected for next release. It was just to save a `merge master` into `for-0.51.0` after that, but that would not have been problematic at all.

Viech:
Rebased the commits on master again, awaiting travis to fail for some reason not related to the code at all. 😃

Viech:
The OS X build (`MODE=1`) fails:

	[640/649] Building CXX object daemon_build/CMakeFiles/engine-lib.dir/src/engine/framework/Crypto.cpp.o
	FAILED: daemon_build/CMakeFiles/engine-lib.dir/src/engine/framework/Crypto.cpp.o 
	/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++   -DBUILD_ENGINE -DDEBUG_BUILD -DMACOS_X -DNACL_ANDROID=0 -DNACL_BUILD_ARCH=x86 -DNACL_BUILD_SUBARCH=64 -DNACL_LINUX=0 -DNACL_OSX=1 -DNACL_WINDOWS=0 -I../daemon/src -I../daemon/libs -ICMakeFiles -I../daemon/libs/nacl -I../daemon/src/engine -D__extern_always_inline=inline  -ffast-math -fno-strict-aliasing -m64 -mtune=generic -mcx16 -fvisibility=hidden -std=gnu++11 -stdlib=libc++ -g -g3   -Wall -Wextra -Werror -Wno-pragmas -Wno-unknown-pragmas -Wno-error=old-style-cast -Woverloaded-virtual -Wno-error=sign-promo -Wno-error=extern-c-compat -Wno-error=missing-field-initializers -Wno-error=sign-compare -Wno-error=switch -Wno-error=unused-function -Wno-error=unused-parameter -Wno-error=unused-variable -MD -MT daemon_build/CMakeFiles/engine-lib.dir/src/engine/framework/Crypto.cpp.o -MF daemon_build/CMakeFiles/engine-lib.dir/src/engine/framework/Crypto.cpp.o.d -o daemon_build/CMakeFiles/engine-lib.dir/src/engine/framework/Crypto.cpp.o -c ../daemon/src/engine/framework/Crypto.cpp
	../daemon/src/engine/framework/Crypto.cpp:156:25: error: no matching function for call to 'nettle_base64_encode_update'
	    int encoded_bytes = nettle_base64_encode_update(
		                ^~~~~~~~~~~~~~~~~~~~~~~~~~~
	/usr/local/include/nettle/base64.h:98:1: note: candidate function not viable: no known conversion from 'value_type *' (aka 'unsigned char *') to 'char *' for 2nd argument
	base64_encode_update(struct base64_encode_ctx *ctx,
	^
	/usr/local/include/nettle/base64.h:47:30: note: expanded from macro 'base64_encode_update'
	#define base64_encode_update nettle_base64_encode_update
		                     ^
	../daemon/src/engine/framework/Crypto.cpp:159:22: error: no matching function for call to 'nettle_base64_encode_final'
	    encoded_bytes += nettle_base64_encode_final( &ctx, output.data() + encoded_bytes );
		             ^~~~~~~~~~~~~~~~~~~~~~~~~~
	/usr/local/include/nettle/base64.h:106:1: note: candidate function not viable: no known conversion from 'value_type *' (aka 'unsigned char *') to 'char *' for 2nd argument
	base64_encode_final(struct base64_encode_ctx *ctx,
	^
	/usr/local/include/nettle/base64.h:48:29: note: expanded from macro 'base64_encode_final'
	#define base64_encode_final nettle_base64_encode_final
		                    ^
	../daemon/src/engine/framework/Crypto.cpp:178:11: error: no matching function for call to 'nettle_base64_decode_update'
	    if ( !nettle_base64_decode_update( &ctx, &decoded_bytes, temp.data(),
		  ^~~~~~~~~~~~~~~~~~~~~~~~~~~
	/usr/local/include/nettle/base64.h:158:1: note: candidate function not viable: no known conversion from 'const value_type *' (aka 'const unsigned char *') to 'const char *' for 5th argument
	base64_decode_update(struct base64_decode_ctx *ctx,
	^
	/usr/local/include/nettle/base64.h:54:30: note: expanded from macro 'base64_decode_update'
	#define base64_decode_update nettle_base64_decode_update
		                     ^
	3 errors generated.

I don't see that this is related to this PR, which only contains gamelogic changes. (@DolceTriade may want to look at the issue.)

PR #1060, opened 2018-01-06
allow custom soundstep in character file for bsuit

illwieckz:
- this fixes new-bsuit soundsteps
- it requires bsuit character file to be updated (using 'custom' profile)
- for mysterious reason, the game code already plays bsuit jump sounds without it

bsuit jump sound worked without that but it was a bug, bsuit jump sound must be
dependent of this character file option too

See UnvanquishedAssets/res-players_src.dpkdir#4


DolceTriade:
LGTM

PR #1062, opened 2018-01-17
Remove the repeater.

DolceTriade:
The model will be reused for the drill. It's not a very useful
structure now that humans no longer require power to build.


Viech:
Looks like you hit all of it! It should be noted that the model is not yet reused for the drill.

illwieckz:
on data side, a wip repeater↔drill swap can be found there:
https://github.com/UnvanquishedAssets/res-buildables_src.dpkdir/commit/9ea761477a3b852d78166e4ce9f3836c2eeda0f6

res-buildables package on http://gg.illwieckz.net/dl/unvanquished/dev/ currently implements this swap, hence the code can be tested against that pak

src/cgame/cg_local.h
```diff
@@ -1681,7 +1681,6 @@ typedef struct
 
 	// TODO: Remove those two.
```
slipher: Now one? :)

src/cgame/cg_servercmds.cpp
```diff
@@ -144,7 +144,6 @@ void CG_ParseServerinfo()
 
 	// TODO: Remove those two.
```
slipher: Also here

slipher:
Does this break cgame<->sgame compatibility due to changing some buildable enum values?

Viech:
@slipher what do you mean by that? cgame and sgame are usually tied together through the dpk that ships both games' VMs. If you would just package either a new cgame or a new sgame into a dpk for a change that affects both sides it is expected to break – you'd get results like when playing old demos in Tremulous, with battlesuits swimming through the floor like sharks (scary!).

slipher:
@Viech You are correct; I just wanted to know to be aware of it while debugging (when I have random combinations of values for cgame.vm.type and sgame.vm.type and stuff built at different times).

illwieckz:
it would be far better to make this PR targets `for-0.51.0` branch (which already brokes compat).

illwieckz:
I fixed a minor error on `res-buildables` side.

Once this code is merged or rebased over `for-0.51.0` branch and run with a `for-0.51.0` engine, you can test it using some pre-0.51.0 package you can get this way:

```sh
wget $(curl http://gg.illwieckz.net/dl/unvanquished/dev/UnvanquishedAssets.get http://gg.illwieckz.net/dl/unvanquished/dev/UnvanquishedMaps.get)
```

illwieckz:
While testing the code I noticed the drill spins while building then stops to spin at idle time.

illwieckz:
any ETA on this? :3

illwieckz:
it works, thanks!

illwieckz:
I just added a commit to edit comments like @slipher asked for. @DolceTriade if you have nothing to add can we merge that?

illwieckz:
I don't see anything that is missing at this point, good job.

PR #1073, opened 2018-04-24
Remove default-team binds when changing keys in the bind menu

slipher:
When you change a bind using the menu, any old keys with the command are removed. But the removal does not succeed if the existing bind is for the "default" team (and the new bind is not). I hesitated to do this before since the existing bind might  be useful for a different team. But I see only two examples of commands used on multiple teams with separate menu items (crouch/wallwalk and armory/evolve) and the pairs are next to each other so it is easy to notice that the other one changes. The current behavior also looks very confusing in the menu, so it definitely makes more sense to purge old default team binds rather than leave them.

### DIFF
--- a/src/cgame/rocket/rocket.cpp
+++ b/src/cgame/rocket/rocket.cpp
@@ -147,7 +147,7 @@ public:
 
 static polyVert_t *createVertexArray( Rocket::Core::Vertex *vertices, int count )
 {
-	polyVert_t *verts = new polyVert_t[ sizeof( polyVert_t ) * count ];
+	polyVert_t *verts = new polyVert_t[ count ];
 
 	for ( int i = 0; i < count; i++ )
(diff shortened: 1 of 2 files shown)


slipher:
Actually, I just thought of one bind that would be broken by this: `+movedown` is also needed for spectators.

src/cgame/rocket/rocket.cpp
```diff
@@ -147,7 +147,7 @@ class DaemonSystemInterface : public Rocket::Core::SystemInterface
 
 static polyVert_t *createVertexArray( Rocket::Core::Vertex *vertices, int count )
 {
-	polyVert_t *verts = new polyVert_t[ sizeof( polyVert_t ) * count ];
```
DolceTriade: ouch....

PR #1076, opened 2018-05-06
Change ordering of lua shutdown to avert crash.

slipher:
Fixes https://github.com/DaemonEngine/Daemon/issues/103.

### DIFF
--- a/src/cgame/rocket/rocket.cpp
+++ b/src/cgame/rocket/rocket.cpp
@@ -386,17 +386,12 @@ void Rocket_Init()
 
 void Rocket_Shutdown()
 {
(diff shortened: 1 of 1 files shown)


src/cgame/rocket/rocket.cpp
```diff
@@ -386,18 +386,13 @@ void Rocket_Init()
 
 void Rocket_Shutdown()
 {
+	// TODO: Consider that it may be more efficient to do nothing at all here, in
```
DolceTriade: The cgame gets restarted several times and in so lib mode, we will leak resources.

PR #1077, opened 2018-05-07
travis needs python

illwieckz:
It's ok now!


DolceTriade:
Thanks! fwiw, theoretically, our python code should work with python2 and python3.

slipher:
Nacl requires py 2

PR #1083, opened 2018-06-29
Remove gamestats

slipher:
No one is using the master server stats collection so why not delete


slipher:
Then our numerous contributors can work on implementing a much better one.

PR #1084, opened 2018-07-15
Randomly reorder librocket destructors some more

slipher:


illwieckz:
I don't know if it's enough but using this patch I have not yet managed to crash the game. Good job.

Issue #909, opened 2016-02-21
Some places use the wrong color code when trying to clear the current color.


mbasaglia:
One place is the print that describes a cvar, it uses ^7 instead of ^*.
There might be other places that do the same.

PR #1092, opened 2018-10-27
Replace ^7 colorcode usage with ^* where it was used as ending null-color

t4im:
Addresses #909 for the sgame/cgame

Also adds a few additional ^* where playernames were involved, to avoid playercolors leaking into text.
Leaves ^7 in front of names and where it might have been intentional.

### DIFF
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -33,6 +33,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 CG_Obituary
 =============
 */
+// end on WHITE instead of ESC for player names following the tags
 static const char teamTag[][8] = { "^2●^7", "^1●^7", "^4●^7" };
 
 #define LONGFORM ">"
@@ -210,7 +211,7 @@ static void CG_Obituary( entityState_t *ent )
 			{
 				if ( meansOfDeath[ mod ].showAssist && assistantInfo )
 				{
-					Log::Notice( "%s (+ %s%s^7) %s %s%s", teamTag[ attackerTeam ], teamTag[ assistantTeam ], assistantName, meansOfDeath[ mod ].icon, teamTag[ ci->team ], targetName );
+					Log::Notice( "%s (+ %s%s^*) %s %s%s", teamTag[ attackerTeam ], teamTag[ assistantTeam ], assistantName, meansOfDeath[ mod ].icon, teamTag[ ci->team ], targetName );
 				}
 				else
 				{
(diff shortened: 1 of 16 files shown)


slipher:
Is ^* even implemented in most places? There was this one time where I implemented it in several places, but after that they were all rewritten and it didn't work anymore.

src/cgame/cg_event.cpp
```diff
@@ -33,6 +33,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 CG_Obituary
 =============
 */
+// end on WHITE instead of ESC for player names following the tags
```
slipher: Why?

src/sgame/sg_spawn_position.cpp
```diff
@@ -93,8 +93,8 @@ void SP_pos_location( gentity_t *self )
 			self->customNumber = 7;
 		}
 
-		message = va( "%c%c%s^7", Color::Constants::ESCAPE, self->customNumber + '0',
-		              self->message );
+		message = va( "%c%c%s%c", Color::Constants::ESCAPE, self->customNumber + '0',
```
slipher: This one doesn't look right.

src/cgame/cg_event.cpp
t4im: Just the cautious assumption, that if the default color of a message would be changed, that player and location names should maybe remain the same.
But i would have no problem with either way.

Might have to rename myself to 7T4im tho D:

src/sgame/sg_spawn_position.cpp
t4im: Oops

src/cgame/cg_event.cpp
slipher: I mean, why did you make this assumption here but not anywhere else?

t4im: Both versions, white-started names and neutral started names, are found throughout the code, i simply left most as they were.

Since players still have to be ported to CBSE, all those names have to be touched again; i suggest using that then to make it consistent via convenience method.

If you prefer the explicit white-prefixes removed till then, i'll gladly go through the code once more.

illwieckz: There is a lot of assumptions everywhere, but perhaps we need to decide something in order to get coherency. Since the nicks are made to be colorful, we can assume _we have to_ set default colors when unspecified, i.e. white there. That's the same issue browser faces with white on white or dark on dark websites issues. That kind of issue can only be fixed with conventions, even if some people will disagree the convention chose something and not something else.

We have a similar issue with icons (= smileys), they are not displayed the same way throughout the ui. Basically there is some places where the selected color is applied on the icon itself, and sometime not. While the ability to color icons is not a bad idea by itself, we must define a a rule and stick to it. Not being able to color icons is not a bad loss btw, and it would avoid people to reset color when _they don't want_ to color icons, and who wants to color them?

slipher:
Is this ready to merge?

t4im:
It is.

PR #1093, opened 2018-10-28
Remove unused gentity fields

t4im:
and other entity related things that can be removed


t4im:
changes are included in #1094

PR #1096, opened 2018-11-08
use ccache with travis

t4im:
Might not store changes from PRs, and as such might not be visible until after a few changes to master.
But if it works, it could potentially speed up the build massively by allowing us to reuse compilations of those parts of the source that haven't changed between CI builds.

The travis setting automatically prefixes the PATH with /usr/lib/ccache/ which contains wrappers around clang and gcc.


t4im:
A few other things to consider:
* `software-properties-common` appears unused and could be removed?
* Allowing the mac build to fail with fast-finish might avoid the false negatives due to staling; At least until the problem is figured out. Deactivating the mac build for the time being might also help freeing up per-account capacities for the other build types.
* Building in unoptimized Debug instead of Release mode might compile faster and run Assertions for us, too. If we want to keep building Releases, we should probably still enable Assertions explicitly somehow.

slipher:
I don't understand the point about the runtime assertions, since no one ever runs the code built on travis. Are you concerned about compile errors in the assert condition?

t4im:
Ah, well, i was thinking of https://docs.travis-ci.com/user/deployment/releases/ some day; but indeed; right now rather pointless. Debug symbols and assertions would be useful though if we chose to make dailies available some day.

PR #1098, opened 2018-11-23
Fix cbse-entity memory leak and mandatory component segfault

t4im:
CBSE Entities and all of their components were only being deleted in `G_FreeEntity`, which isn't used on game shutdowns like map restarts, map changes or actual shutdowns.

To simplify the fix, i also moved the `EmptyEntity` initialization, which also fixes DaemonEngine/CBSE-Toolchain#29

A quick look on the process memory shows that despite the fix we still seem to be leaking some memory somewhere across restarts (although i can't tell if that is cgame or sgame); but this should fix one bigger source.

### DIFF
--- a/src/sgame/sg_entities.cpp
+++ b/src/sgame/sg_entities.cpp
@@ -36,8 +36,6 @@ Maryland 20850 USA.
 #include "sg_entities.h"
 #include "CBSE.h"
 
-static EmptyEntity emptyEntity(EmptyEntity::Params{nullptr});
-
 /*
 =================================================================================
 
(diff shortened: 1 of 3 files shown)


slipher:
The sgame process is restarted on map changes, so like the librocket stuff, it doesn't matter that much.

slipher:
LGTM

illwieckz:
Interesting, I noticed huge memory leak when running for-0.51.0 server on shared lib (because running on gdb to trace anything happening) making it unable to use (it would just make my server oom-kill the other game servers)…

PR #1107, opened 2019-03-04
export game surfaceflags to CustomSurfaceFlags.h

illwieckz:
This is an easy way to fix DaemonEngine/Daemon#107

I just noticed that those surfaceflags either add new values, either replace existing ones, like game's `CONTENTS_NOBUILD` sharing the `0x00004000` value with engine's `CONTENTS_MOVER`.

So in any way, games just want to add values or to override one to do ugly tricks like our game does.

If one day games want a more elaborated behavior, it would be still time to elaborate more.

See also DaemonEngine/Daemon#177 for related PR on the engine side.

### DIFF
--- a/src/sgame/CustomSurfaceFlags.h
+++ b/src/sgame/CustomSurfaceFlags.h
@@ -1,0 +1,45 @@
+/*
+===========================================================================
+
+Unvanquished BSD Source Code
+Copyright (c) 2013-2016, Unvanquished Developers
+All rights reserved.
+
+This file is part of the Unvanquished BSD Source Code (Unvanquished Source Code).
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of the Unvanquished developers nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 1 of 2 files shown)


src/sgame/CustomSurfaceFlags.h
```diff
@@ -0,0 +1,43 @@
+/*
+===========================================================================
+
+Unvanquished BSD Source Code
```
slipher: Should be GPL?

illwieckz: That was a question I wanted to ask. Is a such simple thing as a couple of defines in a header is  copyrightable from the start? If not, it's better to put a very permissive header so in case of doubt the wording is the most permissive one.

In this case it falls more under the database use case than the code one. Is this code? no there is no logic, it's data. Is there a way to write this data in another way? to be grok by C preprocessor, there is no other way, hence the complete absence of originality. So, it's data and data can be copyrighted, but can we copyright the `0x4000` number? Hardly. Can we convince any author to release its code under GPL because he used `0x4000` in a define? Not really. It's start to be easier to claim that a list of 6 numbers may be an original work, but those numbers are determined by the flag mechanism itself. We were not able to chose `0x1234`, we had to chose a special number from a given list, the same way anyone have to compare first digit to `0` to check for parity.

slipher: Aren't we using GPL for the entire gamelogic anyway? In which case the point is moot.

illwieckz: The gamelogic is using GPL but some files that are meant for interaction with engine are in BSD:

- [`src/cgame/cg_api.cpp`](https://github.com/Unvanquished/Unvanquished/blob/master/src/cgame/cg_api.cpp)
- [`src/sgame/sg_api.cpp`](https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_api.cpp)

illwieckz: @DolceTriade or @Viech any word on this?

illwieckz: The engine header file this one is overriding was entirely rewritten from scratch, allowing that engine file to be licensed under a BSD license, which is better for a code that is expected to be shared with game code (reduce the risk of license incompatibilities between engine and game).

This header file overrides that BSD-licensed engine header file, a GPL license for this one would be compatible, but there is absolutely nothing to protect with a GPL license, and this file is likely to be used as a template by third-party projects. Hence a permissive BSD license seems to be very good for it.

PR #1108, opened 2019-03-31
HTML escape key names

slipher:


src/cgame/rocket/rocket.cpp
```diff
@@ -485,6 +485,28 @@ static bool IsInvalidEmoticon( Rocket::Core::String emoticon )
 	return false;
 }
 
+std::string CG_EscapeHTMLText( Str::StringRef text )
```
DolceTriade: This seems useful to include in other places. noteably Rocket_QuakeToRML and rocketChatField.

slipher: Yeah, I noticed there were a couple of other places doing escaping. But they also did other transformations and used different string types, so it didn't seem worth trying to combine them somehow.

PR #1114, opened 2019-07-07
Hack for non-0/1 checkboxes

slipher:
+ misc deleting


illwieckz:
I fixed travis so this branch can be rebased on master.

PR #1118, opened 2019-08-19
import 512×512 unvanquished png icon

illwieckz:
This one is a 512×512 version, upscaled from the 470×470 we had, instead of the 128×128 we currently have. It looks to be the higher-res version available and the original resolution, do someone has the original source for this file?


illwieckz:
So, after having investigated why I found higher resolution icons for our own game on some random brasilian web site by reverse-googling our own icon, I found _this_ `470×470` icon in our own old [0.6.0 linux installer on sourceforge](https://sourceforge.net/projects/unvanquished/files/Linux%20x86_64%20Generic%20Builds/Unvanquished-0.6.0-linux-installer.run):

[unvanquished icon](https://dl.illwieckz.net/b/unvanquished/src/icon/original/470/unvanquished.png)

This looks to be the mother of all our icons.

How did we lost it?

In anyway, _why_ `470×470`?

illwieckz:
So I imported the 470×470 icon, I also included 256/128/64 derivations. Those are known to look good, I compared with various scaling algorithms in GIMP and either I was not able to get better, either the better I got looked as good.

A 470×470 icon resolution may looks weird, but my own Ubuntu distro has a 480×480 filezilla icon and 42×42 imagemagick icon, and there is some other weird occcurrence like an app with 20x20 icon. So non-multiple-of-two icons look to be uncommon but correctly supported, so we don't have to worry about a 470×470 icon.

illwieckz:
Note that we may update our `.ico` and `.icns` files with higher res alternatives, and also redo our updater `.ico` that is very low res.

illwieckz:
original 470×470 icon can be found there:
https://github.com/Unvanquished/UnvanquishedBranding

PR #1125, opened 2019-11-04
add callback to g_BPInitialBudget and g_BPBudgetPerMiner

### LINKED ISSUE #1124: Available buildpoints do not change with cvar change
Changing g_BPBudgetPerMiner or g_BPInitialBudget only take effect after a leech/drill is build.

Gireen:
Fix #1124

### DIFF
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -107,8 +107,22 @@ vmCvar_t           pmove_accurate;
 vmCvar_t           g_minNameChangePeriod;
 vmCvar_t           g_maxNameChanges;
 
-vmCvar_t           g_buildPointInitialBudget;
-vmCvar_t           g_buildPointBudgetPerMiner;
+Cvar::Callback<Cvar::Cvar<int>> g_buildPointInitialBudget(
+		"g_BPInitialBudget",
+		"Initial Budget",
+		Cvar::SERVERINFO,
+		DEFAULT_BP_INITIAL_BUDGET,
+		[](int) {
+			G_UpdateBuildPointBudgets();
+		});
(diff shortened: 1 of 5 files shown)


slipher:
Oh, if it's the case that the calculation works fine but just needs to be updated, you could use a cvar callback. You can make a function get called when the value changes. E.g. 
```
Cvar::Callback<Cvar::Cvar<int>> g_lol("g_lol", "description", 0, 50, [](int lol) { Log::Notice("lol = %d", lol); });
```

Looks like there are 3 obscure features that depend on cvars being in `gameCvarTable`, but none of them are used by the two BP cvars so it's fine to convert them to new-style (C++ API) cvars.

Gireen:
the callback required more changes but that works really better.

src/sgame/sg_main.cpp
```diff
@@ -107,8 +107,22 @@ vmCvar_t           pmove_accurate;
...
 vmCvar_t           g_maxNameChanges;
 
-vmCvar_t           g_buildPointInitialBudget;
-vmCvar_t           g_buildPointBudgetPerMiner;
+Cvar::Callback<Cvar::Cvar<int>> g_buildPointInitialBudget(
+		"g_BPInitialBudget",
+		"Initial Budget",
+		Cvar::SERVERINFO,
+		DEFAULT_BP_INITIAL_BUDGET,
+		[](int budget) {
```
slipher: Remove the name of the unused variable (so like `[](int) {`... )

slipher:
Please squash the commits into one, remove the unused variable names and LGTM otherwise.

slipher:
You can merge it, right?

PR #1126, opened 2019-11-05
always show evolve menu

### LINKED ISSUE #439: No visual feedback is shown for disallowing evolve
If you have 0 morph points and try to evolve, currently only a sound is played. Perhaps the indicator for morph points should flash 0.

Gireen:
show evolve menu even when not enough morph points are available 
Fix #439


illwieckz:
I agree with the idea of being able to list available classes even without evo points. That is especially good for beginners. Default behaviors meant you was not able to check at game startup if your bind was correct (unless you already know the meaning of the bell that is used to notify you don't have enough evo point, which means you're not a beginner)…

PR #1129, opened 2019-11-10
return evovle cost even if not possible

### LINKED ISSUE #1123: Alien evolve menu does not show prices

Gireen:
let BG_ClassCanEvolveFromTo return the price even if its not possible to evolve
fixes #1123 together with UnvanquishedAssets/unvanquished_src.dpkdir#8


slipher:
I suggest to rename ClassCanEvolveFromTo to something like CostToEvolve, remove the credits argument, and give it the single responsibility of determining the cost to evolve between two classes (or -1 if impossible).

Gireen:
ok there is now CostToEvolve for only the costs and ClassCanEvolveFromTo acts as before and uses CostToEvolve

slipher:
LGTM

PR #1130, opened 2019-11-17
reduce impact of close humans on evolution

Gireen:
Instead of a single human entity around it requite 1.5 more human buildings/players than aliens.

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -2345,6 +2345,9 @@ static bool Cmd_Class_internal( gentity_t *ent, const char *s, bool report )
 
 			num = trap_EntitiesInBox( mins, maxs, entityList, MAX_GENTITIES );
 
+			float aliensInRange = 0;
+			float humansInRange = 0;
+
 			for ( i = 0; i < num; i++ )
 			{
 				other = &g_entities[ entityList[ i ] ];
@@ -2353,13 +2356,23 @@ static bool Cmd_Class_internal( gentity_t *ent, const char *s, bool report )
 				     ( other->s.eType == entityType_t::ET_BUILDABLE && other->buildableTeam == TEAM_HUMANS &&
 				       other->powered ) )
 				{
-					if ( report )
-					{
-						G_TriggerMenu( clientNum, MN_A_TOOCLOSE );
-					}
-					return false;
+					humansInRange++;
+
+				}
+				else if ( ( other->client && other->client->pers.team == TEAM_ALIENS ) ||
+				     			( other->s.eType == entityType_t::ET_BUILDABLE && other->buildableTeam == TEAM_ALIENS ) )
+				{
+					aliensInRange++;
 				}
 			}
+			if ( humansInRange > ( aliensInRange * g_evolveArroundHumans.Get() ) ) {
+				if ( report )
+				{
+					G_TriggerMenu( clientNum, MN_A_TOOCLOSE );
+				}
+				return false;
+			}
+
 
 			//check that we are not wallwalking
 			if ( ent->client->ps.eFlags & EF_WALLCLIMB )
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -2345,7 +2345,7 @@ static bool Cmd_Class_internal( gentity_t *ent, const char *s, bool report )
 
 			num = trap_EntitiesInBox( mins, maxs, entityList, MAX_GENTITIES );
 
-			int aliensInRange = 0;
+			int alienBuildingsInRange = 0;
 			int humansInRange = 0;
 			bool visibleToHumans = false;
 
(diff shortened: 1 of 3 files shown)


illwieckz:
This would probably make the game less frustrating, but it would still prevents an alien that is safely hiding in the room next to the human base to evolve, while it looks entirely safe to evolve.

In any way, this is good for bots and would be kept for bots even if we find a way to outsource the "is it safe?" computation to player's brain via specific gameplay mechanism.

Gireen:
That's probably an edge case. With this change aliens can nearly always evolve around there outpost or base even when humans are attacking or buildings are nearby. The decision if it is save to evolve is already by the player and that is also not the meaning of this mechanic. Its only to prevent aliens from evolving directly inside a human base. 
You can also increase the factor with  g_evolveArroundHumans that evolving of aliens is not restricted by humans.

slipher:
As it is, this would allow aliens to evolve during 1-on-1 combat (since the one trying to evolve is counted), which seems wrong to me. To avoid that it could not count the self, or the ratio could be made lower than 1.

To me the main annoying thing about the current system is being unable to evolve in my own base, so another possibility is to just allow evolution when near alien buildings.

Gireen:
So how about we make evolving work like changing teams. A bit like illwieckz's cocoon idea but revers. Aliens will need a certain time without damage to evolve or alternative x alien buildings around.  

That way aliens can:
- not evolve inside a human base or while in combat.  
- evolve near human bases when not attacked.
- always evolve around there own buildings.

slipher:
I like the idea that you can evolve in the human base if you manage to successfully hide.

Evolving whenever you have not recently taken damage could be a bit abusable -- imagine a dretch evolving to tyrant right in front of someone (if there are multiple aliens then you may not be able to shoot all of them soon enough to stop it).

We already have implemented a notion of whether an enemy has seen you, with the beacon auto-tagging. Maybe it could be interesting to incorporate this somehow. Like a sufficient condition to evolve is that you have not recently taken damage AND have not recently been sighted by the enemy. Then you could still have a fun mechanic of evolving right behind someone, but it would take some skill since you have to be sneaky about it.

illwieckz:
> Aliens will need a certain time without damage

Awesome thinking! That's a timer the player has not to wait for if played right! So it's up to the player skill to lose time, and it encourages somewhat realistic behavior: the engine will not arbitrarily prevent you to evolve even if you're well hidden but too much close, and you may pass the human turrets and evolve in the human base's backyard…

And requiring time without damage to evolve is realistic, once can think that before evolving, one has to be healthy and, if not, recover a bit. The idea of slipher about not being seen is good too, but we need to make sure the behavior is obvious.

Allowing aliens to evolve near alien base, at least near overmind looks good. In fact the most occurring situation with aliens are stupidly not being able to evolve is aliens in chasm map. They can't evolve if there is hummies in corridor… That's very frustrating.

src/sgame/sg_cmds.cpp
```diff
@@ -2345,6 +2345,9 @@ static bool Cmd_Class_internal( gentity_t *ent, const char *s, bool report )
 
 			num = trap_EntitiesInBox( mins, maxs, entityList, MAX_GENTITIES );
 
+			float aliensInRange = 0;
+			float humansInRange = 0;
```
DolceTriade: These should both be ints since you can't have partial humans...

src/sgame/sg_main.cpp
```diff
@@ -207,6 +207,8 @@ vmCvar_t           g_emptyTeamsSkipMapTime;
 
 Cvar::Cvar<bool>   g_neverEnd("g_neverEnd", "cheat to never end a game, helpful to load a map without spawn for testing purpose", Cvar::NONE, false);
 
+Cvar::Cvar<float>  g_evolveArroundHumans("g_evolveArroundHumans", "Rate of human to alien entities that prevent evolution", Cvar::NONE, 1.5f);
```
DolceTriade: s/Rate/Ratio

src/sgame/sg_cmds.cpp
Gireen: ok 😅

Gireen:
> Evolving whenever you have not recently taken damage could be a bit abusable ...
yeah that makes sense, by thinking about it poison would also make it a bit annoying 

> Like a sufficient condition to evolve is that you have not recently taken damage AND have not recently been sighted by the enemy

That's a good idea. 
Its now that when you are in line of sight of an human player or building  you can only evolve when 1.5 more alien buildings are around. As before this only applies on a close area around the player, so that on bigger maps aliens still can evolve even if there is no place to hide.

src/sgame/sg_cmds.cpp
```diff
@@ -2345,8 +2345,9 @@ static bool Cmd_Class_internal( gentity_t *ent, const char *s, bool report )
 
 			num = trap_EntitiesInBox( mins, maxs, entityList, MAX_GENTITIES );
 
-			float aliensInRange = 0;
-			float humansInRange = 0;
+			int aliensInRange = 0;
```
slipher: Change variable name to reflect the fact that only buildables are counted?

src/sgame/sg_main.cpp
```diff
@@ -207,7 +207,7 @@ vmCvar_t           g_emptyTeamsSkipMapTime;
 
 Cvar::Cvar<bool>   g_neverEnd("g_neverEnd", "cheat to never end a game, helpful to load a map without spawn for testing purpose", Cvar::NONE, false);
 
-Cvar::Cvar<float>  g_evolveArroundHumans("g_evolveArroundHumans", "Rate of human to alien entities that prevent evolution", Cvar::NONE, 1.5f);
+Cvar::Cvar<float>  g_evolveArroundHumans("g_evolveArroundHumans", "Ratio of alien buildings to human entities that always allow evolution", Cvar::NONE, 1.5f);
```
slipher: around

Gireen: oh.. yarr

illwieckz: Unvanquished is proud to have the best literature reviewing team in the world. One day we will be skilled at C, but at least we are skilled at English. :joy:

slipher:
Fine with me (I intended to express this with my previous comment).

PR #1133, opened 2019-11-19
improve pak downloader script

illwieckz:
- rename `download-dpk` to `download-paks` as it can also download older version with pk3
- do torrent download by default or fallback to http if `aria2c` is missing
- be able to use `aria2c`, `curl` or `wget` for http download
- resume partial http downloaded files instead of restarting entirely their download
- fix a lot of bugs, did you know that a mistake in former torrent downloader
  prevented another bug to wipe your source repository?
 - cycle through HTTP mirrors when one is unavailable
- properly wipe cache directory
- do not clean-up yet files from previous releases

- as a side work, improve the `update-version-number` script
- script extensions considered harmful


illwieckz:
This is part of a huge work I'm doing on scripts, this is first shipping.

Note that the pak downloader is now able to download 0.51.1 paks even if we mistakenly shipped an empty `md5sums` file, since default download mode is the bittorrent one and this one does not require the md5sums file to work.

download-paks
```diff
@@ -0,0 +1,550 @@
...
+
+set -e
+
+# Work around a reported locale problem
+export LANG=C.UTF-8
+
+xdgHomeDir="${XDG_DATA_HOME:-${HOME}/.local/share}/unvanquished"
+
+# Version of Unvanquished for which this script is built
+version=0.51.1
```
DolceTriade: Can we not hard code this? Maybe fetch from latest tag or something?

illwieckz: This is rewritten by `update-version-number` in any way.

Note that this script was shipped by debian package in the past, and it may be used outside of source dir in a legit way (at least downloaded by direct download).

What we can do is to read the git tag if exists, or fallback on this hardcoded one.

illwieckz:
This is current output of: `./download-paks --help`

```
download-paks: download Unvanquished game files

Usage: download-paks [option] [destination directory] [cache directory]

Options:
	--help             Print this help
	--http             Download using HTTP (requires aria2c or curl or wget)
	--torrent          Download using BitTorrent (requires aria2c, default if aria2c is available)
	--no-check         Do not verify downloaded files
	--no-download      Do not download files (only perform file verification)
	--mirror=MIRROR    Download from MIRROR (will not cycle through known mirrors)
	--version=VERSION  Download this VERSION

Default destination directory is /home/illwieckz/.local/share/unvanquished/pkg
Default cache directory is destination directory (files are always downloaded in a .cache subdirectory)

Default version to download:
	0.51.1

Mirror URL examples:
	http://cdn.unvanquished.net
	http://unvcdn.viech.name
	http://cdn.illwieckz.net/unvanquished
```

PR #1137, opened 2019-11-28
Fix Entities::HasFullHealth

Gireen:
This fixes health related bot actions like humans repair and aliens heal.


illwieckz:
When I read on your server the _“Bots repair their base”_ mention, I was like: _“wait, isn't it the expected behavior, wasn't it working before?“_, so maybe it was broken…

illwieckz:
@Gireen is that the only change needed to see alien bots running away in low health situation?

Gireen:
yeah they thought the whole time every building had full health.

This changes let bots retreat
UnvanquishedAssets/unvanquished_src.dpkdir#10

PR #1138, opened 2019-11-28
either display granger or adv granger in evo menu

Gireen:
Only one of them is needed at the same time depending on if the adv granger is unlocked.

### DIFF
--- a/src/cgame/cg_rocket_dataformatter.cpp
+++ b/src/cgame/cg_rocket_dataformatter.cpp
@@ -390,6 +390,13 @@ static void CG_Rocket_DFCMAlienEvolve( int handle, const char *data )
 	const char *action = "";
 	int cost = BG_ClassCanEvolveFromTo( cg.predictedPlayerState.stats[ STAT_CLASS ], alienClass, cg.predictedPlayerState.persistant[ PERS_CREDIT ] );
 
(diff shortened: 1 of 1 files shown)


illwieckz:
looks good

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -1647,6 +1647,12 @@ void CG_Rocket_BuildAlienEvolveList( const char *table )
 		{
 			if ( BG_Class( i )->team == TEAM_ALIENS )
 			{
+				if ( ( i == PCL_ALIEN_BUILDER0 && ( BG_ClassUnlocked( PCL_ALIEN_BUILDER0_UPG ) && !BG_ClassDisabled( PCL_ALIEN_BUILDER0_UPG ) ) )||
```
DolceTriade: I believe that we should not include this change in the "backend". This is a strictly UI change that should be made in the UI...The circleMenu can use a filter function or something that can filter things out like this...

Gireen: Is CG_Rocket_DFCMAlienEvolve the right place for it?

DolceTriade: Yeah, it's fine to put it there.

illwieckz:
@DolceTriade is it ok for you now?

PR #1139, opened 2019-11-29
Make it possible to set console key via bind menus

slipher:


slipher:
See also https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/9.

There's a small bug which is that the menu will allow you to choose a mouse button for the bind, but the code that checks for the console key ignores mouse clicks. So it would be better if the menu would not accept mouse buttons as the chosen key, but I'm not going to worry about it.

PR #1141, opened 2019-12-02
Fix display weapons in scoreboard for spectators

Gireen:


illwieckz:
I have mixed feeling about this but in any way since the spectator can look around there is already no information to hide

Gireen:
this was the normal behavior of the scoreboard before librocket, so i don't know if it was intentional removed or not. But when you follow a player as spectator you can see this things related to his team in the scoreboard, its just inconvenient

illwieckz:
Oh, you're right, that looks more like a regression then. Let's go!

PR #1144, opened 2019-12-26
use default layout instead of builtin if available

Gireen:
If a layout named default is available it will be loaded when no other layout is specified.
The builtin layout is still available as `*BUILTIN*`

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -4268,7 +4268,7 @@ bool G_admin_restart( gentity_t *ent )
 	admin_log( teampref );
 
 	// cvars
-	trap_Cvar_Set( "g_layouts", builtin ? S_BUILTIN_LAYOUT : layout );
+	trap_Cvar_Set( "g_layouts", layout );
 	trap_Cvar_Set( "g_mapRestarted", "y" );
 
 	// handle the flag
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -2229,6 +2229,7 @@ int G_LayoutList( const char *map, char *list, int len )
 void G_LayoutSelect()
 {
 	char fileName[ MAX_OSPATH ];
+	char fileName2[ MAX_OSPATH ];
 	char layouts[ MAX_CVAR_VALUE_STRING ];
 	char layouts2[ MAX_CVAR_VALUE_STRING ];
 	const char *layoutPtr;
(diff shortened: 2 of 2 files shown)


illwieckz:
That's a good idea!

src/sgame/sg_admin.cpp
```diff
@@ -4255,7 +4255,15 @@ bool G_admin_restart( gentity_t *ent )
 	}
 
 	// check that the layout's available
-	builtin = !*layout || !Q_stricmp( layout, S_BUILTIN_LAYOUT );
+	builtin = !Q_stricmp( layout, S_BUILTIN_LAYOUT );
+
+	if( !*layout )
+	{
+		strcpy( layout, "default" );
+		if ( !trap_FS_FOpenFile( va( "layouts/%s/%s.dat", map, layout ), nullptr, fsMode_t::FS_READ ) ) {
```
slipher: I think this return value check is wrong. It returns -1 if the file doesn't exist.

Gireen: this if statement should be true when no file exists

Gireen: this if statement should be true when no file exists

slipher: It's only true if the file exists but has a length of zero.

Gireen: hm i think i get what you mean. i tried the code and strangely this works as intended 🤷‍♂️
but i just realized that this check is not necessary at this place

Gireen:
mappers should place there default layout in builtin.dat 
while server admins can overwrite this with default.dat
and if nothing is available *BUILTIN* from the map is loaded

illwieckz:
interesting, does the game code already knows how to load layouts from dpk?

Gireen:
yes

illwieckz:
Wow, that's amazing. So next step would be to make a cheat command to spawn as builder if there is no spawn. And the only required entity would be player intermission… (we already have `g_neverEnd` to load map without spawns).

src/sgame/sg_admin.cpp
```diff
@@ -4268,7 +4268,7 @@ bool G_admin_restart( gentity_t *ent )
 	admin_log( teampref );
 
 	// cvars
-	trap_Cvar_Set( "g_layouts", builtin ? S_BUILTIN_LAYOUT : layout );
+	trap_Cvar_Set( "g_layouts", !*layout ? "" : layout );
```
slipher: This line is equivalent to setting the cvar to `layout` unconditionally.

Issue #1142, opened 2019-12-12
map rotation `if` action does not support map argument


illwieckz:
According to [wiki](https://wiki.unvanquished.net/wiki/Server/Map_Rotation), `if` can both load map or labels, and this map rotation must be doable:

```python
antares
if numClients > 8 parpax
if numClients < 4 parpax-mini
chasm
forlorn
```

But only labels are supported, so to get the same behavior, this over-complicated rotation has to be done instead:

```python
antares
if numClients > 8 #parpax
if numClients < 4 #parpax-mini
#parpax
parpax
#parpax-mini
parpax-mini
goto chasm
#chasm
chasm
forlorn
```

Gireen:
It seems that maps are treated like goto statements.
The first rotation results in following error
 `Warn: label, map, or rotation parpax-mini not found in rotation1`
if you add that map to the rotation it will jump there without label.

illwieckz:
fixed by #1145

PR #1145, opened 2019-12-26
fix maprotation condition

Gireen:
possible fix for #1142

### DIFF
--- a/src/sgame/sg_maprotation.cpp
+++ b/src/sgame/sg_maprotation.cpp
@@ -404,7 +404,7 @@ static bool G_ParseNode( mrNode_t **node, char *token, const char **text_p, bool
 
 		Q_strncpyz( label->name, token, sizeof( label->name ) );
 	}
-	else if ( *token == '#' || conditional )
+	else if ( *token == '#' )
 	{
 		mrLabel_t *label;
 
@@ -1042,6 +1042,14 @@ Send commands to the server to actually change the map
 static void G_IssueMapChange( int index, int rotation )
 {
 	mrNode_t *node = mapRotations.rotations[ rotation ].nodes[ index ];
+
+	if( node->type == NT_CONDITION )
+	{
+		mrCondition_t *condition;
+		condition = &node->u.condition;
+		node = condition->target;
+	}
+
 	mrMapDescription_t  *map = &node->u.map;
 	char currentMapName[ MAX_STRING_CHARS ];
 


src/sgame/sg_maprotation.cpp
```diff
@@ -1042,6 +1042,14 @@ Send commands to the server to actually change the map
...
 {
 	mrNode_t *node = mapRotations.rotations[ rotation ].nodes[ index ];
+  mrCondition_t *condition;
+
+  if( node->type == NT_CONDITION )
+  {
+    condition = &node->u.condition;
+    node = condition->target;
+  }
+
```
illwieckz: interesting, last time I tried to only remove the `|| conditonal` test I got random ugly crashes, maybe I was missing this.

Gireen: yeah the rotation will try to load the node with the condition as map, which results in random letters as mapname

illwieckz:
on that topic (but that would be another PR), I would like to be able to ship multiple layout in a map dpk archive, and be able to load them. Even being able to override the builtin bsp layout (in that case, asking for builtin would load the one named default in dpk), that would make entity editing and buildable placement an easy job: just create a map with almost nothing, load in engine, make a layout, dump the layout, store it in dpk and voilà.

Gireen:
that is already mostly possible, layouts can be shipped with maps (see parpax mini)
and with #1144 you can set the default layout. What would be needed is a way to distinguish between builtin-default and server default

illwieckz:
hmm, I wanted to write this comment in the #1144 thread :D

src/sgame/sg_maprotation.cpp
slipher: some mixed tabs/spaces indentation or something here

src/sgame/sg_maprotation.cpp
```diff
@@ -1042,6 +1042,14 @@ Send commands to the server to actually change the map
 static void G_IssueMapChange( int index, int rotation )
 {
 	mrNode_t *node = mapRotations.rotations[ rotation ].nodes[ index ];
+  mrCondition_t *condition;
```
slipher: Prefer to declare variables in the narrowest possible scope (inside the `if` block in this case)

illwieckz:
My server is currently using this code and it works, and I see no unattended side effect.

See #1146 for an issue that make this code not really useful: `numClients` currently count bots…

Issue #1146, opened 2019-12-29
`if numClients` in map rotation counts bots


illwieckz:
`if numClients` in map rotation counts bots, which makes it almost useless

illwieckz:
would be better if it does not count bots

illwieckz:
or add `numPlayers` alongside `numClients`

illwieckz:
fixed by #1147

PR #1147, opened 2019-12-31
add "numPlayers" as condition in maprotation

Dimitriio:
Hi,

Issue #1146 : 
I just changed from `numConnectedClients` to `numPlayingPlayers` when the condition is verified.

I am new on the project so I didn't really know if I could test it easily in dev mode.
(I knew it from when I was playing Tremulous times ago).

### DIFF
--- a/src/sgame/sg_struct.h
+++ b/src/sgame/sg_struct.h
@@ -624,6 +624,7 @@ struct level_locals_s
 	bool restarted; // waiting for a map_restart to fire
 
 	int      numConnectedClients; // connected
+	int      numConnectedPlayers; // connected players (not bot) in a team or spec mode
 	int      numAliveClients;     // on a team and alive
 	int      numPlayingClients;   // on a team
 	int      numPlayingPlayers;   // on a team and not a bot
--- a/src/sgame/sg_struct.h
+++ b/src/sgame/sg_struct.h
@@ -623,7 +623,7 @@ struct level_locals_s
 
 	bool restarted; // waiting for a map_restart to fire
 
-	int      numConnectedClients; // connected
+	int      numConnectedClients; // connected clients (players + bots)
 	int      numConnectedPlayers; // connected players (not bot) in a team or spec mode
 	int      numAliveClients;     // on a team and alive
 	int      numPlayingClients;   // on a team
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,10 @@ src/sgame/backend
 
 #ignore clion project files
 .idea/
+
+#ignore visual studio code project files
+.vscode/
+
 #ignore qtcreator project files
 CMakeLists.txt.user
 #ignore external deps
--- a/archlinux/configs/maprotation.cfg
+++ b/archlinux/configs/maprotation.cfg
@@ -8,7 +8,7 @@ rotation1
 	plat23
 	spacetracks
 	station15
-	if numClients < 12 #skip-thunder
+	if numPlayers < 12 #skip-thunder
 		thunder
 	#skip-thunder
 	yocto
--- a/debian/configfiles/maprotation.cfg
+++ b/debian/configfiles/maprotation.cfg
@@ -8,7 +8,7 @@ rotation1
 	plat23
 	spacetracks
 	station15
-	if numClients < 12 #skip-thunder
+	if numPlayers < 12 #skip-thunder
 		thunder
 	#skip-thunder
 	yocto
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -1421,6 +1421,7 @@ void CalculateRanks()
 	char P[ MAX_CLIENTS + 1 ] = "", B[ MAX_CLIENTS + 1 ] = "";
 
 	level.numConnectedClients = 0;
+	level.numConnectedPlayers = 0;
 	level.numPlayingClients   = 0;
 	level.numPlayingPlayers   = 0;
 	level.numPlayingBots      = 0;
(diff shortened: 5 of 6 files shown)


illwieckz:
Hi, welcome! :-)

You can do: `daemonded -set vm.sgame.type 3` to load a shared lib (`.so`/`.dll`) sgame.

You can then set this in `config/server.cfg`:

```
set g_mapConfigs "map"
```

and this in `config/map/default.cfg`:

```
bot fill 3 humans
bot fill 3 aliens
```

To load bots at map startup.

See my work-in-progress [scripts branch](https://github.com/illwieckz/Unvanquished/tree/scripts/dist/configs) for more config samples.

illwieckz:
Also, maybe it would makes sense to rename `numClients` to `numPlayers` or something like that. Because the bug is not that the code does not do what it tells to do, but that it does something that seems useless.

Dimitriio:
@illwieckz do you prefer I totally remove the `numClients` condition as its not really useful or I keep it to not make a breaking change in the configuration of the maprotation ?

illwieckz:
Well, a safe choice may be to add `numPlayers` condition alongside existing `numClients`.
There may be uncommon but useful need for counting bots as players (numClients), I think about some training server we had that relied heavily on bots.

slipher:
`numPlayingPlayers` is not really what you want, since it does not count players on the spectator team. The `CalculateRanks` function has some good information.  Maybe you could try subtracting the values of level.team[*].numPlayers instead.

Dimitriio:
@illwieckz I couldnt test my changes. I tried use the config you told me but I dont really know where to put them. I tried to put it where HOMEPATH is but when I launch the server, it stops with `Idle server with no map — triggering watchdog`.

illwieckz:
You have to load a map, or load a config that loads a map:

```sh
./daemon -set vm.sgame.type 3 -exec server.cfg
```

slipher:
> You have to load a map, or load a config that loads a map:
> 
> ```shell
> ./daemon -set vm.sgame.type 3 -exec server.cfg
> ```

Also make sure to build the `sgame-native-dll` target each time you change the code, else the changes won't be tested.

illwieckz:
Yep, make sure you pass `-DBUILD_GAME_NATIVE_DLL=ON` to first cmake call.

Dimitriio:
I have took the files from your repo and I try `./build/daemonded -set vm.sgame.type 3 +exec <path to server.cfg>` and now i have `couldn't exec '<path to server.cfg' >`

illwieckz:
Paths are:

HOMEDIR/config/server.cfg
HOMEDIR/config/map/default.cfg
HOMEDIR/game/maprotation.cfg

I'm not sure about the path to Unvanquished HOMEDIR is on Windows, it is probably `%HOMEPATH%\Documents\My Games\Unvanquished`

Dimitriio:
I am on Debian system so I don't know neithe hehe.
The root path I use is `/home/dimitri/.local/share/unvanquished`

illwieckz:
so paths are:

```sh
/home/dimitri/.local/share/unvanquished/config/server.cfg
/home/dimitri/.local/share/unvanquished/config/map/default.cfg
/home/dimitri/.local/share/unvanquished/game/maprotation.cfg
```

Dimitriio:
Alright !!! Seems like it's good now !!! 
Thanks for the help ! I was near the solution but a bit lost too. Should be okay now.

And this option have to be passed everytime i rebuild ? or `cmake --build build -- -j4` is enough ?
> Yep, make sure you pass `-DBUILD_GAME_NATIVE_DLL=ON` to first cmake call.

illwieckz:
You do once:

```sh
cmake -H. -Bbuild -DBUILD_GAME_NATIVE_DLL=ON
```

then, as usual, you do any time you want:

```sh
cmake --build build -- -j4
```

Dimitriio:
In [src/sgame/sg_main.cpp](https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_main.cpp) there is this part in the _CalculateRanks_ method: 
```
        // TODO: Use TEAM_ALL or the latter version for this everywhere
	level.team[ TEAM_NONE ].numPlayers += level.numPlayingPlayers;
```
that make my changes not working when there were players in the Human/Alien team as they are count in both spec & the team.

I can try do it if while I am on it. Or change my calculation rule but will need to be changed again when the quoted TODO will be done.

Gireen:
you could add a level.numConnectedPlayers and let it only count humans in CalculateRanks
which could also be useful in other places

src/sgame/sg_struct.h
```diff
@@ -624,6 +624,7 @@ struct level_locals_s
 	bool restarted; // waiting for a map_restart to fire
 
 	int      numConnectedClients; // connected
```
illwieckz: while you're at it, maybe you can also edit that other comment as `// connected clients (players + bots)` or something like that, to be more explicit.

Dimitriio: Sure, I can do that too :)

illwieckz:
That looks great, can you edit the PR title? :-)

Dimitriio:
I change the title to something more explicit. Let me know if it does not suffice.

illwieckz:
that's good!

illwieckz:
my server is running both this change and the one from #1145 and it works as expected:

[numPlayers map rotation condition](https://dl.illwieckz.net/b/unvanquished/numplayers/unvanquished_2020-01-01_193934_000.jpg)

[numPlayers map rotation condition](https://dl.illwieckz.net/b/unvanquished/numplayers/unvanquished_2020-01-01_194159_000.jpg)

[numPlayers map rotation condition](https://dl.illwieckz.net/b/unvanquished/numplayers/unvanquished_2020-01-01_194302_000.jpg)

good job guys!

illwieckz:
@Dimitriio, can you rebase on master and squash those two commits as one (you only need to keep the second message):

- _MapRotation : counting realPlayer in the numClients condition_
- _MapRotation : Add 'numPlayers' condition_

illwieckz:
Thank you for making Unvanquished great again :-)

PR #1158, opened 2020-02-22
display building healthbars for everyone but only when damaged

Gireen:
Display building health bars for everyone not only builders.
Also display them only when buildings are damaged.

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -2130,6 +2130,10 @@ void CG_DrawBuildableStatus()
 {
 	centity_t     *cent;
 	entityState_t *es;
+	const buildableAttributes_t *attr;
+	const playerState_t *ps = &cg.predictedPlayerState;
+	int team = ps->persistant[ PERS_TEAM ];
+
 	int           buildableList[ MAX_ENTITIES_IN_SNAPSHOT ];
     unsigned      buildables = 0;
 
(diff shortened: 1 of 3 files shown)


slipher:
It might be a bit visually cluttered when fighting. I'd have to try it and see.

Gireen:
It runs at the moment on the experimental server. 
To get even close cluttered it would need many buildings very close together and all of them to be damaged at the same time. Which is unlikely but even then only for short until they are repaired/destroyed.
Small change added that builders see always the health of marked buildings.

Gireen:
Now by default only Overmind and Reactor health is shown to the team when damaged.
More can be enabled with cg_drawBuildableHealth howmuch one likes.

slipher:
Setting the cvar to 0 should hided all the health bars.

Is there really a need for builders to hide the health bars of any buildings?

How about this:

```
if ( ( cg_drawBuildableHealth >= 1 && ( isBuilder || ( mainBuilding && damagedBuilding ) ) ) ||
     ( cg_drawBuildableHealth >= 2 && damagedBuilding ) )
```

src/cgame/cg_buildable.cpp
```diff
@@ -2142,8 +2146,22 @@ void CG_DrawBuildableStatus()
...
+		attr = BG_Buildable( es->modelindex );
+
+		if ( es->eType != entityType_t::ET_BUILDABLE || team != attr->team )
+		{
+			continue;
+		}
+
+		bool isBuilder = CG_PlayerIsBuilder( (buildable_t) es->modelindex );
+		bool markedBuilding = es->eFlags & EF_B_MARKED;
+		bool damagedBuilding = ( (float) es->generic1 / (float) attr->health ) < 1;
```
slipher, line 2158: Instead of less than 1 could the threshold be 0.7 or something? It's not very concerning if the buildable lost 1 hp; plus it's not a good practice to be comparing a floating point to the exact value of 1.

Gireen:
> Setting the cvar to 0 should hided all the health bars.

it does already here

https://github.com/Unvanquished/Unvanquished/blob/046faa5d70891353dcedfe52b5f60a614eac4a26/src/cgame/cg_buildable.cpp#L2140-L2143


> Is there really a need for builders to hide the health bars of any buildings?

To make it less cluttered. Is there a need for builders to see the health bars all the time?
In the end it can be decided by the user. what is missing is the combination that the builder sees the bars all the time and others when damaged, that could also be added.

src/cgame/cg_buildable.cpp
Gireen: Is 0.99 or 0.95 also ok?
I mean its the whole point to see if health is less than 100%
To see the OM/RC build progress it would be good to get close to full health with the displayed range.

slipher: I certainly agree that builders should always see the progress bar for a building under construction. But for the case of a damaged building that lost a tiny amount of health, it is not so interesting for players to see. To me it even would be better to see all the health bars rather than only ones less than 99%.

Gireen: Well that's good, the default settings still work like that. And who wants can change it the way they prefer.
I think its useful information to see if a building is attacked, especially for not builders.
To allow more flexibility the threshold could also be made a cvar that leaves this decision to the player.

slipher: > I think its useful information to see if a building is attacked, especially for not builders.

Yes, it could be useful to see this, in order to know if the building is at risk of getting destroyed or needs repairing. But if you have for example a barb splash attack that does 1 hp of damage to all buildings and you see health bars for all of them, I don't see how this is relevant information. For showing damage I would want to have the threshold significantly lower than 100%.

Gireen:
Now every combination is possible
cg_drawBuildableHealth
1 = as before
2 = only for builders when damaged 
3 = always for builders, for others when damaged
4 = for everyone when damaged

and cg_drawBuildableHealthThreshold can be changed to how low one wants.
By default its 0.98 since it only affects the main buildings and for them precise information is important and cant clutter anything.😄

illwieckz:
> cg_drawBuildableHealth
> 1 = as before
> 2 = only for builders when damaged
> 3 = always for builders, for others when damaged
> 4 = for everyone when damaged

I assume `0` is to not draw anything anytime?

otherwise, LGTM.

Gireen:
> I assume 0 is to not draw anything anytime?

correct

slipher:
The meanings of `2` and `3` are swapped in the code vs. the above comment, but I suppose the order doesn't matter. So the code LGTM semantically. Please add a comment to explain the meanings of the possible options for cg_drawBuildableHealth before merging.

bmorel:
This PR requires syncing with current master.

PR #1159, opened 2020-02-24
always display player name in beacon

Gireen:
independent of team

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -1596,13 +1596,8 @@ static void CG_ScanForCrosshairEntity()
 			else if ( targetTeam != TEAM_NONE )
 			{
 				cg.crosshairFoe = true;
-
-				if ( ownTeam == TEAM_NONE )
-				{
-					// spectating, so show the name
-					cg.crosshairClientNum = trace.entityNum;
-					cg.crosshairClientTime = cg.time;
-				}
+				cg.crosshairClientNum = trace.entityNum;
+				cg.crosshairClientTime = cg.time;
 			}
 		}
 	}


illwieckz:
can you make a screenshot?

slipher:
I don't think this information should be available to enemies.

Gireen:
@illwieckz 🤔 on the right side
name

@slipher why? what tactical benefit does it give to enemys? It is also common for shooters that you can see names of the opponents.

DolceTriade:
You can avoid better players.

illwieckz:
If I'm right, in your screenshot the opponent's health is disclosed.

If we want to display player information (name, health, weapon) in beacon, I think it would be better to limit it to team mates.

Unlike usual shooters, Unvanquished is a _strategic_ shooter, and _information_ is a whole battlefield per se.

Gireen:
I mean the beacon text on the right side in this case `[B]Gireen`.
Normal it would say Food instead. And from the human view it shows the alien class name.

This change would only show the name instead these placeholders. 

An other option is to show the name only when seeing the enemy direct with the crosshair instead a beacon with the current position. Like `[B]EmperorJack` in the image but without health.

Its also common for Tactical Shooters and RTS/FPS mixed games to see the name of enemy players. Like Battlefield, Enemy Territory: Quake Wars, Renegade, Natural Selection and Savage to name a few.

slipher:
Yeah, some other games do this... Like in Counter-Strike, if the enemy is hiding in a cloud of smoke, you can look at the cloud, and if you happen to look directly at the enemy, their presence will be revealed by the appearance of their name, and you can shoot them. I think this is stupid :stuck_out_tongue: 

I also don't favor enabling the strategy of seeing which enemy in a group is the most skilled, and targeting them first.

DolceTriade:
I'd prefer we not show the names of the enemy team.

PR #1163, opened 2020-03-01
Allow devolving around Overmind

Gireen:
Allow devolving around the Overmind and return morph points relative to the percent of health left * 0.9.

AdvDragoons spawn only with 1 barb since it could be abused to evolve back and forth for a quick way to ammo.


DolceTriade:
How does allowing 1 ammo prevent abuse? You can still evolve back and forth with one ammo? Otherwise, I think this change looks good.

Gireen:
Yes, but its much less effective. It would require to run back and forth from human base to overmind witch is in the best case a bit faster then just waiting for one barb to regenerate.

PR #1106, opened 2019-02-14
responsive: compute font size from screen size

illwieckz:
This is an attempt to set default size relative to screen size.

Default libRocket font size is `12px`, see [StyleSheetSpecification.cpp:246](https://github.com/Unvanquished/libRocket/blob/7f7c34e67dee17bf2c1cb3e61fe6d847698267a3/Source/Core/StyleSheetSpecification.cpp#L246):

```c++
	RegisterProperty(FONT_SIZE, "12", true, true).AddParser("number");
```

Since FullHD is the most common screen resolution today the game was designed against that resolution, so we must compute a default font size that is 12px on FullHD screen.

That is the compute:

```
font_size = min(width, height)/90
```

This PR is considered WIP because it relies on `r_customWidth/Height` cvars, it would be better to read the screen size directly, especially since that's possible those values are not correctly set if not custom size is set (does it exists anymore, though?). So I need a way to get a proper screen size at this place of the code, other wise it's ok.

Use this branch https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/6 on asset side (required to render it properly).

See also https://github.com/DaemonEngine/Daemon/pull/161 for similar change on engine side (not required).

PR #1167, opened 2020-03-18
macos: attempt to fix partial rendering issue on macOS Catalina, ref #1166

### LINKED ISSUE #1166: The game only draws to the bottom left of the window on macOS Catalina, 10.15.3
I've just managed to build the game on macOS Catalina, and have noticed similar problems to other quake-based games, including Tremulous. When launching the game, it will launch only drawing in the bottom left hand corner of the window with the rest of the window flickering red/black. I believe this is something to do with High DPI mode. I also did a bit of research and found these posts;

https://github.com/redeclipse/base/issues/920
https://github.com/ioquake/ioq3/issues/422
https://github.com/godotengine/godot/issues/32805

Related:
https://hg.libsdl.org/SDL/rev/46b094f7d20e

Screenshot below;
image

illwieckz:
Note that it would only works when started with .app launcher.

Note that if the `NSHighResolutionCapable` is set to `false`,
it does not mean Unvanquished does not handles high resolution.

Unvanquished has responsive UI which scales UI elements whatever
the resolution, see:

- https://github.com/Unvanquished/Unvanquished/pull/1106
- https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/6
- https://github.com/DaemonEngine/Daemon/pull/161

Unvanquished does not need to use any macOS-specific option to handle
high resolution screens, the way things are done accross all platforms
is done the “one ring to rule them all” way.

### DIFF
--- a/macosx/Info.plist
+++ b/macosx/Info.plist
@@ -31,5 +31,7 @@
 	</array>
 	<key>CFBundleVersion</key>
 	<string>0.51</string>
+	<key>NSHighResolutionCapable</key>
+	<false/>
 </dict>
 </plist>


illwieckz:
@cu-kai please test :-)

illwieckz:
Hmm, this looks to only be a workaround, since the game would still render badly when launched from command line or from updater.

DolceTriade:
The updater launches the .app which should respect that.

illwieckz:
At least, is this change useful?

cu-kai:
adding
```
	<key>NSHighResolutionCapable</key>
	<false/>
```

to the `info.plist` _does_ fix it, but a code fix would be more ideal.

illwieckz:
So at least we can do that for next release, then average people will get it right.

I don't know what a _code fix_ is in such case. I almost know nothing on macOS.

illwieckz:
> For what it's worth, if you compile the game on any macOS older than Catalina, it'll scale fine.
> 
> I believe the real fix is here: https://hg.libsdl.org/SDL/rev/46b094f7d20e

Does it mean compiling against a newer SDL fixes the thing?

DolceTriade:
This can't hurt.

PR #1170, opened 2020-04-11
select alien class description independent of display order

### LINKED ISSUE #1165: Evolve menu class descriptions out of order
The class name and description in the evolve menu are out of order
For the Dretch are AdvGranger informations displayed and from there always from the class before.
This is since #1138 
The change should work but it seems that `setDS alienEvolveList` has no effect at all
https://github.com/Unvanquished/Unvanquished/blob/adc6e154435a71828bb26362de6f647d17990c6a/src/cgame/cg_rocket_dataformatter.cpp#L425

Gireen:
fix #1165
this solves the current problem.
setDS is also used on other places where is probably also doesn't work


illwieckz:
Tested it on my end, it works, LGTM.

PR #1172, opened 2020-04-12
PVS Studio fixes

dimhotepus:
Some fixes from PVS Studio static analyzer.

Product page: https://www.viva64.com/en/pvs-studio/
How to use for free: https://www.viva64.com/en/b/0600/


src/cgame/cg_main.cpp
```diff
@@ -989,7 +989,7 @@ void CG_RegisterGrading( int slot, const char *str )
 		return;
 	}
 
-	sscanf(str, "%d %f %s", &model, &dist, texture);
+	sscanf(str, "%d %f %63s", &model, &dist, texture);
```
slipher, line 992: Better not to hard-code the size. You can use `%*s` and pass the number of characters size in an argument.

src/shared/bg_parse.cpp
```diff
@@ -803,7 +803,7 @@ void BG_ParseBuildableAttributeFile( const char *filename, buildableAttributes_t
 	else if ( !( defined & BUILDTIME) ) { token = "buildTime"; }
 	else if ( !( defined & NORMAL) ) { token = "minNormal"; }
 
-	if ( strlen( token ) > 0 )
+	if ( !Util::is_empty(token) )
```
slipher: You could just do `if (*token)` which is a common idiom in the codebase. Also please rebase when fixing this since we prefer not to have any commits which fail to compile.

src/shared/bg_teamprogress.cpp
```diff
@@ -272,8 +272,10 @@ void BG_ImportUnlockablesFromMask( int team, int mask )
 	lastMask = mask;
 	lastTeam = (team_t) team;
 
+#ifdef BUILD_CGAME
 	// no status change yet
 	memset( statusChanges, 0, sizeof( statusChanges ) );
```
slipher, line 277: Also move the declaration of statusChanges inside the ifdef?

src/sgame/sg_buildable.cpp
```diff
@@ -2388,7 +2389,7 @@ void G_LayoutLoad()
 		if ( *layout == '\n' )
 		{
 			i = 0;
-			sscanf( line, "%s %f %f %f %f %f %f %f %f %f %f %f %f\n",
+			sscanf( line, "%1023s %f %f %f %f %f %f %f %f %f %f %f %f\n",
```
slipher, line 2392: Use `%*s`

src/shared/bg_teamprogress.cpp
dimhotepus: Done

src/shared/bg_parse.cpp
dimhotepus: Done

src/cgame/cg_main.cpp
dimhotepus: Mmm, not sure about this.

From http://www.cplusplus.com/reference/cstdio/scanf/:
> * starting asterisk indicates that the data is to be read from the stream  but ignored (i.e. it is not stored in the location pointed by an argument).

Originally i was thinking about this:

```
#define STR2(x) #x
#define STR(X) STR2(X)

#define MAXVAL 5

char string[MAXVAL + 1];
scanf("%" STR(MAXVAL) "s", string);
```

but it looks like such macroses should be stored in daemon, since they may be useful in general. And i don't know either we will update daemon to the latest version or not.

dimhotepus: https://stackoverflow.com/questions/25410690/scanf-variable-length-specifier

slipher: My mistake, apparently it does not work the same way as printf.

The macro way seems not so nice either due to the need to add/subtract 1. However if you wish to pursue it, the stringification macros are [here](https://github.com/DaemonEngine/Daemon/blob/master/src/engine/qcommon/q_shared.h#L173).

I'd say it's OK to merge in the current form with hard-coded sizes. All the code in the Unvanquished repo normally runs in a secure sandbox, so we can have lower standards for secure coding than the code in Daemon.

dimhotepus:
@slipher Looks like all issues are fixed, thank you for suggestions.

src/shared/bg_parse.cpp
```diff
@@ -803,7 +803,7 @@ void BG_ParseBuildableAttributeFile( const char *filename, buildableAttributes_t
 	else if ( !( defined & BUILDTIME) ) { token = "buildTime"; }
 	else if ( !( defined & NORMAL) ) { token = "minNormal"; }
 
-	if ( !Util::is_empty(token) )
+	if ( strlen( token ) > 0 )
```
illwieckz: why doing the change and reverting it? what do you mean by "till daemon is updated" ?

slipher: I guess the idea was to define an `is_empty` function in Daemon.

We could merge using the squash option to avoid putting this into the history. Sounds good?

tsarevichdm: Squash is good.
Now replaced with `if ( *token )`, which doesn't require anything new to be added.

Out of curiosity, when will we update daemon to the upstream version?

dimhotepus: Squash is good.
Now replaced with if ( *token ), which doesn't require anything new to be added.

Out of curiosity, when will we update daemon to the upstream version?

dimhotepus: Oh, i see, @illwieckz already updated daemon. Thank you!

illwieckz: At least, I would move commits in a way they follow immediately (maybe use “revert” word instead of “restore”). Like commit 1, commit 2, commit revert. One day we can revert the revert the day the feature is available. With squash we would lose that.

illwieckz: Note that if you want to implement `is_empty` in engine and thinks it's worth of it, don't hesitate to do it, then it's fine if you update the submodule yourself in the pr.

dimhotepus: Removed affected commits, now history is clean. Now `is_empty` looks too local 4 me to add to daemon. 

But thanks for suggestions!

dimhotepus:
Looks good for me, any issues/suggestions?

DolceTriade:
LGTM. @slipher any final thoughts?

PR #1173, opened 2020-04-13
Build depends on lua, add it to deps

dimhotepus:
See https://github.com/Unvanquished/Unvanquished/blob/master/CMakeLists.txt#L90
for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Visual Studio/MSVC (at least Visual Studio 2017).
 `libjpeg` ≥ 8,
 `libwebp` ≥ 0.2.0,
 `Freetype`,
+`Lua`,
 `OpenAL`,
 `libogg`,
 `libvorbis`,
--- a/debian/control
+++ b/debian/control
@@ -5,6 +5,7 @@ Maintainer: Darren Salt <devspam@moreofthesa.me.uk>
 Build-Depends: debhelper (>= 8.0.0), g++ (>= 4.7) | g++-4.7, po-debconf, cmake,
 	libcurl4-gnutls-dev,
 	libfreetype6-dev,
+	liblua5.2-dev,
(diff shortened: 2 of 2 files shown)


slipher:
I think 5.3 is the required version.

dimhotepus:
> I think 5.3 is the required version.

https://github.com/Unvanquished/Unvanquished/blob/master/.travis.yml#L45

Works with 5.2 for me on Debian 4.19.98

slipher:
I recall there being some user reports of errors with an undefined symbol `lua_rotate` (a function added in version 5.3), although I can't find them now. So to check whether it is working, you need to launch the game with the arguments `-set vm.cgame.type 3`, which runs the client gamelogic as a native DLL (since for NaCl and Windows builds a precompiled version is used rather than the system library).

dimhotepus:
> I recall there being some user reports of errors with an undefined symbol `lua_rotate` (a function added in version 5.3), although I can't find them now. So to check whether it is working, you need to launch the game with the arguments `-set vm.cgame.type 3`, which runs the client gamelogic as a native DLL (since for NaCl and Windows builds a precompiled version is used rather than the system library).

```
$ readelf -d cgame-native-dll.so
0x0000000000000001 (NEEDED)             Shared library: [liblua5.2.so.0]
```

```
Loading VM module /home/dimhotepus/github/Unvanquished/cmake-build-debug/cgame-native-dll.so...
Loaded VM module in 13 msec
]Loading Lua interpreter
Loaded font face Roboto Bold (from fonts/Roboto-Bold.ttf).
...
UnnamedPlayer#1 entered the game
```

Playable for me.

slipher:
So we're good to go, thanks for testing.

DolceTriade:
5.3 works. That lua_rotate bug was a bug in CMake.

PR #1175, opened 2020-04-23
Deletions for version 0.52.0

slipher:


.appveyor.yml
```diff
@@ -33,6 +33,11 @@ build:
 
 before_build:
   - git submodule update --init --recursive
+  - curl -sS https://gitlab.com/illwieckz/git-checkout-modules/raw/master/git-checkout-modules -o git-checkout-modules
+  - bash git-checkout-modules
+    --only-sub=%APPVEYOR_REPO_BRANCH%:has=/sync$
+    --only-sub=%APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH%:has/sync$
```
illwieckz, line 39:
```diff
- --only-sub=%APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH%:has/sync$
+ --only-sub=%APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH%:has=/sync$
```

Anyway that must not be the cause of your issue since this branch does not sync, but there will be trouble in the future because of this mistake.

illwieckz: It still weird to see it manages to checkout something with `for52del:has/sync`:

```
Checkout submodules with reference 'for52del:has/sync$'
Submodule path 'daemon': checked out 'd07f61c37ff5735e8fbda52253c2b62f3990fcbd'
Submodule path 'pkg/unvanquished_src.dpkdir': checked out 'edcaa564d190ab0ef14e758ac9671e2b6fb161cd'
```

It's clearly modifying the references, those may be already right before.

illwieckz: Yep, for some reason I don't know yet it fetch `HEAD` when such garbage is passed. Put a `=` and it will work. At least it worked on my end by doing the exact same commands as appveyor.

illwieckz: OK I now why, in case of mistake like this, it reverts to `HEAD`, because it always run `git submodule update --init --recursive` in submodules before checking out, and if the branch is not found (because wrong), it's expected it does nothing.

PR #1176, opened 2020-04-24
Use git-checkout-modules script for /sync submodules

slipher:
Tested that it works on the 0.52.0/sync branch: https://ci.appveyor.com/project/DolceTriade/unvanquished/builds/32419864

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -32,7 +32,14 @@ build:
   parallel: true
 
 before_build:
-  - git submodule update --init --recursive
+  - curl -sS https://gitlab.com/illwieckz/git-checkout-modules/raw/master/git-checkout-modules -o git-checkout-modules
+  # The git-checkout-modules command below is normally equivalent to 'git submodule update --init --recursive',
+  # but in the case that the branch name ends with '/sync' it tries to find branches with the same name.
+  - bash git-checkout-modules
+    --update
+    --sub-ref=%APPVEYOR_REPO_BRANCH%:has=/sync$
+    --sub-ref=%APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH%:has=/sync$
+    --print
   - ps: if (-not(Test-Path($env:PYTHON))) { & appveyor\install.ps1 }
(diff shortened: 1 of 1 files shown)


slipher:
Apparently it does not work on a non-sync branch if I remove the preliminary `git submodule sync` command. @illwieckz Should this be considered a bug in the script, or is the user expected to ensure that all submodules are checked out before using it?

illwieckz:
It's the **requested feature** to do nothing if the branch name does fit the pattern. You can add the `--update` command before the `has` things to checkout submodules, this way:

```
git-checkout-modules --update --only-sub=%APPVEYOR_REPO_BRANCH%:has=/sync$ --only-sub=%APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH%:has=/sync$ --print
```

Note that `git-checkout-modules --update` does the same as `git submodule update --init --recursive`.

Anyway I recommend to keep the `git submodule update --init --recursive` call in travis/appveyor scripts because this will always work and git has more chance to be maintained longer than my tools. If one day something goes wrong in my tool and there is no one to fix it, the call to my tool would be removable from the build script without breaking the ability to build the tree. Only the sync feature would be lost, but in other case that would still build.

illwieckz:
Note: I reworked a bit my tool, and `--only-sub` had been renamed to `--sub-ref` to be more consistent with other options. Compatibility with `--only-sub` is kept but it would be better to rename it.

.appveyor.yml
```diff
@@ -32,7 +32,14 @@ build:
   parallel: true
 
 before_build:
-  - git submodule update --init --recursive
+  - curl -sS https://gitlab.com/illwieckz/git-checkout-modules/raw/master/git-checkout-modules -o git-checkout-modules
+  # The git-checkout-modules command is normally equivalent to 'git submodule update --init --recursive',
```
illwieckz, line 36: `git-checkout-modules --update` is equivalent to `git submodule update --init --recursive`, not `git-checkout-modules` alone.

slipher: With the 'git-checkout-modules command' I am referring to the entire 5 lines. Maybe I will s/command/command below/.

Issue #1181, opened 2020-04-25
Tutorial: key to follow player in spectator mode is not displayed until a player is followed


illwieckz:
See:

[![follow key](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial/unvanquished_2020-04-24_211753_001.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial/unvanquished_2020-04-24_211753_001.jpg)

Note: #1182 is gonna fix it.

illwieckz:
Fixed by  #1191.

Issue #544, opened 2015-01-30
Create new default key bindings.


Viech:
The old bindings are both outdated (e.g. item selection) and inefficient.

lamefun:
The truth has finally been seen!

Veyrdite:
Using 'x' for sprint is very awkward if you want to strafe (ie hold W and A) at the same time.  Moving this to the shift key make it much easier.

I remember you voicing issues about using the Y/Z keys before due to them being swapping in some keyboard layouts.  Do we want to officially support non-QWERTY layouts in the single set of default binds?

Viech:
I think QWERTZ is quite common, though I don't know any numbers.

mbasaglia:
If we can access the raw key id we could generate bindings depending on the physical key position instead if the associated character

illwieckz:
it has to be done for next release.

this is a proposition:

```
[‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾]
[ ~      ] [ 1      ] [ 2      ] [ 3      ] [ 4      ] [ 5      ] [ 6      ]
[ consol ] [        ] [        ] [ barb   ] [        ] [        ] [        ]
[        ] [ blastr ] [ weapon ] [ nade   ] [        ] [        ] [        ]
[________] [________] [________] [________] [________] [________] [________]

[‾‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾] [‾‾‾‾‾‾‾‾]
[ Tab     ] [ Q      ] [ W      ] [ E      ] [ R      ] [ T      ] [ Y      ]
[ score   ] [        ] [ forwd  ] [        ] [ evolve ] [ talk   ] [ ttalk  ]
[         ] [ medkit ] [        ] [ reload ] [ buy    ] [        ] [        ]
[_________] [________] [________] [________] [________] [________] [________]
 
[...]

PR #1182, opened 2020-04-25
Fix and use attack3 for barb/spit, mark and deconstruct with mouse, cycle weapons, rework binds and tutorial, fix #1179, #1180, #1181, #1183, ref #544

### LINKED ISSUE #1179: Do not hijack CG_NextWeapon_f / CG_PrevWeapon_f and weapnext/weapprev for inventory selection
This is a standalone issue related to https://github.com/Unvanquished/Unvanquished/issues/544#issuecomment-619299752

Historically the code was hijacking `CG_NextWeapon_f` and CG_PrevWeapon_f` functions to select the next or previous inventory item, whatever it's a weapon or not. For example buying a weapon would call `CG_NextWeapon_f` at some point.

Also, the code hijacked the `weapnext`/`weapprev` commands to select inventory items in game.

To be able to implement true `CG_NextWeapon_f` and `CG_PrevWeapon_f` and give `weapnext`/`weapprev` a behavior that fits the name, we need to rename legacy `CG_NextWeapon_f` and `CG_PrevWeapon_f` to some other name like `CG_SelectNextInventoryItem_f`/`CG_SelectPrevtInventoryItem_f`.

### LINKED ISSUE #1180: Do not hijack useitem for attack3, use attack3
This issue is related to #544.

The code is currently able to handle `attack3` fire events, `attack3` animation etc.
Aliens already use attack3, for example the dragoon uses `attack1` to bite, `attack2` to bounce, and attack3 to barb, while the granger uses `attack1` to build, `attack2` to bite, `attack3` to spit, etc.

But there is no command for this. Instead, the code abuses `useitem`. As soon as we create an attack3 command properly those `attack3` actions works.

Even the spectator follow mode abuses `useitem` to make an `attack3`, the bind is for `useitem` (while there is no `attack3` command), but the code reads `attack3` events.

On the other hand the human grenade uses `itemact grenade`, not `attack3` (yet), but it does not prevent to fix attack3, at least it is required to fix `attack3` to make the grenade a third weapon like the dragoon barb.

Like #1179, I'm looking at the code for ways to implement workarounds for #544, and I find workarounds to remove instead.

illwieckz:
Requires similar branches on `Dæmon`, `unvanquished_src.dpkdir` and `res-weapons_src.dpkdir`.

- https://github.com/DaemonEngine/Daemon/pull/330  
- https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/15

Fix #1179, #1180, #1181, #1183, ref #544.

### DIFF
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -581,8 +581,8 @@ void CG_InitConsoleCommands()
 
 	// Defined in src/engine/qcommon/q_shared.h, see BUTTON_ATTACK etc.
 	trap_RegisterButtonCommands(
-	    // 0      1       2       3     45      6        7       89ABCD           E      <- bit nos.
-	      "attack,attack2,attack3,taunt,,sprint,activate,useitem,,,,,,deconstruct,rally"
+	    // 0      1       2       3     45      6        789ABCD           E      <- bit nos.
+	      "attack,attack2,attack3,taunt,,sprint,activate,,,,,,,deconstruct,rally"
 	    );
 }
 
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -581,8 +581,8 @@ void CG_InitConsoleCommands()
 
 	// Defined in src/engine/qcommon/q_shared.h, see BUTTON_ATTACK etc.
 	trap_RegisterButtonCommands(
-	    // 0      1       2       3     45      6        789ABCD           E      <- bit nos.
-	      "attack,attack2,attack3,taunt,,sprint,activate,,,,,,,deconstruct,rally"
+	    // 0      1       2       3     45      6        789ABCD           E    F  <- bit nos.
+	      "attack,attack2,attack3,taunt,,sprint,activate,,,,,,,deconstruct,rally,"
 	    );
 }
 
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -579,9 +579,10 @@ void CG_InitConsoleCommands()
 		trap_AddCommand( commands[ i ].cmd );
 	}
 
+	// Defined in src/engine/qcommon/q_shared.h, see BUTTON_ATTACK etc.
 	trap_RegisterButtonCommands(
-	    // 0      12       3     45      6        78       9ABCD           E      <- bit nos.
-	      "attack,,useitem,taunt,,sprint,activate,,attack2,,,,,deconstruct,rally"
+	    // 0      1       2       3     45      6        7       89ABCD           E      <- bit nos.
+	      "attack,attack2,attack3,taunt,,sprint,activate,useitem,,,,,,deconstruct,rally"
 	    );
 }
 
--- a/src/cgame/cg_tutorial.cpp
+++ b/src/cgame/cg_tutorial.cpp
@@ -449,7 +449,7 @@ static void CG_HumanText( char *text, playerState_t *ps )
 	if ( upgrade != UP_NONE )
 	{
 		Q_strcat( text, MAX_TUTORIAL_TEXT, va( _( "Press %s to throw the %s\n" ),
-			CG_KeyNameForCommand( "+itemact grenade" ),
+			CG_KeyNameForCommand( "itemact grenade" ),
 			_( BG_Upgrade( upgrade )->humanName ) ));
 	}
 
--- a/src/cgame/cg_tutorial.cpp
+++ b/src/cgame/cg_tutorial.cpp
@@ -102,7 +102,7 @@ static const char *CG_KeyNameForCommand( const char *command )
 			if ( binding.command == OPEN_CONSOLE_CMD )
 			{
 				// Hard-coded console toggle key binding
-				keyNames = "Shift-Escape";
+				keyNames = "SHIFT-ESCAPE";
 				// cl_consoleKeys is yet another source of keys for toggling the console,
 				// but it is omitted out of laziness.
 			}
--- a/src/cgame/cg_tutorial.cpp
+++ b/src/cgame/cg_tutorial.cpp
@@ -504,6 +504,10 @@ static void CG_HumanText( char *text, playerState_t *ps )
 	Q_strcat( text, MAX_TUTORIAL_TEXT,
 	          va( _( "Press %s and any direction to sprint\n" ),
 	              CG_KeyNameForCommand( "+sprint" ) ) );
+
+	Q_strcat( text, MAX_TUTORIAL_TEXT,
+	          va( _( "Press %s to crouch\n" ),
+	              CG_KeyNameForCommand( "+movedown" ) ) );
 }
 
 /*
--- a/src/cgame/cg_tutorial.cpp
+++ b/src/cgame/cg_tutorial.cpp
@@ -664,6 +664,8 @@ const char *CG_TutorialText()
 
 	if ( !cg.demoPlayback )
 	{
+		Q_strcat( text, MAX_TUTORIAL_TEXT, "\n" );
+
 		if ( cgs.clientinfo[ cg.clientNum ].team == TEAM_NONE )
 		{
 			Q_strcat( text, MAX_TUTORIAL_TEXT, va( _( "Press %s to chat\n" ), CG_KeyNameForCommand( "message_public" ) ) );
@@ -672,6 +674,9 @@ const char *CG_TutorialText()
 		{
 			Q_strcat( text, MAX_TUTORIAL_TEXT, va( _( "Press %s to chat or %s to chat to your team\n" ), CG_KeyNameForCommand( "message_public" ),  CG_KeyNameForCommand( "message_team" ) ) );
 		}
+
+		Q_strcat( text, MAX_TUTORIAL_TEXT, "\n" );
+
 		Q_strcat( text, MAX_TUTORIAL_TEXT, va( _( "Press %s to open the console\n" ), CG_KeyNameForCommand( "toggleconsole" ) ) );
 		Q_strcat( text, MAX_TUTORIAL_TEXT, _( "Press ESCAPE for the menu" ) );
 	}
--- a/src/cgame/cg_tutorial.cpp
+++ b/src/cgame/cg_tutorial.cpp
@@ -52,8 +52,7 @@ static bind_t bindings[] =
 	{ "buy ammo",       N_( "Buy Ammo" ),                              {} },
 	{ "itemact medkit", N_( "Use Medkit" ),                            {} },
 	{ "+activate",      N_( "Use Structure/Evolve" ),                  {} },
-	{ "modcase alt \"/deconstruct marked\" /deconstruct",
-	                    N_( "Deconstruct Structure" ),                 {} },
+	{ "+deconstruct",   N_( "Deconstruct Structure" ),                 {} },
 	{ "weapprev",       N_( "Previous Weapon" ),                       {} },
 	{ "weapnext",       N_( "Next Weapon" ),                           {} },
 	{ "message_public", N_( "Global chat" ),                           {} },
@@ -195,7 +194,7 @@ static void CG_BuilderText( char *text, playerState_t *ps )
 
 	if ( ( es = CG_BuildableInRange( ps, nullptr ) ) )
 	{
-		const char *key = CG_KeyNameForCommand( "modcase alt \"/deconstruct marked\" /deconstruct" );
+		const char *key = CG_KeyNameForCommand( "+deconstruct" );
 
 		if ( es->eFlags & EF_B_MARKED )
 		{
@@ -207,6 +206,9 @@ static void CG_BuilderText( char *text, playerState_t *ps )
 			Q_strcat( text, MAX_TUTORIAL_TEXT,
 					  va( _( "Press %s to mark this structure for replacement\n" ), key ) );
 		}
+
+		Q_strcat( text, MAX_TUTORIAL_TEXT,
+				  va( _( "Hold %s to deconstruct this structure\n" ), key ) );
 	}
 }
 
--- a/src/cgame/cg_tutorial.cpp
+++ b/src/cgame/cg_tutorial.cpp
@@ -41,13 +41,13 @@ static const char* const OPEN_CONSOLE_CMD = "toggleconsole";
 
 static bind_t bindings[] =
 {
-	{ "+useitem",       N_( "Activate Upgrade" ),                      {} },
 	{ "+speed",         N_( "Run/Walk" ),                              {} },
 	{ "+sprint",        N_( "Sprint" ),                                {} },
 	{ "+moveup",        N_( "Jump" ),                                  {} },
 	{ "+movedown",      N_( "Crouch" ),                                {} },
 	{ "+attack",        N_( "Primary Attack" ),                        {} },
 	{ "+attack2",       N_( "Secondary Attack" ),                      {} },
+	{ "+attack3",       N_( "Tertiary Attack" ),                       {} },
 	{ "reload",         N_( "Reload" ),                                {} },
 	{ "buy ammo",       N_( "Buy Ammo" ),                              {} },
 	{ "itemact medkit", N_( "Use Medkit" ),                            {} },
@@ -227,8 +227,8 @@ static void CG_AlienBuilderText( char *text, playerState_t *ps )
 	if ( ps->stats[ STAT_CLASS ] == PCL_ALIEN_BUILDER0_UPG )
 	{
 		Q_strcat( text, MAX_TUTORIAL_TEXT,
-		          va( _( "Press %s to launch a projectile\n" ),
-		              CG_KeyNameForCommand( "+useitem" ) ) );
+		          va( _( "Press %s to spit\n" ),
+		              CG_KeyNameForCommand( "+attack3" ) ) );
 
 		Q_strcat( text, MAX_TUTORIAL_TEXT,
 		          va( _( "Press %s to walk on walls\n" ),
@@ -307,8 +307,8 @@ static void CG_AlienLevel3Text( char *text, playerState_t *ps )
 	if ( ps->stats[ STAT_CLASS ] == PCL_ALIEN_LEVEL3_UPG )
 	{
 		Q_strcat( text, MAX_TUTORIAL_TEXT,
-		          va( _( "Press %s to launch a projectile\n" ),
-		              CG_KeyNameForCommand( "+useitem" ) ) );
+		          va( _( "Press %s to launch a barb\n" ),
+		              CG_KeyNameForCommand( "+attack3" ) ) );
 	}
 
 	Q_strcat( text, MAX_TUTORIAL_TEXT,
@@ -351,19 +351,6 @@ CG_HumanText
 */
 static void CG_HumanText( char *text, playerState_t *ps )
 {
-	const char *name;
-	upgrade_t upgrade = UP_NONE;
-
-	if ( cg.weaponSelect < 32 )
-	{
-		name = cg_weapons[ cg.weaponSelect ].humanName;
-	}
-	else
-	{
-		name = cg_upgrades[ cg.weaponSelect - 32 ].humanName;
-		upgrade = (upgrade_t) ( cg.weaponSelect - 32 );
-	}
-
 	if ( !ps->ammo && !ps->clips && !BG_Weapon( ps->weapon )->infiniteAmmo )
 	{
 		//no ammo
@@ -444,6 +431,22 @@ static void CG_HumanText( char *text, playerState_t *ps )
 		}
 	}
 
+	upgrade_t upgrade = UP_NONE;
+	for ( const auto u : { UP_GRENADE, UP_FIREBOMB } )
+	{
+		if ( BG_InventoryContainsUpgrade( u, ps->stats ) )
+		{
+			upgrade = u;
+		}
+	}
+
+	if ( upgrade != UP_NONE )
+	{
+		Q_strcat( text, MAX_TUTORIAL_TEXT, va( _( "Press %s to throw the %s\n" ),
+			CG_KeyNameForCommand( "+itemact grenade" ),
+			_( BG_Upgrade( upgrade )->humanName ) ));
+	}
+
 	// Find next weapon in inventory.
 	weapon_t nextWeapon = CG_FindNextWeapon( ps );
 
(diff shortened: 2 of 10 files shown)


illwieckz:
Note: this PR targets `0.52.0/sync` even if it builds on `master` because it relies on some modified engine stuff and maybe it breaks compatibility.

illwieckz:
There is now a real functional next/previous weapon cycling.

The tutorial also prints the next weapon nicely:

[tutorial](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial-behind-inventory/unvanquished_2020-04-26_052015_000.jpg)

[tutorial](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial-behind-inventory/unvanquished_2020-04-26_052048_000.jpg)

illwieckz:
Along with instruction to switch weapon, tutorial for mantis dash and human jetpack is added, the tutorial for the grenade now prints the real grenade name (Grenade / Fire Bomb) based on the inventory.

illwieckz:
The mechanism to cycle through selectable items from the inventory and to activate them does not exist anymore. Weapons are cyclable directly using `weapprev`/`weapnext` commands or `itemact` commands. Other items from the inventory are directly accessible via dedicated keys. The `+useitem` command does not exist anymore.

Note: the +useitem flag was also deleted on qcommon code stored engine side, maybe some games like Smokin Guns would ask to restore it one day because they also use a selectable inventory. If the need is real, it would not be hard to re-add it. See https://github.com/DaemonEngine/Daemon/pull/330.

illwieckz:
I added to tutorial some instructions for chatting, and some vertical spacers:

[tutorial](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial/unvanquished_2020-04-26_171249_000.jpg)

[tutorial](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial/unvanquished_2020-04-26_172024_000.jpg)

illwieckz:
I implemented “_mark for deconstruction_” and “_deconstruct marked_” on mouse left and mouse middle. When playing the Granger, the related attack mode (swipe, spit) are disabled when close enough to mark/deconstruct a buildable. The tutorial handles it well:

[tutorial](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial/unvanquished_2020-04-27_011516_000.jpg)

[tutorial](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial/unvanquished_2020-04-27_011534_000.jpg)

[tutorial](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial/unvanquished_2020-04-27_011542_000.jpg)

Because of that I dropped my implementation of human grenade on attack3. It would be much harder to retain the grenade (which is not part of the weapon used), dropping a grenade mistakenly in base is disastrous, not like a lost swipe or spit, and well, making a grenade the global third attack mode of all weapon is very hacky. People can still bind grenade on middle click and those who knows to not carry grenade when deconstructing.

src/cgame/cg_weapons.cpp
```diff
@@ -2207,9 +2207,34 @@ void CG_NextWeapon_f()
 		return;
 	}
 
-	// HACK: This only works with two weapons.
-	// See: https://github.com/Unvanquished/Unvanquished/issues/544#issuecomment-619299752
-	trap_SendClientCommand( "itemtoggle blaster\n" );
+	// Find next weapon in inventory.
+	weapon_t nextWeapon = WP_NONE;
```
slipher, line 2212: This doesn't really work in the general case of more than two weapons. So if you don't want to spend time fixing it, I'd rather keep `itemtoggle blaster` which is simpler.

src/cgame/cg_tutorial.cpp
```diff
@@ -444,13 +444,32 @@ static void CG_HumanText( char *text, playerState_t *ps )
 		}
 	}
 
-	if ( upgrade == UP_NONE ||
-	     ( upgrade > UP_NONE && BG_Upgrade( upgrade )->usable ) )
+	// Find next weapon in inventory.
+	// FIXME: factorize code with CG_NextWeapon_f() from src/cgame/cg_weapons.cpp
+	weapon_t nextWeapon = WP_NONE;
+	for ( weapon_t w = WP_NONE; w < WP_NUM_WEAPONS; w = weapon_t( w + 1 ) )
```
slipher, line 463: In this case since we do need to know the weapon, a simpler algorithm would be to find the first weapon that is not the current wepano.

src/cgame/cg_weapons.cpp
illwieckz: Why? It is expected to remember the first weapon it sees before finding the current weapon (to handle the case of the current weapon being the last one) until it finds the current weapon and, if it's not the end of the list, remember the next weapon and stop.

src/cgame/cg_tutorial.cpp
illwieckz: That's what the code does, but it also takes care of the case of the current weapon is the last one of the list.

illwieckz: - if current weapon: continue
- if first weapon before finding current weapon: remember
- if first weapon after finding current weapon: remember, stop

src/cgame/cg_weapons.cpp
slipher: Doing what you describe would work, but what the code does is always return the 2nd other weapon. You'd need to add some condition like `w > cg.snap->ps.weapon`.

illwieckz: hmmm

illwieckz: I rewrote it entirely, also made a reusable function, can you proof read that again?

src/cgame/cg_tutorial.cpp
illwieckz: This code now reuses the function from `cg_weapons.cpp`.

src/cgame/cg_tutorial.cpp
```diff
@@ -444,13 +444,15 @@ static void CG_HumanText( char *text, playerState_t *ps )
 		}
 	}
 
-	if ( upgrade == UP_NONE ||
-	     ( upgrade > UP_NONE && BG_Upgrade( upgrade )->usable ) )
+	// Find next weapon in inventory.
+	weapon_t nextWeapon = CG_FindNextWeapon( ps );
```
slipher, line 461: commits need to be reordered

src/cgame/cg_consolecmds.cpp
```diff
@@ -579,9 +579,10 @@ void CG_InitConsoleCommands()
 		trap_AddCommand( commands[ i ].cmd );
 	}
 
+	// Defined in src/engine/qcommon/q_shared.h, see BUTTON_ATTACK etc.
 	trap_RegisterButtonCommands(
-	    // 0      12       3     45      6        78       9ABCDEF      <- bit nos.
-	      "attack,,useitem,taunt,,sprint,activate,,attack2,,,,,,rally"
+	    // 0      1       2       3     45      6        7       89ABCDE     F <- bit nos.
+	      "attack,attack2,attack3,taunt,,sprint,activate,useitem,,,,,,,rally,"
```
slipher: useitem should just be deleted, right?

src/cgame/cg_tutorial.cpp
```diff
@@ -462,6 +462,13 @@ static void CG_HumanText( char *text, playerState_t *ps )
 		              _( BG_Weapon( nextWeapon )->humanName ) ) );
 	}
 
+	if ( BG_InventoryContainsUpgrade( UP_JETPACK, ps->stats ) )
+	{
+		Q_strcat( text, MAX_TUTORIAL_TEXT, va( _( "Press %s to use the %s\n" ),
```
slipher, line 473: "Hold" instead of "Press"

src/cgame/cg_weapons.cpp
```diff
@@ -2261,6 +2324,61 @@ void CG_SelectNextInventoryItem_f()
...
+CG_FindPrevWeapon
+Find previous weapon in inventory.
+===============
+*/
+weapon_t CG_FindPrevWeapon( playerState_t *ps )
+{
+	weapon_t prevWeapon = WP_NONE;
+
+	// For each weapon in the known weapon list from the end to the beginning:
+	for ( weapon_t w = WP_NUM_WEAPONS; w > WP_NONE; w = weapon_t( w - 1 ) )
```
slipher, line 2338: WP_NUM_WEAPONS is itself not a valid weapon number, so the loop should start at one less.

src/cgame/cg_tutorial.cpp
```diff
@@ -545,15 +549,13 @@ static void CG_SpectatorText( char *text, playerState_t *ps )
...
+			          va( _( "Press %s to return to first-person spectator mode\n" ),
 			              CG_KeyNameForCommand( "+attack3" ) ) );
 		}
+
+		Q_strcat( text, MAX_TUTORIAL_TEXT,
+		          va( _( "Press %s to follow the next player\n" ),
+		              CG_KeyNameForCommand( "weapnext" ) ) );
 	}
 	else
 	{
```
slipher, line 565: Looks like this case needs to be updated.

src/cgame/cg_view.cpp
```diff
@@ -922,6 +921,15 @@ static int CG_CalcFov()
 		}
```
slipher, line 921: This is out of place now. I guess the follow and unfollow cases could be unified since it is the same command either way.

slipher:
Still need to review the builder controls changes. I don't understand, if attack3 becomes a deconstruction command, then which command is granger spit?

illwieckz:
> Still need to review the builder controls changes. I don't understand, if attack3 becomes a deconstruction command, then which command is granger spit?

This is contextual. Usually, attack2 is swipe and attack3 is spit, but if you're close enough to deconstruct a friendly structure, attack2 is mark for replacement and attack3 is deconstruct marked. That's similar to the behaviour we already have when attack2 cancels the build of a structure, even if that key is used to wipe if no structure is selected to swipe.

See this video for a demonstration:

[Unvanquished granger contextual mouse behaviour](https://dl.illwieckz.net/b/unvanquished/wip/granger-attack3-deconstruction/20200427-165753-000.unvanquished-granger-contextual-mouse-behaviour.webm)

slipher:
I'm concerned that having the hard-deconstruct key shared with something else makes it too easy to do by accident. Also, I think the current setup was deliberately designed to make it harder to hard-deconstruct something. The mark system is a bit harder to understand than simple deconstruction, so a new player finds the hard-deconstruct key, it seems likely they will use that and ignore the mark system.

BTW, in your video there seems to be a bug that mark deconstruct still makes the swipe sound.

illwieckz:
> The mark system is a bit harder to understand than simple deconstruction, so a new player finds the hard-deconstruct key, it seems likely they will use that and ignore the mark system.

The hard-deconstruct key requires the buildable to be marked first, as before.

> BTW, in your video there seems to be a bug that mark deconstruct still makes the swipe sound.

The bug already exists in master, the swipe sound is made when cancelling the build. So it's out of topic.

src/cgame/cg_tutorial.cpp
illwieckz: Are you sure? Beware that GitHub just makes garbage about commit ordering, it sometime lists them by date (and maybe the commit date that is not the author date) instead of the way they are chained together.

src/cgame/cg_tutorial.cpp
illwieckz: When spectating, it's `attack1` to join a team anytime, `attack3` to start to follow a player, then `attack3` to cycle between 1st person view or chase, `weapnext`/`weapprev` to switch player to follow, and `attack2` to cancel following and return to free fly spectating.

src/cgame/cg_view.cpp
illwieckz: It's the same command (it's togglable), but with this design, follow starts with middle mouse and stops with mouse right (cancel).

This is to make possible to cycle between 1st person/chase view while never stopping to follow a given player. Before, once you have switched from 1st person to chase, it was not possible to return to 1st person view without unfollowing entirely, restarting to follow, and recycling to find the player again.

illwieckz:
I forgot to explain my modification the the spectator thing.

Before, the middle mouse button was used to:

- start to follow
- switch from first person view to chase cam
- stop to follow

It made impossible to return to first person view when chasing, without stopping to follow first.

Once modified the behaviour is this one:

Player starts to middle mouse as usual (attack3) and stops with mouse right (cancel, like builder cancel).

While following, it is possible to cycle between 1st person/chase by clicking on middle mouse, and it cycles between 1st person view and chasing infinitely, never returning to free fly.

While following it is also possible to cycle between players with the mouse wheel (next/previous player).

src/cgame/cg_consolecmds.cpp
illwieckz: Yes, but at this point it is done iteratively, `attack3` needs to be added before we can drop `useitem`, and it's better to put `attack3` at the right place since the beginning, hence the temporary move of `useitem` to keep each commit usable before being able to drop it entirely.

src/cgame/cg_tutorial.cpp
illwieckz: Fixed

src/cgame/cg_weapons.cpp
illwieckz: Fixed

src/cgame/cg_view.cpp
illwieckz: Note: we may add non-togglable commands (to start and stop following) to make the code explicit and remove some test that only exists because the command is a togglable command. But that's out of topic.

src/cgame/cg_weapons.cpp
```diff
@@ -2189,16 +2189,68 @@ void CG_DrawItemSelectText()
...
+CG_FindNextWeapon
+Find next weapon in inventory.
+===============
+*/
+weapon_t CG_FindNextWeapon( playerState_t *ps )
+{
+	weapon_t nextWeapon = WP_NONE;
+
+	// For each weapon in the known weapon list from the beginning to the end:
+	for ( weapon_t w = WP_NONE; w < WP_NUM_WEAPONS; w = weapon_t( w + 1 ) )
```
illwieckz, line 2203: @slipher I guess we can also start from `WP_NONE + 1`, what do you think about it?

slipher: That's OK but not too important. Using w = WP_NUM_WEAPONS is much worse, since some things use the weapon number as an index into an array of length WP_NUM_WEAPONS.

src/cgame/cg_tutorial.cpp
slipher: Are you telling me it's actually impossible to see the order of the commits? Surely things cannot be so bad...

src/cgame/cg_view.cpp
slipher: I did not realize this is intentional. I don't like this change - attack3 is too obscure to use for the most important control when spectating.

illwieckz: I did not chose or changed anything about the fact `attack3` starts to follow. I made `attack2` cancels the following.

On a side note, the alternative to `attack3` to start to follow would be `attack2` but then you would return to the first issue when you can't come back to 1st person view to chase or you would cancel with `attack3` which is not consistent with the builder cancel, and this time, absolutely not intuitive. The other alternative to start follow would be `attack1` which is something I am strongly against, because that means we can't join a game with attack1. Some games uses `space` (like xonotic) to join because `attack1` is used to follow, that means a spectator cannot fly up when free flying (and if the user mistakenly presses space by muscle reflex, he joins and lose his position, which is very frustrating. That alternative is so frustrating I developed a special phobia to press space when spectating in other games even ours.

So, I did not chose `attack3` to start to follow. That always worked like this. What I did was to move the follow cancel on `attack2` like the builder cancel (which is intuitive). Also when a player clicks `attack1` to join it also stops to follow.

This implementation looks to be the best implementation: everything is available on the mouse, that does not prevent to freefly. I would like to see that in other games instead of developing phobias. I just made it better by removing the “stop following” command from the follow view mode cycle.

src/cgame/cg_weapons.cpp
illwieckz: Yes, in this case that's just an optimization (saving one loop) but not a mistake by itself.

src/cgame/cg_tutorial.cpp
illwieckz: I stopped to trust Github ordering of the commits long time ago.

illwieckz:
I fixed a small bug where right clicking to cancel a buildable would also mark the already-built buildable in range to deconstruct, this because even if the code running in one frame does not mark for deconstruction if cancellation is done first, the next frame is played before the player releases the key…

illwieckz:
I added a commit that just reword some code in `shared/bg_pmove.cpp`, basically by doing a switch/case instead of testing multiple time the same variable, and reuse the result of an operation multiple time, instead of redoing the same operation again… Also some goon-related tests are only done in goon-related block, not for every weapon. And also some luci-related code are now only don in luci-related block.

That changes nothing in the behaviour, but that may help to implement one day "hold attack2 to deconstruct".

illwieckz:
I implemented "hold `attack2` to deconstruct", so commits after “_commands: make deconstruction code reusable_” are either new either deeply rewritten.

This keep available `attack3` so human players can bind grenade on it.

illwieckz:
Oh no, I overflowed the stat index…

illwieckz:
I removed the "hold attack2 to deconstruct" part to put it in another PR to come.

UnvanquishedAssets/unvanquished_src.dpkdir#15 is not yet ready but that's not very important (it does not prevent to build), both Unvanquished and Dæmon part must be OK.

src/cgame/cg_tutorial.cpp
```diff
@@ -545,15 +549,13 @@ static void CG_SpectatorText( char *text, playerState_t *ps )
 		else if ( cgs.clientinfo[ cg.clientNum ].team == TEAM_NONE )
```
slipher, line 549: I think this should just be `else {` ? Since you can switch camera type regardless of whether you are on the spectator team.

src/cgame/cg_view.cpp
slipher: I like attack2 for start following. Then we have attack2 for cycling following vs. free-flying, and attack3 for cycling view mode. To me this is a good kind of consistency. If we do that I would be very happy with how it turns out, since it keeps the nice property of being able to use all following features with only the mouse, while necessitating the awkward mouse3 for only the least important operation.

src/cgame/cg_tutorial.cpp
```diff
@@ -265,6 +265,10 @@ static void CG_AlienLevel1Text( char *text )
 	          va( _( "Press %s to swipe\n" ),
 	              CG_KeyNameForCommand( "+attack" ) ) );
 
+	Q_strcat( text, MAX_TUTORIAL_TEXT,
+	          va( _( "Press %s to dash\n" ),
```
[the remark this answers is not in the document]
illwieckz: I used `dash` word because that's the word used in War§ow for a similar movement, and I lack imagination.

src/cgame/cg_tutorial.cpp
```diff
@@ -462,6 +462,13 @@ static void CG_HumanText( char *text, playerState_t *ps )
 		              _( BG_Weapon( nextWeapon )->humanName ) ) );
 	}
 
+	if ( BG_InventoryContainsUpgrade( UP_JETPACK, ps->stats ) )
+	{
+		Q_strcat( text, MAX_TUTORIAL_TEXT, va( _( "Hold %s to use the %s\n" ),
```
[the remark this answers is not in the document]
illwieckz: The name lookup seems to be more future-proof regarding translation.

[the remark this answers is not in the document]
slipher: Actually I was thinking the opposite regarding translation. The more complete of a sentence you have, the more likely the translation is to be correct. For example in a language with a case system, a noun by itself (__jetpack__) is likely to have a different translation than a noun which is the object of a verb (use __jetpack__).

src/cgame/cg_tutorial.cpp
```diff
@@ -444,6 +431,22 @@ static void CG_HumanText( char *text, playerState_t *ps )
...
+		if ( BG_InventoryContainsUpgrade( u, ps->stats ) )
+		{
+			upgrade = u;
+		}
+	}
+
+	if ( upgrade != UP_NONE )
+	{
+		Q_strcat( text, MAX_TUTORIAL_TEXT, va( _( "Press %s to throw the %s\n" ),
+			CG_KeyNameForCommand( "+itemact grenade" ),
```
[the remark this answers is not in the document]
illwieckz: done

src/cgame/cg_tutorial.cpp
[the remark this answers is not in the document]
illwieckz: done

src/cgame/cg_tutorial.cpp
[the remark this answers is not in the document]
illwieckz: we can think about it later, this PR is already big, getting it merged is better than a perfect one not merged.

src/shared/bg_misc.cpp
```diff
@@ -602,6 +602,19 @@ weapon_t BG_WeaponNumberByName( const char *name )
 	return ( weapon_t )0;
 }
 
+const char* BG_WeaponNameByNumber( weapon_t number )
```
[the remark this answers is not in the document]
illwieckz: done

src/cgame/cg_weapons.cpp
```diff
@@ -2189,6 +2189,62 @@ void CG_DrawItemSelectText()
...
+/*
+===============
+CG_FindNextWeapon
+Find next weapon in inventory.
+===============
+*/
+weapon_t CG_FindNextWeapon( playerState_t *ps )
+{
+	weapon_t nextWeapon = WP_NONE;
+
```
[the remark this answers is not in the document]
illwieckz: @slipher: what if we merge that working code and you submit such optimization after that? I know it may be easy to do but my brain is already full of things from that PR, I'm afraid dealing with that would be too much for me at this point (even if, taken separately, it may looks obvious). This works anyway and my brain is full.

src/cgame/cg_view.cpp
illwieckz: done, it works well.

illwieckz:
The game now fails to look-up teambind keys for display in tutorial. This worked before, as seen on screenshot on https://github.com/Unvanquished/Unvanquished/pull/1182#issuecomment-619475060

This is probably related to some other work elsewhere that was done in the meantime. The code in this PR worked before. This can be tracked down in another PR later.

src/cgame/cg_tutorial.cpp
illwieckz: done

illwieckz:
OK, everything from that PR is fully functional now. There is that little controversy about modifying the button array in engine, but I think it's not bad enough to prevent the merge. Making things better (even redesigning that whole stuff) can be done later, even for another release.

slipher:
Merged as #1208.

PR #1188, opened 2020-05-01
Hold attack2 to deconstruct (requires to split weapon charge and build timer in engine)

illwieckz:
This is the PR to hold attack2 to deconstruct a buildable. Unfortunately I was not able to rebase it on `0.52.0` without conflict so it includes #1182 for the moment.

Requires https://github.com/DaemonEngine/Daemon/pull/338 on engine side.

Note: do not substract yet the time spent to hold attack2 to the default deconstruction time.


illwieckz:
This PR is incomplete: does not feature the substraction thing, does not protect protected buildable yet etc. See #1192 for a PR that is complete but rely on an ugly hack to avoid the modification of the engine. If one day someone wants to fix the engine and want to look at this implementation, please remind this code is not complete.

PR #1191, opened 2020-05-02
Fix confusion about different team numbers related to binds

### LINKED ISSUE #1187: Spectator tutorial fails to get key names from its team (works if global)
See the "follow" bind advice:

[spectator bind](https://dl.illwieckz.net/b/unvanquished/bugs/spectator-bind/unvanquished_2020-04-24_211753_000.jpg)

Note that even as spectator it is displayed correctly when following an human or an alien:

[spectator bind](https://dl.illwieckz.net/b/unvanquished/bugs/spectator-bind/unvanquished_2020-04-24_211814_000.jpg)

This bind works in all case:

```
bind  MOUSE3        +attack3
```

This reproduces the bug:

```
teambind aliens     MOUSE3 +attack3
teambind humans     MOUSE3 itemact grenade
teambind spectators MOUSE3 +attack3
```

This issue is blocking for `0.52`.

slipher:
Fixes #1187.

### DIFF
--- a/src/cgame/cg_key_name.cpp
+++ b/src/cgame/cg_key_name.cpp
@@ -23,6 +23,7 @@ along with Unvanquished. If not, see <http://www.gnu.org/licenses/>.
 
 #include "cg_key_name.h"
 
+#include "common/Log.h"
 #include "common/String.h"
 #include "engine/client/cg_api.h"
 
@@ -46,4 +47,25 @@ std::string CG_KeyDisplayName(Key key) {
     default:
         return "???";
     }
+}
+
+static int bindTeam;
+
+void CG_SetBindTeam(team_t team) {
+    switch (team) {
+    case TEAM_NONE:
+        bindTeam = 3; // BIND_TEAM_SPECTATORS
+        break;
+    case TEAM_HUMANS:
+    case TEAM_ALIENS:
+        bindTeam = static_cast<int>(team);
+        break;
+    default:
+        Log::Warn("Invalid team %d", team);
+    }
+}
+
+// Get the current team according to the Keyboard::BindTeam type in the engine
+int CG_CurrentBindTeam() {
+    return bindTeam;
 }
--- a/src/cgame/cg_key_name.h
+++ b/src/cgame/cg_key_name.h
@@ -25,7 +25,12 @@ along with Unvanquished. If not, see <http://www.gnu.org/licenses/>.
 #define CG_KEY_NAME_H_
 
 #include "common/KeyIdentification.h"
+#include "shared/bg_public.h"
 
 std::string CG_KeyDisplayName(Keyboard::Key key);
+
+void CG_SetBindTeam(team_t team);
+
+int CG_CurrentBindTeam();
 
 #endif // CG_KEY_NAME_H_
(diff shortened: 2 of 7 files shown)


illwieckz:
What is making those build bots crazy?

slipher:
> What is making those build bots crazy?

One of those things that only shows up when you build without the precompiled header (which I didn't think was used for gamelogic...).

illwieckz:
I don't know why, but I don't get it working when merged on master, but when I merge it on my `binds/sync` branch it works. Thanks.

illwieckz:
There is still the issue of following players not using (and not displaying) spectator binds. I'm looking at it.

slipher:
> There is still the issue of following players not using (and not displaying) spectator binds.

What do you mean? If you are referring to the situation of being on a team but dead and following teammates, then spectator teambinds cannot be used. Only the binds for the team you are on.

illwieckz:
Yes, the thing is that dead players can follow their teammates, it works. That's why I'm also looking for spectator binds for dead players.

I managed to fix `CG_SetBindTeam` so the tutorial display spectator binds to dead people, but unfortunately the bind does not work.

slipher:
You can't use spectator binds as a dead player. When you are on the alien team, you can only use binds for the alien team and the default team, regardless of whether you're alive. That's working as intended and as it should.

illwieckz:
> That's working […] as it should.

Why? I would like to provide a bind to follow players when I'm dead. Currently it works for aliens but not for humans because they don't have 3rd attack weapon.

illwieckz:
Well… I can imagine player having `vsay` teambinds when following a teammate.

slipher:
The current set of default binds has <kbd>Enter</kbd> for follow. What's wrong with that?

illwieckz:
> The current set of default binds has <kbd>Enter</kbd> for follow. What's wrong with that?

That looks to be a leftover. The code intercepts `useitem` (<kbd>Mouse 3</kbd>) to call `follow`.
The tutorial talk about `useitem` (<kbd>Mouse 3</kbd>) everywhere.

Also <kbd>Mouse 3</kbd> is a very bad key regarding usability: it's far from both left hand (on <kbd>W</kbd><kbd>A</kbd><kbd>S</kbd><kbd>D</kbd>) and right hand (on mouse). The thing is that I'm removing `useitem`.

The key to follow in Unvanquished was is expected to be <kbd>Mouse 3</kbd> and is advertised this way.

---
Note: @Gireen had the idea to make <kbd>Enter</kbd> an alternative bind for global talk. Since you have to put your right hand on keyboard to chat, <kbd>Enter</kbd> is next to right hand when talking.

illwieckz:
> The code intercepts useitem (Mouse 3) to call follow.

Having proper spectator key binds would allow use to remove those hacks (same with `followprev`/`follownext`, to stop abusing `weapprev`/`weapnext`).

But nevermind, that's another job. This PR is ready.

illwieckz:
Just for reference, that is my fix to get spectator key names when following, whatever the team:

[0001-Binds-true-spectator-binds-in-tutorial.patch.txt](https://dl.illwieckz.net/b/unvanquished/wip/true-spectator-binds/0001-Binds-true-spectator-binds-in-tutorial.patch)

It's not enough though, since keys are still reads as human/alien.

PR #1192, opened 2020-05-02
Hold attack2 to deconstruct (uses negative build timer ugly hack to know it's not a weapon charge, better fix the engine)

illwieckz:
This is a PR to implement “_hold `attack2` to deconstruct a buildable_”. Unlike #1188, it does not require to extend `playerstat` in engine. This is done by leveraging a very-intrusive and ugly hack, storing build timer as negative to know it's a build timer and not a weapon charge. The hack assumes we will never have to use them at the same time.

To quote the commit introducing the hack:

> Since weapon charge require build timer to expire first,
> we can wait for build timer to not be smaller than zero
> then store the weapon charge in the same field.
>
> That's very ugly and will bite someone hard in the back in the future,
> please fix engine instead. The day a bug surfaces because of this hack,
> more people would be hurt than the day we would fix the engine.
>
> Not fixing the engine is like feeding an idol who does not mind about us
> and does not deserve it.

This PR is complete, rebased on `0.52.0`, does not require anything else. _Edit: it requires an updated `ui/rockethud.txt`, see comment below._

It adjusts tutorial accordingly and and do nice things to display the message for protected buildables as soon as possible. The time spent to hold the key is substracted to the build timer that comes after the deconstruction, this way players holding key or having a keybind on key press have to wait the same time between buildables when deconstructing multiple buildables in a row.

@slipher has some idea to make this entirely client side. In all case, this works and is ready, also, it is strongly recommended that alternative implementation leverages as much as possible some bits implemented there, like the early message for protected buildables, and the very complex state machine to handle both key press and key hold on the same key to do a different action.


illwieckz:
Note: the PR includes a debug commit that must be removed before merging, this commit is there to help testing, especially to avoid some build timer code to be disabled when using `devmap`.

illwieckz:
I forgot to say that `ui/rockethud.txt` must be edited to add `hud_chargebar` to `ckit.hudgroup`, `abuild.hudgroup` and `abuildupg.hudgroup` to get the charging bar displayed.

illwieckz:
Obsoleted by #1205.

TODO: check if @slipher's code also does things like _“display the protected buildable message as soon as possible”_?
TODO: also merge _“bikeshedding: add some comments to help to understand or fix the code”_ separately.

PR #1197, opened 2020-05-05
Gamelogic-configurable netcode for playerState_t

slipher:
See https://github.com/DaemonEngine/Daemon/pull/341


illwieckz:
I now see no issue on my end. I'm not sure to understand everything but I'm confident with what you've done.

PR #1198, opened 2020-05-10
rocket/progressbar: do not divide by zero, fix #1189

### LINKED ISSUE #1189: Bizarre code calling for trouble in CG_Rocket_GetBuildTimerProgress
1. What happens if this test fails and `max` is undefined?
2. Does `max` can be `0` (then divide by `0`)?

https://github.com/Unvanquished/Unvanquished/blob/f79ea6e9b10f6d0d0c32b59f9ed371e099c52b9a/src/cgame/cg_rocket_progressbar.cpp#L59-L60

https://github.com/Unvanquished/Unvanquished/blob/f79ea6e9b10f6d0d0c32b59f9ed371e099c52b9a/src/cgame/cg_rocket_progressbar.cpp#L70-L78

illwieckz:
rocket/progressbar: do not divide by zero, fix #1189

- initialize `max` explicitly, set to zero
- make sure to not divide by zero

### DIFF
--- a/src/cgame/cg_rocket_progressbar.cpp
+++ b/src/cgame/cg_rocket_progressbar.cpp
@@ -57,7 +57,7 @@ static float CG_Rocket_GetOverallLoadProgress()
 static float CG_Rocket_GetBuildTimerProgress()
 {
 	static int misc = 0;
-	static int max;
+	static int max = 0;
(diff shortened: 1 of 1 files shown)


illwieckz:
Note that I have tested in game to make sure the game behaves the same after this is merged, and it behaves the same.

PR #1201, opened 2020-05-19
+deconstruct: hold to force-decon

slipher:
My attempt at a press-and-hold-to-deconstruct implementation, based on the work by @illwieckz.

I'm not comfortable overloading `+attack2` with deconstruct, since I think it interferes with use of the granger swipe attack too much. The build weapon's range is fairly long, so I think it is too easy to end up issuing deconstruct marks by mistake if you are trying to attack a human inside the base.

So instead, I introduced a new `+deconstruct` command which has that behavior of mark-deconstructing when short pressed and force-deconstructing when held.

This is based on the playerstate/sync branches, so I chose that as the target branch to hide the diffs.

You can check out https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/17 to see the charge bar in the UI.


illwieckz:
The logic works on my end but the charge bar is not displayed.

slipher:
> The logic works on my end but the charge bar is not displayed.

Even using the UnvanquishedAssets/unvanquished_src.dpkdir#17 branch?

illwieckz:
Hmm, sorry, it looks like I checked out the branch in the wrong repository. The charge bar works.

Anyway, your implementation does not prevent to decon the last spawn or the rc/overmind. The test must be done at the start of the charge procedure (to not charge on last spawn or rc/overmind) and at the end of the charge procedure (for when the player charge on a buildable but fire on another). See my commits about “protected buildables” in #1192, and commits making use of `G_IsProtectedBuildable`.

slipher:
> Anyway, your implementation does not prevent to decon the last spawn or the rc/overmind.

The decon protection continues to work for me. Probably you are running into the unintended side effects of `devmap` again :weary:. I'll try to create a PR for more fine-grained cheat cvars later today.

> The test must be done at the start of the charge procedure

Well, currently there is no concept of "locking on" to a buildable, so there is nothing to be done at the start. Do you think there should be a lock-on? I.e. to begin charging you must be pointing at a buildable, and the charge cancels if you move away from it?

illwieckz:
> The decon protection continues to work for me. Probably you are running into the unintended side effects of devmap again weary. I'll try to create a PR for more fine-grained cheat cvars later today.

Ah, yes, probably…

> Well, currently there is no concept of "locking on" to a buildable, so there is nothing to be done at the start. Do you think there should be a lock-on? I.e. to begin charging you must be pointing at a buildable, and the charge cancels if you move away from it?

Oh, I did not noticed it was possible to charge before reaching a buildable (now I remember you talked about it). I'm mixed on the fact a buildable does not have to be in front of the player before charging, I prefer when it is required, this way the charge bar is a effective feedback the buildable will be deconstructed. For the protected decon prevention, I think it have to be done the sooner as possible, because it's frustrating to do a temporized action to only know at the end it was not possible from the start. Also if we test for decon protection at charge startup we would have implemented 90% of decon charge on buildable only.

In all case I think decon charging on buildable only and decon protection as soon as possible is better for user interaction and autodidaxis. Also it would be more coherent with the tutorial that activates when player is facing a buildable.

slipher:
I implemented a form of target lock-on (supporting the early error message when the target is not valid for deconstruction) and allowed deconstruction during buildable placement.

The target lock requires you to stay near the targeted buildable when the charge completes. It intentionally does not require you to keep aiming at the buildable. I want to make it actually usable during combat, so you can move around a bit after selecting the target.

illwieckz:
> The target lock requires you to stay near the targeted buildable when the charge completes. It intentionally does not require you to keep aiming at the buildable.

Does it remind the buildable selected to deconstruct, preventing to load on a buildable and fire on another?

illwieckz:
Nice.

slipher:
Had to move this to https://github.com/Unvanquished/Unvanquished/pull/1205 due to Github bugs.

PR #1205, opened 2020-06-21
+deconstruct: hold to force-decon

slipher:
Same as https://github.com/Unvanquished/Unvanquished/pull/1201 which Github closed of its own accord, and mysteriously refuses to let me reopen.


illwieckz:
Can we merge this? if there is missing things, we may do another PR.

PR #1208, opened 2020-09-01
#1182 with a few fixes

slipher:
- Simplify algorithm for selecting next/previous weapon
- Changed the button stuff a bit (This PR does not depend on an engine change )
- Squash together some commits that redid stuff from previous ones (e.g. adding then removing an `itemtoggle blaster` hack)

### DIFF
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -579,9 +579,10 @@ void CG_InitConsoleCommands()
 		trap_AddCommand( commands[ i ].cmd );
 	}
 
+	// Defined in src/engine/qcommon/q_shared.h, see BUTTON_ATTACK etc.
 	trap_RegisterButtonCommands(
-	    // 0      12       3     45      6        78       9ABCD           E      <- bit nos.
-	      "attack,,useitem,taunt,,sprint,activate,,attack2,,,,,deconstruct,rally"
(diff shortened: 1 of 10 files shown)


illwieckz:
I tested and it works.

Note: I wanted to ask if it's better to write `SHIFT-ESCAPE` or `SHIFT+ESCAPE`?

slipher:
> Note: I wanted to ask if it's better to write SHIFT-ESCAPE or SHIFT+ESCAPE?

I think either is OK. I tried googling "ctrl alt delete", and got 6 results with `-`es and 5 with `+`es.

illwieckz:
For some reason I feel `+` expresses better the idea of a simultaneous use.

illwieckz:
Feel free to merge when you want to, note that for such branch with many commits I find better to properly rebase on latest upstream and merge with a merge commit (`--no-ff`), this way it's easy to look at the whole branch later without polluting that much history.

illwieckz:
Interestingly, with your implementation the weapon roll does not roll forever, my implementation restarted the lookup from zero when the last weapon was used, which is a common behaviour in FPS.

slipher:
> Interestingly, with your implementation the weapon roll does not roll forever

I tested this and it works for me.

illwieckz:
Huuu, I may be fooled by some leftover binds, testing with a brand new profile fixes things.

I noticed another issue: F11 screenshot does not work with a brand new profile, maybe I broke something in `default.cfg`.

PR #1210, opened 2020-09-05
Use a specific prefix for bots so they are not named UnnamedPlayer

illwieckz:
Use a specific prefix for bots so they are not named `UnnamedPlayer#1`:

[](https://dl.illwieckz.net/b/unvanquished/shots/unnamed-bots/unvanquished_2020-09-05_220125_000.jpg)

There is a `g_unnamedBotNamePrefix` cvar to customize the bot name prefix but even if I'm not happy with the name (it's not named the `g_bot_` way but like existing `g_unnamedBotNamePrefix`) it does the job.

The default bot name prefix is `[bot] Bot#`.

### DIFF
--- a/src/sgame/sg_bot.h
+++ b/src/sgame/sg_bot.h
@@ -26,6 +26,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #ifndef __BOT_HEADER
 #define __BOT_HEADER
 
+#define UNNAMED_BOT "[bot] Bot"
+
 struct botEntityAndDistance_t
 {
 	gentity_t *ent;
--- a/src/sgame/sg_client.cpp
+++ b/src/sgame/sg_client.cpp
@@ -644,6 +644,14 @@ bool G_IsUnnamed( const char *name )
 		return true;
 	}
 
+	length = strlen( g_unnamedBotNamePrefix.string );
+
+	if ( g_unnamedNumbering.integer && length &&
+	     !Q_strnicmp( testName, g_unnamedBotNamePrefix.string, length ) )
+	{
+		return true;
+	}
+
 	return false;
 }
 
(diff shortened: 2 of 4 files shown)


slipher:
Remove the first commit, since you ended up not using and it also breaks Travis.

illwieckz:
The first commit is used and required, I was just a bit generous in copy-pasting some code to keep the behavior : some variables were defined but not used in one function in three I modified.

illwieckz:
Without first commit, `G_UnnamedClientName` can't be aware of `ent` and then can't test for `ent->r.svFlags & SVF_BOT`.

slipher:
OK, my bad. In that case you can still get rid of the first commit, and get the client num from `gclient_t* client` by `client->ps.clientNum`.

illwieckz:
> get the client num from `gclient_t* client` by `client->ps.clientNum`.

Oh nice, thanks, I was first looking for something like that and missed it… That will be better indeed.

illwieckz:
@slipher, now that I edited the code to use `client->ps.clientNum` it does not work anymore. It builds but bot are named `UnnamedPlayer` instead of `[bot] Bot`.

slipher:
OK, looks like this information is not filled in yet before the first frame with the bot is run. `int clientNum = client - level.clients;` should be more reliable.

illwieckz:
Thanks, it now works.

illwieckz:
Can we merge this?

slipher:
This prefix should be checked in `G_IsUnnamed` to prevent two players from having the same name.

illwieckz:
I added a check in `G_IsUnnamed`, hoping that was what was asked.

I also added a commit to move the `#` out of the default prefix but to always add it to the prefix when adding numbers, this way no one has to add this `#` in default prefixes, and it does not sound bad to always use the `prefix#number` scheme.

slipher:
> I added a check in `G_IsUnnamed`, hoping that was what was asked.

Yes, thanks.

> I also added a commit to move the `#` out of the default prefix but to always add it to the prefix when adding numbers, this way no one has to add this `#` in default prefixes, and it does not sound bad to always use the `prefix#number` scheme.

This one is not to my taste; it seems less configurable for no reason.

illwieckz:
Yep, but does it makes sense to not use a `#` separator between prefix and numbers? This even looks language agnostic… Well, I maybe see one issue with right-to-left languages, I don't know… Another idea would be to make the separator a cvar but that looks overkill…

slipher:
Why does there need to be a concept of a "separator"? You have the prefix and the number, that's all.

necessarily-equal:
IMO, it's better not to have a separator: if you don't have a separator, you can say prefix="bot#". When it's the other way around and you don't want the separator (for example, for localisation purpose), you're stuck.

illwieckz:
I removed the commit removing the separator from the default names and doing it anytime.

Anyway, it may be good to make at least a define for it, it's like hardcoded 5 times… But that can be done later.

PR #1221, opened 2020-09-26
sample config: sv_dedicated is now server.private

illwieckz:
sample config: sv_dedicated is now server.private


slipher:
The server config change looks good. The submodule update should be changed or removed since it is pointing to the 0.52/sync branch.

illwieckz:
Right, fixed the ref.

Issue #1220, opened 2020-09-24
Bots ignore dynamic map elements


Gireen:
it seens that entities func_spawn and func_destructable are ignored by bots.

illwieckz:
Bots have been reported to miss `func_sarah_connor` repeatedly! :robot:

bmorel:
If this is fixed, maybe it can be closed?

Gireen:
nope. The merged pr only fixed part of the problem. Witch is enough for me but to solve this completely more investigation and changes would be needed.
 
> doesn't work on walls that are spawned on map start

and func_destructable would require the same changes

bmorel:
Yes, it's clearly not fixed.

> But unlike doors bots currently ignore this objects and just try to run trough them

This is even wrong: bots also ignore doors and try to run through them. They do *not* pause but run until the door opens and let them pass, basically. This can be seen notably in usstremor.

slipher:
> bots also ignore doors and try to run through them

That's basically what I do too. Especially with high ping 😛

bmorel:
Won't be easy to fix, especially with recast. Moving to Q3's original AAS might help a lot though.

bmorel:
Is there is a map with those entities? I'd be curious to know what they are meant to do, and see if I can fix that, one way or another.
For now, I'd likely just make them consider those as obstacles, even if that's not a proper implementation.

bmorel:
I'm tempted to close this issue for 3 reasons:

1. it's vague. Yes, bots ignore some stuff, in many cases, and those are all different feature requests (waiting for an event, have subgoals to activate a door before going through it, etc)
2. all of the bot-specific problems I could find in actual maps (more than 50!) were fixed
3. no map was named, so there is no way to easily reproduce the issue

Gireen:
func_destructable can be any brush. it has health and disappears when it received enough damage
but can also be reset by the map

func_spawn can be any brush and be toggled by the map

1. its precise. bots ignore this elements cause they don't update the navmesh
and the related PR shows parts of a solution for this
2. I don't its a good reason to close a issue without fixing, just to close as many as possible. The bug wont go away and when it is found again it requires unnecessary time to investigate its source again. 
3. There are probably plenty of old tremulous maps that use this. Any map where destructible elements are or objects like walls instantly appear or disappear should be affected.

bmorel:
> its precise. bots ignore this elements cause they don't update the navmesh
[...]

PR #1229, opened 2020-10-11
make toggleable walls bot obstacles

Gireen:
doesn't work on walls that are spawned on map start

relevant for #1220

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -2583,22 +2583,48 @@ Use_func_spawn
 */
 void func_spawn_act( gentity_t *self, gentity_t*, gentity_t *activator )
 {
-  if( self->r.linked )
-    trap_UnlinkEntity( self );
-  else
-  {
-    trap_LinkEntity( self );
-    if( !( self->spawnflags & 2 ) )
-      G_KillBrushModel( self, activator );
-  }
+	vec3_t    mins, maxs;
+
+	if( self->r.linked )
+	{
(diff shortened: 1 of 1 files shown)


necessarily-equal:
just a style thing: this mixes tabs and spaces in the files it touches.

illwieckz:
I'm not sure to understand what it is about, do you have a simple in-game example?

Gireen:
func_spawn is an object/brush/mapentity that can be toggled. to appear or disappear by a trigger. 
I don't think that this is used in any of our maps so far.
Imagine a box or wall that disappears instantly when a button is pushed and reappears if it is pushed again.
Similar to doors but without moving. But unlike doors bots currently ignore this objects and just try to run trough them. 

With this changes the navmesh is updated when such objects appear or disappear.

illwieckz:
Anything making bots better is welcome! =)

LGTM but I'm not a wizard on such topic.

slipher:
LGTM; I too have no idea how this part of the code works.

How did you run into this issue when there are no maps that use it?

Gireen:
I'm trying mapping and look what interesting things are possible. 
And thereby came up with a [silly method to change maps](https://forums.unvanquished.net/viewtopic.php?f=9&t=2196). 😀

necessarily-equal:
That makes me think… We could have force fields in games! Kinda like 
those in the Darth Maul duel in Star Wars 1 or the cargo bay access to 
space in the USS Tremor alien base.

Le dim 18 oct. 2020 à 20:10, Gireen <notifications@github.com> a 
écrit :

PR #1231, opened 2020-11-09
BUGFIX: back slashes needed for the brackets in the regex

taylor:
This fixes the regex so the space and tab are treated as a character set as intended.
The regex was stripping all the 't's from the URLs

### DIFF
--- a/download-paks
+++ b/download-paks
@@ -52,7 +52,7 @@ colorPrintf () {
 }
 
 listCdn () {
-	(sed -e 's/#.*//g;s/[ \t]*//g' | grep -v '^$') <<-EOF
+	(sed -e 's/#.*//g;s/\[ \t\]*//g' | grep -v '^$') <<-EOF
 	# official
 	http://cdn.unvanquished.net
 	# Viech


illwieckz:
What is your system? can you print the output of:

```sh
printf '===\t===\n'
printf '===\t===\n' | sed -e 's/\[\t\]//'
printf '===\t===\n' | sed -e 's/[\t]//'
```

taylor:
I tried this on an Ubuntu system and on a Mac.

On the Ubuntu system I see:

```
➔ uname -a
Linux dev1.vulk.co 3.13.0-88-generic #135-Ubuntu SMP Wed Jun 8 21:10:42 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
➔ printf '===\t===\n'
===	===
➔ printf '===\t===\n' | sed -e 's/\[\t\]//'
===	===
➔  printf '===\t===\n' | sed -e 's/[\t]//'
======
```

On the Mac I see:

```
$uname -a
Darwin iMac 19.6.0 Darwin Kernel Version 19.6.0: Thu Jun 18 20:49:00 PDT 2020; root:xnu-6153.141.1~1/RELEASE_X86_64 x86_64

$printf '===\t===\n'
===	===
$printf '===\t===\n' | sed -e 's/\[\t\]//'
===	===
$printf '===\t===\n' | sed -e 's/[\t]//'
===	===
```

Strange... I tried using -E option for sed and that makes no difference.  Yet it makes a different in the bash script and works on both Linux and Mac.

taylor:
Using a "hard" tab on cli does work (for Mac)

```
printf '===\t====\n' | sed -E 's/[\    ]//g'
=======
```

slipher:
Perhaps, you could put an `echo` with each URL, instead of solving the hard problem of making a tab character on Mac.

illwieckz:
So, the regular expression did not have a bug, this is the macOS `sed` that lacks features.

I've pushed a fix to both `master` and upcoming `0.52.0/sync` branch.

Thanks for the report.

Note that because of a mistake we did when we released 0.51, you'll have to install `aria2c` to get the script working correctly (`wget` and `curl` backends will fail).

taylor:
Not sure if this is worth another PR, but after thinking about all of this, using tr might be a good choice and it works on BSD-based and Linux platforms. Here is an example with a tab and space:

```
printf '===\t= ==\n' | tr -d '[:blank:]'
```

illwieckz:
Right, it's a nice way to do it! Maybe it's not that useful to add one more commit for something that was already over-engineered at first. XD But if I have to make more edits on this file maybe I'll ship this!

PR #1234, opened 2020-11-16
pause empty servers

### LINKED ISSUE #1227: Pause the game at startup until a real player joins
It would be cool if on a server the game is paused at startup until a real player joins. This will save a lot of energy on idling servers.

Gireen:
fix #1227


Gireen:
well it works as intended except bots disconnect after 4 minutes when the game is paused.
autofill bots reconnect if a client connects to the server.

slipher:
Could it be due to the inactivity timer? For example if you set `g_inactivity 1` (i.e. 1 second) do they get kicked right away?

In any case it seems sensible to disable the activity timer for bots.

Gireen:
no g_inactivity has no effect on the bots. but it was something similar.
in daemon sv_timeout is 240 seconds

slipher:
LGTM, nice and simple implentation.

Gireen:
small additions. now it should work correctly.
(update matchTime that timelimits works, and wait 1 second before pause or else humans win instantly)

illwieckz:
Yay!

PR #1235, opened 2020-11-23
add devteam command

### LINKED ISSUE #1091: implement devteam
It's now possible to load a map that lacks spawn entities by setting `g_neverEnd` to `1`, it allows anyone to load an incomplete or foreign map in engine while mapping/porting. Adding a `devteam` comment would allow mapper to spawn in map to build structures and save it as a layout.

I think about this usage:

- `devteam alien`: spawn as advanced granger at the position the current player is, whatever the current team, including spectator
- `devteam human`: spawn as human with ckit and jetpack at the position the current player is, whatever the current team, including spectator

Would require cheats being enabled of course.

The idea to not require a previous `team` call then a `devteam` one without argument is to prevent the player to return to origin, hence the mapper would be able to freefly in the map and spawn in-place.

Gireen:
fix #1091 
lets player change team and spawn as builder at the current position

Works not 100%. I have here a problem with the view direction, it changes when spawning and makes player turn 180°.

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -520,6 +520,58 @@ char *ConcatArgsPrintable( int start )
 
 /*
 ==================
+Cmd_Devteam_f
+
+Change team and spawn as builder at the current position
+==================
+*/
+void Cmd_Devteam_f( gentity_t *ent )
+{
+	gentity_t *spawn;
+
+	if ( trap_Argc() < 2 )
+	{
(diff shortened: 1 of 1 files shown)


illwieckz:
Oh, that's really helpful! Thanks for taking care of this!

slipher:
LGTM

illwieckz:
> I have here a problem with the view direction, it changes when spawning and makes player turn 180°.

We can investigate this later.

PR #1241, opened 2020-12-09
Display names of left/right function key scancodes

slipher:
It is still not possible to bind these using the menu, as the keyNum_t
one (e.g. SHIFT) comes before the scancode one (e.g. hw:LSHIFT).


slipher:
https://github.com/DaemonEngine/Daemon/pull/404 is the corresponding engine PR.

TODO: update the submodule reference before merging.

PR #1242, opened 2020-12-12
require time without damage for devolving

Gireen:
slipher discovered that devolving can turn adv dragoons into deadly machinegun turrets. 🤯
With this change devolving is only possible when no damage was given or taken for the set amount.
That is 5 seconds by default, witch i think is already enough to prevent abuse and not affect normal usage.

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -2407,12 +2407,27 @@ static bool Cmd_Class_internal( gentity_t *ent, const char *s, bool report )
 				if ( cost != CANT_EVOLVE )
 				{
 
-					if ( ( cost < 0 || ( cost == 0 && currentClass == PCL_ALIEN_LEVEL0 ) )	&& ( G_DistanceToBase( ent ) >= g_devolveMaxBaseDistance.Get() ) ) {
-						if ( report )
-						{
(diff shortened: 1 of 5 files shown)


src/sgame/sg_cmds.cpp
```diff
@@ -2407,12 +2407,27 @@ static bool Cmd_Class_internal( gentity_t *ent, const char *s, bool report )
...
+								G_TriggerMenu( clientNum, MN_A_NOTINBASE );
+							}
+							return false;
+						}
+						if ( ent->client->lastDamageTime &&
+						     ent->client->lastDamageTime + g_devolveDamageCooldown.Get() * 1000 > level.time) {
+							if ( report )
+ 							{
+								float remaining = ( ( ent->client->lastDamageTime + g_devolveDamageCooldown.Get() * 1000 ) - level.time ) / 1000;
+								char* msg = va( "%s %i %.0f", QQ( N_("You cannot devolve until $1$ after combat. Try again in $2$s.") ),
```
illwieckz, line 2423: I guess you forgot a `s` after `$1$`.

Otherwise, that works well !

slipher:
As an alternative, I have made the attack cooldown extend across evolutions in #1243.

Gireen:
Yeah the alternative works better without a waiting time.

PR #1243, opened 2020-12-13
Preserve attack cooldown timer across evolutions

slipher:


DolceTriade (approved):

slipher:
I implemented a gradual return of evolution points after devolving (0.4 points/second by default), as suggested by @Gireen in IRC. This will discourage both devolving right before you would be killed, and devolving in order to replenish barbs. Hence, I have removed the changes to make returned points proportional to health, and to give only 1 barb on spawning, which were meant to discourage those respective things.

slipher:
Put the health fraction multiplier back because it is needed to prevent devolve being too advantageous for faster healing.

PR #1248, opened 2020-12-29
Fix UI papercuts

necessarily-equal:
### Always allow buying upgrade when funds would be sufficient

Sometimes you have enough money to buy something but first need to sell
what conflicting gear you already have. For example when you want to
replace your heavy armor to put on on a light armor and have 100 credits
in your pocket. The game didn't allow you to buy it and required you to
sell it explicitely first. This commit fixes the annoyance.

Example test process:

* /devmap usstremor
* /give momentum 300 
* /give funds 400 
* *buy power armor*
* *try to buy anything else without selling your power armor*
note that if you give yourself 1000 credits on the 0.51, you can buy 
every update, even conflicting, but you can't do that if you have 400 
credits.

### Fix rounding annoyance in displayed morph points

as said on IRC:
    afontain:
            do you have some other ui papercuts in mind?
    peertubed:
            wrong rounding for evos
            like if you have 5.99 and you need 6 to evolve
            then it displays 6.0, but you can't [buy a dragoon]
            should round down

    With this commit, it now rounds down as it should.

Example test process:
 * `/devmap antares`
 * join aliens
 * `/give funds 1.99`
 * look at what's displayed on bottom-right

### DIFF
--- a/src/cgame/cg_rocket_dataformatter.cpp
+++ b/src/cgame/cg_rocket_dataformatter.cpp
@@ -172,14 +172,33 @@ static void CG_Rocket_DFCMArmouryBuyWeapon( int handle, const char *data )
 	Rocket_DataFormatterFormattedData( handle, va( "<button class='armourybuy %s' onMouseover='Events.pushevent(\"setDS armouryBuyList weapons %s\", event)' %s>%s<img src='/%s'/></button>", Class, Info_ValueForKey( data, "2" ), action, Icon, CG_GetShaderNameFromHandle( cg_weapons[ weapon ].ammoIcon )), false );
 }
 
+// inspired by Cmd_Sell_upgrades in src/sgame/sg_cmds.cpp
+static bool CanAffordUpgrade(upgrade_t upgrade, int stats[])
+{
+	playerState_t *ps = &cg.snap->ps;
+	int credits = ps->persistant[ PERS_CREDIT ];
+
+	const int slots = BG_Upgrade( upgrade )->slots;
+
+	for ( int i = UP_NONE; i < UP_NUM_UPGRADES; i++ )
+	{
+		bool usesTheSameSlot = slots & BG_Upgrade( i )->slots;
+		if ( BG_InventoryContainsUpgrade( i, stats ) && usesTheSameSlot )
+		{
+			// conflicting item that will be replaced, add it to funds
+			credits += BG_Upgrade( i )->price;
+		}
+	}
+
+	return BG_Upgrade( upgrade )->price <= credits;
+}
+
 static void CG_Rocket_DFCMArmouryBuyUpgrade( int handle, const char *data )
 {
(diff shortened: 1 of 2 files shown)


illwieckz:
Oh! Thank-you for fixing something I even haven't reported yet (was like… low on my priority list)! :+1:
I confirm the issue, such fix is welcome!

necessarily-equal:
Oh, right, I remember that I wasn't sure about the proper prefix name to give to the new function. Is `CG_Rocket_DFCMCanAffordUpgrade` appropriate? or is the prefix reserved for functions that will be used directly by librocket?

illwieckz:
I don't know! 😅

src/cgame/cg_rocket_dataformatter.cpp
```diff
@@ -172,14 +172,37 @@ static void CG_Rocket_DFCMArmouryBuyWeapon( int handle, const char *data )
 	Rocket_DataFormatterFormattedData( handle, va( "<button class='armourybuy %s' onMouseover='Events.pushevent(\"setDS armouryBuyList weapons %s\", event)' %s>%s<img src='/%s'/></button>", Class, Info_ValueForKey( data, "2" ), action, Icon, CG_GetShaderNameFromHandle( cg_weapons[ weapon ].ammoIcon )), false );
 }
 
+// inspired by Cmd_Sell_upgrades in src/sgame/sg_cmds.cpp
+static bool something_CanAffordUpgrade(upgrade_t upgrade, int stats[])
```
slipher, line 176: For a static function I usually don't use any prefix.

src/cgame/cg_rocket_dataformatter.cpp
```diff
@@ -172,14 +172,37 @@ static void CG_Rocket_DFCMArmouryBuyWeapon( int handle, const char *data )
...
+	for ( int i = UP_NONE; i < UP_NUM_UPGRADES; i++ )
+	{
+		if ( i != upgrade && BG_InventoryContainsUpgrade( i, stats ) )
+		{
+			int slot = BG_Upgrade( i )->slots;
+
+			if ( slots & slot ) // conflicting item
+			{
+				// the item will be replaced: add it to funds
+				credits += BG_Upgrade( upgrade )->price;
```
slipher, line 192: Shouldn't this be `BG_Upgrade( i )` ?

slipher:
Looks good generally. 

`DF` apparently means dataformatter. Not sure about `CM`, maybe context menu.

src/cgame/cg_rocket_dataformatter.cpp
necessarily-equal: good catch, fixed

necessarily-equal:
I've simplified it a bit and rebased on `0.52.0/sync`. Should be good

necessarily-equal:
I've added another commit and updated the description accordingly

PR #1251, opened 2020-12-30
Allow visualisation of 2× healing

necessarily-equal:
it happens when an alien is not too far from another. It's a feature
that wasn't discoverable at all, and most people don't even know it
exists. Hopefully this change could make it a bit more obvious, even if
I still find it a bit confusing.

"Wait, why am I healing now? I didn't even move"

I think it adds a bit of noise, so we may want to look for a better solution instead.

otherwise, to be merged alongside https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/33.

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -1552,6 +1552,7 @@ struct cgMedia_t
 	qhandle_t   healthCross;
 	qhandle_t   healthCross2X;
 	qhandle_t   healthCross3X;
+	qhandle_t   healthCross4X;
 	qhandle_t   healthCrossMedkit;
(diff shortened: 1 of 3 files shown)


illwieckz:
Yeah, git stores no information of moving, so if you want to keep moves in git, you need to do dedicated commits for moves and dedicated commits for edits.

necessarily-equal:
Well, do _you_ want to separate commits? I know the two files are 
renamed already :)

illwieckz:
I guess that file is not that precious to keep track of its move…

DolceTriade:
Looks ok to me.

PR #1252, opened 2020-12-30
Fix context menu regression

necessarily-equal:
see the commit messages

### DIFF
--- a/src/cgame/cg_rocket_dataformatter.cpp
+++ b/src/cgame/cg_rocket_dataformatter.cpp
@@ -404,44 +404,48 @@ static void CG_Rocket_DFCMHumanBuildables( int handle, const char *data )
 static void CG_Rocket_DFCMAlienEvolve( int handle, const char *data )
 {
 	class_t alienClass = (class_t) atoi( Info_ValueForKey( data, "1" ) );
-	const char *Class = "";
+	const char *FunctionalClass = "";
+	const char *CosmeticClass = "";
 	const char *Icon = "";
 	const char *action = "";
 	int cost = BG_ClassCanEvolveFromTo( cg.predictedPlayerState.stats[ STAT_CLASS ], alienClass, cg.predictedPlayerState.persistant[ PERS_CREDIT ] );
 
-	if ( ( alienClass == PCL_ALIEN_BUILDER0 && ( BG_ClassUnlocked( PCL_ALIEN_BUILDER0_UPG ) && !BG_ClassDisabled( PCL_ALIEN_BUILDER0_UPG ) ) )||
-			 ( alienClass == PCL_ALIEN_BUILDER0_UPG && ( !BG_ClassUnlocked( PCL_ALIEN_BUILDER0_UPG ) ) ) )
-	{
-		Rocket_DataFormatterFormattedData( handle, "", false );
-		return;
-	}
-
 	if( cg.predictedPlayerState.stats[ STAT_CLASS ] == alienClass )
 	{
-		Class = "active";
+		FunctionalClass = "active";
 		//Check mark icon. UTF-8 encoding of \uf00c
 		Icon = "<icon class=\"current\">\xEF\x80\x8C</icon>";
 	}
 	else if ( !BG_ClassUnlocked( alienClass ) || BG_ClassDisabled( alienClass ) )
 	{
-		Class = "locked";
+		FunctionalClass = "locked";
 		//Padlock icon. UTF-8 encoding of \uf023
 		Icon = "<icon>\xEF\x80\xA3</icon>";
 	}
 	else if ( cost == CANT_EVOLVE )
 	{
 
-		Class = "expensive";
+		FunctionalClass = "expensive";
 		//$1 bill icon. UTF-8 encoding of \uf0d6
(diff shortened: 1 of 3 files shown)


Gireen:
ha. i never noticed that despite playing also humans 😅
the change you did was also my first approach but later changed cause https://github.com/Unvanquished/Unvanquished/pull/1138#discussion_r351928691

src/cgame/rocket/rocketCircleMenu.h
```diff
@@ -313,7 +313,7 @@ class RocketCircleMenu : public Rocket::Core::Element, public Rocket::Controls::
 					{
 						if ( GetChild( i ) == button )
 						{
-							parameters.Set( "index", button->GetAttribute< int >( "alienclass", 0 ) );
+							parameters.Set( "index", va( "%d", i - 1 ) );
```
Gireen, line 316: how about using the index if alienclass is not set as attribute?
and changing the alienclass to something more generic witch could be used for every cyrclemenu witch has hidden elements in its datasource.

necessarily-equal:
I'm not fond of that, to be honest. It creates a special exception that 
will probably create confusion later on. And possibly bugs like the one 
we just saw :X

Maybe we can have another attribute that every menu would have? I'm 
still not fond of that, to be honest.

necessarily-equal:
With `display: none;`

Screenshot from 2020-12-30 21-08-41

so that's not a solution either.

necessarily-equal:
Here, new version that doesn't mix data and presentation. It's a bit overengineered, but at least it's consistent?

needs https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/34 to work.

(btw, if you could not forget to update the submodule reference when merging, that would be great! I had to bisect to figure this out, and this was made more difficult because I had to manually find the corresponding commits)

illwieckz:
The thing about replacing granger by advanced granger is that it makes no sense to devolve from advanced granger to granger (given we can now devolve near overmind).

Gireen (approved):
nice solution. looks good to me

src/cgame/cg_rocket_dataformatter.cpp
```diff
@@ -409,14 +409,12 @@ static void CG_Rocket_DFCMAlienEvolve( int handle, const char *data )
...
 	int cost = BG_ClassCanEvolveFromTo( cg.predictedPlayerState.stats[ STAT_CLASS ], alienClass, cg.predictedPlayerState.persistant[ PERS_CREDIT ] );
 
-	if ( ( alienClass == PCL_ALIEN_BUILDER0 && ( BG_ClassUnlocked( PCL_ALIEN_BUILDER0_UPG ) && !BG_ClassDisabled( PCL_ALIEN_BUILDER0_UPG ) ) )||
-			 ( alienClass == PCL_ALIEN_BUILDER0_UPG && ( !BG_ClassUnlocked( PCL_ALIEN_BUILDER0_UPG ) ) ) )
+	if ( ( alienClass == PCL_ALIEN_BUILDER0 && ( BG_ClassUnlocked( PCL_ALIEN_BUILDER0_UPG ) && !BG_ClassDisabled( PCL_ALIEN_BUILDER0_UPG ) ) )
+			|| ( alienClass == PCL_ALIEN_BUILDER0_UPG && ( !BG_ClassUnlocked( PCL_ALIEN_BUILDER0_UPG ) ) ) )
 	{
-		Rocket_DataFormatterFormattedData( handle, "", false );
-		return;
+		Class = "hidden";
```
slipher, line 415: It would be nice for the class name to be something more unique instead of `hidden`, since it is necessary, to understand the code in one repository, to find the corresponding code in the other repository.

necessarily-equal:
hmm, maybe? The other .hidden in 
pkg/unvanquished_src.dpkdir/ui/shared/basics.rcss works equally well to 
understand what's happening, and I'm lacking inspiration for another 
sensible name. Maybe you have a suggestion?

Gireen:
doublegranger 😄
What also good is, if named more specific, that custom huds could change it.

necessarily-equal:
.hidden-entry?

necessarily-equal:
note that in the current state, a custom css couldn't show the entry 
again: it's not click action.

Gireen:
I mean the fact that its hidden is defined in css?
CSS classes should be named after what they affect not what they do. There can be exceptions sure but this is the best practice. In this case the class would be set only for the granger entry that is currently not necessarily needed. 
For example we would decide to have it no more hidden but instead bright green, the class name wouldn't match anymore. ( i know crazy example 😂 )
Or if we would decide to also hide locked entrys the class locked could be added to the display none css.

At the moment i don't think its possible to have custom css jet. but mods could still use this.

necessarily-equal:
If I switch to the `doublegranger` naming, I'll drop the changes in the other stylesheets https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/34/files. Also, the hidden button will still not be clickable. Is that ok for you?

Gireen:
Oh now i see what you mean with not clickable. 
How about moving the granger if statement on its own at the end and add doublegranger to the class string?
Then it should act normal and only add the additional class.

Gireen (approved):
awesome 🎉

Issue #1226, opened 2020-10-08
Medistation can still heal between death and explosion?


illwieckz:
While playing chasm, at some point an mantis killed the medistation, I came with a construction kit to deconstruct it before it explodes. While I was close enough to be a bit above the medistation, the medistation animation was played, even if the buildable was already dead. Because I was too much slow to operate, the medistation exploded.

I don't know if I got healed between medistation death and explosion, but for sure, the animation played and it's unexpected.

bmorel:
I believe this is fixed in 0.52.0/sync?

PR #1254, opened 2021-01-01
medistation: don't heal people when destroyed

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1226.

### DIFF
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -609,7 +609,7 @@ void HMedistat_Think( gentity_t *self )
 	}
 
 	// clear target on power loss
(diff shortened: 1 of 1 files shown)


src/sgame/sg_buildable.cpp
```diff
@@ -609,7 +609,7 @@ void HMedistat_Think( gentity_t *self )
 	}
 
 	// clear target on power loss
-	if ( !self->powered )
+	if ( !self->powered || Entities::HealthOf(self) <= 0.0f )
```
slipher, line 612: I think you can use `Entities::IsDead`

illwieckz (approved):
LGTM

PR #1262, opened 2021-01-03
Fixes and cleanups

necessarily-equal:


src/cgame/cg_drawtools.cpp
```diff
@@ -396,16 +394,12 @@ float CG_FadeAlpha( int startMsec, int totalMsec )
 	// fade out
 	if ( totalMsec - t < FADE_TIME )
 	{
-		fade = ( totalMsec - t ) * 1.0 / FADE_TIME;
+		fade = (float) ( totalMsec - t ) / FADE_TIME;
```
DolceTriade, line 397: nit: Use C++ style casts for new code.

src/cgame/cg_view.cpp
```diff
@@ -94,7 +94,7 @@ void CG_TestModel_f()
 
 	if ( trap_Argc() == 3 )
 	{
-		cg.testModelEntity.backlerp = atof( CG_Argv( 2 ) );
+		cg.testModelEntity.backlerp = (float)atof( CG_Argv( 2 ) );
```
DolceTriade, line 97: nit: ditto

DolceTriade (changes requested):
Thanks for the clean ups! Some minor nits:

Please include rationale for making attrib a float. Also for the last commit, it's actually something like "Replace Q_Clamp with Math::Clamp" so please commit it as such so people aren't confused when going through git logs.

src/cgame/cg_drawtools.cpp
necessarily-equal: I don't really get the point of these fancy casts, and find them unreadable, but :man_shrugging:

Fixed. I've converted one or two extra for good mesure

PR #1264, opened 2021-01-04
Building fixes

necessarily-equal:
magic


illwieckz:
The _“Drop dead code”_ commit looks to be code (hack? workaround?) for legacy Tremulous MD3 turret. Can you rename the title like `Drop dead code for Tremulous MD3 turret (hack?)` or something like that?

The `tag_` thing in Quake 3 MD3 models are special points in multi-mesh models to rotate things for example or attach things. Like the weapon would be attached on a hand, the torso would rotate over legs, etc. So basically the Tremulous turret was expected to be multi-mesh and rotate the canon on its pod.

DolceTriade (approved):

PR #1270, opened 2021-01-06
Refactor the evolve logic

necessarily-equal:
Fixes the displayed cost in evolve list, before it was displayed that evolveCost is equal to 0 when it was impossible to evolve.

Includes a second commit for cleaner display of evo points (along with https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/35).

### DIFF
--- a/src/cgame/cg_rocket_datasource.cpp
+++ b/src/cgame/cg_rocket_datasource.cpp
@@ -1661,7 +1661,13 @@ void CG_Rocket_BuildAlienEvolveList( const char *table )
 			Info_SetValueForKey( buf, "num", va( "%d", i ), false );
 			Info_SetValueForKey( buf, "name", BG_ClassModelConfig( i )->humanName, false );
 			Info_SetValueForKey( buf, "description", BG_Class( i )->info, false );
-			Info_SetValueForKey( buf, "price", va( "%.1f", price / CREDITS_PER_EVO ), false );
+			if (price >= 0.0f) {
+				Info_SetValueForKey( buf, "price", va( "Price: %.1f", price / CREDITS_PER_EVO ), false );
+			}
+			else
+			{
+				Info_SetValueForKey( buf, "price", va( "Returned: %.1f", -price / CREDITS_PER_EVO ), false );
+			}
 
 			Rocket_DSAddRow( "alienEvolveList", "default", buf );
 
--- a/src/cgame/cg_rocket_datasource.cpp
+++ b/src/cgame/cg_rocket_datasource.cpp
@@ -1653,11 +1653,9 @@ void CG_Rocket_BuildAlienEvolveList( const char *table )
 			}
 
 			buf[ 0 ] = '\0';
-			price = BG_CostToEvolve( cg.predictedPlayerState.stats[ STAT_CLASS ], i );
-			if( price == CANT_EVOLVE){
-				price = 0;
-			}
-			if( price < 0 ){
+			price = static_cast<float>( BG_ClassEvolveInfoFromTo(
+						cg.predictedPlayerState.stats[ STAT_CLASS ], i ).evolveCost );
+			if( price < 0.0f ){
 				price *= (( float ) cg.predictedPlayerState.stats[ STAT_HEALTH ] / ( float ) BG_Class( cg.predictedPlayerState.stats[ STAT_CLASS ] )->health ) * DEVOLVE_RETURN_FRACTION;
 			}
 			Info_SetValueForKey( buf, "num", va( "%d", i ), false );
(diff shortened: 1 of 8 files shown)


Gireen:
interesting. When does this problem occur? It should (and does for me) only display 0 for granger and dretch.

necessarily-equal:
I did display 0 from granger to dretch and from dretch to granger, but 
evolving to, say, tyrant did default to 0 because CANT_EVOLVE isn't a 
proper cost.

src/shared/bg_public.h
```diff
@@ -133,6 +133,12 @@ struct playerState_t
 	int           misc[ MAX_MISC ]; // misc data
 };
 
+// the possibily and the cost of evolving to an alien form
+struct evolveInfo_t {
+	bool canEvolve;
```
slipher, line 138: The name seems misleading. `formUnlocked` would make more sense

src/shared/bg_misc.cpp
```diff
@@ -412,63 +412,54 @@ bool BG_ClassHasAbility( int pClass, int ability )
...
 
-	return evolveCost;
+	return { canEvolve, evolveCost };
 }
 
 /*
 ==============
 BG_AlienCanEvolve
+answers true if the alien can evolve to any other form.
+FIXME:	this function will always return true because it will notice you can
```
slipher, line 450: weird spacing

src/sgame/sg_cmds.cpp
```diff
@@ -2452,39 +2452,34 @@ static bool Cmd_Class_internal( gentity_t *ent, const char *s, bool report )
...
 
 			if ( G_RoomForClassChange( ent, newClass, infestOrigin ) )
 			{
-				if ( cost != CANT_EVOLVE )
+				if ( evolveInfo.canEvolve && ent->client->pers.credit >= evolveInfo.evolveCost )
 				{
-
-					if ( ( cost < 0 || ( cost == 0 && currentClass == PCL_ALIEN_LEVEL0 ) )	&& ( G_DistanceToBase( ent ) >= g_devolveMaxBaseDistance.Get() ) ) {
+					bool devolving = evolveInfo.evolveCost < 0 ||
+						( evolveInfo.evolveCost == 0 && currentClass == PCL_ALIEN_LEVEL0 );//CHECKME!
```
slipher, line 2462: checked

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -1653,11 +1653,9 @@ void CG_Rocket_BuildAlienEvolveList( const char *table )
...
 
 			buf[ 0 ] = '\0';
-			price = BG_CostToEvolve( cg.predictedPlayerState.stats[ STAT_CLASS ], i );
-			if( price == CANT_EVOLVE){
-				price = 0;
-			}
-			if( price < 0 ){
+			price = static_cast<float>( BG_ClassEvolveInfoFromTo(
+						cg.predictedPlayerState.stats[ STAT_CLASS ], i ).evolveCost );
+			if( price < 0.0f ){
```
slipher, line 1658: Nit: also check that the class is unlocked (according to the evolveInfo_t) instead of assuming the value is nonnegative in that case?

src/shared/bg_public.h
```diff
@@ -133,6 +133,12 @@ struct playerState_t
 	int           misc[ MAX_MISC ]; // misc data
 };
 
+// the possibily and the cost of evolving to an alien form
```
slipher, line 136: spelling: possibility

src/cgame/cg_rocket_datasource.cpp
necessarily-equal: seems fine to me. Let's say I'm a tyrant and that we lost a lot of momentum so that adv goon is locked again. It would still make sense to say that I will recover 1.8 morph point if I devolve to adv goon.

src/shared/bg_public.h
necessarily-equal: updated the name

slipher:
Evolution changes LGTM. We should think more about evolving between classes of the same cost though. Do we allow unlimited no-cost evolution between adv. marauder and dragoon? (That wasn't allowed before). Also, dretch->granger evolution is gameplay-affecting since the non-advanced granger can now reach places that require wallwalking. But probably still can't place buildings on the walls/ceiling.

necessarily-equal:
> Do we allow unlimited no-cost evolution between adv. marauder and dragoon? (That wasn't allowed before).

I've added a commit so that this is considered to be devolving, so only allowed next to the overmind. That seems reasonable.

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -1661,7 +1661,13 @@ void CG_Rocket_BuildAlienEvolveList( const char *table )
 			Info_SetValueForKey( buf, "num", va( "%d", i ), false );
 			Info_SetValueForKey( buf, "name", BG_ClassModelConfig( i )->humanName, false );
 			Info_SetValueForKey( buf, "description", BG_Class( i )->info, false );
-			Info_SetValueForKey( buf, "price", va( "%.1f", price / CREDITS_PER_EVO ), false );
+			if (price >= 0.0f) {
+				Info_SetValueForKey( buf, "price", va( "Price: %.1f", price / CREDITS_PER_EVO ), false );
```
Gireen, line 1665: should this better be done with an dataformatter?

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -1661,7 +1661,13 @@ void CG_Rocket_BuildAlienEvolveList( const char *table )
...
+			}
+			else
+			{
+				Info_SetValueForKey( buf, "price", va( "Returned: %.1f", -price / CREDITS_PER_EVO ), false );
```
Gireen, line 1669: is returned right in this case as the opposite to price?
maybe something like refund matches better.

necessarily-equal:
I've used returned to avoid refunded actually: humans have a monetary 
unit (credits) and I tried to avoid reminding people that the machanism 
is actually pretty similar.

Maybe "recycled" would be a better term?

src/sgame/sg_cmds.cpp
```diff
@@ -2458,7 +2458,7 @@ static bool Cmd_Class_internal( gentity_t *ent, const char *s, bool report )
 			{
 				if ( evolveInfo.classIsUnlocked && ent->client->pers.credit >= evolveInfo.evolveCost )
 				{
-					bool devolving = evolveInfo.evolveCost < 0 ||
```
slipher: Doesn't this make granger->dretch "devolving"? We don't want that.

slipher:
I like "returned"

necessarily-equal:
I've added code for the dretch->granger special cases

src/cgame/cg_rocket_datasource.cpp
necessarily-equal: probably, though it seems pretty complex for a simple thing

PR #1273, opened 2021-01-07
load defaultmaprotation.cfg

### LINKED ISSUE #1268: Add default configuration to run local public servers
It would be nice if the content of dist/configs would be used by default if running a server

Gireen:
fixes #1268
requires a defaultmaprotation.cfg 
with a simple rotation named defaultRotation1


src/sgame/sg_maprotation.cpp
```diff
@@ -1468,15 +1468,13 @@ bool G_MapRotationActive()
...
 ===============
-G_InitMapRotations
+G_LoadMaprotaion
 
-Load and initialise the map rotations
+Load a maprotation file if it exists
 ===============
 */
-void G_InitMapRotations()
+void G_LoadMaprotaion( const char *fileName )
```
illwieckz, line 1476: You mean

```diff
- G_LoadMaprotaion
+ G_LoadMaprotation
```
right? =)

src/sgame/sg_maprotation.cpp
```diff
@@ -1489,12 +1487,29 @@ void G_InitMapRotations()
...
+/*
+===============
+G_InitMapRotations
+
+Load and initialise the map rotations
+===============
+*/
+void G_InitMapRotations()
+{
+	G_LoadMaprotaion( "defaultmaprotation.cfg" );
```
illwieckz, line 1501: What if we do: `default/maprotation.cfg`?

That's not required to make this working, but maybe we would have more default files in the future and a prefix can be useful?

src/sgame/sg_maprotation.cpp
```diff
@@ -1489,12 +1487,29 @@ void G_InitMapRotations()
...
+===============
+G_InitMapRotations
+
+Load and initialise the map rotations
+===============
+*/
+void G_InitMapRotations()
+{
+	G_LoadMaprotation( "default/maprotation.cfg" );
+	G_LoadMaprotation( "maprotation.cfg" );
```
bmorel, line 1502: won't this overwrite default/maprotation.cfg?

Gireen: yes and no.
there can be multiple rotation files loaded.
but if there is a defaultRotation1 in maprotation.cfg it could be overwritten.

illwieckz:
I guess we can merge that? that looks good to me.

src/sgame/sg_maprotation.cpp
```diff
@@ -1505,7 +1505,11 @@ void G_InitMapRotations()
 	{
 		if ( g_initialMapRotation.string[ 0 ] != 0 )
 		{
-			G_StartMapRotation( g_initialMapRotation.string, false, true, false, 0 );
+			if( !G_StartMapRotation( g_initialMapRotation.string, false, true, false, 0 ) )
+			{
+				Log::Warn( "failed to load g_initialMapRotation: %s ", g_initialMapRotation.string );
+				G_StartMapRotation( "defaultRotation1", false, true, false, 0 );
```
slipher, line 1511: Can we lose the "1" i.e. `defaultRotation`?

src/sgame/sg_maprotation.cpp
```diff
@@ -1505,7 +1505,11 @@ void G_InitMapRotations()
 	{
 		if ( g_initialMapRotation.string[ 0 ] != 0 )
 		{
-			G_StartMapRotation( g_initialMapRotation.string, false, true, false, 0 );
+			if( !G_StartMapRotation( g_initialMapRotation.string, false, true, false, 0 ) )
+			{
+				Log::Warn( "failed to load g_initialMapRotation: %s ", g_initialMapRotation.string );
```
slipher, line 1510: extra space after `%s`

slipher (commented):
What's the plan for packaging? Make another copy of the `dist/` rotation in unvanquished_src.dpkdir?

Gireen:
> What's the plan for packaging? Make another copy of the `dist/` rotation in unvanquished_src.dpkdir?

yes, or moving instead copying

PR #1275, opened 2021-01-08
Wording fix in server.cfg sample

Pretuer:
Simplified some language to allow "Dummies" (reffering to the book sieres) to understand.

Also certain text were pointing the person to another direction so I changed them

Above to below eg.

### DIFF
--- a/dist/configs/config/server.cfg
+++ b/dist/configs/config/server.cfg
@@ -1,21 +1,21 @@
 // Example Unvanquished dedicated server config
 
 // advertise the server on public server list
-// 0 - advertise everything
-// 1 - don't advertise but reply to status queries
-// 2 - don't reply to status queries but accept connections
-// 3 - only accept LAN connections
+// 0 - advertise everything - this will make the server viewable to everyone looking at the list
+// 1 - don't advertise but reply to status queries - this allows players to join and see what map is being played, how many people are playing etc. 
+// 2 - don't reply to status queries but accept connections - players can connect but not see server info 
+// 3 - only accept LAN connections 
 set server.private 0
 
 // server can be joined without password
-// set 1 to require password (see g_password above)
+// set 1 to require password (see g_password below)
 set g_needpass 0
 
 // password to join server
-//set g_password "secret"
+//set g_password "(your password here)"
 
 // rcon password to remotely manage server
-//set rcon.server.password "secret"
+//set rcon.server.password "(admin password here) "
 
 // loading screen
 set sv_hostname "^NUnvanquished ^3alpha server"
@@ -60,7 +60,7 @@ set g_emptyTeamsSkipMapTime 15
 // force each team to have a similar number of players
 set g_teamForceBalance 1
 
-// enable warmup for 20 second
+// enable warmup for 20 second - the game will do a countdown before starting 
 //set g_doWarmup 1
 //set g_warmup 20
 
--- a/dist/configs/config/server.cfg
+++ b/dist/configs/config/server.cfg
@@ -1,7 +1,7 @@
 // Example Unvanquished dedicated server config
 
 // advertise the server on public server list
-// 0 - advertise everything - this will make the server viewable to everyone looking at the list
+// 0 - advertise everything, the server will be added to the public server list and information about it will be displayed to players
 // 1 - don't advertise but reply to status queries - this allows players to join and see what map is being played, how many people are playing etc. 
 // 2 - don't reply to status queries but accept connections - players can connect but not see server info 
 // 3 - only accept LAN connections 
--- a/dist/configs/config/server.cfg
+++ b/dist/configs/config/server.cfg
@@ -2,9 +2,9 @@
 
 // advertise the server on public server list
 // 0 - advertise everything, the server will be added to the public server list and information about it will be displayed to players
-// 1 - don't advertise but reply to status queries - this allows players to join and see what map is being played, how many people are playing etc. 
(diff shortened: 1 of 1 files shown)


dist/configs/config/server.cfg
```diff
@@ -1,21 +1,21 @@
...
+// set 1 to require password (see g_password below)
 set g_needpass 0
 
 // password to join server
-//set g_password "secret"
+//set g_password "(your password here)"
 
 // rcon password to remotely manage server
-//set rcon.server.password "secret"
+//set rcon.server.password "(admin password here) "
```
illwieckz, line 18: unneeded extra whitespace before the ending double-quote

```diff
- //set rcon.server.password "(admin password here) "
+ //set rcon.server.password "(admin password here)"
```

dist/configs/config/server.cfg
```diff
@@ -1,21 +1,21 @@
 // Example Unvanquished dedicated server config
 
 // advertise the server on public server list
-// 0 - advertise everything
-// 1 - don't advertise but reply to status queries
-// 2 - don't reply to status queries but accept connections
-// 3 - only accept LAN connections
+// 0 - advertise everything - this will make the server viewable to everyone looking at the list
```
illwieckz, line 4: Better use a comma instead of ` - ` (same for other lines).

« make the server viewable » is an ambiguous concept.

Maybe something like this is better:

> 0 - advertise everything, the server will be added to the public server list and information about it will be displayed to players

dist/configs/config/server.cfg
```diff
@@ -1,21 +1,21 @@
...
+// 1 - don't advertise but reply to status queries - this allows players to join and see what map is being played, how many people are playing etc.
```
illwieckz, line 5: The correct typography for _et cætera_ abbrevation is:

```
[…] playing, etc.
```

Because it's part of the enumeration.

So this looks better this way:

```diff
- // 1 - don't advertise but reply to status queries - this allows players to join and see what map is being played, how many people are playing etc.
+ // 1 - don't advertise but reply to status queries, this allows players to join and see what map is being played, how many people are playing, etc.
```

dist/configs/config/server.cfg
```diff
@@ -1,21 +1,21 @@
...
 
 // advertise the server on public server list
-// 0 - advertise everything
-// 1 - don't advertise but reply to status queries
-// 2 - don't reply to status queries but accept connections
-// 3 - only accept LAN connections
+// 0 - advertise everything - this will make the server viewable to everyone looking at the list
+// 1 - don't advertise but reply to status queries - this allows players to join and see what map is being played, how many people are playing etc. 
+// 2 - don't reply to status queries but accept connections - players can connect but not see server info 
+// 3 - only accept LAN connections
```
illwieckz, line 7: Useless trailing whitespaces there and there.

illwieckz:
Here is your suggestion with some fixes (commas, white space, etc.):

```
// Example Unvanquished dedicated server config

// advertise the server on public server list
// 0 - advertise everything, the server will be added to the public server list and information about it will be displayed to players
// 1 - don't advertise but reply to status queries, this allows players to join and see what map is being played, how many people are playing, etc.
// 2 - don't reply to status queries but accept connections, players can connect but not see server information
// 3 - only accept LAN connections
set server.private 0

// server can be joined without password
// set 1 to require password (see g_password below)
set g_needpass 0

// password to join server
//set g_password "(your password here)"

// rcon password to remotely manage server
//set rcon.server.password "(admin password here)"

// loading screen
set sv_hostname "^NUnvanquished ^3alpha server"
set g_motd "^2get news on ^5unvanquished.net"

// disallow downloads from the server
set sv_allowdownload 0

// direct and http download can be setup this way
//set sv_dl_maxRate "1000"
//set sv_wwwBaseURL "example.com/unvanquished/pkg"
//set sv_wwwDlDisconnected "0"
//set sv_wwwDownload "1"
//set sv_wwwFallbackURL "dl.unvanquished.net/pkg"

// a stat web page can be advertized this way
// the link will be displayed in compatible server browsers
//set sv_statsURL "http://example.com/unvanquished/stats"

// maximum number of clients
// greater than 24 is not recommended from a gameplay point of view
set sv_maxclients 24

// set 25 slot with 1 private, provides 24 public slots
// while admin can still connect and fill the extra slot
// while server is full
//set sv_maxclients 25
//set sv_privateclients 1
//set sv_privatePassword "secret"

// no time limit
set timelimit 0

// end the game as a draw after 45 minutes
//set timelimit 45

// load next map if no people is playing on server at this time
// it ensures money, evolution and build points are regularly reset
// on fresh map so real player are not welcomed on a stalled bot match
set g_emptyTeamsSkipMapTime 15

// force each team to have a similar number of players
set g_teamForceBalance 1

// enable warmup for 20 second, the game will do a countdown before starting
//set g_doWarmup 1
//set g_warmup 20

// load per map config file from this subdirectory
// it will load map/<mapname>.cfg like map/plat23.cfg
// or fallback on map/default.cfg
// map config files can be used to fill game with bots
set g_mapConfigs "map"

// following the first map, start this rotation
// rotation is described in game/maprotation.cfg
set g_initialMapRotation rotation1

// start this map first, usually the last of the rotation
// since rotation will start from the first one after that
map yocto
```

I also fixed a typo I probably introduced myself before (`slotes`).

If you replace the complete file by this it may be OK to merge.

Pretuer:
I have edited it

PR #1276, opened 2021-01-08
Rework team damage handling

necessarily-equal:
Fixes https://github.com/Unvanquished/gameplay/issues/10, at least partially.

Note that the team damage settings still does apply only to the *other* members of a team. Do we really want to keep that? IMO it makes no sense, as if a lucifer cannon is less dangerous for your friends, it makes sense if it's less dangerous for you. Idem for the barbs. Should we change this?

The first commit comes with the corresponding change in daemon: https://github.com/DaemonEngine/Daemon/pull/416.

### DIFF
--- a/src/sgame/components/HealthComponent.cpp
+++ b/src/sgame/components/HealthComponent.cpp
@@ -3,6 +3,10 @@
 
 static Log::Logger healthLogger("sgame.health");
 
+static Cvar::Cvar<float>  g_friendlyFireAlienMultiplier("g_friendlyFireAlienMultiplier", "Damage multiplier for alien-to-alien damage (doesn't include buildings, or self)", Cvar::NONE, 0.5f);
+static Cvar::Cvar<float>  g_friendlyFireHumanMultiplier("g_friendlyFireHumanMultiplier", "Damage multiplier for human-to-human damage (doesn't include buildings, or self)", Cvar::NONE, 1.0f);
+static Cvar::Cvar<bool>   g_friendlyBuildableFire("g_friendlyBuildableFire", "Can one hurt the team's buildable?", Cvar::NONE, true);
+
 HealthComponent::HealthComponent(Entity& entity, float maxHealth)
 	: HealthComponentBase(entity, maxHealth), health(maxHealth)
 {}
@@ -76,9 +80,6 @@ Util::optional<Vec3> direction, int flags, meansOfDeath_t meansOfDeath) {
 	if (!(flags & DAMAGE_NO_PROTECTION)) {
 		// Check for protection from friendly damage.
 		if (entity.oldEnt != source && G_OnSameTeam(entity.oldEnt, source)) {
-			// Check if friendly fire has been disabled.
-			if (!g_friendlyFire.integer) return;
-
 			// Never do friendly damage on movement attacks.
 			switch (meansOfDeath) {
 				case MOD_LEVEL3_POUNCE:
@@ -111,7 +112,7 @@ Util::optional<Vec3> direction, int flags, meansOfDeath_t meansOfDeath) {
 		if (entity.oldEnt->s.eType == entityType_t::ET_BUILDABLE && source->client &&
 		    meansOfDeath != MOD_DECONSTRUCT && meansOfDeath != MOD_SUICIDE &&
 		    meansOfDeath != MOD_REPLACE) {
-			if (G_OnSameTeam(entity.oldEnt, source) && !g_friendlyBuildableFire.integer) {
+			if (G_OnSameTeam(entity.oldEnt, source) && !g_friendlyBuildableFire.Get()) {
 				return;
 			}
 		}
@@ -119,9 +120,32 @@ Util::optional<Vec3> direction, int flags, meansOfDeath_t meansOfDeath) {
 
 	float take = amount;
 
+	// reduce team damage according to the team, ignoring buildings.
+	// Note that team damage is reduced only for the *other* members of the
+	// team, otherwise you could simply, say, sit on a grenade and lure
+	// aliens to come to you.
+	if (entity.oldEnt != source && G_OnSameTeam(entity.oldEnt, source)
+			&& entity.oldEnt->s.eType != entityType_t::ET_BUILDABLE) {
+		if ( G_Team(source) == TEAM_ALIENS )
+		{
+			take *= g_friendlyFireAlienMultiplier.Get();
+		}
+		else
+		{
+			take *= g_friendlyFireHumanMultiplier.Get();
+		}
+	}
+
+
 	// Apply damage modifiers.
 	if (!(flags & DAMAGE_PURE)) {
 		entity.ApplyDamageModifier(take, location, direction, flags, meansOfDeath);
+	}
+
+	// if we don't actually take damage, stop here
+	if ( take == 0.0f )
+	{
+		return;
 	}
 
 	// Update combat timers.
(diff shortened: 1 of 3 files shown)


slipher:
Is this acid tube damage curve thing discussed somewhere? It seems unrelated to the linked issue.

necessarily-equal:
No, it hasn't been discussed before. I can move it to another PR if 
desired.

slipher:
Yes please move, it is OK to lump together some unrelated technical fixes, but this is not good with novel gameplay proposals.

necessarily-equal:
move the acid tube thing to https://github.com/Unvanquished/Unvanquished/pull/1285

src/sgame/sg_main.cpp
```diff
@@ -365,8 +366,10 @@ static cvarTable_t gameCvarTable[] =
 
 	// gameplay: basic
 	{ &g_timelimit,                   "timelimit",                     "45",                               CVAR_SERVERINFO,                                 0, true     , nullptr       },
-	{ &g_friendlyFire,                "g_friendlyFire",                "1",                                CVAR_SERVERINFO,                                 0, true     , nullptr       },
-	{ &g_friendlyBuildableFire,       "g_friendlyBuildableFire",       "1",                                CVAR_SERVERINFO,                                 0, true     , nullptr       },
+	{ &g_friendlyFireAlienMultiplier, "g_friendlyFireAlienMultiplier", "0.5",                              0,                                 0, true     , nullptr       },
```
slipher, line 369: Please use the C++ cvars (e.g. `Cvar::Cvar<float>`) for new ones

src/sgame/components/HealthComponent.cpp
```diff
@@ -119,11 +116,32 @@ Util::optional<Vec3> direction, int flags, meansOfDeath_t meansOfDeath) {
...
 	// Apply damage modifiers.
 	if (!(flags & DAMAGE_PURE)) {
 		entity.ApplyDamageModifier(take, location, direction, flags, meansOfDeath);
 	}
 
+	// if we don't actually take damage, stop here
+	if ( take == 0.0f )
+	{
+		return;
+		// FIXME: this disables the damage indicator, but still shows blood
```
slipher, line 142: Not a bad thing IMO, to show what you are hitting.

src/sgame/components/HealthComponent.cpp
```diff
@@ -119,11 +116,32 @@ Util::optional<Vec3> direction, int flags, meansOfDeath_t meansOfDeath) {
 
 	float take = amount;
 
+	// halve team damage according to the team, ignoring buildings
```
slipher, line 119: Comment shouldn't say "halve" since it's configurable?

slipher:
I don't favor the change to apply friendly-fire modifiers to self-damage. Particularly in the case of grenades.

necessarily-equal:
I've removed the last commit entirely for now

src/sgame/sg_main.cpp
necessarily-equal: Maybe it makes more sense to convert them in a latter PR so that they are updated in bulk?

src/sgame/components/HealthComponent.cpp
necessarily-equal: updated

src/sgame/components/HealthComponent.cpp
necessarily-equal: Well, I would disagree, but I've removed the comment for now to avoid pushing that way. If someone gets annoyed by it they can discuss it.

src/sgame/sg_main.cpp
slipher: It's not like this is the first time C++ cvars are being introduced in the sgame. There are many new-style cvars already, so just old ones that haven't been converted are here.

necessarily-equal: updated

slipher:
Looks like there are compile errors.

src/sgame/sg_extern.h
```diff
@@ -172,6 +170,9 @@ extern  vmCvar_t g_instantBuilding;
 
 extern  Cvar::Cvar<float> g_evolveAroundHumans;
 extern  Cvar::Cvar<float> g_devolveMaxBaseDistance;
+extern  Cvar::Cvar<float> g_friendlyFireAlienMultiplier;
```
slipher, line 173: Why not put these in HealthComponent.cpp so you don't need to forward-declare them?

src/sgame/components/HealthComponent.cpp
```diff
@@ -111,19 +108,42 @@ Util::optional<Vec3> direction, int flags, meansOfDeath_t meansOfDeath) {
...
 		}
 	}
 
 	float take = amount;
 
+	// reduce team damage according to the team, ignoring buildings.
+	// Note that team damage is reduced only for the *other* members of the
+	// team, otherwise you could simply, say, sit on a grenade and lure
+	// aliens to come to you.
+	if (entity.oldEnt != source && G_OnSameTeam(entity.oldEnt, source)
```
slipher, line 123: This stuff should go inside `if (!(flags & DAMAGE_NO_PROTECTION))`

necessarily-equal: To raise this back from the dead, why should it be moved inside? The only documentation about it I've found is this:
```C
#define DAMAGE_NO_PROTECTION 0x00000004 /**< Game settings don't prevent damage. */
```

Also note these two:
```C
#define DAMAGE_PURE          0x00000001 /**< Amount won't be modified. */
#define DAMAGE_NO_LOCDAMAGE  0x00000008 /**< Don't apply locational modifier. */
```

Do you mean I should move it inside the `if (!(flags & DAMAGE_PURE))` instead?

To go back on `DAMAGE_NO_PROTECTION`, I feel like its both unused, and broken: the only user is the overmind, and it suggests that is supposed to deal as much damage to an armored opponent as it does to an unarmored one, but this is not what is happening: armor protects anyway. Maybe if you can explain the correct behavior or some documentation about the intended usage, I can correct it or document it.

necessarily-equal: To raise this back from the dead, why should it be moved inside? The only documentation about it I've found is this:
```C
#define DAMAGE_NO_PROTECTION 0x00000004 /**< Game settings don't prevent damage. */
```

Also note these two:
```C
#define DAMAGE_PURE          0x00000001 /**< Amount won't be modified. */
#define DAMAGE_NO_LOCDAMAGE  0x00000008 /**< Don't apply locational modifier. */
```

Do you mean I should move it inside the `if (!(flags & DAMAGE_PURE))` instead?

To go back on `DAMAGE_NO_PROTECTION`, I feel like its both unused, and broken: the only user is the overmind, and it suggests that is supposed to deal as much damage to an armored opponent as it does to an unarmored one, but this is not what is happening: armor protects anyway. Maybe if you can explain the correct behavior or some documentation about the intended usage, I can correct it or document it.

bmorel:
This patch seems fine to me, I like the idea of having more per-team configs instead of whole config, since the game tries to have a different gameplay for each time, this is really desirable.
One thing though:  it would be better to move the cvar into an array (would prevent checking for teams to know which cvar to change).

src/sgame/components/HealthComponent.cpp
slipher: DAMAGE_NO_PROTECTION means the damage must not be reduced by friendly fire cvars. It is used by other things besides the overmind attack. For example, the patch in its current form causes telefragging not to work - instead of killing the target it only removes 50% of the health.

slipher:
241afdeef5cff3615d7318d20bd3aab67fa4076a

PR #1280, opened 2021-01-09
Update sg_cmds.cpp

Pretuer:
Reduced the kill-time for the suicide in 2 from 20 seconds to 10 seconds. as previously it was long. also grammar.

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -788,8 +788,8 @@ void Cmd_Kill_f( gentity_t *ent )
 	{
 		if ( ent->suicideTime == 0 )
 		{
-			trap_SendServerCommand( ent - g_entities, "print_tr \"" N_("You will suicide in 20 seconds") "\"" );
-			ent->suicideTime = level.time + 20000;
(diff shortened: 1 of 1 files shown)


illwieckz:
We can't easily rename the `kill` command as this name is established since 1999 and is widely used that way in probably all quake3 derivatives, so players from the past 22 years expect this command to be named `kill`, even if I agree that name is too much ambiguous.

I'm not saying we will never rename things (we already renamed things and will rename more) but… well, this is not something without consequences.

Pretuer:
I've removed the rename. I'm more focused on the cool down/time to die

slipher:
The `kill` timeout must be long so that it can't easily be abused to build bases faster.

Pretuer:
> must be long

Okay sure, 10 seconds is long enough 15 is okay anything below 10 is short

bmorel:
I think it would be a good idea to reduce it, but I also think it should be configurable by server owner (for parkours, too small values might be not that nice).
Also, I'd be curious to see if the travis build is really broken or not (I think: it's not.)

PR #1281, opened 2021-01-10
Fix some bugs with corpses

necessarily-equal:


DolceTriade (approved):

src/shared/bg_utilities.cpp
```diff
@@ -165,3 +165,9 @@ void RemoveFlag(int &flags, int flag) {
 void ToggleFlag(int &flags, int flag) {
 	flags ^= flag;
 }
+
+// sets the first argument the new anim, but toogling the toogle bit
+void BG_NewAnim(int &anim, int newAnim) {
```
slipher, line 170: It is not good style to use a reference here - just take one input and return something.

But anyway I don't think this commit is all that helpful in general. It's hard to tell what this function does looking at the name; I'd rather just have the code right there at the callsite, as it is now.

necessarily-equal:
I've removed the commit for now

slipher:
It seems, you removed the wrong commit.

necessarily-equal:
fixed

PR #1282, opened 2021-01-10
WIP: port old-style vm commands to C++ Command

necessarily-equal:
Asking for a review: does this way of rewriting commands seems to be the way to go? Is there a modernization in the way CVars can be set? Anything else?

Should I convert command names to camelCase in the process?

Also, scaleup and scaledown seems to be pretty weird commands, that don't have any purpose. Should we drop them?

### DIFF
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -31,27 +31,39 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 /*
 =================
-CG_SizeUp_f
+SizeUp
 
 Keybinding command
 =================
 */
-static void CG_SizeUp_f()
-{
-	trap_Cvar_Set( "cg_viewsize", va( "%i", std::min( cg_viewsize.integer + 10, 100 ) ) );
-}
+class SizeUpCmd : public Cmd::StaticCmd {
+	public:
+		SizeUpCmd(): StaticCmd("sizeup", Cmd::KEY_BINDING, "Scale up the view size") {
+		}
+
+		virtual void Run(const Cmd::Args&) const override {
+			trap_Cvar_Set( "cg_viewsize", va( "%i", std::min( cg_viewsize.integer + 10, 100 ) ) );
+		}
+};
+static SizeUpCmd sizeUpRegistration;
 
 /*
 =================
-CG_SizeDown_f
+SizeDown
 
 Keybinding command
 =================
 */
-static void CG_SizeDown_f()
-{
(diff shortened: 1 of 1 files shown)


src/cgame/cg_consolecmds.cpp
```diff
@@ -31,27 +31,39 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
-static void CG_SizeUp_f()
-{
-	trap_Cvar_Set( "cg_viewsize", va( "%i", std::min( cg_viewsize.integer + 10, 100 ) ) );
-}
+class SizeUpCmd : public Cmd::StaticCmd {
+	public:
+		SizeUpCmd(): StaticCmd("sizeup", Cmd::KEY_BINDING, "Scale up the view size") {
+		}
+
+		virtual void Run(const Cmd::Args&) const override {
```
slipher, line 44: do not write `virtual` on an override

src/cgame/cg_consolecmds.cpp
```diff
@@ -31,27 +31,39 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
 Keybinding command
 =================
 */
-static void CG_SizeUp_f()
-{
-	trap_Cvar_Set( "cg_viewsize", va( "%i", std::min( cg_viewsize.integer + 10, 100 ) ) );
-}
+class SizeUpCmd : public Cmd::StaticCmd {
+	public:
+		SizeUpCmd(): StaticCmd("sizeup", Cmd::KEY_BINDING, "Scale up the view size") {
```
slipher, line 41: `KEY_BINDING` means a command for managing binds, not a command used as the target of a keybind.

src/cgame/cg_consolecmds.cpp
```diff
@@ -31,27 +31,39 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 /*
 =================
-CG_SizeUp_f
+SizeUp
```
slipher, line 34: I'd drop the banner; they are usually seen only on functions.

slipher:
> Is there a modernization in the way CVars can be set?

`Cvar::Set` is the direct replacement for `trap_Cvar_Set`. When possible, it is preferable to set the cvar using the cvar object (whether old- or new-style), rather than supplying the name as a string.

necessarily-equal:
By Set, you mean that SetValue that takes a string, right?

bmorel:
The only problem I see in that PR is that it does not follows globally used codestyle, plus slipher's comments.

necessarily-equal:
I'm not working on this at the moment

PR #1283, opened 2021-01-12
Fix compilation error with mismatch type in Math::Clamp

Kangz:


necessarily-equal:
Looks good to me, though maybe with `b->size = Math::Clamp( cgs.bc.hudSize / static_cast<float>( sqrt( b->dist ) ), …` so that the division is made with a float? (for consistency)

Kangz:
Done

DolceTriade (approved):

PR #1285, opened 2021-01-25
Zone damage (acid tube) damage rework

necessarily-equal:
graph

I find this makes the acid tubes quite more dangerous. So we may want to
reduce the base damage a bit more, then.

The intent of this commit is to avoid the situation where an human sits
at the border of the zone of action, and it looks like the acid tube is
doing something, while it really deals 1 damage per minute.


necessarily-equal:
CC @Viech, I think you are interested in gameplay and balancing discussions.

Viech (approved):

PR #1287, opened 2021-01-26
Reactor zap rework

necessarily-equal:
see commits.

@Gireen: do you think buildings like the armory should block the zap too?


Gireen:
Yes. I think everything that can block bullets should also block the reactor zapping. 
So that the poor aliens can hide from the evil reactor. 😢

necessarily-equal:
One PR I've been thinking about was to allow marauder zap to pass through allies to help team fight. So maybe only buildings and walls should block reactor zap for consistency ?

Gireen:
Yes. That should covers most cases.
I don't think that humans exactly between reactor and aliens happens that often to make it a problem.

PR #1289, opened 2021-01-26
don't wallrun while firing jetpack, fix #1288

### LINKED ISSUE #1288: disable wall jump while firing jetpack
Jetpack is hard to use because as soon as you touch a wall (or even the platform you want to land on), you're kicked very fast some meters away, even faster than the jetpack can move you, because of wall jump effect.

Maybe we can keep wall jump mechanism but disable it when firing the jetpack, to keep the best of them without them conflicting together?

illwieckz:
Before:

[jetpack walljump](https://dl.illwieckz.net/b/unvanquished/bugs/jetpack-walljump/unvanquished_2021-01-26_164703_000-jetpack-wall-jump.jpg)

After:

[jetpack walljump](https://dl.illwieckz.net/b/unvanquished/bugs/jetpack-walljump/unvanquished_2021-01-26_164703_000-jetpack-hit-and-fall.jpg)

### DIFF
--- a/src/shared/bg_pmove.cpp
+++ b/src/shared/bg_pmove.cpp
@@ -1052,6 +1052,8 @@ static bool PM_CheckPounce()
 /*
 =============
 PM_CheckWallJump
+
+Used by marauders.
 =============
 */
 static bool PM_CheckWallJump()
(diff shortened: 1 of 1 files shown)


slipher:
According to the `res-players` files, `wallJumper` is used for Marauders while `wallRunner` is used for Humans. So only the wallrun one needs to be changed.

illwieckz:
OK, I edited the patch to only modify the WallRun code, and added comments to help future readers about which function is used by which team.

slipher:
LGTM

PR #1306, opened 2021-02-01
minor readability changes

bmorel:
commit 1: minor changes to make code easier to read
commit 2: same, but by introducing a macro (could be rejected depending on policy, which is why is a 2nd commit)

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1198,21 +1198,21 @@ void BotGetIdealAimLocation( gentity_t *self, botTarget_t target, vec3_t aimLoca
 {
 	//get the position of the target
 	BotGetTargetPos( target, aimLocation );
-	bool target_buildable = BotGetTargetType( target ) == entityType_t::ET_BUILDABLE;
+	bool isTargetBuildable = BotGetTargetType( target ) == entityType_t::ET_BUILDABLE;
 	//this retrieves the target's species, to aim at weak point:
 	// * for humans, it's the head (but code only applies an offset here, with the hope it's the head)
 	// * for aliens, there is no weak point, and human bots try to take missile's speed into consideration (for luci)
-	team_t target_team = BotGetTargetTeam( target );
+	team_t targetTeam = BotGetTargetTeam( target );
 
-	if ( !target_buildable && BotTargetIsEntity( target ) && target_team == TEAM_HUMANS )
+	if ( !isTargetBuildable && BotTargetIsEntity( target ) && targetTeam == TEAM_HUMANS )
 	{
 
 		//aim at head
-		//FIXME: do not rely on hard-coded offset but evaluate which point have lower arm
+		//FIXME: do not rely on hard-coded offset but evaluate which point have lower armor
 		aimLocation[2] += target.ent->r.maxs[2] * 0.85;
 
 	}
-	else if ( target_buildable || target_team == TEAM_ALIENS )
+	else if ( isTargetBuildable || targetTeam == TEAM_ALIENS )
 	{
 		//make lucifer cannons aim ahead based on the target's velocity
 		if ( self->client->ps.weapon == WP_LUCIFER_CANNON && self->botMind->botSkill.level >= 5 )
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1198,15 +1198,21 @@ void BotGetIdealAimLocation( gentity_t *self, botTarget_t target, vec3_t aimLoca
 {
 	//get the position of the target
 	BotGetTargetPos( target, aimLocation );
+	bool target_buildable = BotGetTargetType( target ) == entityType_t::ET_BUILDABLE;
+	//this retrieves the target's species, to aim at weak point:
+	// * for humans, it's the head (but code only applies an offset here, with the hope it's the head)
+	// * for aliens, there is no weak point, and human bots try to take missile's speed into consideration (for luci)
+	team_t target_team = BotGetTargetTeam( target );
 
-	if ( BotGetTargetType( target ) != entityType_t::ET_BUILDABLE && BotTargetIsEntity( target ) && BotGetTargetTeam( target ) == TEAM_HUMANS )
+	if ( !target_buildable && BotTargetIsEntity( target ) && target_team == TEAM_HUMANS )
 	{
 
 		//aim at head
+		//FIXME: do not rely on hard-coded offset but evaluate which point have lower arm
 		aimLocation[2] += target.ent->r.maxs[2] * 0.85;
 
 	}
-	else if ( BotGetTargetType( target ) == entityType_t::ET_BUILDABLE || BotGetTargetTeam( target ) == TEAM_ALIENS )
+	else if ( target_buildable || target_team == TEAM_ALIENS )
 	{
 		//make lucifer cannons aim ahead based on the target's velocity
 		if ( self->client->ps.weapon == WP_LUCIFER_CANNON && self->botMind->botSkill.level >= 5 )
--- a/src/shared/bg_public.h
+++ b/src/shared/bg_public.h
@@ -320,11 +320,11 @@ void Pmove( pmove_t *pmove );
 enum statIndex_t
 {
   STAT_HEALTH,
-  STAT_ITEMS,
+	STAT_ITEMS,      // bitfield of all possible upgrades, probably only for humans.
   STAT_ACTIVEITEMS,
-  STAT_WEAPON,     // current primary weapon
+  STAT_WEAPON,     // current primary weapon. Works for humans and aliens. For aliens, is related to STAT_CLASS.
   STAT_MAX_HEALTH, // health limit
-  STAT_CLASS,      // player class
+  STAT_CLASS,      // player class. Human's armor or alien's form. Do not use to guess team (which is done in some places), instead use PERS_TEAM
   STAT_STATE2,     // more client states
   STAT_STAMINA,    // humans: stamina
   STAT_STATE,      // client states
--- a/src/shared/bg_public.h
+++ b/src/shared/bg_public.h
@@ -320,7 +320,7 @@ void Pmove( pmove_t *pmove );
 enum statIndex_t
 {
   STAT_HEALTH,
-	STAT_ITEMS,      // bitfield of all possible upgrades, probably only for humans.
+  STAT_ITEMS,      // bitfield of all possible upgrades, probably only for humans.
   STAT_ACTIVEITEMS,
   STAT_WEAPON,     // current primary weapon. Works for humans and aliens. For aliens, is related to STAT_CLASS.
   STAT_MAX_HEALTH, // health limit


src/sgame/sg_bot_util.cpp
```diff
@@ -317,35 +317,39 @@ void BotGetDesiredBuy( gentity_t *self, weapon_t *weapon, upgrade_t *upgrades, i
 	int credits = self->client->ps.persistant[PERS_CREDIT];
 	int usableCapital = credits + equipmentPrice;
 
+	//could not use inline function since using "too_expensive" value (usableCapital) inside.
```
slipher, line 320: You could use a lambda function. `auto CanBuy = [usableCapital](weapon_t w) { return ...; };`

src/sgame/sg_bot_ai.cpp
```diff
@@ -746,21 +746,25 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 
 			if ( myTeam == TEAM_HUMANS )
 			{
-				if ( self->botMind->botSkill.level >= 3 && DistanceToGoalSquared( self ) < Square( MAX_HUMAN_DANCE_DIST )
```
slipher, line 749: This change seems less readable to me since it creates variables named "dist" which are actually squared distances.

bmorel:
I've used a lambda instead of a macro: good idea, didn't thought about it (well, 2 lambdas, no choice, not building otherwise) and have reverted the changes you disagreed with.

src/sgame/sg_bot_util.cpp
```diff
@@ -317,35 +317,36 @@ void BotGetDesiredBuy( gentity_t *self, weapon_t *weapon, upgrade_t *upgrades, i
 	int credits = self->client->ps.persistant[PERS_CREDIT];
 	int usableCapital = credits + equipmentPrice;
 
+	//TODO: ensure that adding usableCapital + 1 several times can never overflow
+	//  (overflow an int by summing less than 10 values below than 2000
+	//  is more than unlikely though, but that code is not future-proof)
+	auto CanBuyWeapon = [usableCapital](weapon_t  item) { return BG_WeaponUnlocked(  item ) ? BG_Weapon(  item )->price : usableCapital + 1; };
```
slipher, line 323: Sorry I wasn't paying attention the first time, and assumed these were returning booleans rather than an amount of money... As it is, the functions are misleadingly named, and the logic seems more contorted compared to the original. What's wrong with the original code? The conditions in the original code seem most straightforward.

bmorel: I can change the name. WeaponCost or UpgradeCost would also lie, somehow, since it would return a value which is not real when the item is not unlocked (yet).

What I am trying to improve here is the fact that, to check for any combination, you must check 2 things per equipment, and I suspect that there is even a bug here when one of those equipment is forbidden by configuration, because otherwise you probably need to add a 3rd check for stuff which can be disabled (actually only weapons).
So, to check for a combination like: pulse rifle + medium armor + radar, current code checks for 6 conditions (paired), should probably check for 7 (since it should check if weapon is authorized I think) and if someday radar, jetpack, grenades, armors are added to the list of equipment which can be forbidden (why not?) it would require 3 checks per item. In previous case, that's a total of 9 checks.
Checks which are multiplied by the number of combinations.

I don't think the changes I propose are ideal, far from it (would be better to have some sort of encapsulation somewhere else), but I believe they improve the situation and makes further changes easier to do, more obvious.

slipher: The concept of the code is to buy a given level of armor/upgrades only if there will be enough money left over to buy a reasonably good gun. The bsuit+painsaw is performed to ensure that if the bot buys a bsuit, it will be able to buy something *at least* as expensive as a painsaw. So I think the code is working as intended. 

There is one possible scenario that could be considered sort of a bug - let's say, all guns cheaper than pulse rifle are disabled, and the bot has more money than cost(bsuit) + cost(painsaw) and less money than cost(bsuit) + cost(pulse rifle). Then the bot will end up with bsuit+rifle. In that case it might be better to not buy the bsuit unless it can also afford the pulse rifle.

I don't think this change improves the situation as it just obfuscates the code without changing functionality.

bmorel: Ok, I'll remove those changes from the patch, then.

src/sgame/sg_bot_util.cpp
```diff
@@ -1198,15 +1198,18 @@ void BotGetIdealAimLocation( gentity_t *self, botTarget_t target, vec3_t aimLoca
 {
 	//get the position of the target
 	BotGetTargetPos( target, aimLocation );
+	bool target_buildable = BotGetTargetType( target ) == entityType_t::ET_BUILDABLE;
+	//would be better to just check if "not same team" but dunno how. Also, might try to kill "neutral"?
```
slipher, line 1202: `G_Team( self )` to get the bot's team.

src/shared/bg_public.h
```diff
@@ -320,11 +320,11 @@ void Pmove( pmove_t *pmove );
...
 {
   STAT_HEALTH,
-  STAT_ITEMS,
+  STAT_ITEMS,      // bitfield of all possible upgrades, probably only for humans.
   STAT_ACTIVEITEMS,
-  STAT_WEAPON,     // current primary weapon
+  STAT_WEAPON,     // current primary weapon. Works for humans and aliens. For aliens, is related to STAT_CLASS.
   STAT_MAX_HEALTH, // health limit
-  STAT_CLASS,      // player class
+  STAT_CLASS,      // player class. Human's armor or alien's form. Can also get used to guess team.
```
slipher, line 327: The usual way to get the team from playerstate is with PERS_TEAM, so I'd remove the bit about guessing the team.

slipher (commented):
Please squash the commits so there aren't any reverts.

bmorel:
> Please squash the commits so there aren't any reverts.

Will do when there will be no other issues, I find it easier to track my changes without the squash.

src/sgame/sg_bot_util.cpp
bmorel: In the end, this code retrieves target's species, not really the team. I've updated comment to make that more clear.

src/shared/bg_public.h
bmorel: I think I've seen it in code, using STAT_CLASS to guess species... I've updated the comment to say "don't do that, use PERS_TEAM instead".

src/sgame/sg_bot_util.cpp
```diff
@@ -1198,15 +1198,20 @@ void BotGetIdealAimLocation( gentity_t *self, botTarget_t target, vec3_t aimLoca
...
 	//get the position of the target
 	BotGetTargetPos( target, aimLocation );
+	bool target_buildable = BotGetTargetType( target ) == entityType_t::ET_BUILDABLE;
+	//this retrieves the target's species, to aim at weak point:
+	// * for humans, it's the head (but code only applies an offset here, with the hope it's the head)
+	// * for aliens, there is no weak point, and human bots try to take missile's speed into consideration (for luci)
+	team_t target_team = BotGetTargetTeam( target );
 
-	if ( BotGetTargetType( target ) != entityType_t::ET_BUILDABLE && BotTargetIsEntity( target ) && BotGetTargetTeam( target ) == TEAM_HUMANS )
+	if ( target_buildable == false && BotTargetIsEntity( target ) && target_team == TEAM_HUMANS )
```
necessarily-equal, line 1207:
```suggestion
	if ( !target_buildable && BotTargetIsEntity( target ) && target_team == TEAM_HUMANS )
```

src/sgame/sg_bot_util.cpp
```diff
@@ -1198,15 +1198,20 @@ void BotGetIdealAimLocation( gentity_t *self, botTarget_t target, vec3_t aimLoca
...
 	{
 
 		//aim at head
 		aimLocation[2] += target.ent->r.maxs[2] * 0.85;
 
 	}
-	else if ( BotGetTargetType( target ) == entityType_t::ET_BUILDABLE || BotGetTargetTeam( target ) == TEAM_ALIENS )
+	else if ( target_buildable == true || target_team == TEAM_ALIENS )
```
necessarily-equal, line 1214:
```suggestion
	else if ( target_buildable || target_team == TEAM_ALIENS )
```

But I'm really wondering if that one shouldn't be 
```suggestion
	else if ( !target_buildable && target_team == TEAM_ALIENS )
```

src/sgame/sg_bot_util.cpp
```diff
@@ -1198,15 +1198,20 @@ void BotGetIdealAimLocation( gentity_t *self, botTarget_t target, vec3_t aimLoca
 {
 	//get the position of the target
 	BotGetTargetPos( target, aimLocation );
+	bool target_buildable = BotGetTargetType( target ) == entityType_t::ET_BUILDABLE;
+	//this retrieves the target's species, to aim at weak point:
```
necessarily-equal, line 1202:
```suggestion
	// this uses the target's information and species, to aim at a judicious spot:
```

src/sgame/sg_bot_util.cpp
```diff
@@ -1198,15 +1198,20 @@ void BotGetIdealAimLocation( gentity_t *self, botTarget_t target, vec3_t aimLoca
...
+	// * for humans, it's the head (but code only applies an offset here, with the hope it's the head)
+	// * for aliens, there is no weak point, and human bots try to take missile's speed into consideration (for luci)
```
necessarily-equal, line 1204:
```suggestion
	// * for aliens, there is no weak point, and human bots try to take missile's speed into consideration (for luci, pulse…)
```

necessarily-equal (changes requested):

src/sgame/sg_bot_util.cpp
bmorel: Well, this patch set only aim at changing nothing but make things more readable (and I learn a bit about desired code style).
Current code does not handle pulse rifle, blaster, flamer, barbs, pounces, ... at all, so changing the comment this way would result in what I hate the most: comments which lie.
When this code will be fixed so that it handle other weapons, then the comment shall be updated to reflect that, yes.

src/sgame/sg_bot_util.cpp
bmorel: right... considering the fact it's just "apply hard-coded offset" instead of "aim for the point with the less armor which can be reached"

src/sgame/sg_bot_util.cpp
bmorel: I don't know the code enough yet, so I tried to change nothing in this set (but yes, it's very tempting to integrate some changes in there)

src/sgame/sg_bot_util.cpp
bmorel: dunno if the '!' is preferred over "==false". I personally prefer the '!' variant, but have to admit that it's a bit less readable. @slipher : what is unvanquished team's opinion on that?

src/shared/bg_public.h
```diff
@@ -320,11 +320,11 @@ void Pmove( pmove_t *pmove );
 enum statIndex_t
 {
   STAT_HEALTH,
-  STAT_ITEMS,
+	STAT_ITEMS,      // bitfield of all possible upgrades, probably only for humans.
```
slipher, line 323: fix indentation

src/sgame/sg_bot_util.cpp
```diff
@@ -1198,15 +1198,21 @@ void BotGetIdealAimLocation( gentity_t *self, botTarget_t target, vec3_t aimLoca
...
+	team_t target_team = BotGetTargetTeam( target );
 
-	if ( BotGetTargetType( target ) != entityType_t::ET_BUILDABLE && BotTargetIsEntity( target ) && BotGetTargetTeam( target ) == TEAM_HUMANS )
+	if ( !target_buildable && BotTargetIsEntity( target ) && target_team == TEAM_HUMANS )
 	{
 
 		//aim at head
+		//FIXME: do not rely on hard-coded offset but evaluate which point have lower arm
```
slipher, line 1211: I don't understand what "which point have lower arm" means

bmorel:
Le Tue, 09 Feb 2021 16:47:05 -0800,
slipher <notifications@github.com> a écrit :

> I don't understand what "which point have lower arm" means

Me neither... because "armor" was truncated (or maybe I didn't write
the word entirely, who knows?).
So, the point is, aim at the head because head have lower armor than
other parts.

PR #1308, opened 2021-02-02
Mac continous build on Azure Pipelines

slipher:
See https://github.com/DaemonEngine/Daemon/pull/421 for more information.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,6 @@ cache: ccache
 addons:
   apt:
     update: true
-  homebrew:
-    packages:
-    - ninja
-    - ccache
 
 matrix:
   fast_finish: true
@@ -26,16 +22,6 @@ matrix:
       os: linux
       compiler: clang
       env: MODE=2
-  allow_failures:
-    - name: "macOS game-native-dll (clang)"
-      os: osx
-      compiler: clang
-      env: MODE=1
-    - name: "macOS nacl-nexe (pnacl-clang)"
-      os: osx
-      compiler: clang
-      env: MODE=2
-    # osx links gcc to clang; so skip
 
 install:
   - if [ "$TRAVIS_OS_NAME" == "linux" ]; then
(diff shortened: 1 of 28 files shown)


slipher:
I need to fix the branch name for non-PR builds.

slipher:
Got the branch stuff to work finally. @illwieckz does azure_pipelines.yml look good in terms of `git-checkout-modules` usage?

illwieckz:
> illwieckz does azure_pipelines.yml look good in terms of `git-checkout-modules` usage?

That looks OK to me.

illwieckz:
Hmm, one thing that triggers a doubt: `refs/heads/users`.

I know `pull` heads like `pull/1308/head` for pull request, I don't know the `users` one.

If it's about getting PR branch from forks, currently travis and appveyor configuration does not look at them.

slipher:
Yeah, the case where `refs/heads/users/` is stripped probably won't often be useful. I only got started down that route because the documentation gave an example containing it, and failed to mention that the prefix is absent in cases where the source branch is in the same repository.

Maybe it would actually be more logical to use only the target branch in all cases, ignoring the source branch? Like for the case of merging 0.52.0/sync into master, it would be better to check that master build as-is rather than trying to sync any submodules based on the name.

illwieckz:
If I'm right, what we did for the others is that if a PR wants to merge branch A in branch B, we first check for `/sync` in B (and checkout submodules with this branch name if there is) then check for `/sync` in A (and checkout submodules with this branch name if there is).

illwieckz:
> Maybe it would actually be more logical to use only the target branch in all cases, ignoring the source branch?

If I'm right the most important thing is the source branch, for example you make a PR in Unvanquished targeting master with `feature/sync` source branch, you want daemon submodule to be checked out with `feature/sync` too.

The fact we also check for target branch is for having work-in-progress version branches like `0.52.0/sync`.

We can have feature branch targeting version branch, that's why we do both.

PR #1309, opened 2021-02-02
Some warning fixes

slipher:


necessarily-equal (approved):
From a "I don't know the code well" point of view, it looks good to me.

PR #1310, opened 2021-02-07
Add some actions and conditions to bots

Gireen:


src/sgame/sg_bot_ai.cpp
```diff
@@ -621,10 +621,25 @@ AINodeStatus_t BotActionClassDodge( gentity_t *self, AIGenericNode_t* )
...
+		botEntityAndDistance_t e = AIEntityToGentity( self, et );
+		if ( !BotChangeGoalEntity( self, e.ent ) )
+		{
+			return STATUS_FAILURE;
+		}
+	}
+	else if( a->nparams == 3 )
+	{
+		vec3_t pos = { AIUnBoxFloat(a->params[0]), AIUnBoxFloat(a->params[1]), AIUnBoxFloat(a->params[2]) };
+		if ( !BotChangeGoalPos( self, pos ) )
```
slipher, line 645: What is the use case of this?

src/sgame/sg_bot_ai.cpp
```diff
@@ -540,6 +540,14 @@ AINodeStatus_t BotActionFireWeapon( gentity_t *self, AIGenericNode_t* )
 	return STATUS_SUCCESS;
 }
 
+AINodeStatus_t BotActionTeleport( gentity_t *self, AIGenericNode_t *node )
```
slipher, line 543: Is this for debugging or what?

src/sgame/sg_bot.h
```diff
@@ -92,6 +92,8 @@ struct botMemory_t
 	int lastThink;
 	int stuckTime;
 	vec3_t stuckPosition;
+
+	int aliveTime;
```
slipher, line 96: `spawnTime` would be more descriptive

src/sgame/sg_admin.cpp
```diff
@@ -5733,6 +5733,22 @@ bool G_admin_bot( gentity_t *ent )
 			G_BotDel( clientNum ); //delete the bot
 		}
 	}
+	else if ( !Q_stricmp( arg1, "behavior" ) && args.Argc() == 4 )
```
slipher, line 5736: Add the new subcommand to `BotUsage`

src/sgame/sg_admin.cpp
```diff
@@ -5733,6 +5733,22 @@ bool G_admin_bot( gentity_t *ent )
...
+
+		char err[MAX_STRING_CHARS];
+		const char *name = args[2].data();
+		int clientNum = G_ClientNumberFromString( name, err, sizeof( err ) );
+		if ( clientNum == -1 ) //something went wrong when finding the client Number
+		{
+			ADMP( va( "%s %s %s", QQ( "^3$1$:^* $2t$" ), "bot", Quote( err ) ) );
+			return false;
+		}
+		const char *bahavior = args[3].data();
```
slipher, line 5748: spelling

src/sgame/sg_bot.cpp
```diff
@@ -205,7 +205,28 @@ bool G_BotSetDefaults( int clientNum, team_t team, int skill, const char* behavi
 			return false;
 		}
 	}
+	return true;
```
slipher, line 208: If you are changing the behavior of an existing bot, don't you need to do the stuff with clearing the current/running nodes?

Gireen:
Teleporting or vectors as goals can be used for testing but also any other "scripted" actions like on mission maps or in a tutorial.

src/sgame/sg_bot.cpp
Gireen: I don't know. It seems to work without, but you are probably right. 😄

illwieckz:
> Teleporting or vectors as goals can be used for […] tutorial.

That's good!

illwieckz:
Oh, I forgot about that PR, what's missing for merging?

src/sgame/sg_bot.h
Gireen: im not sure but i think alivetime matches better.
the time a bot is alive. like stucktime is the time that a bot is well stuck.
spawntime could maybe also be misunderstood as time it takes to spawn or similar.

illwieckz: I agree with `aliveTime` being more explicit.

slipher: aliveTime is good for the function which can be used in the behavior tree - amount of time the bot has been alive. I think spawnTime is better just for this variable, which contains the time the bot spawned.

Gireen: oh yes. that's absolute correct. 😲

slipher:
Everything looks good!

illwieckz:
@Gireen, I believe the behavior tree does not make use of the feature to delay bot evolve yet?

PR #1311, opened 2021-02-08
do not do picmip on more UI elements

illwieckz:
Similar to c30a4acd3e7c3f7c00f967306d1a3b6db01cf449, those were missing.


slipher:
LGTM

PR #1313, opened 2021-02-09
Small fixes and improvements

necessarily-equal:


illwieckz:
I guess “the area displayed around eggs, overmind and reactor” was about power range but that does not exist anymore as you noticed.

src/cgame/cg_buildable.cpp
```diff
@@ -2229,11 +2229,12 @@ void CG_Buildable( centity_t *cent )
 
 	if ( !( es->eFlags & EF_B_SPAWNED ) )
 	{
-		sfxHandle_t prebuildSound = cgs.media.humanBuildablePrebuild;
+		sfxHandle_t prebuildSound;
 
 		switch ( team )
 		{
 		case TEAM_HUMANS:
+		default:
```
illwieckz, line 2222: maybe:

```
default: // make the compiler happy by making it knows prebuildSound is always set
```

or something like that (if that was the issue)?

necessarily-equal: I've dropped the commit for now, I guess it will need proper multiple team handling anyway.

slipher:
I would rather drop the indentation-only commits. We have thousands and thousands of places with ugly indentation so it is a waste of time to be manually fixing them piecemeal, and meanwhile it breaks blame.

Other commits look good.

Issue #1314, opened 2021-02-09
[0.52] “growing alien buildable” animation when a buildable is built is not displayed


illwieckz:
I wonder what introduced this regression.

illwieckz:
[![broken growing alien buildable animation](https://dl.illwieckz.net/b/unvanquished/bugs/broken-growing-alien-buildable-animation/unvanquished_2021-02-09_083344_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/broken-growing-alien-buildable-animation/unvanquished_2021-02-09_083344_000.jpg)

[![broken growing alien buildable animation](https://dl.illwieckz.net/b/unvanquished/bugs/broken-growing-alien-buildable-animation/unvanquished_2021-02-09_083347_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/broken-growing-alien-buildable-animation/unvanquished_2021-02-09_083347_000.jpg)

[![broken growing alien buildable animation](https://dl.illwieckz.net/b/unvanquished/bugs/broken-growing-alien-buildable-animation/unvanquished_2021-02-09_083355_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/broken-growing-alien-buildable-animation/unvanquished_2021-02-09_083355_000.jpg)

necessarily-equal:
Having a look

necessarily-equal:
it was fairly subtle, https://github.com/Unvanquished/Unvanquished/pull/1317

illwieckz:
Fixed by #1317.

PR #1317, opened 2021-02-11
Fix buildable anims

necessarily-equal:


illwieckz:
I confirm it fixed #1314 on my end. =)

PR #1320, opened 2021-02-11
Only display the devolve button as activatable near the overmind

necessarily-equal:
See the commits.

For the "too far from overmind" logo, I don't have a good one. Suggestions welcome.

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -2557,3 +2557,32 @@ void CG_Buildable( centity_t *cent )
 		CG_EndShadowCaster( );
 	}
 }
+
+// maybe move this to shared/ and add a prototype?
+static bool IsMainBuildable(buildable_t buildable)
+{
+	return buildable == BA_A_OVERMIND || buildable == BA_H_REACTOR;
+}
+
+const centity_t *CG_LookupMainBuildable()
+{
+	for( int beaconNum = 0; beaconNum < cg.beaconCount; beaconNum++ ) {
+		const auto b = cg.beacons[ beaconNum ];
+		if ( b->type == BCT_TAG && !(b->flags & (EF_BC_TAG_PLAYER|EF_BC_ENEMY))
+				&& IsMainBuildable( static_cast<buildable_t>(b->data) ) )
+		{
+			return &cg_entities[ b->target ];
+		}
+	}
+
+	return nullptr;
+}
+
+// keep in sync with G_DistanceToBase
+float CG_DistanceToBase()
+{
+	const centity_t *ent = CG_LookupMainBuildable();
+	if (!ent)
+		return 1e+37f; // in accordance to sgame
+	return Distance(cg.predictedPlayerEntity.lerpOrigin, ent->lerpOrigin);
+}
--- a/src/cgame/cg_animdelta.cpp
+++ b/src/cgame/cg_animdelta.cpp
@@ -51,7 +51,7 @@ bool AnimDelta::ParseConfiguration(clientInfo_t* ci, const char* token2, const c
 	char* token = COM_Parse2( data_p );
 	if ( !token || token[0] != '{' )
 	{
-		Log::Notice( "^1ERROR^*: Expected '{' but found '%s' in %s's character.cfg", token, ci->modelName );
+		Log::Warn( "Expected '{' but found '%s' in %s's AnimDelta", token, ci->modelName );
 		return true;
 	}
 	while ( 1 )
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -1617,6 +1617,8 @@ struct cgs_t
 	int      maxclients;
 	char     mapname[ MAX_QPATH ];
 
+	float    devolveMaxBaseDistance; // used for evolve/devolve ui
+
 	float    momentumHalfLife; // used for momentum bar (un)lock markers
 	float    unlockableMinTime;  // used for momentum bar (un)lock markers
 
@@ -1999,6 +2001,8 @@ void     CG_InitBuildables();
 void     CG_HumanBuildableDying( buildable_t buildable, vec3_t origin );
 void     CG_HumanBuildableExplosion( buildable_t buildable, vec3_t origin, vec3_t dir );
 void     CG_AlienBuildableExplosion( vec3_t origin, vec3_t dir );
+const centity_t *CG_LookupMainBuildable();
+float    CG_DistanceToBase();
 
 //
 // cg_animation.c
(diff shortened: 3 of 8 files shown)


src/cgame/cg_buildable.cpp
```diff
@@ -1338,9 +1338,12 @@ CG_BuildableParticleEffects
 static void CG_BuildableParticleEffects( centity_t *cent )
 {
 	entityState_t *es = &cent->currentState;
-	team_t        team = BG_Buildable( es->modelindex )->team;
-	int           health = es->generic1;
-	float         healthFrac = ( float ) health / BG_Buildable( es->modelindex )->health;
+
+	ASSERT(es->eType == entityType_t::ET_BUILDABLE);
```
slipher, line 1342: Use `ASSERT_EQ`

slipher, line 1342: Use `ASSERT_EQ`

src/cgame/cg_ents.cpp
```diff
@@ -1381,3 +1381,59 @@ void CG_AddPacketEntities()
 		}
 	}
 }
+
+// see also G_Team in src/sgame/sg_ents.cpp
+team_t CG_Team(const centity_t *ent)
```
slipher, line 1386: Why don't you use this function in any of these places where the team needs to be determined?

src/cgame/cg_ents.cpp
```diff
@@ -1381,3 +1381,59 @@ void CG_AddPacketEntities()
...
+			return (team_t) ci.team;
+		}
+		case entityType_t::ET_BUILDABLE:
+			return (team_t) ent->currentState.modelindex2;
+		default:
+			return TEAM_NONE;
+	}
+}
+
+/*
```
slipher, line 1404: delete

src/cgame/cg_ents.cpp
```diff
@@ -1381,3 +1381,59 @@ void CG_AddPacketEntities()
...
+		case entityType_t::ET_PLAYER:
+		case entityType_t::ET_CORPSE:
+		{
+			clientInfo_t &ci = cgs.clientinfo[es.clientNum];
+			return (team_t) ci.team;
+		}
+		case entityType_t::ET_BUILDABLE:
+			return (team_t) ent->currentState.modelindex2;
+		default:
+			return TEAM_NONE;
```
slipher, line 1400: Maybe there should be an assert or something, if it was used on an entity of the wrong type.

src/cgame/cg_ents.cpp
necessarily-equal: I've moved the CG_Team work into another PR as it can wait for the 0.53 "AnyTeam" milestone. If anyone wants to tag it, it's https://github.com/Unvanquished/Unvanquished/pull/1337 (leet!)

slipher:
Hmm, it might be possible for this to forbid you from evolving even when it is possible. The cgame is not guaranteed to have locations for all entities, only ones in certain visibility regions. So like if the OM is close to you but on the other side of a wall, this might tell you you can't evolve.

illwieckz:
We can then reword that the overmind must be in sight.

slipher:
> 
> 
> We can then reword that the overmind must be in sight.

You mean rewrite the sgame code?

illwieckz:
Yeah, ultimately both. If the code does not fit what we say, we may fix what we say, if what we say still does not fit the code, we may fix the code… until we get something that fist perfectly. 😁

If we can't ensure being close to the overmind works everytime or in a way we can't really describe to the user, we may find a similar mechanism we can ensure it works everytime and can be easily described.

Like, if the condition is to be in sight of the overmind and close from a given distance, I don't see why the specific limitations and other optimizations of the engine can prevent to have the words in the gameplay guide to fit the behavior the player experiences.

Gireen:
The beacon for base/overmind could maybe used for this. The client should know its position and distance.
And if there is no beacon the overmind is down. Depending on when the beacon is created, on placement or finished growth, it could also used to know if evolving is possible.

necessarily-equal:
@Gireen here's how it looks like when looking it up from the beacon:
```C++
        for( int beaconNum = 0; beaconNum < cg.beaconCount; beaconNum++ ) {
                auto b = cg.beacons[ beaconNum ];
                if ( b->type == BCT_TAG && !(b->flags & (EF_BC_TAG_PLAYER|EF_BC_ENEMY))
                                && IsMainBuildable( static_cast<buildable_t>(b->data) ) )
                {
                        return &cg_entities[ b->target ];
                }
        }
```
Do you prefer it that way? I personally don't really see the point over iterating on cg_entities directly, since you need this array anyway.

Gireen:
For me both are fine. But iterating only the beacons feels more right. 😄

illwieckz:
@afontain, can you rebase on latest `0.52.0/sync` branch?

necessarily-equal:
Rebased

necessarily-equal:
uh, rebased, but broken

slipher:
What was the solution to the problem that the cgame is not guaranteed to have knowledge of all entities? I don't like the idea that the menu can forbid you from doing an action, because the cgame tried incorrectly to imitate the logic of the sgame.

The menu code is not working because it changes the now-dead code for the old version of circlemenus.

Gireen:
The beacon for mainbase is the overmind position. It should be the same as sgame.

For the lua circle menu there are only small changes required
This
https://github.com/Unvanquished/Unvanquished/pull/1320/files#diff-0dd2ea7342970b29ba0e63dd7492fd87a3e236870dfbc0ba86c322e771835b4aR432-R439

has just to be split up these places
https://github.com/Unvanquished/Unvanquished/pull/1348/files#diff-8281311e17d0f35249779e2ccc208333cc2024762edc8c5716c2bf4588fbb7d4R1734-R1749 

https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/40/files#diff-d94057619f7ef02c147e9a4ba26d0479b88c74ea7cdefcdafc042102ba9af496R47-R58

necessarily-equal:
Thanks @slipher and @Gireen. Rebased.

illwieckz:
About that:

> What was the solution to the problem that the cgame is not guaranteed to have knowledge of all entities

I assume that if under some obscure circumstance the cgame can't know where is the overmind, you can't be close enough to validate the condition for being close enough to the overmind to evolve anyway…

illwieckz:
Hmm, @necessarily-equal it looks like it is requiring a rebase again… :-/

necessarily-equal:
rebased

illwieckz:
That seems to work perfectly! When using the UI there is a lock icon preventing to select the lower classes, and if using the console, a message is still displayed on screen telling why the action was not done, very good! 👍

PR #1321, opened 2021-02-12
Commented  thunder

Pretuer:
I commented the region about thunder, because it currently doesn't work. I checked if it still parses it and it does!

### DIFF
--- a/dist/configs/game/maprotation.cfg
+++ b/dist/configs/game/maprotation.cfg
@@ -8,9 +8,9 @@ rotation1
 	plat23
 	spacetracks
 	station15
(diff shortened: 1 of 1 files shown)


illwieckz:
Hi, the code in the master branch knows how to read those lines, the sample rotation map in the master branch is on-par wth the code.

The thing is that the released code is older than the master branch. So when using released build, just remove the thunder lines on your end but nothing has to be fixed in repository.

I don't think the way to comment those lines is to use `/* */`, when doing that those lines are probably just ignored as bad map names or something like that.

PR #1330, opened 2021-02-21
Team genericity

bmorel:
Preliminary work to make it easier to untie notions of team and species/sides.
Things are mostly:

* put hard coded values into arrays
* move code into `for( int i = TEAM_NONE; i < NUM_TEAMS; ++i )...` loops
* use some (sometimes already existing) functions instead of writing specific code (deduplication)

But there are some bigger things too, which needs real peer review. This is mostly WiP, I even have some local commits I need to clean up.

### DIFF
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -617,19 +617,7 @@ void CG_OnPlayerWeaponChange()
 	// Change the HUD to match the weapon. Close the old hud first
 	Rocket_ShowHud( ps->weapon );
 
-	// Rebuild weapon lists if UI is in focus.
-	if ( trap_Key_GetCatcher() == KEYCATCH_UI && ps->persistant[ PERS_TEAM ] == TEAM_HUMANS )
-	{
-		CG_Rocket_BuildArmourySellList( "default" );
-		CG_Rocket_BuildArmouryBuyList( "default" );
-
-		Rocket::Core::ElementDocument* document = menuContext->GetDocument( rocketInfo.menu[ ROCKETMENU_ARMOURYBUY ].id );
-		if ( document->IsVisible() )
-		{
-			document->DispatchEvent( "refreshdata", {} );
-		}
-	}
-
+	CG_Rocket_UpdateArmouryMenu();
 	cg.weaponOffsetsFilter.Reset( );
 
 	cg.predictedPlayerEntity.pe.weapon.animationNumber = -1; //force weapon lerpframe recalculation
@@ -637,13 +625,13 @@ void CG_OnPlayerWeaponChange()
 
 /*
 =========================
-CG_OnPlayerUpgradeChange
+CG_Rocket_UpdateArmouryMenu
 
 Called on upgrade change
 =========================
 */
 
-void CG_OnPlayerUpgradeChange()
+void CG_Rocket_UpdateArmouryMenu()
 {
 	playerState_t *ps = &cg.snap->ps;
 
--- a/src/sgame/sg_bot_parse.cpp
+++ b/src/sgame/sg_bot_parse.cpp
@@ -265,14 +265,19 @@ static AIValue_t botCanEvolveTo( gentity_t *self, const AIValue_t *params )
 	return AIBoxInt( BotCanEvolveToClass( self, c ) );
 }
 
-static AIValue_t humanMomentum( gentity_t*, const AIValue_t* )
+// Returns a team's momentum for use in behavior trees.
+// If the team number is invalid, debug builds will have assertion failure
+// and release builds will return the value 0, so that it's clear invalid
+// team won't have bought anything
+static AIValue_t momentum( gentity_t* self, const AIValue_t *params )
 {
-	return AIBoxInt( level.team[ TEAM_HUMANS ].momentum );
-}
-
-static AIValue_t alienMomentum( gentity_t*, const AIValue_t* )
-{
-	return AIBoxInt( level.team[ TEAM_ALIENS ].momentum );
+	int requestedTeam = AIUnBoxInt( params[ 0 ] ); //is really a team_t
+	ASSERT( G_IsPlayableTeam( requestedTeam ) );
+	if( !G_IsPlayableTeam( requestedTeam ) )
+	{
+		return AIBoxInt( 0 );
+	}
+	return AIBoxInt( level.team[ requestedTeam ].momentum );
 }
 
 static AIValue_t aliveTime( gentity_t*self, const AIValue_t* )
@@ -339,7 +344,7 @@ static const struct AIConditionMap_s
 } conditionFuncs[] =
 {
 	{ "alertedToEnemy",    VALUE_INT,   alertedToEnemy,    0 },
-	{ "alienMomentum",     VALUE_INT,   alienMomentum,     0 },
+	{ "momentum",          VALUE_INT,   momentum,          1 },
 	{ "aliveTime",         VALUE_INT,   aliveTime,         0 },
 	{ "baseRushScore",     VALUE_FLOAT, baseRushScore,     0 },
 	{ "buildingIsDamaged", VALUE_INT,   buildingIsDamaged, 0 },
@@ -356,7 +361,6 @@ static const struct AIConditionMap_s
 	{ "haveUpgrade",       VALUE_INT,   haveUpgrade,       1 },
 	{ "haveWeapon",        VALUE_INT,   haveWeapon,        1 },
 	{ "healScore",         VALUE_FLOAT, healScore,         0 },
-	{ "humanMomentum",     VALUE_INT,   humanMomentum,     0 },
 	{ "inAttackRange",     VALUE_INT,   inAttackRange,     1 },
 	{ "isVisible",         VALUE_INT,   isVisible,         1 },
 	{ "matchTime",         VALUE_INT,   matchTime,         0 },
--- a/src/sgame/sg_team.cpp
+++ b/src/sgame/sg_team.cpp
@@ -210,7 +210,7 @@ void G_LeaveTeam( gentity_t *self )
 	gentity_t *ent;
 	int       i;
 
-	if ( TEAM_ALIENS == team || TEAM_HUMANS == team )
+	if ( G_IsPlayableTeam( team ) )
 	{
 		G_RemoveFromSpawnQueue( &level.team[ team ].spawnQueue, self->client->ps.clientNum );
 	}
@@ -526,8 +526,7 @@ void CheckTeamStatus()
 				continue;
 			}
 
-			if ( ent->inuse && ( ent->client->pers.team == TEAM_HUMANS ||
-			                     ent->client->pers.team == TEAM_ALIENS ) )
+			if ( ent->inuse && G_IsPlayableTeam( ent->client->pers.team ) )
 			{
 				loc = GetCloseLocationEntity( ent );
 
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -2041,7 +2041,7 @@ void CG_CheckEvents( centity_t *cent );
 void CG_EntityEvent( centity_t *cent, vec3_t position );
 void CG_PainEvent( centity_t *cent, int health );
 void CG_OnPlayerWeaponChange();
-void CG_OnPlayerUpgradeChange();
+void CG_Rocket_UpdateArmouryMenu();
 void CG_OnMapRestart();
 
 //
--- a/src/cgame/cg_snapshot.cpp
+++ b/src/cgame/cg_snapshot.cpp
@@ -210,7 +210,7 @@ static void CG_TransitionSnapshot()
 	}
 
 	if (ops->stats[STAT_ITEMS] != ps->stats[STAT_ITEMS]) {
-		CG_OnPlayerUpgradeChange();
+		CG_Rocket_UpdateArmouryMenu();
 	}
 }
 
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -552,7 +552,7 @@ void SpectatorThink( gentity_t *ent, usercmd_t *ucmd )
 
 	// Check to see if we are in the spawn queue
 	// Also, do some other checks and updates which players need while spectating
-	if ( team == TEAM_ALIENS || team == TEAM_HUMANS )
+	if ( G_IsPlayableTeam( team ) )
 	{
 		client->ps.persistant[ PERS_UNLOCKABLES ] = BG_UnlockablesMask( client->pers.team );
 		queued = G_SearchSpawnQueue( &level.team[ team ].spawnQueue, ent - g_entities );
@@ -576,7 +576,7 @@ void SpectatorThink( gentity_t *ent, usercmd_t *ucmd )
 		}
 
 		//be sure that only valid team "numbers" can be used.
-		ASSERT(team == TEAM_ALIENS || team == TEAM_HUMANS);
+		ASSERT( G_IsPlayableTeam( team ) );
 		G_RemoveFromSpawnQueue( &level.team[ team ].spawnQueue, client->ps.clientNum );
 
 		client->pers.classSelection = PCL_NONE;
@@ -663,7 +663,7 @@ void SpectatorThink( gentity_t *ent, usercmd_t *ucmd )
 		if ( client->ps.pm_flags & PMF_QUEUED )
 		{
 			/* team must exist, or there will be a sigsegv */
-			ASSERT(team == TEAM_HUMANS || team == TEAM_ALIENS);
+			ASSERT( G_IsPlayableTeam( team ) );
 			client->ps.persistant[ PERS_SPAWNQUEUE ] = level.team[ team ].numSpawns;
 			client->ps.persistant[ PERS_SPAWNQUEUE ] |= G_GetPosInSpawnQueue( &level.team[ team ].spawnQueue,
 			                                                                  client->ps.clientNum ) << 8;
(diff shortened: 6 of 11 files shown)


slipher:
When developing the anyteams mod, I saw 3 common idioms for looping over the playable teams:
- `for ( int i = TEAM_NONE + 1; i < NUM_TEAMS; i++ )`
- `for (team_t team = TEAM_NONE; (team = G_IterateTeams(team)); )`
- Just writing the same statement twice, once with `TEAM_ALIENS` and once with `TEAM_HUMANS`

In pratically all cases, the loop needed to run for each group of players fighting for a common objective (`TeamIndex` in the mod), *not* for each species (`race_t` in the mod).

I suggest we replace all of these with a range-for iterator that iterates over the playable teams. Something like

```
inline std::array<team_t, 2> BG_PlayableTeams() {  // i.e. non-spectator
    return { 1, 2 };
}
```
would work. It can be used like `for (team_t team : BG_PlayableTeams()) ...`. Similarly there could be `BG_AllTeams()` or something which would include spectator as a team.

In the [anyteams tree](https://github.com/slipher/Unvanquished/tree/anyteam), I suggest grepping for (besides `G_IterateTeams`) `TI_1` and `TI_2`, which are constants of type `TeamIndex`. I think a large fraction of occurrences thereof can be replaced with the range-for idiom.

bmorel:
> I suggest we replace all of these with a range-for iterator that iterates over the playable teams. Something like
> 
> ```
> inline std::array<team_t, 2> BG_PlayableTeams() {  // i.e. non-spectator
>     return { 1, 2 };
> }
> ```
> 
> would work. It can be used like `for (team_t team : BG_PlayableTeams()) ...`. Similarly there could be `BG_AllTeams()` or something which would include spectator as a team.

Good idea. I think both BG_PlayableTeams() and BG_AllTeams() are needed, or at least can be.

> In the [anyteams tree](https://github.com/slipher/Unvanquished/tree/anyteam), I suggest grepping for (besides `G_IterateTeams`) `TI_1` and `TI_2`, which are constants of type `TeamIndex`. I think a large fraction of occurrences thereof can be replaced with the range-for idiom.

Same in that bunch of patches tbh. I'll try to find time and motivation in the week to apply my patchs to your anyteam branch, see where there are conflicts, and solve them when it does not affect gameplay (the goal of this patchset is not to change gameplay, just to make it easier to do so).

src/sgame/components/OvermindComponent.cpp
```diff
@@ -72,35 +72,40 @@ Entity* OvermindComponent::FindTarget() {
...
+	team_t selfTeam = GetTeamComponent().Team();
 	team_t aTeam = G_Team(a.oldEnt);
 	team_t bTeam = G_Team(b.oldEnt);
 
-	// Prefer humans.
-	if (aTeam == TEAM_HUMANS && bTeam != TEAM_HUMANS) return true;
-	if (aTeam != TEAM_HUMANS && bTeam == TEAM_HUMANS) return false;
+	//don't attack friends or non-players
+	if ( G_OnSameTeam( selfTeam, aTeam ) ) return false;
+	if ( G_OnSameTeam( selfTeam, bTeam ) || !G_IsPlayableTeam( bTeam ) ) return true;
```
necessarily-equal, line 93:
```suggestion
	if ( !G_IsPlayableTeam(aTeam) ) return false; // ignore spectators

	if ( !G_OnSameTeam( aTeam, bTeam) )
	{
		// prefer to look at your enemies
		if ( G_OnSameTeam( selfTeam, aTeam ) ) return false;
		if ( G_OnSameTeam( selfTeam, bTeam ) ) return true;
	} // it's a tie: both are in your team, or in another team
```

src/sgame/components/OvermindComponent.cpp
```diff
@@ -72,35 +72,40 @@ Entity* OvermindComponent::FindTarget() {
...
-	}
+	// The overmind senses weakness! (but not armor, so is kind of a joke)
+	HealthComponent* aHealthComponent = a.Get<HealthComponent>();
+	HealthComponent* bHealthComponent = b.Get<HealthComponent>();
 
-	// Break tie by preferring smaller distance.
-	return aDistance < bDistance;
+	// same health: attack the closer one.
+	if (aHealthComponent == bHealthComponent) return aDistance < bDistance;
+	return aHealthComponent->HealthFraction() < bHealthComponent->HealthFraction();
```
necessarily-equal, line 108:
```suggestion
	// none have health, look at the closest one
	if (!aHealthComponent && !bHealthComponent) return aDistance < bDistance;
	// prefer looking at the one that can be hurt: you don't want to beat a dead horse
	if (!aHealthComponent) return false;
	if (!bHealthComponent) return true;
	// The overmind senses weakness!
	return aHealthComponent->HealthFraction() < bHealthComponent->HealthFraction();
```

src/sgame/components/OvermindComponent.cpp
```diff
@@ -72,35 +72,40 @@ Entity* OvermindComponent::FindTarget() {
...
+	float bDistance = G_Distance(entity.oldEnt, b.oldEnt);
+	if (bDistance > ATTACK_RANGE) return true;
 
-		// The overmind senses weakness!
-		if (aHealthComponent && bHealthComponent) {
-			if (aHealthComponent->HealthFraction() < bHealthComponent->HealthFraction()) return true;
-			if (aHealthComponent->HealthFraction() > bHealthComponent->HealthFraction()) return false;
-		}
-	}
+	// The overmind senses weakness! (but not armor, so is kind of a joke)
```
necessarily-equal, line 102:
```suggestion
```

necessarily-equal (changes requested):

src/sgame/sg_team.cpp
```diff
@@ -569,23 +564,44 @@ void CheckTeamStatus()
...
-		else
-		{
-			level.numTeamImbalanceWarnings = 0;
-		}
 	}
 }
+
+int G_NumberPlayableTeams()
+{
+	return TEAM_HUMANS - TEAM_ALIENS + 1;
```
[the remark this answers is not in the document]
bmorel: humans are 2, aliens are 1, current code thus returns 2 -1 + 1 = 2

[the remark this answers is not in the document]
necessarily-equal: and NUM_TEAMS is 3, thus 3-1 = 2. I'm not saying it's wrong right now, but if someone were to write this
```diff
 enum team_t
 {
   TEAM_ALL = -1,
   TEAM_NONE,
   TEAM_ALIENS,
   TEAM_HUMANS,
+  TEAM_ZOMBIES,
 
   NUM_TEAMS
 };
```
it would break :-/

bmorel:
I believe that subset could be merged, as it only contains really trivial changes.

src/shared/bg_public.h
```diff
@@ -151,6 +151,14 @@ enum team_t
   NUM_TEAMS
 };
```
slipher, line 153: For stuff that is only used in one file, I'd rather see it defined next to the point of use, as it is now.

What exactly is your goal with this commit? I mean, in what aspect should the code be better than it was before?

src/sgame/sg_bot_parse.cpp
```diff
@@ -339,7 +334,7 @@ static const struct AIConditionMap_s
 } conditionFuncs[] =
 {
 	{ "alertedToEnemy",    VALUE_INT,   alertedToEnemy,    0 },
-	{ "alienMomentum",     VALUE_INT,   alienMomentum,     0 },
+	{ "momentum",          VALUE_INT,   momentum,          0 },
```
slipher, line 337: What if you wanted to use the enemy team's momentum in the behavior tree? I guess that's the point of having both.

src/cgame/cg_event.cpp
```diff
@@ -617,19 +617,7 @@ void CG_OnPlayerWeaponChange()
...
-		CG_Rocket_BuildArmouryBuyList( "default" );
-
-		Rocket::Core::ElementDocument* document = menuContext->GetDocument( rocketInfo.menu[ ROCKETMENU_ARMOURYBUY ].id );
-		if ( document->IsVisible() )
-		{
-			document->DispatchEvent( "refreshdata", {} );
-		}
-	}
-
+	CG_OnPlayerUpgradeChange();
```
slipher, line 620: Rename this to "rebuild armoury menu" or something like that

src/sgame/sg_team.cpp
```diff
@@ -287,19 +287,11 @@ void G_ChangeTeam( gentity_t *ent, team_t newTeam )
...
-		  ( int )( ent->client->pers.credit *
-		           ALIEN_MAX_CREDITS / HUMAN_MAX_CREDITS + 0.5f );
-	}
-	else if ( oldTeam == TEAM_ALIENS && newTeam == TEAM_HUMANS )
-	{
-		// Convert from alien to human credits
-		ent->client->pers.credit =
-		  ( int )( ent->client->pers.credit *
-		           HUMAN_MAX_CREDITS / ALIEN_MAX_CREDITS + 0.5f );
+		  static_cast<int>( ceil( ent->client->pers.credit *
```
slipher, line 293: This calculation is equivalent to `ent->client->pers.credit*g_teamMaxCredits[new_team]/g_teamMaxCredits[oldTeam] + 1`, which I suspect is no what you wanted. (The old calculation didn't make much sense either - it is equivalent to `ent->client->pers.credit*ALIEN_MAX_CREDITS/HUMAN_MAX_CREDITS` and vice versa.)

src/sgame/sg_bot_parse.cpp
bmorel: Good point, but in that case, it'd be better to add an argument I think. Currently, none of those are used, though, and I fail to see how they could be used considering current capabilities of the BT mechanisms (can't do math operations notably).

src/shared/bg_public.h
bmorel: The point is to prepare a more uniform way to configure a whole team/species characteristics. I think it would be better to move stuff in an arrray of struct, really, and I'd do it without problems, but I wanted to know where and how it should be done to follow unv's style.

src/cgame/cg_event.cpp
bmorel: It's just reusing existing code I thing, I don't think I came with such a name myself? Or do you mean renaming "CG_OnPlayerWeaponChange" and not "CG_OnPlayerUpgradeChange" ? I thought you'd prefer minimal changes.

src/sgame/sg_team.cpp
slipher: My point was that the math does not make sense - it is doing useless extra operations.

> In my opinion, this whole idea of converting credits is bad, though.

HUMAN_MAX_CREDITS and ALIEN_MAX_CREDITS are the same, so you could certainly consider the conversion idea.

src/cgame/cg_event.cpp
slipher: Rename CG_OnPlayerUpgradeChange because it's going to be used for things other than upgrade changes.

src/sgame/sg_bot_parse.cpp
slipher: I imagine the use case for momentum would be to check whether some weapon/class/structure is unlocked. You wouldn't need any math for that - it's just comparing greater or less than a constant number. And a team-generic one wouldn't be particularly useful since the different teams have different unlocks.

src/sgame/sg_team.cpp
bmorel: you meant, consider the removal of conversion idea? Looks like I didn't noticed at that point that the constants were the same... dunno why, since I think I merged them into something too in some commit? maybe one not candidate for push? Don't remember, but this really looks like code which could be avoided.

necessarily-equal:
It's a bit weird, but appreciable, to be able to keep some credits you 
already have when changing team. That's something I do from time to 
time to help the loosing team (say humans got reinforcement), and you 
definitely don't want to be a dretch against battlesuit in such a case.

src/sgame/sg_bot_parse.cpp
bmorel: Right, but it would allow to move the non-generic part into behavior trees, which have to be specific anyway (since different gameplay for team is the goal). The C++ code could benefit from a `momentum( some_team )` function, so that there is less code there. Plus, I don't think currently a team can know opponent's momentum for players, so why bots should be allowed to do so?

src/shared/bg_public.h
slipher: IMO this centralized location only makes refactoring more difficult: it's impossible to tell what the fields are used for, so you have to jump back and forth between the global arrays and the point of use. It's more efficient when the stuff is defined at the point of use.

Let's say you're making a mod where aliens are replaced with kittens, then what you can do is rename `TEAM_ALIENS` to `TEAM_KITTENS`. Then you will get a compile error for each place `TEAM_ALIENS` was mentioned, which helps you to identify all the things that needed to be changed. The refactoring with `G_IsPlayableTeam` and team iterators will help to eliminate false positives for this method.

bmorel:
> It's a bit weird, but appreciable, to be able to keep some credits you already have when changing team. That's something I do from time to time to help the loosing team (say humans got reinforcement), and you definitely don't want to be a dretch against battlesuit in such a case.

Yes, but since the max credit values are the same for both teams, it'd be easier to simply not convert. My attempt was, as slipher pointed out, to have a more generic code, but he's right to point at the fact that since both teams have same max values, that's basically useless.

necessarily-equal:
One example reason is « because humans can infer that pretty easily, 
while bots have no such understanding. »

When you see a dragoon, you *remember* aliens have that. On the other 
hand, implementing that for bots would a lot of work for not much 
benefit.

Lastly, not all functions use are about cheating. The person designing 
the bot use should keep in mind what is fair and what's entertaining. 
To take an extreme example, we have a bot teleport action now (I think 
that was merged). Does that mean bots will now teleport?

bmorel:
> One example reason is « because humans can infer that pretty easily, while bots have no such understanding. » When you see a dragoon, you *remember* aliens have that. On the other hand, implementing that for bots would a lot of work for not much benefit. Lastly, not all functions use are about cheating. The person designing the bot use should keep in mind what is fair and what's entertaining. To take an extreme example, we have a bot teleport action now (I think that was merged). Does that mean bots will now teleport?

Please quote what you're answering to when using mail reply, since otherwise there's no context. I guess here you're answering to:

> Right, but it would allow to move the non-generic part into behavior trees, which have to be specific anyway (since different gameplay for team is the goal). The C++ code could benefit from a momentum( some_team ) function, so that there is less code there. Plus, I don't think currently a team can know opponent's momentum for players, so why bots should be allowed to do so?

Having bots remembering the last thing the team has seen in enemy teams is something I'm planning to implement in the botMemory_t structure, which implies there is no need to know exactly a team's momentum.
Plus, your explanation does not explains how having a single "momentum" function would be bad, if (and it's true current proposed patch does not) there's an argument to pass the team's name

src/sgame/sg_bot_parse.cpp
```diff
@@ -265,14 +265,19 @@ static AIValue_t botCanEvolveTo( gentity_t *self, const AIValue_t *params )
...
+static AIValue_t momentum( gentity_t* self, const AIValue_t *params )
 {
-	return AIBoxInt( level.team[ TEAM_HUMANS ].momentum );
-}
-
-static AIValue_t alienMomentum( gentity_t*, const AIValue_t* )
-{
-	return AIBoxInt( level.team[ TEAM_ALIENS ].momentum );
+	int requestedTeam = AIUnBoxInt( params[ 0 ] ); //is really a team_t
+	ASSERT( G_IsPlayableTeam( requestedTeam ) );
```
slipher, line 275: `ASSERT` should only be used for programming errors (in C++). This is an issue with user input, so it should have a log warning instead.

src/sgame/sg_team.cpp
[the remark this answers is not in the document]
bmorel: right, shuold use `NUM_TEAMS - 1` even if that would also break under some conditions, but without making team_t (or another struct for less problems) a structure that describe teams, and have a list of those somewhere, there's no real solution.

src/sgame/components/OvermindComponent.cpp
bmorel: at that point, I think that targets without HealthComponent should not trigger calls to this function, I'll do some testing to see if an ASSERT() will make the code crash or not.

src/sgame/components/OvermindComponent.cpp
bmorel: thinking anew about it, it also does not use remaining health, so if at some point it have enemies with different amounts of health, the bigger one have 150HP out of 350HP and the 2nd one have 30HP out of 30HP, the bigger one will be targeted, so I indeed think it's not working (generically, that is), even without the armor problem.
This commit is still labelled "bunch of changes" because it needs lot of rework anyway. (I'm going to cleanup that branch to see if some if this stuff can be merged soon)

src/sgame/sg_bot_parse.cpp
bmorel: I thought the ASSERT would not be compiled in release builds (like the standard one) which is why I still handled the case the line after for release builds. Guess I had not thought of searching for unv's warning facility.

slipher: It's true, ASSERT is not compiled in release builds. But again, assertions should only be used to catch C++ programming errors within the sgame module (see e.g. https://stackoverflow.com/a/253219). The behavior tree is a user input; sgame should not crash due to a mistake in the behavior tree, regardless of build type. Nor should it silently ignore the mistake with no log message in release mode.

slipher:
Merged as https://github.com/Unvanquished/Unvanquished/commit/fd8c6a5208de20741e2b863bc457ea45f05fe413.

PR #1332, opened 2021-02-22
Partial bot improvements

bmorel:
Commit 1: improve human bot aiming. Something should probably be done for aliens, too.
Commit 2: provide bots some primitives to later improve team-play

Both have been tested locally and are known to work (through a more advanced branch which introduces for-check changes) but commit 2 probably can use improvements.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1214,10 +1214,29 @@ void BotGetIdealAimLocation( gentity_t *self, botTarget_t target, vec3_t aimLoca
 	}
 	else if ( isTargetBuildable || targetTeam == TEAM_ALIENS )
 	{
-		//make lucifer cannons aim ahead based on the target's velocity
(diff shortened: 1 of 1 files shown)


slipher:
Commit 1 looks good to merge on its own.

I'd prefer to see Commit 2 integrated with some code which actually uses it. When I come across an unused function, I'm apt to delete it :)

bmorel:
> Commit 1 looks good to merge on its own.
> 
> I'd prefer to see Commit 2 integrated with some code which actually uses it. When I come across an unused function, I'm apt to delete it :)

Understandable. I'd like to expose them to behavior trees, but I don't know how. Any actual change using that in C++ side would be arguable, at best, because it depends on map, people's opinion and other things I don't think about right now.

bmorel:
I've removed 2nd commit for now, as you pointed out in IRC, some stuff is already there, so it needs more work from me to avoid code duplication, and to have code actually used anyway.

Issue #1304, opened 2021-01-29
bot dretch will never attack turrets


bmorel:
But they can theoretically damage them.

https://github.com/Unvanquished/Unvanquished/blob/0.52.0/sync/src/sgame/sg_bot_util.cpp#L668
https://github.com/Unvanquished/Unvanquished/blob/0.52.0/sync/src/sgame/sg_bot_util.cpp#L2093

should probably check if the amount of damages dealt is non-zero and if building is from enemy team instead of checking for class == dretch.

illwieckz:
Hmm, that's probably a relic from the time dretch could not attack anything.

necessarily-equal:
I've got something in the works

PR #1336, opened 2021-02-22
Dretch attack condition rework

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1304.

to be merged along with https://github.com/UnvanquishedAssets/res-buildables_src.dpkdir/pull/5.


slipher:
Merging this would be problematic since then we won't be able to easily test the 0.52 branch using the 0.51 packages (except `unvanquished`), since it will be broken without an updated `res-weapons`. You see, this is why I was saying yesterday that we should get rid of attribute files and put the attribute tables in the code...

slipher:
Maybe this could be moved forward by transferring all of the attr.cfg files into the unvanquished dpk repo - it seems like everyone agrees that would be an improvement.

illwieckz:
Hmm, I have missed that thread.

Basically the reason why res-weapons, res-models, res-buildables and unvanquished are split is… repository management.

Git repository size issues plus the fact those repositories were not split is the reason why most of those assets were not tracked in repositories *at all*, because at some point the problems related to big repositories encourages to not track assets at all (bad short-term solution). It required years of work to go back from this.

Outside of git better working with small modules (this is not only a limitation, this is vertuous, like avoiding random never merged forks living in subversion subdirectories), merging thoes repositories brings many fancy issue.

Just to name a few:

- Merging those repositories can be an immediate enforcement to move from GitHub to GitLab (without keeping asset repository mirrors). I'm not against a move to GitLab, but I'm not sure we can sustain this in an unplanned way.
- Xonotic uses merged repositories and face specific issues like public repositories not being clonable over https because of some timelimits, so no public clone of the repository, registering is required, etc.
- Also you want to just update one model and clone gigabytes… meh.

As freem noticed in IRC, maybe the main problem is the models/*.cfg, and maybe just the models/*.attr.cfg files. Maybe we can keep .model.cfg in resources repositories?

At the time I designed the repository layout, I split unvanquished into unvanquished and res-models then res-models in res-weapons/buildables/players. Comes the question about those models/*.cfg files. I had no idea about the best place for them, so for a start, I've stored them in repositories where the models and textures lived. But I know maybe that would not be the best.

Basically, the place to put those .cfg files is the place where it is more likely to be modified.

Better not merge the whole repositories if we can.

illwieckz:
Note: some of the issues related to split repositories can also be tampered if not fixed by tools (that's why [Urcheon](https://github.com/DaemonEngine/Urcheon), [git-checkout-modules](https://gitlab.com/illwieckz/git-checkout-modules)).

I also see improvements that are wanted for the tooling:

- Merging multiple dpks to produce a mod dpk (also needed to merge gamecode)
- Writing the DEPS versions of the dependencies (not only the self ancestor the package extends)

slipher:
@illwieckz No one suggested merging the entire asset repositories. Just the small attr.cfg text files.

illwieckz:
Well, in IRC there was talks of merging the entire assets from the quoted repositories. 😁

bmorel:
If everyone agrees on merging some of those files in unvanquished.dpkdir repo, it will need PRs related to those files to be merged before, and the repos in which they are initially stored to be frozen.

FYI: here is the list of file cfg type files and what I guess what they are for considering their content:

* `attr.cfg`: gameplay, physics
* `beacon.cfg`: user interface
* `iqe.cfg`: graphics
* `locdamage.cfg`: gameplay, collisions
* `missile.cfg`: gameplay, collisions, graphics (animations)
* `model.cfg`: graphics

src/sgame/sg_buildable.cpp
```diff
@@ -181,6 +181,25 @@ bool G_InsideBase(gentity_t *self) {
 	return trap_InPVSIgnorePortals(self->s.origin, mainBuilding->s.origin);
 }
 
+bool G_DretchCanDamageEntity( const gentity_t *self, const gentity_t *ent )
```
bmorel, line 184: why not start by an assert that checks that ent has correct type (pre-conditions) and then do a single "return player || ( building && attackable )"?

I think it would be easier to read (and use less lines)?

Also, why using a pointer for ent and self, instead of references?

src/sgame/sg_bot_util.cpp
```diff
@@ -2115,23 +2079,31 @@ bool BotEnemyIsValid( gentity_t *self, gentity_t *enemy )
...
 	}
 
+	// ignore neutrals
 	if ( BotGetEntityTeam( enemy ) == TEAM_NONE )
 	{
 		return false;
 	}
 
+	// ignore teamates
+	if ( BotGetEntityTeam( enemy ) == BotGetEntityTeam( self ) )
```
bmorel, line 2101: There is a G_OnSameTeam() function which could be used, here.
Also, is it really useful to split this check and previous one?

illwieckz:
> `iqe.cfg`

Those are `iqm` command files. They are used by `iqm` to produce `.iqm` files (models) from `.iqe` files, that command file tells if there is rotation/translation/scale to apply on conversion from `.iqe` to `.iqm`, so the `.iqe.cfg` file must always live in the same repository of `.iqe` models.

illwieckz:
> `presets/*.cfg`, `default.cfg`

Those are presets of game configuration (like, graphical options, default binds…), best place is `unvanquished_src.dpkdir` without doubt.

src/sgame/sg_bot_util.cpp
necessarily-equal: I think the `G_OnSameTeam(team_t, team_t)` isn't merged yet?

bmorel:
Le Mon, 12 Apr 2021 11:03:19 -0700,
necessarily-equal ***@***.***> a écrit :

> @necessarily-equal commented on this pull request.
> 
> 
> 
> >  	if ( BotGetEntityTeam( enemy ) == TEAM_NONE )  
>  	{
>  		return false;
>  	}
>  
> +	// ignore teamates
> +	if ( BotGetEntityTeam( enemy ) == BotGetEntityTeam( self ) )
> 
> I think the `G_OnSameTeam(team_t, team_t)` isn't merged yet?
> 

There is one here:
https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_team.cpp#L137

src/shared/bg_parse.cpp
```diff
@@ -607,6 +607,7 @@ void BG_ParseBuildableAttributeFile( const char *filename, buildableAttributes_t
 
 	text = text_buffer;
 
+	ba->dretchAttackable = false; //FIXME
```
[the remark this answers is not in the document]
illwieckz: > Fix what?

I agree, please tell what is to be fixed, why it's perfectible.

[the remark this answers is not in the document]
necessarily-equal: Fixed: basically this is already the default, because the struct is initilised to 0 (I don't remember if that's with an explicit memset or with a `= {};`, but it doesn't matter)

src/sgame/sg_buildable.cpp
necessarily-equal: about the pointers, basically because the code around it uses pointer already, so it's more consistent.

I've reworked the `if(cond) return true; else return false;`, but kept the switch. I still find it more readable that way.

necessarily-equal:
Should be ready. I find it makes alien bots much more dangerous against bases in early game :)

illwieckz:
> Merging this would be problematic since then we won't be able to easily test the 0.52 branch using the 0.51 packages.

Thanks to RC we now have everything required to test it.

illwieckz:
That looks to work as expected and I haven't seen regression.

Human bots did not have hope in 0.52RC before that, I'm afraid real humans against bots may not have hope either?

illwieckz:
I just started a 5v5 bot match on plat23, and after 3min, not only no one dretch killed one turret, but all alien bots reaching human base had evolved to more evolved form. Marauder killed two turrets at 4min.

PR #1337, opened 2021-02-23
Introduce CG_Team

necessarily-equal:


slipher:
When introducing a new function, make sure to actually use it somewhere, so we don't have dead code.

necessarily-equal:
Ok, I've updated this PR and rebased. It should be ready for review.

src/cgame/cg_local.h
```diff
@@ -2050,6 +2050,20 @@ void CG_PositionRotatedEntityOnTag( refEntity_t *entity, const refEntity_t *pare
                                     qhandle_t parentModel, const char *tagName );
 void CG_TransformSkeleton( refSkeleton_t *skel, const vec_t scale );
 
+team_t CG_Team(const entityState_t &es);
```
slipher, line 2053: It would be good to have a comment about whether these functions always relate to the client, or if they take on the role of the watched player in case of a spectator client.

src/cgame/cg_ents.cpp
```diff
@@ -1379,3 +1379,21 @@ void CG_AddPacketEntities()
...
 	}
 }
+
+// see also G_Team in src/sgame/sg_ents.cpp
+team_t CG_Team(const entityState_t &es)
+{
+	switch (es.eType)
+	{
+		case entityType_t::ET_PLAYER:
+		case entityType_t::ET_CORPSE:
```
slipher, line 1389: What's the sense of a corpse being on a team? They don't even have any effect on gameplay.

src/cgame/cg_local.h
```diff
@@ -2064,6 +2064,17 @@ inline bool CG_IsOnMyTeam(const entityState_t &es) {
 	return CG_Team(es) == CG_MyTeam();
 }
 
+int CG_Health(const entityState_t &es);
+inline int CG_Health(const centity_t *ent) {
+	return CG_Health(ent->currentState);
+}
+inline bool CG_IsAlive(const entityState_t &es) {
+	return CG_Health(es) == 0;
```
slipher, line 2072: This doesn't look right

slipher (commented):
Are you really sure it's always OK to swap `cg.snap->ps` for `cg.predictedPlayerState` or vice versa?

src/cgame/cg_ents.cpp
necessarily-equal: I don't think it's used in cgame, but it allows for more consistency with sgame, that does this too.

src/cgame/cg_local.h
necessarily-equal: Good point. They take the role of the watched player, for HUD purposes I guess.

src/cgame/cg_local.h
necessarily-equal: Indeed, fixed.

src/cgame/cg_local.h
necessarily-equal: I've added a comment and an alternative for those who could want it.

slipher:
Put the evolving displacement feature in a separate PR

slipher (commented):
Reviewed the commits except the two about evolving near walls. I'm still nervous about swapping the cg.snap state with the cg.predictedPlayerState, in regards to CG_MyTeam.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -1545,18 +1543,21 @@ static void CG_ScanForCrosshairEntity()
...
 	}
 
-	ownTeam = ( team_t ) cg.snap->ps.persistant[ PERS_TEAM ];
-	targetState = &cg_entities[ trace.entityNum ].currentState;
+	entityState_t &targetState = cg_entities[ trace.entityNum ].currentState;
+
+	team_t ownTeam    = CG_MyTeam();
+	team_t targetTeam = CG_Team(targetState);
+
+	if (!CG_IsAlive(targetState))
```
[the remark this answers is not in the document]
necessarily-equal: I think it's only used when in a team playing, so that you should not have odd conditions where you don't have an entity.

Generally speaking if you are dead, you still have an entity and you do get an entity as soon as you spawn. I think you normally have an entity at all time?

necessarily-equal:
> Put the evolving displacement feature in a separate PR

indeed, that's an infortunate result of `git rebase master` when a commit sneaked there.

src/cgame/cg_rocket_draw.cpp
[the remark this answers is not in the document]
necessarily-equal: also, if you want to harden this function, I can add this in CG_MyTeam and other teams functions

```C
if (team <= TEAM_NONE || team >= NUM_TEAMS) return TEAM_NONE;
```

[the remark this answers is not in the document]
slipher: No I mean the target entity. Maybe it could be a mover or something.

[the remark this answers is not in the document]
necessarily-equal: I mean, if it's a mover then I'd be surprised if it's in a team, but yeah, that changed. The whole point of this code is to "tag" the crosshair depending if it's a friend or foe though.

Maybe one day we could make it react to destructable wall, etc, but those don't currently work in unv.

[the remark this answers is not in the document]
slipher: I tried it out and you can indeed inspect non-alive entities. Hence the `!CG_IsAlive` check disrupts the functionality.
unvanquished_2021-09-20_021849_000
.

PR #1345, opened 2021-03-09
Improve human bot shoping

### LINKED ISSUE #1344: human bots will never buy light armor at stage 0
Human bots will never buy light armor in stage 0.

bmorel:
* fixes #1344
* also fixes bots taking machinegun despite it being disabled
* makes code safer by providing an array size
* move long list of checks into a for loop
* adds some unrelated comments

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1292,7 +1292,7 @@ AINodeStatus_t BotActionBuy( gentity_t *self, AIGenericNode_t *node )
 	if ( buy->nparams == 0 )
 	{
 		// equip action
-		BotGetDesiredBuy( self, &weapon, upgrades, &numUpgrades );
+		numUpgrades = BotGetDesiredBuy( self, weapon, upgrades, 4 );
 	}
 	else
 	{
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -321,6 +321,13 @@ int BotValueOfUpgrades( gentity_t *self )
 // * "upgrades" contains a list of upgrades to use
 // * "weapon" contains weapon to use
 // * Returns number of upgrades to buy (does not includes weapons)
+//
+// will favor better armor above everything else, then team-utilies
+// and finally the most expensive gun possible.
+// TODO: allow bots to buy jetpack and grenades, despite the fact
+//   they can't use them (yet. For grenades it should not be that
+//   hard?): since default cVar prevent bots to buy those, that
+//   will be one less thing to change later and would have no impact.
 int BotGetDesiredBuy( gentity_t *self, weapon_t &weapon, upgrade_t upgrades[], size_t upgradesSize )
 {
 	ASSERT( self && upgrades );
@@ -340,36 +347,18 @@ int BotGetDesiredBuy( gentity_t *self, weapon_t &weapon, upgrade_t upgrades[], s
 		upgrades[i] = UP_NONE;
 	}
 
-	if ( BG_UpgradeUnlocked( UP_BATTLESUIT ) && usableCapital >= BG_Upgrade( UP_BATTLESUIT )->price )
+	// manually sorted by preference, hopefully a future patch will have a much smarter way to select weapon
+	struct
 	{
-		upgrades[numUpgrades] = UP_BATTLESUIT;
-	}
-	else if ( BG_UpgradeUnlocked( UP_MEDIUMARMOUR ) && usableCapital >= BG_Upgrade( UP_MEDIUMARMOUR )->price )
+		int authorized;
+		int price;
+		upgrade_t upgrade;
+	} armors[] =
 	{
-		upgrades[numUpgrades] = UP_MEDIUMARMOUR;
-	}
-	else if ( usableCapital >= BG_Upgrade( UP_LIGHTARMOUR )->price )
-	{
-		upgrades[numUpgrades] = UP_LIGHTARMOUR;
-	}
-	usableCapital -= BG_Upgrade( upgrades[numUpgrades] )->price;
-	numUpgrades += upgrades[numUpgrades] == UP_NONE ? 0 : 1;
-
-	//TODO this really needs more generic code, but that would require
-	//deeper refactoring (probably move equipments and classes into structs)
-	//and code to make bots _actually_ use other equipments.
-	unsigned int alliesNumbers[MAX_CLIENTS] = {};
-	int nbAllies = FindAllies( alliesNumbers, MAX_CLIENTS, G_Team( self ) );
-	int nbRadars = numTeamUpgrades[UP_RADAR];
-	bool teamNeedsRadar = 100 * ( 1 + nbRadars ) / ( 1 + nbAllies ) < 75;
-	if ( teamNeedsRadar && g_bot_radar.integer
-			&& BG_UpgradeUnlocked( UP_RADAR ) && usableCapital >= BG_Upgrade( UP_RADAR )->price
-			&& ( BG_Upgrade( upgrades[0] )->slots & BG_Upgrade( UP_RADAR )->slots ) == 0 )
-	{
-		upgrades[numUpgrades] = UP_RADAR;
-		usableCapital -= BG_Upgrade( upgrades[numUpgrades] )->price;
-		numUpgrades ++;
-	}
+		{ g_bot_battlesuit.integer  , BG_Upgrade( UP_BATTLESUIT )->price, UP_BATTLESUIT },
+		{ g_bot_mediumarmour.integer, BG_Upgrade( UP_MEDIUMARMOUR )->price, UP_MEDIUMARMOUR },
+		{ g_bot_lightarmour.integer , BG_Upgrade( UP_LIGHTARMOUR )->price, UP_LIGHTARMOUR },
+	};
 
 	// manually sorted by preference, hopefully a future patch will have a much smarter way to select weapon
 	struct
@@ -390,6 +379,34 @@ int BotGetDesiredBuy( gentity_t *self, weapon_t &weapon, upgrade_t upgrades[], s
 		{ g_bot_painsaw.integer , BG_Weapon( WP_PAIN_SAW )->price      , WP_PAIN_SAW },
 		{ g_bot_rifle.integer   , BG_Weapon( WP_MACHINEGUN )->price    , WP_MACHINEGUN },
 	};
+
+	for ( auto const &armor : armors )
+	{
+		if ( armor.authorized
+				&& BG_UpgradeUnlocked( armor.upgrade ) && usableCapital >= armor.price )
+		{
+			upgrades[numUpgrades] = armor.upgrade;
+			usableCapital -= armor.price;
+			numUpgrades++;
+			break;
+		}
+	}
+
+	//TODO this really needs more generic code, but that would require
+	//deeper refactoring (probably move equipments and classes into structs)
+	//and code to make bots _actually_ use other equipments.
+	unsigned int alliesNumbers[MAX_CLIENTS] = {};
+	int nbAllies = FindAllies( alliesNumbers, MAX_CLIENTS, G_Team( self ) );
+	int nbRadars = numTeamUpgrades[UP_RADAR];
+	bool teamNeedsRadar = 100 * ( 1 + nbRadars ) / ( 1 + nbAllies ) < 75;
+	if ( teamNeedsRadar && g_bot_radar.integer
+			&& BG_UpgradeUnlocked( UP_RADAR ) && usableCapital >= BG_Upgrade( UP_RADAR )->price
+			&& ( BG_Upgrade( upgrades[0] )->slots & BG_Upgrade( UP_RADAR )->slots ) == 0 )
+	{
+		upgrades[numUpgrades] = UP_RADAR;
+		usableCapital -= BG_Upgrade( upgrades[numUpgrades] )->price;
+		numUpgrades ++;
+	}
 
 	for ( auto const &wp : weapons )
 	{
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -375,35 +375,35 @@ int BotGetDesiredBuy( gentity_t *self, weapon_t &weapon, upgrade_t upgrades[], s
 	struct
 	{
 		int authorized;
+		int price;
 		weapon_t weapon;
 	} weapons[] =
 	{
-		{ g_bot_lcannon.integer , WP_LUCIFER_CANNON },
-		{ g_bot_flamer.integer  , WP_FLAMER },
+		{ g_bot_lcannon.integer , BG_Weapon( WP_LUCIFER_CANNON )->price, WP_LUCIFER_CANNON },
+		{ g_bot_flamer.integer  , BG_Weapon( WP_FLAMER )->price        , WP_FLAMER },
 		// pulse rifle has lower priority to keep previous "correct" behavior
-		{ g_bot_prifle.integer  , WP_PULSE_RIFLE },
-		{ g_bot_chaingun.integer, WP_CHAINGUN },
-		{ g_bot_mdriver.integer , WP_MASS_DRIVER },
-		{ g_bot_lasgun.integer  , WP_LAS_GUN },
-		{ g_bot_shotgun.integer , WP_SHOTGUN },
-		{ g_bot_painsaw.integer , WP_PAIN_SAW },
-		{ g_bot_rifle.integer   , WP_MACHINEGUN },
+		{ g_bot_prifle.integer  , BG_Weapon( WP_PULSE_RIFLE )->price   , WP_PULSE_RIFLE },
+		{ g_bot_chaingun.integer, BG_Weapon( WP_CHAINGUN )->price      , WP_CHAINGUN },
+		{ g_bot_mdriver.integer , BG_Weapon( WP_MASS_DRIVER )->price   , WP_MASS_DRIVER },
+		{ g_bot_lasgun.integer  , BG_Weapon( WP_LAS_GUN )->price       , WP_LAS_GUN },
+		{ g_bot_shotgun.integer , BG_Weapon( WP_SHOTGUN )->price       , WP_SHOTGUN },
+		{ g_bot_painsaw.integer , BG_Weapon( WP_PAIN_SAW )->price      , WP_PAIN_SAW },
+		{ g_bot_rifle.integer   , BG_Weapon( WP_MACHINEGUN )->price    , WP_MACHINEGUN },
 	};
 
 	for ( auto const &wp : weapons )
 	{
-		if ( wp.authorized && BG_WeaponUnlocked( wp.weapon ) && usableCapital >= BG_Weapon( wp.weapon )->price )
+		if ( wp.authorized && BG_WeaponUnlocked( wp.weapon ) && usableCapital >= wp.price )
 		{
 			if ( wp.weapon == WP_FLAMER && numTeamWeapons[WP_FLAMER] > numTeamWeapons[WP_PULSE_RIFLE] )
 			{
 				continue;
 			}
 			weapon = wp.weapon;
+			usableCapital -= wp.price;
 			break;
 		}
 	}
-
-	usableCapital -= BG_Weapon( weapon )->price;
 
 	//now test to see if we already have all of these items
 	//check if we already have everything
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -331,6 +331,10 @@ int BotGetDesiredBuy( gentity_t *self, weapon_t &weapon, upgrade_t upgrades[], s
 	int usableCapital = credits + equipmentPrice;
 	int numUpgrades = 0;
 
+	unsigned int numTeamUpgrades[UP_NUM_UPGRADES] = {};
+	unsigned int numTeamWeapons[WP_NUM_WEAPONS] = {};
+
+	ListTeamEquipment( self, numTeamUpgrades, UP_NUM_UPGRADES, numTeamWeapons, WP_NUM_WEAPONS );
 	for ( size_t i = 0; i < upgradesSize; ++i )
 	{
 		upgrades[i] = UP_NONE;
@@ -383,6 +387,10 @@ int BotGetDesiredBuy( gentity_t *self, weapon_t &weapon, upgrade_t upgrades[], s
 	{
 		if ( wp.authorized && BG_WeaponUnlocked( wp.weapon ) && usableCapital >= BG_Weapon( wp.weapon )->price )
 		{
+			if ( wp.weapon == WP_FLAMER && numTeamWeapons[WP_FLAMER] > numTeamWeapons[WP_PULSE_RIFLE] )
+			{
+				continue;
+			}
 			weapon = wp.weapon;
 			break;
 		}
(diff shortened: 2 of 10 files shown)


src/sgame/sg_bot_util.cpp
```diff
@@ -1355,6 +1358,31 @@ float BotAimAngle( gentity_t *self, vec3_t pos )
...
 }
 
+// fills allyEntityNumbers with players of a team, and return the number of entities found.
+// pre-condition:
+// * team is playable
+// * allyEntityNumbers is an array of at least maxAllies unsigned integers
+// * allyEntityNumbers memory is initialized
+// * there are less than MAX_CLIENTS players in team
+// * there is enough space to store all team's player's references into allyEntityNumbers
+unsigned int FindAllies( unsigned int *allyEntityNumbers, unsigned int maxAllies, team_t team )
```
necessarily-equal, line 1381: I think it would help to use `unsigned int allyEntityNumbers[]` or even `unsigned int allyEntityNumbers[maxAllies]` here, though the later could exist in C only (not C++).

src/sgame/sg_bot_util.cpp
```diff
@@ -332,6 +332,12 @@ int BotGetDesiredBuy( gentity_t *self, weapon_t *weapon, upgrade_t *upgrades, in
 	int usableCapital = credits + equipmentPrice;
 	int numUpgrades = 0;
 
+	unsigned int numTeamUpgrades[UP_NUM_UPGRADES];
+	unsigned int numTeamWeapons[WP_NUM_WEAPONS];
+	memset( numTeamUpgrades, 0, sizeof( numTeamUpgrades ) );
+	memset( numTeamWeapons , 0, sizeof( numTeamWeapons  ) );
```
necessarily-equal, line 338: erm, how about replacing these with `= {};` on the above lines?

src/sgame/sg_bot_util.cpp
```diff
@@ -352,7 +358,10 @@ int BotGetDesiredBuy( gentity_t *self, weapon_t *weapon, upgrade_t *upgrades, in
 	usableCapital -= BG_Upgrade( upgrades[numUpgrades] )->price;
 	numUpgrades += upgrades[numUpgrades] == UP_NONE ? 0 : 1;
 
-	if ( BG_UpgradeUnlocked( UP_RADAR )
+	const team_t team = static_cast<team_t>( self->client->pers.team );
```
necessarily-equal, line 361: `team = G_Team( self );`

src/sgame/sg_bot_util.cpp
necessarily-equal: Also, shouldn't this should take itself as argument to avoid including itself? Then `self->ent.OnSameTeam(other)` or `G_OnSameTeam` could be used.

src/sgame/sg_bot_util.cpp
```diff
@@ -310,8 +310,18 @@ int BotValueOfUpgrades( gentity_t *self )
...
+// * weapon is a valid pointer
+// * "upgrades" is an array of at least maxUpgrades elements
+// post-conditions:
+// * "upgrades" contains a list of upgrades to use
+// * "weapon" contains weapon to use
+// * Returns number of upgrades to buy (does not includes weapons)
 int BotGetDesiredBuy( gentity_t *self, weapon_t *weapon, upgrade_t *upgrades, int maxUpgrades )
 {
+	ASSERT( self && weapon && upgrades );
+	ASSERT( self->client->pers.team == TEAM_HUMANS ); // only humans can buy
```
necessarily-equal, line 327: `G_Team(self)`? Otherwise, I think the modern way is `self->ent.Get<TeamComponent>()->Team()` :P

src/sgame/sg_bot_util.cpp
```diff
@@ -310,96 +310,70 @@ int BotValueOfUpgrades( gentity_t *self )
...
+		{ g_bot_prifle.integer  , WP_PULSE_RIFLE },
+		{ g_bot_chaingun.integer, WP_CHAINGUN },
+		{ g_bot_mdriver.integer , WP_MASS_DRIVER },
+		{ g_bot_lasgun.integer  , WP_LAS_GUN },
+		{ g_bot_shotgun.integer , WP_SHOTGUN },
+		{ g_bot_painsaw.integer , WP_PAIN_SAW },
+		{ g_bot_rifle.integer   , WP_MACHINEGUN },
+	};
+
+	for ( auto const &wp : weapons )
```
necessarily-equal, line 392: nice!

src/sgame/sg_bot_ai.cpp
```diff
@@ -1262,7 +1262,7 @@ AINodeStatus_t BotActionBuy( gentity_t *self, AIGenericNode_t *node )
 	if ( buy->nparams == 0 )
 	{
 		// equip action
-		BotGetDesiredBuy( self, &weapon, upgrades, &numUpgrades );
+		numUpgrades = BotGetDesiredBuy( self, &weapon, upgrades, 4 );
```
necessarily-equal, line 1265: > 4

magic number

src/sgame/sg_bot_util.cpp
```diff
@@ -1459,6 +1486,50 @@ bool PlayersBehindBotInSpawnQueue( gentity_t *self )
...
+		if ( team == TEAM_HUMANS )
+		{
+			// for all possible upgrades, check if current player already have it.
+			// this is very fragile, at best, since it will break if TEAM_ALIENS becomes
+			// able to have upgrades. Class is not considered, neither.
+			// last but not least, it's playing with bits.
+			int * stat = &ally->client->ps.stats[STAT_ITEMS];
+			for ( int up = UP_NONE + 1; up < UP_NUM_UPGRADES; ++up )
+			{
+				if ( *stat & ( 1 << up ) )
```
necessarily-equal, line 1524: If the `*stat` is constant, why check it on each loop? Also, I don't see what it purpose is, which means it could use a better name (but maybe it shouldn't exist at all?).

src/sgame/sg_bot_util.cpp
```diff
@@ -1459,6 +1486,50 @@ bool PlayersBehindBotInSpawnQueue( gentity_t *self )
...
+// * numWeapons memory is initialized
+// * self's team have less than MAX_CLIENTS players
+void TeamEquipment( gentity_t *self, unsigned int* numUpgrades, size_t maxUpgrades, unsigned int* numWeapons, size_t maxWeapons )
+{
+	ASSERT( self );
+	ASSERT( numUpgrades );
+	ASSERT( numWeapons );
+	const team_t team = static_cast<team_t>( self->client->pers.team );
+	unsigned int alliesNumbers[MAX_CLIENTS];
+	memset( alliesNumbers, 0, sizeof( alliesNumbers ) );
```
necessarily-equal, line 1504: `= {};` is better than memset here too

src/sgame/sg_bot_util.cpp
```diff
@@ -1459,6 +1486,50 @@ bool PlayersBehindBotInSpawnQueue( gentity_t *self )
...
+// * numUpgrades memory is initialized
+// * numWeapons is an array of at least maxWeapons unsigned integers
+// * numWeapons memory is initialized
+// * self's team have less than MAX_CLIENTS players
+void TeamEquipment( gentity_t *self, unsigned int* numUpgrades, size_t maxUpgrades, unsigned int* numWeapons, size_t maxWeapons )
+{
+	ASSERT( self );
+	ASSERT( numUpgrades );
+	ASSERT( numWeapons );
+	const team_t team = static_cast<team_t>( self->client->pers.team );
```
necessarily-equal, line 1502: `G_Team( self );`

src/sgame/sg_bot_util.cpp
```diff
@@ -1382,12 +1408,13 @@ int FindBots( int *botEntityNumbers, int maxBots, team_t team )
 	int numBots = 0;
 	int i;
 	memset( botEntityNumbers, 0, sizeof( int )*maxBots );
```
necessarily-equal, line 1410: `= {};`

src/sgame/sg_bot_util.cpp
```diff
@@ -1382,12 +1408,13 @@ int FindBots( int *botEntityNumbers, int maxBots, team_t team )
...
-	for ( i = 0; i < MAX_CLIENTS; i++ )
+	for ( i = 0; i < MAX_CLIENTS && numBots < maxBots; i++ )
 	{
 		testEntity = &g_entities[i];
+		// testEntity->r.svFlags => "SVF_NOCLIENT, SVF_BROADCAST, etc..." good thing grep exists!
 		if ( testEntity->r.svFlags & SVF_BOT )
 		{
-			if ( testEntity->client->pers.team == team && numBots < maxBots )
+			if ( testEntity->client->pers.team == team )
```
necessarily-equal, line 1417: could use `G_Team()` too

necessarily-equal (changes requested):
Neat, here's a review

src/sgame/sg_bot_util.cpp
```diff
@@ -310,96 +310,70 @@ int BotValueOfUpgrades( gentity_t *self )
 	return worth;
 }
 
-void BotGetDesiredBuy( gentity_t *self, weapon_t *weapon, upgrade_t *upgrades, int *numUpgrades )
+int BotGetDesiredBuy( gentity_t *self, weapon_t *weapon, upgrade_t *upgrades, int maxUpgrades )
```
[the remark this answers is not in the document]
necessarily-equal: upgrades could be marked as an array `upgrade_t upgrades[]` and weapon could avoid an assert by using a reference `weapon_t &weapon`.

src/sgame/sg_bot_util.cpp
bmorel: Ideally, I'd use a G_IsOnSameTeam() or alike that is pending in another branch, maybe I'll see if I can cherry pick it.

src/sgame/sg_bot_util.cpp
```diff
@@ -1444,6 +1472,50 @@ bool PlayersBehindBotInSpawnQueue( gentity_t *self )
...
 
+// Initializes numUpgrades and numWeapons arrays with team's current equipment
+// pre-condition:
+// * self points to a player which is inside a playable team
+// * numUpgrades is an array of at least maxUpgrades unsigned integers
+// * numUpgrades memory is initialized
+// * numWeapons is an array of at least maxWeapons unsigned integers
+// * numWeapons memory is initialized
+// * self's team have less than MAX_CLIENTS players
+void TeamEquipment( gentity_t *self, unsigned int* numUpgrades, size_t maxUpgrades, unsigned int* numWeapons, size_t maxWeapons )
```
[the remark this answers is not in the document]
bmorel: I've gone for ListTeamEquipement, and used arrays there, too.

bmorel:
Note that I have a pending commit to integrate UI elements for the new cVar, but I will wait for https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/39 and this one to be merged before doing a PR for that. Messing with branches for trivial patches is super annoying, and it would make no sense without this PR anyway (worse: it would be a broken PR).

bmorel:
New set of commits also fix #1347

src/sgame/sg_bot_util.cpp
```diff
@@ -316,26 +316,24 @@ int BotValueOfUpgrades( gentity_t *self )
...
 // * "weapon" contains weapon to use
 // * Returns number of upgrades to buy (does not includes weapons)
-int BotGetDesiredBuy( gentity_t *self, weapon_t *weapon, upgrade_t *upgrades, int maxUpgrades )
+int BotGetDesiredBuy( gentity_t *self, weapon_t &weapon, upgrade_t upgrades[], size_t upgradesSize )
 {
-	ASSERT( self && weapon && upgrades );
+	ASSERT( self && upgrades );
 	ASSERT( self->client->pers.team == TEAM_HUMANS ); // only humans can buy
-	ASSERT( maxUpgrades >= 2 ); // because human bots will want up to 2 upgrades...
+	ASSERT( upgradesSize >= 2 ); // because human bots will want up to 2 upgrades...
```
necessarily-equal, line 328: I don't understand the comment here, "up to 2"?

bmorel: armor and radar. Other upgrades can not, in current state, be used by bots.

slipher:
Can this be split up? 14 is too many commits.

necessarily-equal:
I don't think it should be split up, the more recent commits can simply be squashed instead

necessarily-equal:
I don't think this should be split, IMO the more recent commits (fixes) shou'd be squashed

bmorel:
I avoided squashing to make it easier to move between my branches, and it's true that I added more changes after initial review.
Some commits can probably be squashed and keep a meaningful history.

src/sgame/sg_bot_util.cpp
```diff
@@ -1459,6 +1497,50 @@ bool PlayersBehindBotInSpawnQueue( gentity_t *self )
...
 
+// Initializes numUpgrades and numWeapons arrays with team's current equipment
+// pre-condition:
+// * self points to a player which is inside a playable team
+// * numUpgrades is an array of at least numUpgradesSize unsigned integers
+// * numUpgrades memory is initialized
+// * numWeapons is an array of at least numWeaponsSize unsigned integers
+// * numWeapons memory is initialized
+// * self's team have less than MAX_CLIENTS players
+void ListTeamEquipment( gentity_t *self, unsigned int numUpgrades[], size_t numUpgradesSize, unsigned int numWeapons[], size_t numWeaponsSize )
```
slipher, line 1508: use `static` for functions used only within the same file

src/sgame/sg_bot_util.cpp
```diff
@@ -1370,6 +1382,31 @@ float BotAimAngle( gentity_t *self, vec3_t pos )
...
+// * allies is an array of at least alliesSize unsigned integers
+// * allies memory is initialized
+// * there are less than MAX_CLIENTS players in team
+// * there is enough space to store all team's player's references into allies
+unsigned int FindAllies( unsigned int allies[], size_t alliesSize, team_t team )
+{
+	ASSERT( G_IsPlayableTeam( team ) );
+	gentity_t *testEntity;
+	unsigned int numAllies = 0;
+	for ( size_t i = 0; i < MAX_CLIENTS && numAllies < alliesSize; i++ )
```
slipher, line 1397: You could use `ForEntities<ClientComponent>` and then check the `TeamComponent`.

Also it seems like you wouldn't want to count yourself as an ally.

slipher (commented):
With the strict ordering of radar before armor before weapons, it will do stuff like buy radar with naked rifle, or rifle + bsuit. This seems like a regression compared to the old algorithm, which at least ensured that it had enough money left over for a decent weapon before buying some armor.

bmorel:
> With the strict ordering of radar before armor before weapons, it will do stuff like buy radar with naked rifle, or rifle + bsuit. This seems like a regression compared to the old algorithm, which at least ensured that it had enough money left over for a decent weapon before buying some armor.

This is an accidental change made by the rebasing, I admit I only "tested" things by compiling after, since behavior was verified before.
Armors should be bought 1st, then radar, then weapon, so that it should be close to old (but without the "no armor at stage 0" bug, ability to spare some money by not buying "useless" radar and ability to mix flamer/pulse, of course)

src/sgame/sg_bot_util.cpp
bmorel: > You could use ForEntities<ClientComponent> and then check the TeamComponent.

I'm reluctant to do that, for several reasons, some of those being:

* it is not documented, the only "documentation" in there is: `// Definitions of ForEntities //`
* it's a variadic template, without static_assert, which means it will end in me wasting lot of time debugging cryptic error messages, to replace a single, easy to read for anyone, line.
* from what I could see, it relies on a std::set. I'm not sure using a red-black tree to store (sorted...) pointers is a good idea, for a start, nor that it is a good idea to use that to iterate over a flat array. From what I understand by reading code, performance-wise, it would be a bad idea (this function is likely to be called several times per "server frame")

> Also it seems like you wouldn't want to count yourself as an ally.

Right, the name is probably not nicely chosen. I've renamed it as ListTeamMembers.

slipher: > 
> 
> > You could use ForEntities and then check the TeamComponent.
> 
> I'm reluctant to do that, for several reasons, some of those being:
> 
>     * it is not documented, the only "documentation" in there is: `// Definitions of ForEntities //`
>     * it's a variadic template, without static_assert, which means it will end in me wasting lot of time debugging cryptic error messages, to replace a single, easy to read for anyone, line.

If you refuse to use an undocumented function, you're not going to get far in this codebase. It should be doable to figure out the usage from examples (which invariably have a single argument).

> 
>     * from what I could see, it relies on a std::set. I'm not sure using a red-black tree to store (sorted...) pointers is a good idea, for a start, nor that it is a good idea to use that to iterate over a flat array. From what I understand by reading code, performance-wise, it would be a bad idea (this function is likely to be called several times per "server frame")

Since we are iterating over all entities, rather than looking one up, it is similar to iterating a linked list. Yes, there is room for performance improvements, but the penalty of a linked list over an array is not at  enough to justify spurning a clean solution for one that is, in your own words, "pure guessing and luck" :smile: 

> > Also it seems like you wouldn't want to count yourself as an ally.
> 
> Right, the name is probably not nicely chosen. I've renamed it as ListTeamMembers.
> I'm not sure it should be changed to skip self.

OK, but you still need to exclude the bot's own weapons somehow when doing the calculations.

slipher (commented):
The order is now armor,radar,weapons rather than radar,armor,weapons; but it can still buy radar with naked rifle, or rifle + bsuit. Which is a regression compared to the old algorithm, which (ignoring the possibility of disabled weapon cvars) ensured that it had enough money left over for a decent weapon before buying some armor.

bmorel:
Le Sun, 21 Mar 2021 19:49:26 -0700,
slipher ***@***.***> a écrit :

> @slipher commented on this pull request.
> 
> The order is now armor,radar,weapons rather than radar,armor,weapons;
> but it can still buy radar with naked rifle, or rifle + bsuit. Which
> is a regression compared to the old algorithm, which (ignoring the
> possibility of disabled weapon cvars) ensured that it had enough
> money left over for a decent weapon before buying some armor.

A decent weapon?
Smg have the 2nd stage 0 DpS, and have a fire rate which makes it
usable both in melee and at distance.
All human weapons are decent, except maybe lasgun, which deals less
damages per shot than mass driver while requiring constant, perfect
aiming.

Even if I was wrong and SMG was not a decent weapon, the old
algorithm is worse in many other points. The old algorithm:

* is unable to buy armors at stage 0, which is *huge* bug (probably the
  1st bug I've seen directly favoring alien side though).
* would never have saved money when all other teammates were equipped
  with a radar (this is a very arguable thing, which is why it's still
  in a different commit).
* would never have let bot buy a pulse rifle (same as above, but you
  will need quite some time to make me agree that bots not using pulse
  rifle is a good thing for human team).
* would sometime favor shotgun+radar+light armor over medium armor.

The algo I push clearly indicates what it does (unlike previous one),
is easy to maintain (unlike previous one), works according to
description (like previous one I guess, since previous one had no
description) which is reasonably close to what one would also expect
from average player (this is purely my opinion, I admit it): sure,
sometimes it might makes sense to favor light armor over medium,
because you want to run faster, but that's not something you'll see
often. Average player would also buy grenades and jetpack, depending on
various parameters, and I could make bots buy those (it's on my
todo-list) but for now, they don't know how to use them.

As for rifle+bsuit... sure it can, and? Old algo also authorized that.

> > > You could use ForEntities and then check the TeamComponent.
> > 
> > I'm reluctant to do that, for several reasons, some of those being:
> 
> If you refuse to use an undocumented function, you're not going to
> get far in this codebase. It should be doable to figure out the usage
> from examples (which invariably have a single argument).

True about not using undocumented stuff.
My point still stands, which is: I think this is an overkill and
undocumented function, that is hard to understand how it works.
Unlike a for loop, that _any_ developer will immediately understand
what it implies, this function "requires reading other examples".
Those examples "invariably use a single argument"?
Why using variadic templates + std::set (a B-tree) then? Because C++
can?

> >     * from what I could see, it relies on a std::set. I'm not sure
> > using a red-black tree to store (sorted...) pointers is a good
> > idea, for a start, nor that it is a good idea to use that to
> > iterate over a flat array. From what I understand by reading code,
> > performance-wise, it would be a bad idea (this function is likely
> > to be called several times per "server frame")
> 
> Since we are iterating over all entities, rather than looking one up,
> it is similar to iterating a linked list. Yes, there is room for
> performance improvements, but the penalty of a linked list over an
> array is not at  enough to justify spurning a clean solution for one
> that is, in your own words, "pure guessing and luck" :smile: 

My own words are because the code base lack documentation (which is
something I also improve locally in this PR, so I'm not just whining
for the beauty of whining).
They are also because I was able to find that solution myself at a
moment I knew nothing about the code base (rebasing made information
disappear, but it's really old comment, ~2 months ago I suppose) and had
doubts it would do the correct behavior, yet I thought it had
reasonable hopes to work.

Those are untrue for ForEntities(), for which I've read the code. I
needed "a little more" (huge under-statement) than 15 minutes to have a
grasp at how it __might__ work. It still works over the same kind of
stuff, so understanding it still requires to understand the mechanism
I've used.
Currently, I probably understand how it works in the big lines, and I
disagree it's a good idea to use it in it's current state. Not because
it would not work, but because it is not clear what it really does, and
such trivial thing such as iterating over an array (not even
containing instances of polymorphic stuff) should be... well, trivial,
and not rely on variadic templates and std::set and whatever dirt I have
not seen because it's too far under the carpet.

To put it in a single word, in my opinion, this mechanism is
over-engineered.
I'm ok to use over-engineered solutions when the alternatives are
worse, like, I could be fired otherwise, or there are no other way.
Here, those points don't apply.

> > > Also it seems like you wouldn't want to count yourself as an ally.
> > 
> > Right, the name is probably not nicely chosen. I've renamed it as
> > ListTeamMembers. I'm not sure it should be changed to skip self.
> 
> OK, but you still need to exclude the bot's own weapons somehow when
> doing the calculations.

This list the team's equipment. The bot should be able to know what it
currently have, thus if needed, this should be done in caller.
I will patch GetDesiredBuy later accordingly, to avoid bots switching
pulse/flamer or selling radar in favor of bigger gun (and thus going
back to armory for nothing in the process).

slipher (commented):
For every CBSE component, there is a list of entities which have that component. `ForEntities` iterates over that list for a given component type. The point is that we shouldn't have to loop over all 1024 entities to get just the human buildings or whatever. It uses `std::set` because it is *under*-engineered if anything - a more efficient but more complex implementation could be written, but no one has done that work. What `std::set` gives us, for now, is the ability to add or remove entities during iteration without crashing (I believe this indeed happens, during the "thinking" loop).

> The old algorithm:
> * is unable to buy armors at stage 0

Ha, nice find. Probably the painsaw was available from the start when the code was written...

It's true that lasgun and mdriver combine poorly with the battlesuit, perhaps even worse than rifle. IMO only painsaw, shotgun, or one of the 400+ credit weapons should be combined with bsuit.

Radar with no armor seems like a very poor choice in any situation.

bmorel:
This PR should probably be  closed, my betterai branch have received various improvements since that, and with the regular rebases to keep history cleaner, it's likely to just be annoying to resync everything on top of this.
There are still few improvements I want to put into AI before a "first iteration" should be merged, plus those changes also interact with UnvanquishedAssets/unvanquished_src.dpkdir.git repo, so things are likely to be simpler to review when those repos will be merged.

illwieckz:
@bmorel said this is obsoleted by work in progress branch:
https://github.com/bmorel/Unvanquished/commits/betterai

PR #1348, opened 2021-03-17
Support code for circle menu rewrite

slipher:


slipher:
Stuff needed for https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/40.

I haven't deleted any of the stuff for the old circle menus yet, so they will still work with this gamelogic.

PR #1349, opened 2021-03-18
Secfixes and related cleanups

necessarily-equal:


src/shared/parse.cpp
```diff
@@ -3182,7 +3180,7 @@ static int Parse_Directive_error( source_t *source )
 {
 	token_t token;
 
-	strcpy( token.string, "" );
+	*token.string = '0';
```
slipher, line 3088: should be `\0`

src/shared/parse.cpp
```diff
@@ -3547,10 +3545,8 @@ static int Parse_Directive_define( source_t *source )
 	}
 
 	//allocate define
-	define = ( define_t * ) Z_Malloc( sizeof( define_t ) + strlen( token.string ) + 1 );
-	Com_Memset( define, 0, sizeof( define_t ) );
-	define->name = ( char * ) define + sizeof( define_t );
-	strcpy( define->name, token.string );
+	define = (define_t *) calloc( sizeof( define_t ), 1 );
```
slipher, line 3453: The allocation function probably shouldn't be changed, since I don't see any corresponding change to a freeing function.

src/shared/parse.cpp
```diff
@@ -4113,7 +4109,7 @@ static define_t *Parse_CopyDefine( define_t *define )
 	newdefine = ( define_t * ) Z_Malloc( sizeof( define_t ) + strlen( define->name ) + 1 );
 	//copy the define name
 	newdefine->name = ( char * ) newdefine + sizeof( define_t );
-	strcpy( newdefine->name, define->name );
+	memcpy( newdefine->name, define->name, sizeof newdefine );
```
slipher, line 4017: size is wrong

src/shared/parse.cpp
```diff
@@ -4354,7 +4350,7 @@ bool Parse_ReadTokenHandle( int handle, pc_token_t *pc_token )
 Parse_SourceFileAndLine
 ===============
 */
-int Parse_SourceFileAndLine( int handle, char *filename, int *line )
+int Parse_SourceFileAndLine( int handle, char filename[ MAX_QPATH ], int *line )
```
slipher, line 4257: You could type the argument as `char(&filename)[MAX_QPATH]`, to enforce that it is actually an array of that length.

src/shared/parse.cpp
```diff
@@ -4335,7 +4331,7 @@ bool Parse_ReadTokenHandle( int handle, pc_token_t *pc_token )
 	}
 
 	ret = Parse_ReadToken( sourceFiles[ handle ], &token );
-	strcpy( pc_token->string, token.string );
+	memcpy( pc_token->string, token.string, sizeof pc_token->string );
```
slipher, line 4238: Should be `Q_strncpyz`? Copying to the maximum buffer length is undefined behavior if it reads unset bytes at the end, as well as less efficient.

src/shared/parse.cpp
```diff
@@ -4366,7 +4362,7 @@ int Parse_SourceFileAndLine( int handle, char *filename, int *line )
 		return false;
 	}
 
-	strcpy( filename, sourceFiles[ handle ]->filename );
+	memcpy( filename, sourceFiles[ handle ]->filename, MAX_QPATH );
```
slipher, line 4269: Q_strncpyz

slipher (commented):
There's not much of a security concern with this code, as it's located within the cgame, which is untrusted code running in a secure sandbox, and the server controls what content is parsed. But in any case it's nice to have these fixes for correctness - thanks!

Actually I recently moved this code out of the engine and into the gamelogic, partially inspired by the large amount of memory safety bugs. But also it should be more efficient to have it there, since it prevents a lot of unnecessary IPC calls between the engine and gamelogic.

src/shared/parse.cpp
necessarily-equal: that's a really neat trick. I was really surprised to see there was no way to get even a warning from the compiler with the sized array.

src/shared/parse.cpp
```diff
@@ -3547,10 +3545,8 @@ static int Parse_Directive_define( source_t *source )
 	}
 
 	//allocate define
-	define = ( define_t * ) Z_Malloc( sizeof( define_t ) + strlen( token.string ) + 1 );
-	Com_Memset( define, 0, sizeof( define_t ) );
-	define->name = ( char * ) define + sizeof( define_t );
-	strcpy( define->name, token.string );
+	define = (define_t *) Z_Malloc( sizeof( define_t ) );
+	define->name = strdup( token.string );
```
slipher, line 3452: `strdup` is added, without any new freeing functions, so a memory leak must be introduced

necessarily-equal: Right, fixed.

PR #1363, opened 2021-04-11
provide a hint to fix a build error likely caused by submodules

bmorel:
This commit tries to give a useful hint to people which are switching branches about why something build-related broke.


slipher:
This error is not likely to be caused by submodules. It happens when there's a problem with compiler flags. See https://github.com/DaemonEngine/Daemon/pull/443

slipher:
Closing since G_Team constness was already merged, and the compiler flag thing was fixed.

PR #1364, opened 2021-04-12
Reuse recently-freed entities slots when necessary

peoro:
The code was always meant to reuse recently-freed entity slots when necessary. However it never did so as detailed in the commit message.

The affected piece of code dates back to the original Q3A (https://github.com/id-Software/Quake-III-Arena/blob/master/code/game/g_utils.c#L374-L434) and it's pretty critical as well as fiddly.
We may want to send the patch further upstream since many projects are affected and Google has more than a thousand results for "G_Spawn: no free entities".

### DIFF
--- a/src/sgame/sg_entities.cpp
+++ b/src/sgame/sg_entities.cpp
@@ -61,6 +61,11 @@ void G_InitGentity( gentity_t *entity )
 	entity->s.number = entity - g_entities;
 	entity->r.ownerNum = ENTITYNUM_NONE;
 	entity->creationTime = level.time;
+	
+	if ( g_debugEntities.integer > 2 )
+	{
+		Log::Debug("Initing Entity %i", entity - g_entities );
+	}
 }
 
 /*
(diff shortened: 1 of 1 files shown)


src/sgame/sg_entities.cpp
```diff
@@ -80,45 +80,43 @@ angles and bad trails.
...
 		}
 
-		if ( i != MAX_GENTITIES )
+		// the first couple seconds of server time can involve a lot of
+		// freeing and allocating, so relax the replacement policy
+		if ( newEntity->freetime > level.startTime + 2000 && level.time - newEntity->freetime < 1000 )
 		{
-			break;
+			// we could look for the oldest one, but who cares, really
+			forcedEnt = newEntity;
```
slipher, line 101: Maybe assign to `forcedEnt` only if it is null? That way the lowest possible entity num is used, consistent with the non-forced case.

slipher (commented):
Ah, so the problem with the old code is where it says `if ( i != MAX_ENTITIES )` it should be `if ( i != ENTITYNUM_MAX_NORMAL )`, which would be the one-line fix. The proposed version seems more readable though, and avoids looping twice which is good.

src/sgame/sg_entities.cpp
peoro: I thought it was OK to avoid 3 more lines of code and not worry much about which entity to reuse in rare and extreme cases. But you're right really, it's cleaner to choose a slot with some logic rather than picking the last one.  
I would suggest to pick the slot that was freed the longest ago rather than the lowest one: this would reduce the amount of glitches players may see when slots are reused too quickly.

Let me know which policy you prefer and should I amend and force-push a single commit, or push this changes in a separate one?

By the way, a small note about this whole function: if it were ever to need further optimization, a much better approach would be to use a circular buffer of freed entity slots, instead of iterating through the entities array every time. I was unsure about changing this part of the code that much though.

slipher: > I would suggest to pick the slot that was freed the longest ago rather than the lowest one: this would reduce the amount of glitches players may see when slots are reused too quickly.

OK sounds good

peoro: Just pushed the change. This time I included some debug messages too (hidden behind `g_debugEntities`, like they did in `G_FreeEntity`).

src/sgame/sg_entities.cpp
```diff
@@ -80,45 +85,49 @@ angles and bad trails.
...
+			continue;
 		}
 
-		if ( i != MAX_GENTITIES )
+		// the first couple seconds of server time can involve a lot of
+		// freeing and allocating, so relax the replacement policy
+		if ( newEntity->freetime > level.startTime + 2000 && level.time - newEntity->freetime < 1000 )
 		{
-			break;
+			// we could look for the oldest one, but who cares, really
```
slipher, line 105: remove comment

PR #1369, opened 2021-04-13
Display emoticons in server name (<hostname/>)

slipher:


illwieckz:
I have noticed it didn't worked on map load script and scoreboard, does it fixes the issue for both?

slipher:
> 
> 
> I have noticed it didn't worked on map load script and scoreboard, does it fixes the issue for both?

Yes

illwieckz:
LGTM

PR #1370, opened 2021-04-14
Undefined behaviour fixes

necessarily-equal:


src/sgame/sg_combat.cpp
```diff
@@ -158,19 +158,25 @@ static const gentity_t *G_FindKillAssist( const gentity_t *self, const gentity_t
...
+	if ( killer - g_entities >= MAX_CLIENTS )
 	{
 		return nullptr;
 	}
 
-	// Require that the assist was for, at least, 25% of the damage or
+	// Require that the assist was for, at least, 25% of the health or
 	// as much damage as the killer did, whichever is lower
 	damage = self->entity->Get<HealthComponent>()->MaxHealth() / 4.0f;
 	if ( killer )
```
slipher, line 174: I don't see why there shouldn't be assists when the killer is a non-client (or the same player who died for that matter...). Why not just change `if ( killer )` to `if ( killer && killer->client )`

necessarily-equal: Fair enough, it makes sense to have kill assist messages when something is killed by, say, a turret. Fixed.

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -765,8 +765,14 @@ static void AddToAlOutputs( char *name )
 
 void CG_Rocket_SetAlOutputsOutput( const char*, int index )
 {
+	const char *device = "";
+	if (index >= -1 && index < rocketInfo.data.alOutputsCount )
+	{
+		device = rocketInfo.data.alOutputs[ index ];
```
slipher, line 771: I think `index >= -1` is not what you wanted

necessarily-equal: Indeed

slipher:
Thanks!

PR #1372, opened 2021-04-14
Fix animation load warning in 0.52 RC

slipher:
A new log message appeared: `Failed to load animation file models/weapons/lgun/lgun_view.iqm:reload`

But the lasgun cannot actually be reloaded. So, don't look for reload animations for guns without multiple clips.


necessarily-equal:
I thought @illwieckz pushed a fix for that one?

illwieckz:
I don't believe I took care of that myself.

illwieckz:
LGTM and I haven't seen any regression.

PR #1394, opened 2021-04-26
Rewrite the behavior tree loading

necessarily-equal:
Use more C++ and fix a double free (where an included tree would be freed when removed from the cached list of behaviour trees, in addition to be freed as usual).

The file surely could use even more C++, but it's something. And the part of the file that's still to rewrite seems to be working fine

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -28,7 +28,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "Entities.h"
 
 static botMemory_t g_botMind[MAX_CLIENTS];
-static AITreeList_t treeList;
+static AITreeList treeList;
 
 /*
 =======================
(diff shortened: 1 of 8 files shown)


necessarily-equal:
I think this is ready for review. Here's some context and questioning on the various commits:

* *Rewrite behaviour tree loading*: I turned the parsing functions into constructors because that's what they effectively are, but it's quite ugly IMO and gives constructors that can `throw`. I can split it back into a parsing function and a regular constructor if desired.
* *Fix a memory leak in behavior trees*: in a separate commit to avoid forgetting it in case the rest isn't desired.
* *Minor cleanups*: no comment.
* *Rename some types to remove the `_t`*: I'd like a comment on that. Do I remove the `_t` everywhere? I find it quite ugly, especially in constructors like `AIValue_t::AIValue_t`.
* *Rewrite behaviour tree expressions to `AIExpression_t`*: More C++ing.

slipher:
It's not a good idea to introduce exceptions because you can't use them in the available standalone WASM engines.

bmorel:
I see some comments in the PR. Could the PR be split to get the doc merged, and let the not-ready stuff as it is (or maybe close it, it has been months already, and while I'd love a BT rewrite, I guess this PR will not be merged anyway)?

Note that, I'm also quite tired by the actual BT parser and am planning to work on this, too. So, the most stuff from this PR merged, the better it would be, and it might even make me forget about working on that thing. FWIW, my plan is to use coco/R to generate a parser and a lexer, and to redo the whole thing. I don't have necessarily-equal's patience, and the BT system's lack of maintainability is a real problem when working on AI.

bmorel:
Are you still planning to get this merged? Otherwise it can probably be closed.

PR #1400, opened 2021-05-03
removal of unused cVars, fix of config inconsistencies and minor refactoring

bmorel:


bmorel:
companion PR of https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/53, since it removes unused cVar from the codebase.

src/shared/bg_parse.cpp
```diff
@@ -1182,11 +1182,12 @@ void BG_ParseClassAttributeFile( const char *filename, classAttributes_t *ca )
 			ca->jumpMagnitude = atof( token );
 			defined |= JUMPMAGNITUDE;
 		}
-		else if ( !Q_stricmp( token, "cost" ) )
+		// "cost" is legacy. It has been replaced by "price" to increase consistency in interfaces.
+		else if ( !Q_stricmp( token, "cost" ) || !Q_stricmp( token, "price" ) )
```
slipher, line 1186: If you must use the same word everywhere, why not use the one which is already in config files, so that changes to packages are not needed?

src/sgame/sg_bot_util.cpp
```diff
@@ -801,8 +794,8 @@ void BotGetTargetPos( botTarget_t target, vec3_t rVec )
 void BotTargetToRouteTarget( gentity_t *self, botTarget_t target, botRouteTarget_t *routeTarget )
 {
 	vec3_t mins, maxs;
-	int i;
 
+	routeTarget->type = botRouteTargetType_t::BOT_TARGET_STATIC;
```
slipher, line 798: The original way, without double-assignment, is better.

slipher (commented):
Good to see these unused cvars finally being gotten rid of.

Target branch should be changed to `master`

bmorel:
Le Wed, 05 May 2021 17:20:42 -0700,
slipher ***@***.***> a écrit :

> @slipher commented on this pull request.
> > @@ -1182,11 +1182,12 @@ void BG_ParseClassAttributeFile( const char
> > *filename, classAttributes_t *ca )
>  			ca->jumpMagnitude = atof( token );
>  			defined |= JUMPMAGNITUDE;
>  		}
> -		else if ( !Q_stricmp( token, "cost" ) )
> +		// "cost" is legacy. It has been replaced by "price"
> to increase consistency in interfaces.
> +		else if ( !Q_stricmp( token, "cost" )
> || !Q_stricmp( token, "price" ) )
> 
> If you must use the same word everywhere, why not use the one which
> is already in config files, so that changes to packages are not
> needed?

Currently, in config files, depending on which file (upgrade or
weapon IIRC, not looking at code rn) you look into, you can have either
"price" or "cost".
If config files where in the same package, I would have simply removed
"cost" and fixed the config in the same commit. I am still in favor to
do that, this is just a commit I had in tree since long (and since
anyway this PR requires acceptance of another one, that would not hurt).

illwieckz:
I modified target to `master` now that we have merged `0.52.0/sync` to `master`, maybe a rebase is needed.

src/shared/bg_parse.cpp
bmorel: As said through the mail reply (which is not nicely integrated):

I would prefer to get rid completely of one of the terms, but that requires patching 2 repos, which is why I kept compat.
There are probably other things like that, I'm almost sure I remember things having same effect but different wording between aliens and humans weapons, with no real reason.
I would like to at least get rid of those things at some point, but the price/cost is one which striked me more I suppose.

bmorel:
In my opinion, those changes should wait for UnvanquishedAssets/unvanquished_src.dpkdir and Unvanquished/Unvanquished to be merged, since it will be easier to keep things synced then.

Plus, I might reintroduce those cVars at some point, with an actual implementation behind, and introduce other changes which are all part of my work on AI improvements, so this probably just does not make sense to keep this PR opened.
I'm only commenting in case one disagree with those points and prefers to keep it open.

illwieckz:
@bmorel said this is obsoleted by work in progress branch:
https://github.com/bmorel/Unvanquished/commits/betterai

PR #1412, opened 2021-05-17
rewrite download-paks <3 Viech

illwieckz:
- make a difference between temporary folder and (persistent) cache folder
- detect version from repository (can still be overridden from command line option)
- harden it a bit (use `set -u`)
- also re-add @Kangz's mirror

Tested with `aria2c`, `curl` and `wget` with `torrent` and `http` methods on Linux and macOS with versions `0.52.0` (dpk), `0.51.1` (dpk, empty `md5sums`), `0.50.0` (pk3).

This is expected to make @Viech happy.

```
$ ./download-paks -h
download-paks: download Unvanquished dpk files

Usage: download-paks [option] [destination directory]

Options:
	--help             Print this help
	--no-check         Do not verify downloaded files
	--no-download      Do not download files (only perform file verification)
	--http             Download using HTTP (requires aria2c or curl or wget)
	--torrent          Download using BitTorrent (requires aria2c, default if aria2c is available)
	--tool=TOOL        Download using this tool (aria2c, curl or wget)
	--mirror=MIRROR    Download from this mirror (will not cycle through known mirrors)
	--version=VERSION  Download this version
	--cache=PATH       Cache downloaded files in this directory
	--temp=PATH        Download temporary files in this directory
	--list-mirrors     List known mirrors

Default destination directory is /home/illwieckz/.local/share/unvanquished/pkg
Default cache directory is /home/illwieckz/.cache/unvanquished/download-paks/pkg

Default version to download:
	0.52.0

Known mirror URLs:
	http://cdn.unvanquished.net
	http://unvcdn.viech.name
	http://cdn.illwieckz.net/unvanquished
	https://webseed.unv.kangz.net
```

### DIFF
--- a/download-paks
+++ b/download-paks
@@ -61,8 +61,8 @@ listCdn () {
 	http://unvcdn.viech.name
 	# illwieckz
 	http://cdn.illwieckz.net/unvanquished
-	# kangz
-	# http://webseed.unv.kangz.net/unv-launcher
+	# Kangz
+	https://webseed.unv.kangz.net
 	# Amanieu
 	# http://vanquished.zapto.org/~amanieu/unv-launcher
 	EOF
--- a/download-paks
+++ b/download-paks
@@ -130,7 +130,7 @@ colorPrintfLn () {
 listCdn () {
 	(sed -e 's/#.*//g;s/['$'\t'' ]*//g' | grep -Ev '^$') <<-EOF
 	# Official
-	http://cdn.unvanquished.net
+	https://cdn.unvanquished.net
 	# Viech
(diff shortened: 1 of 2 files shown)


slipher:
What's wrong with the hard-coded version number? Having the dependency on Git doesn't seem like an improvement.

Viech:
You can give the version as an argument now, like with the previous script. Defaulting the version via git sounds fine to me because you either use this script within your source directory to get assets or within a packaging script (where giving the version explicitly is preferable to having it hardcoded, imo). I don't know of any other use case.

PR #1419, opened 2021-06-01
cmake: bring back old CMake compatibility

illwieckz:
Counterpart of DaemonEngine/Daemon#476

Reverts f8c0e253ea5099acc73b42e583abbc66b8f49f4e

Now the CMake code in engine is testing for CMake version
to use newer features when available.


illwieckz:
There is solid solutions to get 3.12 CMake on older system, both Ubuntu repositories and universal installer scripts that are verified to work on Xenial (16.04) and Trusty (14.04) that are still under ESM. So we can avoid making the CMake files more complicated and keep CMake 3.12 as an hard requirement.

See details in DaemonEngine/Daemon#475.

PR #1422, opened 2021-06-06
Add appstream data

necessarily-equal:
Not sure what the way to go is with this. Should we merge this and add the copy the appstream data on flathub to get it released faster?

Also, should we prepare a blogpost maybe?


dist/net.unvanquished.Unvanquished.appstream.xml
```diff
@@ -0,0 +1,85 @@
...
+      <image type="thumbnail" width="768" height="432">https://unvanquished.net/wp-content/uploads/2021/05/20210504-072808-002.unvanquished-0.52-human-team-768x432.jpg</image>
+    </screenshot>
+  </screenshots>
+
+  <url type="homepage">https://unvanquished.net</url>
+  <url type="bugtracker">https://github.com/Unvanquished/Unvanquished/issues</url>
+  <url type="help">https://wiki.unvanquished.net/wiki/Main_Page</url>
+
+  <releases>
+    <release version="0.52.0" date="2021-05-14">
```
slipher, line 59: Have this be updated by the `update-version-number` script

dist/net.unvanquished.Unvanquished.appstream.xml
```diff
@@ -0,0 +1,85 @@
...
+      <description>
+        <p>
+          The first beta release, coming after years of continuous work
```
slipher, line 62: Does there really need to be a release-specific description or URL? It seems annoying to have to update them.

dist/net.unvanquished.Unvanquished.appstream.xml
necessarily-equal: The entries shouldn't be updated, we need to add new ones instead. I don't think this is really automatable though

dist/net.unvanquished.Unvanquished.appstream.xml
necessarily-equal: It needs not, but then you loose the version information in the store, and the mini changelog

necessarily-equal:
To give an idea how it may look, here's Shattered Pixel Dungeon in GNOME Software

image
image
image

illwieckz:
My main issue is that's basically a blog, I'm not sure the Unvanquished repository is the best place to host such thing. There is always a chicken and egg problem with such things, like the summary being written after the tag being set, but then the packaging script fetches the tag that does not contain the summary… And I don't want to wait on the blog post introduction to be ready to tag a release…

illwieckz:
> Also, should we prepare a blogpost maybe?

About announcing Flatpak, we may just announce it over social media for 0.52.0 flatpak.

On engine side we have a fix for the bad locale crash, a fix for the nvidia bug, on game side we have a fix for the trail bug, and a fix for a keybind mistake. We can then publish a point release (0.52.1) and announce Flatpak support in the same time.

dist/net.unvanquished.Unvanquished.appstream.xml
```diff
@@ -0,0 +1,85 @@
...
+    <memory>3072</memory> <!-- 3GiB RAM -->
+  </recommends>
+
+  <requires>
+	  <control>pointing</control>
+  </requires>
+
+  <requires>
+	  <control>keyboard</control>
+  </requires>
```
bilelmoussaoui, line 80:
```suggestion
  <requires>
	  <control>pointing</control>
	  <control>keyboard</control>
  </requires>
```

necessarily-equal: Would that mean both are needed?

bilelmoussaoui: see https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-requires-recommends-control

necessarily-equal: well, I've read it, three times, before making this PR, but apparently I got it wrong anyway ^^'

dist/net.unvanquished.Unvanquished.appstream.xml
Vanadiae: It's nice to have them because they act like a changelog but shown in stores. If Unvanquished had a changelog/NEWS file somewhere, then `appstream-util news-to-appdata` could have automated the process of updating the releases shown in this appstream file.

dist/net.unvanquished.Unvanquished.appstream.xml
```diff
@@ -0,0 +1,85 @@
...
+      Become a marine, handle your big weapon and dress that
+      impressive exoskeleton to save the human race like a boss. Or if
+      you prefer, be a monstrous beast and save those cute grangers
+      from those two-legged invaders who only know how to turn
+      beautiful organic equilibrium into lifeless machines and dead
+      concreted worlds.
+    </p>
+  </description>
+
+  <launchable type="desktop-id">net.unvanquished.Unvanquished.desktop</launchable>
```
AsciiWolf, line 22: Isn't upstream [desktop file](https://github.com/Unvanquished/Unvanquished/search?q=.desktop) name just unvanquished.desktop? Also, this desktop file is not present in the _dist_ directory, only in distro specific directories.

slipher: I believe the Debian package is dead, but @Viech still updates the Arch one.

necessarily-equal:
> I believe the Debian package is dead, but @Viech still updates the 
> Arch one.
> 

That's my understanding as well

illwieckz:
The Debian package scripts are expected to be up to date and working (see https://github.com/Unvanquished/Unvanquished/pull/1135 ). I fixed them. I not only fixed Debian stuff, I improved some of those files to be reusable by other distros.

For example, Arch package has yet to be updated to the new distro-independant running scripts, which are meant to ensure to reduce duplication of efforts and feature fragmentation.

illwieckz:
For the appstream file, if we gonna host it here, we can also do a `flatpak/0.52.0-1` tag and other things like that that can be tagged _after_ the `0.52.0` if things were missing at tagging time (like the blabla thing) (and then a `flatpak/0.52.0-2` if we need a flatpak-specific bugfix).

This method was used for Archlinux, for example we have [`v0.51.1`](https://github.com/Unvanquished/Unvanquished/releases/tag/v0.51.1) and [archlinux/0.51.1-1](https://github.com/Unvanquished/Unvanquished/releases/tag/archlinux%2F0.51.1-1).

Viech:
> This method was used for Archlinux, for example we have [`v0.51.1`](https://github.com/Unvanquished/Unvanquished/releases/tag/v0.51.1) and [`archlinux/0.51.1-1`](https://github.com/Unvanquished/Unvanquished/releases/tag/archlinux%2F0.51.1-1).

That was my stupid way of hosting distro-specific files on the game's repository though. I don't want to do these tags in the future. In fact my current `PKGBUILD`s already use the `unvanquished/` tags despite still using the distribution files in the `archlinux` folder:

```shell
_archive="archive/refs/tags/unvanquished/${pkgver}.tar.gz"
[…]
source=("unvanquished.install"
        "unvanquished_${pkgver}.tar.gz::https://github.com/Unvanquished/Unvanquished/${_archive}"
        "daemon_${pkgver}.tar.gz::https://github.com/DaemonEngine/Daemon/${_archive}"
        "breakpad_${pkgver}.tar.gz::https://github.com/DaemonEngine/breakpad/${_archive}"
        "crunch_${pkgver}.tar.gz::https://github.com/DaemonEngine/crunch/${_archive}"
        "recastnavigation_${pkgver}.tar.gz::https://github.com/DaemonEngine/recastnavigation/${_archive}"
        "naclsdk_${_naclsdk_ver}.tar.bz2::https://dl.unvanquished.net/deps/${_naclsdk_ver}.tar.bz2")
```

PR #1426, opened 2021-06-09
Ignore bots when checking team balance

slipher:
So if you have bots and team balance enforcement enabled, people can't all join the same team.


illwieckz:
LGTM

PR #1434, opened 2021-06-15
Fall back to old paths of missile configs

slipher:
88850e97 broke compatibility with the 0.52 asset packages. This is bad for development - see [Compatibility](https://wiki.unvanquished.net/wiki/Compatibility) "Develop Unvanquished on master, using latest-release DPKs".

### DIFF
--- a/src/shared/bg_misc.cpp
+++ b/src/shared/bg_misc.cpp
@@ -825,9 +825,17 @@ void BG_InitMissileAttributes()
 		ma->name   = md->name;
 		ma->number = md->number;
 
-		// for simplicity, read both from a single file
-		BG_ParseMissileAttributeFile( va( "configs/missiles/%s.attr.cfg", ma->name ), ma );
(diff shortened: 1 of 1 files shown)


illwieckz:
Can we just release 0.52.1 on Sunday 20 instead?

slipher:
Why not both? I can have a working dev environment for the next few days, and also release something :)

slipher:
This issue can be seen on @bmorel 's bot development server, which shows the value of keeping master compatible with the latest release's assets.

bmorel:
Or maybe have some hint at server hosters so that they know about the breakage and how to fix it?
Just checked, it's in the git log, but I admit that I mostly read it with `--oneline` and here nothing makes it possible to guess for the breakage (only «88850e97b split .missile.cfg files into .attr.cfg and .model.cfg files» shows there).

Sure it broke my server for some time, and confused me, but on the other hand I don't think there's many servers hosting frequently updated master (I'm not running master neither, since I bring my own changeset, though) so maybe it's not *that* important as long as there's something which makes it clear and obvious there's a breaking thing.

PR #1442, opened 2021-06-24
Disable autopause + stuff

slipher:


Gireen:
The time can only get as high as the timelimit. If one sets it to 0 then this can happen.

slipher:
Yeah, I guess this stuff is more of a problem of not resetting the game, than anything autopause specifically does. But from the choice of on or off, off still seems better for the moment because at least bots can fulfill a function of resetting the game every 15 minutes or so by winning...

src/sgame/sg_main.cpp
```diff
@@ -2741,12 +2741,6 @@ void G_RunFrame( int levelTime )
 				G_FreeEntity( ent );
 				continue;
 			}
-			else if ( ent->unlinkAfterEvent )
-			{
-				// items that will respawn will hide themselves after their pickup event
```
necessarily-equal, line 2746: I would keep this as this can prove useful for mods. In fact, I was looking for a way to do that for the zombie mod and the juggernaut mod (where bases would be replaced by loots here and there)

Otherwise, looks good to me

necessarily-equal (changes requested):

src/sgame/sg_main.cpp
slipher: This is a bit of dead code left over from Quake. You can hardly expect that it is going to work or do what you want.

necessarily-equal:
Bringing it back will need further work I guess. At least I'll know where to find it now

Le 27 juin 2021 16:43:06 GMT+02:00, slipher ***@***.***> a écrit :
***@***.*** commented on this pull request.

PR #1446, opened 2021-06-29
Further simplifications

necessarily-equal:
This builds upon and includes commits from https://github.com/Unvanquished/Unvanquished/pull/1366. You'll want to review that one first I think.


slipher:
LGTM

Issue #1429, opened 2021-06-09
The pertinence of unvanquished-migrate-profile.


illwieckz:
We have a [`unvanquished-migrate-profile`](https://github.com/Unvanquished/Unvanquished/blob/6ac003a0ceee239fa56a97b0b2de782b6f91be53/dist/scripts/unvanquished-migrate-profile) script. To my knowledge, it works (I was the last one making it working, *after* the last layout change).

This script is meant to migrate old Unvanquished profiles to actual one, on Linux.

At first we had a profile that was `${HOME}/.Unvanquished` (maybe with some Wolf:ET-ism).

Then we had a profile that was `${HOME}/.unvanquished` with some file reorganisation.

Now we have `${XDG_DATA_HOME:-${HOME}/.local/share}/unvanquished` (and standard XDG data folder).

Given we now follow the standard, it is not meant to be changed one more time (and if it happens one day, that would probably not be before _multiple decades_). So we can assume the current path is now stable.

I don't remember when the first move happened. If I'm right the second move happened for 0.51. At the time the engine even had code to do that second move (dropped prior 0.52).

The updater still has code to do the second move if I'm right.

We are now on the release _after_ the one doing the change.

The `unvanquished-migrate-profile` works, so the cost of distributing it when doing distro packages is low. But we can also question its pertinence.

slipher:
We removed homepath migration code from Daemon a while back: https://github.com/DaemonEngine/Daemon/commit/7c44effd6e3075a4874486d2966cabcf4a068938

illwieckz:
For our users playing released build it was only removed some days ago (and was only kept for one version, but that lasted 2 years and an half).

slipher:
This needs to be nuked in the [updater](https://github.com/Unvanquished/updater/blob/1b45e83df7d6b63e46cc65b8e67d3e2c061dca86/linux.cpp#L57) too :)

Issue #153, opened 2012-12-27
CG_RegisterClientModelname failed.


Gummar:
Same issue as: https://github.com/Unvanquished/Unvanquished/issues/122

It is happening with me now. I'm using the bot version, which the last commit was on 21/10/12.
http://puu.sh/1EJCi.

My "autogen.cfg" file inside the profile folder: http://pastebin.com/pyMuwB5r
I tested the Unvanquished in the "0.6.0-1-deb-294 (Compiled) folder" this version does work.

PR #1449, opened 2021-07-08
Remove dist/scripts/unvanquished-migrate-profile

necessarily-equal:
* It's not used anywhere
* Most users that had a config will likely either have run 0.51 or 0.52 once in the past 4-5 years
* If they don't, do we even know the dotfile format is still compatible?
* The code that was actually used to do a migration has been removed (https://github.com/DaemonEngine/Daemon/commit/7c44effd6e3075a4874486d2966cabcf4a068938)

I think this all makes a point that we don't want to start using this now.

Closes https://github.com/Unvanquished/Unvanquished/issues/1429.


illwieckz:
> * It's not used anywhere

it is used here: https://github.com/Unvanquished/Unvanquished/blob/master/dist/scripts/unvanquished-helper#L15

i.e. in `unvanquished`, `unvanquished-tty` and `unvanquished-server` wrappers for distro packages.

> * Most users that had a config will likely either have run 0.51 or 0.52 once in the past 4-5 years

Yes.

> * If they don't, do we even know the dotfile format is still compatible?

All the migrations paths were verified to be working before removing the similar code in engine. I would have fixed them otherwise because this is because this script exist I did not oppose to the removal of the code in engine, see below.

> * The code that was actually used to do a migration has been removed ([DaemonEngine/Daemon@7c44eff](https://github.com/DaemonEngine/Daemon/commit/7c44effd6e3075a4874486d2966cabcf4a068938))

The more recent migration path is still not removed from updater:

https://github.com/Unvanquished/updater/blob/a1b783da721a96ced2191ea12225062a39751671/unix.cpp#L42-L69

The only reason I did not oppose to that engine commit in #153 is that because we had already three migration codes, and as long the migration is done by the launcher (whatever the launcher), we don't need to do the migration in engine.

This is also related to the fact we can't support running the game outside of a launcher anyway (game is not integrated, people just mess with it, and on some OS it would just not run at all) so we must assume we have a launcher, the launcher being the updater, or the distro-script calling `unvanquished-migrate-profile`.

So this is because both migration code is not removed from updater and from `unvanquished-migrate-profile` expected to be used in distro launchers I did not oppose to the removal of the code in engine.

So this is not “_we can delete `unvanquished-migrate-profile` because similar code was removed in engine_”, it is “_we could delete similar code in engine because `unvanquished-migrate-profile` is kept_”.

> I think this all makes a point that we don't want to start using this now.

So the relevant point is about deciding to get rid of homepath migration mechanism, decision which was never made (the removal of the engine code is not a sign of such decision).

As long as the updater has similar code and this distro script works, I don't see a reason to remove this distro script.

This does not prevent us to decide to not care about such migration in non-updater launchers if it's for a better good, like if distro packagers have more trouble dealing with it than without.

Did you get special trouble from using this file while doing a distro package?

necessarily-equal:
> it is used here: https://github.com/Unvanquished/Unvanquished/blob/master/dist/scripts/unvanquished-helper#L15
>
> i.e. in unvanquished, unvanquished-tty and unvanquished-server wrappers for distro packages.

That is true, but *that* is not used anywhere. So in the end the migration script is not used anywhere.

> The only reason I did not oppose to that engine commit in #153 is that because we had already three migration codes, and as long the migration is done by the launcher (whatever the launcher), we don't need to do the migration in engine.

That's a fair point, so let's drop that argument.

> This is also related to the fact we can't support running the game outside of a launcher anyway (game is not integrated, people just mess with it, and on some OS it would just not run at all) so we must assume we have a launcher, the launcher being the updater, or the distro-script calling unvanquished-migrate-profile.

Why can't we? what do you mean with "people just mess with it"? I don't see how any distro script would help here.

About the "and on some OS it would just not run at all", that's unrelated to the distro scripts: the distro will make it work.

> As long as the updater has similar code and this distro script works, I don't see a reason to remove this distro script.

Well, I guess. It's more than I don't see a reason to use it.

If we focus on the specific case of this migration script, it had no purpose anyway for the two packages I made (NixOS and Flatpak), because nobody could have installed the game before. I guess it may make sense for the people on Debian that did install the package between 2012 and 2014. Or for people that downloaded the game manually more than 4 years ago and that do not want a fresh start. I would think we would may want a fresh start in such a case. Wouldn't we risk borked key bindings, for example? I would be quite surprised if a 5 year-old config still works flawlessly today.

If I look at the broader thing, I mostly see it does is provide a --gdb flag, and allows for some weird indirection where you could do a possibly confusing `unvanquished-server --client` or `unvanquished --tty` instead of  `unvanquished-tty`. I don't see how this level of indirection is better than having a slim `/usr/bin/unvanquished-server` that contains `/usr/lib/daemon -pakpath /usr/share/unvanquished/pkg`. But we should aim to do without.

The concept of wrapper script could be obsoleted with a cmake parameter or with some finding code in dæmon. It would allow not having a wrapper at all. Not having a wrapper is the ideal situation IMO. When we do that, `gdb unvanquished` is the only thing you need to launch the game in gdb; it's standard, and it works for other debugging tools too! (`strace`, intel vtune, etc.)

> This does not prevent us to decide to not care about such migration in non-updater launchers if it's for a better good, like if distro packagers have more trouble dealing with it than without.
>
> Did you get special trouble from using this file while doing a distro package?

In the current state you need to notice that there are scripts you need to use, try to get how they are structured, add 10 lines of code to put each file in the right place and then write a custom file in /etc to get it to work. This is quite a bit of work for something that seems to be lacking a purpose, so people will probably not use it. (Also the file doesn't even belong in /etc in the first place. I'm sure putting a file in /etc to explain to the package where its own files are would not have passed the code review for nixpkgs).

In the end, understanding how it works and ensuring it works correctly is certainly more trouble than doing the same as the arch package and having it work right out of the box. I guess maybe having a nice CMake install target would avoid that. But I fail to see how these scripts can help improve the experience for either users, packagers, or upstream maintainers.

illwieckz:
> Why can't we? what do you mean with "people just mess with it"? I don't see how any distro script would help here.

I talk about people using the unizip and then extract their OS-specific folder and get a subfolder created in their back by their file-manager, and attempting to fix it by doing weird things not following the readme instructions. Or just getting insulted by their OS (macOS). So, because we can't support this way of running the game, we don't have to care about this and _distro script does not have to help_. :-)

The point is just, since game is expected to be launched with a launcher in all case (being our updater or a distro script), we don't have to care about the case where an user does not use a launcher, hence we don't need that much migration path in engine. We don't have to care that much about non-supported usage.

> Wouldn't we risk borked key bindings, for example? I would be quite surprised if a 5 year-old config still works flawlessly today.

We make special effort to add a migration code in the engine, for that special case of bindings.

> The concept of wrapper script could be obsoleted with a cmake parameter or with some finding code in dæmon. It would allow not having a wrapper at all.

Yes, this can be a better route. Better decide to drop migration before since there would not migration script anyway.

> But I fail to see how these scripts can help improve the experience for either users, packagers, or upstream maintainers.

It improves compared to what was there before: you just has to copy paste scripts without to modify them, you only have to edit a config file with distro-specific paths. Reminder: at this point NixOS is not on topic, because the way it works may be as much different as between Linux and macOS for example. Those scripts were thought for things like Debian, Ubuntu, Arch, Slackware, Fedora, Mageia, SUSE… Not for things like snap, flatpak and nix.

necessarily-equal:
¯\_(ツ)_/¯

PR #1452, opened 2021-07-19
minor changes to make level_locals_s more constrained

bmorel:
Set several variables related to time as private and give them proper access functions, in the hope of making it easier to add time-related features.
Removed some never read code.
Fixed some -Wpadding warnings.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -2141,15 +2141,15 @@ bool G_admin_time( gentity_t *ent )
 	case 1:
 		Com_GMTime( &qt );
 
-		gameMinutes = level.matchTime()/1000 / 60;
-		gameSeconds = level.matchTime()/1000 % 60;
+		gameMinutes = level.ageMinutes();
+		gameSeconds = level.ageSeconds() % 60;
 
-		timelimit = level.timelimit * 60000; // timelimit is in minutes
+		timelimit = level.timeLimitMin() * 60;
 
-		if(level.matchTime() < timelimit)
+		if(level.ageSeconds() < timelimit)
 		{
-			remainingMinutes = (timelimit - level.matchTime())/1000 / 60;
-			remainingSeconds = (timelimit - level.matchTime())/1000 % 60 + 1;
+			remainingMinutes = (timelimit - level.ageSeconds()) / 60;
+			remainingSeconds = (timelimit - level.ageSeconds()) % 60;
 
 			ADMP( va( "%s %02i %02i %02i %02i %02i %i %02i", QQ( N_("^3time: ^*time is ^d$1$:$2$:$3$^* UTC; game time is ^d$4$:$5$^*, reaching time limit in ^d$6$:$7$^*") ),
 			  qt.tm_hour, qt.tm_min, qt.tm_sec, gameMinutes, gameSeconds, remainingMinutes, remainingSeconds ) );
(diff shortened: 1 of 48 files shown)


Gireen (commented):
TeamLocation is not useless

bmorel:
> TeamLocation is not useless

What is it for, then? And, I kept this one separated from previous commit because I though it might have some actual usage, but that it's buggy. The previous commit contains a fix: it currently does not handles UP_MEDIUMARMOUR, and when one reads the code, it really feels like a bug.
At first, I thought it was the scoreboard update or something like that, but no. So I removed the code, to see what would break, started a game and... nothing broke?

In conclusion, I would really like to know which purpose this code I removed here serves, so that I can know if it is indeed buggy or not.

Gireen:
I think you got something wrong, the code you remove is to share the location of teammates.

Center Left
https://youtu.be/2doFwEKBcUQ?t=744

bmorel:
@Gireen : I removed the offending commit IIRC (was long ago), could you re-check that everything's fine?

slipher:
We discussed this kind of change on IRC and agreed that a way forward is to a few members into a sub-struct and make some things private within the substruct (assuming the things in the substruct are actually somehow related and interact with each other).

I don't think it is an improvement to make some members private within the huge global struct. Maybe it seems attractive now, since the number of functions which can set `time` or whatever becomes smaller. But taking this approach to its logical conclusion, most sgame functions would be members of level_locals_t, meaning they are accessible by any part of this code. So this approach doesn't actually improve encapsulation in the end.

bmorel:
Exact, I forgot it.
To be honest, this have not much practical interest, IIRC, except making it easier to follow the "thread" (as in cloth) of memory writes where reading the code.
I think it can help others, but honestly, I also think we have more useful things to discuss, so I'm ok to close it.

Just a side note though: on my bigger branch, I did that (moving only partial members of structs as private) on parts which affect a lot of my commits.
This made it easier to me to improve AI's individual behaviors, and the situation might be more complex than here. So if this have to be a rule, then I suppose I should avoid pushing by AI changes _before I rewrite the whole botMemory_t structure_. I think this is a bit extreme.

src/sgame/sg_main.cpp
```diff
@@ -2831,3 +2803,49 @@ void G_PrepareEntityNetCode() {
 		entity.PrepareNetCode();
 	});
 }
+
+bool level_locals_s::warmingUp( void ) const
+{
+  return g_doWarmup.integer && warmupTime > m_time;
```
necessarily-equal, line 2809: The equivalent code would be `&& warmupTime > m_time + 1`, I guess

src/sgame/sg_admin.cpp
```diff
@@ -2141,15 +2141,15 @@ bool G_admin_time( gentity_t *ent )
 	case 1:
 		Com_GMTime( &qt );
 
-		gameMinutes = level.matchTime/1000 / 60;
-		gameSeconds = level.matchTime/1000 % 60;
+		gameMinutes = level.ageMinutes();
+		gameSeconds = level.ageSeconds() % 60;
 
-		timelimit = level.timelimit * 60000; // timelimit is in minutes
+		timelimit = level.timeLimitMin() * 60;
```
necessarily-equal, line 2147:
```suggestion
		timelimit = level.timeLimitMin() * 60; // 60 minutes
```

src/sgame/sg_struct.h
```diff
@@ -645,11 +645,29 @@ struct level_locals_s
...
+
+	int time( void ) const { return m_time; }
+	void updateTime( int levelTime ){ m_time = levelTime; m_matchTime = levelTime - m_startTime; }
+	int startTime( void ) const { return m_startTime; }
+	int matchTime( void ) const { return m_matchTime; }
+
+	int lastFrameDelay( void ) const { return m_time - previousTime; }
+	int ageMinutes( void ) const { return ( m_time - m_startTime ) / 60000; }
+	int ageSeconds( void ) const { return ( m_time - m_startTime ) / 1000; }
+	bool warmingUp( void ) const;
```
necessarily-equal, line 682: +1 for making this a function, this can reduce code duplication here and there. For the record, I think I had to duplicate the condition for the juggernaut mod

src/sgame/sg_spawn.cpp
```diff
@@ -1081,14 +1081,9 @@ void SP_worldspawn()
 	// see if we want a warmup time
 	trap_SetConfigstring( CS_WARMUP, "-1" );
 
-	if ( g_doWarmup.integer )
-	{
-		level.warmupTime = level.matchTime + ( g_warmup.integer * 1000 );
-		trap_SetConfigstring( CS_WARMUP, va( "%i", level.warmupTime ) );
-		G_LogPrintf( "Warmup: %i\n", g_warmup.integer );
-	}
+	level.initWarmUp();
```
necessarily-equal, line 1084: not fond of that one from a style point of view, but shrug

necessarily-equal (changes requested):
I'm in favor of that kind of PR. I've left some comments when it changed the current behavior though.

src/sgame/sg_struct.h
slipher: That kind of thing should be a function e.g. `int G_MatchTimeSeconds()` for stylistic consistency with existing code.

slipher:
Closed by author request.

PR #1453, opened 2021-07-19
Const fix

bmorel:
mostly add the const keyword in few places + minor cleanup.


necessarily-equal (approved):
:+1:

slipher:
Good cleanup, thanks

PR #1454, opened 2021-07-20
Fix medikit healing shader for real this time

necessarily-equal:


necessarily-equal:
Proposed testing protocol:
1. join humans
2. get hurt and use medikit
3. notice the medikit healing shader on the bottom left
4. get hurt again and go back to a medistation
5. notice the regular healing shader on the bottom left
6. join aliens
7. walk on creep
8. notice the regular healing shader on the bottom left
9. give yourself momentum and build a booster
10. notice the 3-cross healing shader on the bottom left

PR #1456, opened 2021-07-21
refactoring

bmorel:


src/sgame/sg_bot_util.cpp
```diff
@@ -1649,6 +1650,19 @@ void BotFireWeaponAI( gentity_t *self )
 	}
 }
 
+void BotChangeClass( gentity_t *self, class_t newClass, vec3_t newOrigin )
```
illwieckz, line 1653: maybe `inline`?

necessarily-equal (dismissed):

src/sgame/sg_bot_util.cpp
slipher: It never makes sense to use `inline` in a non-header

src/sgame/sg_bot_util.cpp
```diff
@@ -1844,39 +1858,15 @@ void BotBuyUpgrade( gentity_t *self, upgrade_t upgrade )
...
-			if ( !G_RoomForClassChange( self, PCL_HUMAN_LIGHT, newOrigin ) )
-			{
-				return;
-			}
-			VectorCopy( newOrigin, self->client->ps.origin );
-			self->client->ps.stats[ STAT_CLASS ] = PCL_HUMAN_LIGHT;
-			self->client->pers.classSelection = PCL_HUMAN_LIGHT;
-			BotSetNavmesh( self, PCL_HUMAN_LIGHT );
-			self->client->ps.eFlags ^= EF_TELEPORT_BIT;
+			BotChangeClass( self, PCL_HUMAN_LIGHT, newOrigin );
```
slipher, line 1861: The new code is not equivalent. It does not return from `BotBuyUpgrade` when there is no room.

src/sgame/sg_bot_util.cpp
```diff
@@ -1578,7 +1579,7 @@ void BotFireWeaponAI( gentity_t *self )
 		case WP_ALEVEL0:
 			break; //auto hit
 		case WP_ALEVEL1:
-			if ( distance < LEVEL1_CLAW_RANGE )
+			if ( distance < LEVEL1_CLAW_RANGE && readyFire )
```
slipher, line 1582: Why is this used only for level1? Also as a non-refactoring change it should be in a separate commit. `git add -p` is good.

src/shared/bg_public.h
```diff
@@ -131,6 +131,8 @@ struct playerState_t
 	int           clips; // clips held
 	int           tauntTimer; // don't allow another taunt until this runs out
 	int           misc[ MAX_MISC ]; // misc data
+
+	bool IsWeaponReady( void ) const;
```
slipher, line 135: Do not put `void` in the parentheses - it is meaningless in C++

src/sgame/sg_cmds.cpp
```diff
@@ -228,10 +227,10 @@ int G_ClientNumberFromString( const char *s, char *err, int len )
...
 		{
-			if ( p )
+			if ( err )
 			{
-				l = Q_snprintf( p, l2, "%-2d — %s^*", i, cl->pers.netname );
-				p += l;
+				l = Q_snprintf( err, l2, "%-2d — %s^*", i, cl->pers.netname );
+				err += l;
 				l2 -= l;
 			}
```
slipher, line 236: Original value of `err` is used at the end of the function

src/sgame/sg_bot_util.cpp
bmorel: so I guess it's fine as it is?

slipher: fine as it is

src/sgame/sg_admin.cpp
```diff
@@ -80,6 +80,35 @@ struct g_admin_command_t
...
+	char                 name[ MAX_NAME_LENGTH ];
+	char                 flags[ MAX_ADMIN_FLAGS ];
+	char                 pubkey[ RSA_STRING_LENGTH ];
+	char                 msg[ RSA_STRING_LENGTH ];
+	char                 msg2[ RSA_STRING_LENGTH ];
+	qtime_t              lastSeen;
+	int                  counter;
+};
+
+int G_admin_joindelay( g_admin_spec_t const* ptr )
```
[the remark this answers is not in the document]
bmorel: tbh, I do not really understand how the delay stuff works exactly, so it's a bit hard to document. In this case, it's only a setter, but "C-style", that is, the struct is defined in the cpp file, thus it's definition is not accessible from outside. Instead, functions which only exposes what is really used from the outside are exposed. This is comparable to a "private" in C++-style code.
The name might be badly chosen thought. It's only "inspired" from what I suppose the member accessed really is (`expires`) from what I understood of context.

This commit is really about having everything which does the admin auth stuff in a single place instead of spread into 3 cpp files.

src/shared/bg_public.h
bmorel: It's just an (old, dating to my pre-c++ times indeed) habit, which also makes it easier (for me at least) to distinguish between calls and declarations.
I will remove this since that's a problem.

src/sgame/sg_cmds.cpp
```diff
@@ -952,16 +951,16 @@ void Cmd_Team_f( gentity_t *ent )
...
 			trap_SendServerCommand( ent - g_entities,
 			                        "print_tr \"" N_("You cannot join a team until the next game.") "\"" );
 			return;
 		}
 
-		if ( specOnly->expires - t > 0 )
+		if ( G_admin_joindelay( specOnly ) - t > 0 )
 		{
-			int remaining = specOnly->expires - t;
+			int remaining = G_admin_joindelay( specOnly );
```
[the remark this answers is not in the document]
bmorel: unlikely intended. Fixed.

src/sgame/sg_cmds.cpp
bmorel: Exact. I did not understood that, in fact, here an error messaged is prepared in case more than one approximative matches are found, which would then be overwritten if no match at all can be.

src/sgame/sg_bot_util.cpp
```diff
@@ -1647,6 +1648,20 @@ void BotFireWeaponAI( gentity_t *self )
 	}
 }
 
+bool BotChangeClass( gentity_t *self, class_t newClass, vec3_t newOrigin )
+{
+	if ( !G_RoomForClassChange( self, newClass, newOrigin ) )
+	{
+		return true;
```
necessarily-equal, line 1655: probably should return `false` here.

```suggestion
		return false;
```

The expectation is `if (BotChangeClass(…)) { /* worked */ }`

src/sgame/sg_bot_util.cpp
```diff
@@ -1647,6 +1648,20 @@ void BotFireWeaponAI( gentity_t *self )
...
+	}
+	VectorCopy( newOrigin, self->client->ps.origin );
+	self->client->ps.stats[ STAT_CLASS ] = newClass;
+	self->client->pers.classSelection = newClass;
+	BotSetNavmesh( self, newClass );
+	self->client->ps.eFlags ^= EF_TELEPORT_BIT;
+	return false;
```
necessarily-equal, line 1662:
```suggestion
	return true;
```

src/sgame/sg_bot_util.cpp
```diff
@@ -1800,93 +1815,75 @@ void BotBuyWeapon( gentity_t *self, weapon_t weapon )
...
-			self->client->pers.classSelection = PCL_HUMAN_MEDIUM;
-			BotSetNavmesh( self, PCL_HUMAN_MEDIUM );
-			self->client->ps.eFlags ^= EF_TELEPORT_BIT;
-		}
-		else if ( upgrade == UP_BATTLESUIT )
-		{
-			if ( !G_RoomForClassChange( self, PCL_HUMAN_BSUIT, newOrigin ) )
+			break;
+		case UP_BATTLESUIT:
+			if ( BotChangeClass( self, PCL_HUMAN_BSUIT, newOrigin ) )
```
necessarily-equal, line 1875:
```suggestion
			if ( !BotChangeClass( self, PCL_HUMAN_BSUIT, newOrigin ) )
```

src/sgame/sg_bot_util.cpp
```diff
@@ -1800,93 +1815,75 @@ void BotBuyWeapon( gentity_t *self, weapon_t weapon )
...
-			self->client->pers.classSelection = PCL_HUMAN_LIGHT;
-			BotSetNavmesh( self, PCL_HUMAN_LIGHT );
-			self->client->ps.eFlags ^= EF_TELEPORT_BIT;
-		}
-		else if ( upgrade == UP_MEDIUMARMOUR )
-		{
-			if ( !G_RoomForClassChange( self, PCL_HUMAN_MEDIUM, newOrigin ) )
+			break;
+		case UP_MEDIUMARMOUR:
+			if ( BotChangeClass( self, PCL_HUMAN_MEDIUM, newOrigin ) )
```
necessarily-equal, line 1869:
```suggestion
			if ( !BotChangeClass( self, PCL_HUMAN_MEDIUM, newOrigin ) )
```

src/sgame/sg_bot_util.cpp
```diff
@@ -1800,93 +1815,75 @@ void BotBuyWeapon( gentity_t *self, weapon_t weapon )
...
+	}
 
-		if ( upgrade == UP_LIGHTARMOUR )
-		{
-			if ( !G_RoomForClassChange( self, PCL_HUMAN_LIGHT, newOrigin ) )
+	vec3_t newOrigin;
+	switch( upgrade )
+	{
+		case UP_LIGHTARMOUR:
+			if ( BotChangeClass( self, PCL_HUMAN_LIGHT, newOrigin ) )
```
necessarily-equal, line 1863:
```suggestion
			if ( !BotChangeClass( self, PCL_HUMAN_LIGHT, newOrigin ) )
```

necessarily-equal (changes requested):

src/sgame/sg_bot_util.cpp
```diff
@@ -1647,6 +1648,20 @@ void BotFireWeaponAI( gentity_t *self )
 	}
 }
 
+bool BotChangeClass( gentity_t *self, class_t newClass, vec3_t newOrigin )
```
necessarily-equal, line 1651:
```suggestion
static bool BotChangeClass( gentity_t *self, class_t newClass, vec3_t newOrigin )
```
The function is used only in this file

src/sgame/sg_bot_util.cpp
bmorel: If there is not enough room to change class, return that there's an error. This allows to test directly for `error` instead of checking for `!error`, thus:

```
if ( call_1() ) {
  //error handling
  return;
}
//no problem, continue
if ( call_2() ) {
  //error handling
  return;
}
//no problem, continue
```

instead of:

```
if ( call_1() ) {
  //no problem
  if ( call_2() ) {
    //no problem
  }
  return;
}
//error handling
```

or
```
if ( !call_1() ) {
  //error handling
  return;
}
//no problem, continue
if ( !call_2() ) {
  //error handling
  return;
}
//no problem, continue
```

src/sgame/sg_bot_util.cpp
necessarily-equal: still not `static`?

necessarily-equal (changes requested):

necessarily-equal (approved):

PR #1458, opened 2021-07-23
Cleanups

necessarily-equal:
This PR needs https://github.com/DaemonEngine/Daemon/pull/516 to be merged first, and submodules synced.

### DIFF
--- a/src/sgame/sg_cm_world.cpp
+++ b/src/sgame/sg_cm_world.cpp
@@ -891,9 +891,9 @@ void G_CM_Trace( trace_t *results, const vec3_t start, const vec3_t mins2, const
 		maxs2 = vec3_origin;
 	}
 
-    vec3_t mins, maxs;
-    VectorCopy(mins2, mins);
-    VectorCopy(maxs2, maxs);
+	vec3_t mins, maxs;
+	VectorCopy(mins2, mins);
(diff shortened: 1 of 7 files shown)


src/sgame/sg_cm_world.cpp
```diff
@@ -113,14 +113,8 @@ clipHandle_t G_CM_ClipHandleForEntity( const gentity_t *ent )
 		return CM_InlineModel( ent->s.modelindex );
 	}
 
-	if ( ent->r.svFlags & SVF_CAPSULE )
-	{
-		// create a temp capsule from bounding box sizes
-		return CM_TempBoxModel( ent->r.mins, ent->r.maxs, true );
-	}
-
 	// create a temp tree from bounding box sizes
```
slipher, line 116: I don't understand this comment but maybe it's better to get rid of it since it was on only one of the cases?

necessarily-equal: I've updated the comment to say "tree/capsule"

slipher:
Looks good, just waiting on the submodule thing.

slipher:
Hmm actually it's the other way around, this needs to be merged before Daemon#516. Apparently we did not need to wait on Daemon.

PR #1459, opened 2021-07-23
Round displayed human credits to the lower dozen (1299->1290)

necessarily-equal:
This should be a slight improvement in readability

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -885,7 +885,8 @@ public:
 		if ( credits != value )
 		{
 			credits = value;
-			SetText( va( "%d", credits ) );
+			// print it, rounded down to the dozen
+			SetText( va( "%d", (credits/10)*10 ) );
 		}
 	}
 


slipher:
Why? I don't get it.

necessarily-equal:
I think the last digit is just noise and doesn't bring any actual information. This makes it slightly easier to read in my opinion.

slipher:
Hmm well to me a stuck digit would look bad...

Gireen:
I think that rounding would be weird and create inconsistent results for players. Cause received credits would sometimes be more or less depending on rounding. Its also not more readable imo. 

Despite that, it doesn't hurt to have it optional, with a cvar to enable for whoever wants to use it.

necessarily-equal:
I don't think it would create any noticeable inconsistency: if you kill a dretch all by yourself, you still get +200creds. The rounding happens when you hurt some alien, in which case the number is often really hard to estimate anyway. That's why I don't think there is a sensible loss of information.

necessarily-equal:
I don't want to have an option for this kind of things, the solution is usually to just have a default that works™. If too many people think this is not a good change, let's close this for now. (I still like it though)

PR #1461, opened 2021-07-24
Fix Warn: Cvar cannot be registered twice

Gireen:
for custom admin commands

If a custom command was used the second time it would show this error message
Warn: Cvar arg_all cannot be registered twice 
for every one of this cvars

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -1583,17 +1583,13 @@ bool G_admin_cmd_check( gentity_t *ent )
 	if ( ( c = G_admin_command( command ) ) )
 	{
 		int j;
-		trap_Cvar_Register( nullptr, "arg_all", "", CVAR_TEMP | CVAR_ROM | CVAR_USER_CREATED );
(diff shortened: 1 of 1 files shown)


slipher:
What's a custom admin command?

Gireen:
Server admin can create there own commands in admin.dat
for example
```
[command]
command = fillbots
exec    = bots fill $arg_1
desc    = fillbots <number> 
flag    = fillbots
```

slipher:
The variable substitution stuff doesn't work for me. For example if I have `exec    = echo $arg_1$` it just prints `$arg_1$` verbatim, without substitution.

Gireen:
indeed. but arg_1 is set, cause the warning is displayed. it seems that cvars are not accessible there (anymore?).
despite that its still possible in executed files so this does work

game/admin.dat
```
[command]
command = fillbots
exec    = exec fillbots.cfg
desc    = fillbots <number> 
flag    = fillbots
```
config/fillbots.cfg
```
echo  we need $arg_1$ more bots
```

PR #1462, opened 2021-07-28
Add autocompletion of the command name for devteam

necessarily-equal:
https://github.com/Unvanquished/Unvanquished/pull/1235 missed this. A bit above the list this PR touched, there is a `// keep the list synchronized with the list in cg_consolecmds for completion.`


necessarily-equal:
This will only autocomplete the command name, yes. I don't know if there is already a function to complete teams, but at least the function will answer "/devteam [h|a]", and this should be documentation enough.

necessarily-equal:
Example test protocol:

1. launch
2. /dev<tab>
3. notice how `/devteam` is only completed on this branch

PR #1463, opened 2021-07-28
Prevent bot dretch from biting turrets when they shouldn't damage structures

necessarily-equal:


slipher:
The concept seems wrong to me. The bot attack cvar should mean that the AI doesn't seek out turrets and try to aim at them. Not that the gameplay code behaves differently.

necessarily-equal:
Agreed. On the other hand, the current behavior seems like a bug to me, 
as bots will actually destroy turrets. On a small map, they are even 
quite good at it.

(For the record, this patch is part of me trying to make a tutorial 
map, and the current dretch behavior basically breaks it. I can 
workaround it if there is a real need, but this PR seems the best way 
to resolve this to me.)

Gireen:
How about making a separate cvar for g_bot_damageStruct ?
Could be interesting as a damage modifier, that a tutorial map could have bots do different amounts of damage to buildings.

illwieckz:
I agree a separate cvar may be better.

slipher:
> On the other hand, the current behavior seems like a bug to me,
as bots will actually destroy turrets. On a small map, they are even
quite good at it.

Like they are running into turrets by accident while trying to do other stuff? Or deliberately seeking them out?

necessarily-equal:
> Like they are running into turrets by accident while trying to do other stuff? Or deliberately seeking them out?

They are essentially running into the turret and biting it incidentally to get the human snack on the other side. Since they sometimes get stuck running *into* the turret to get behind it, they easily grab 3 hits — half the turret health.

bmorel:
I also noticed that fact in my bot testings, where I usually do not want them to destroy buildings, so that I can observe their behaviors more easily.
Note that this problem does not only affect dretches, but all bots. It's more noticeable with dretches and (obviously) area damages (barbs, zap, flamer, luci, shotgun mostly) but it can also happen that bots miss shots and those lost shots can land on buildables, too.

Gireen:
Then how about g_bot_dretchDamageStruct ?

Gireen:
or maybe better move this directly into the health component
HealthComponent::HandleDamage
there happens godmode checks

necessarily-equal:
So, what I now have in mind is a `botAvoidStruct` cvar which should be nicer overall. I'm keeping this open to hopefully not forget to do it one day :)

Issue #1260, opened 2021-01-02
Corpse death animations need a different `deadMins` value


necessarily-equal:
In practice, for the armorless human, this means that the dead body will teleport a bit higher once replaced.

DolceTriade:
Yeah, we can have a config file with variable z offsets for the different death anims I guess... I wonder if we can re-export iqms with this set somehow if we parse the iqe file.

illwieckz:
is it a duplicate of:

- #1058

?

necessarily-equal:
#1058 was to be fixed by https://github.com/necessarily-equal/res-players_src.dpkdir/commit/7dc41cca04b8aef04d74c4d8b982f5dd5e613d70, but since https://github.com/Unvanquished/Unvanquished/commit/5de0b9289281793307e2bcc7aeb74779a866d678, something more elaborate is needed, to avoid (say) having one corpse over the ground and one under.

PR #1464, opened 2021-07-28
Model debugging fixes

slipher:


necessarily-equal:
> Bounding box of a corpse makes no sense since it has no gameplay functionality.

It's not quite true. It doesn't impact the gameplay, but it has visual implications. And it needs debugging! For example, see https://github.com/Unvanquished/Unvanquished/issues/1260.

(Since it's broken though, this isn't a regression or anything, and doesn't block this PR. But it would be nice to have it.)

illwieckz:
Would there be a similar fix for `misc_anim_model` (models not being baked in bsp)? The md3 ones work properly (see sandals in chasm), but the md5 or iqm ones seems to fail, there should be 4 decorative turrets in outside part of the stairs dome in parpax (`/setviewpos 125 -3366 1261 -35 25`) but they are not rendered.

slipher:
> It's not quite true. It doesn't impact the gameplay, but it has visual implications. And it needs debugging!

As I understand it, cg_drawBBOX is for debugging the bounding boxes of things that are involved in traces. AFAIK no tracing is done with corpses. So in principle it could be useful if there were a trace affecting how it melts into the ground or something, but currently there is nothing to show.

> Would there be a similar fix for misc_anim_model (models not being baked in bsp)? The md3 ones work properly (see sandals in chasm), but the md5 or iqm ones seems to fail, there should be 4 decorative turrets in outside part of the stairs dome in parpax

Parpax (2.6.1) doesn't seem to have any `misc_anim_model` - SP_gfx_animated_model is not called. This function is called as expected for the sandals in Chasm.

illwieckz:
> Parpax (2.6.1) doesn't seem to have any `misc_anim_model` - SP_gfx_animated_model is not called. This function is called as expected for the sandals in Chasm.

Ah, my bad, I experimented `misc_anim_model` but either reverted it, either never pushed it. I'll upload a build using `misc_anim_model` so we can test.

Now that I think about it, I'm less sure if md3 work with `misc_anim_model`, I expect it to be working but my memories are also mixed with other stuff.

slipher:
The misc_anim_model works on Chasm.

illwieckz:
@slipher: https://dl.illwieckz.net/b/unvanquished/bugs/iqm-map-animated-model/map-parpax_2.7.1-dirty.dpk

scene: `/setviewpos 125 -3366 1261 -35 25`

There should be decorative turrets on each four pods. Note: they are not flying anymore, they must be right on the pods.

slipher:
All right I added the line for the Parpax ones to work.

illwieckz:
Thanks!

I confirm it works, before:

[iqm map animated model](https://dl.illwieckz.net/b/unvanquished/bugs/iqm-map-animated-model/unvanquished_2021-08-02_052606_000.jpg)

after:

[iqm map animated model](https://dl.illwieckz.net/b/unvanquished/bugs/iqm-map-animated-model/unvanquished_2021-08-02_052832_000.jpg)

I don't know what happens with the `ET_CORPSE` thing, for the rest, that looks good to me.

PR #1468, opened 2021-07-30
Drop old things

necessarily-equal:
Comes along https://github.com/DaemonEngine/Daemon/pull/518 but doesn't depend on it.

### DIFF
--- a/src/cgame/cg_drawtools.cpp
+++ b/src/cgame/cg_drawtools.cpp
@@ -551,7 +551,7 @@ void CG_DrawRangeMarker( rangeMarker_t rmType, const vec3_t origin, float range,
 
 		pcsh = cgs.media.plainColorShader;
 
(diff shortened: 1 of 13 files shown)


slipher:
Inline linkage should never be used in a .cpp. So all of these `INLINE`s should be deleted rather than replaced with `inline`.

necessarily-equal:
I've dropped the `inline` in cpp files. I don't think they are very useful, even if I think they didn't hurt either.

src/sgame/sg_bot.h
```diff
@@ -48,12 +48,7 @@ struct enemyQueueElement_t
 	int        timeFound;
 };
 
-struct enemyQueue_t
-{
-	enemyQueueElement_t enemys[ MAX_ENEMY_QUEUE ];
-	int front;
-	int back;
-};
+using enemyQueue_t = Queue<enemyQueueElement_t, MAX_CLIENTS+1>;
```
slipher, line 51: compile error here

necessarily-equal:
whoops, I didn't mean to commit this.

Issue #1469, opened 2021-07-31
map loading screen: give user feedback of engine being compiling GLSL shaders at map load when using `r_lazyShaders 1`


illwieckz:
map loading screen: give user feedback of engine being compiling GLSL shaders at map load when using `r_lazyShaders 1`

illwieckz:
One thing that can confuse users is that the GLSL compilation happens after 100% is reached, so the game is stuck at 100% without any other feedback.

slipher:
I've been thinking we need this anyway (for shader compilation on engine startup). It takes ~60 seconds on my machine if using Intel drivers. Doing it on game load would be nice since it is easier to draw some text using the cgame.

necessarily-equal:
I think this has been fixed?

PR #1470, opened 2021-08-01
ui: attempt to re-enable the loading messages

illwieckz:
We have dead messages that are expected to be displayed when loading the game data. If I'm right it was not ported when switching from Q3 ui to libRocket. My main purpose to revive this is to be able to add a message saying the GLSL shaders are compiling, see:

- #1469

Though the messages were not up to actual gameplay (jetpackers can't camp in the air anymore, the batterypack is removed, etc.) so I have rewrote some sentences and added new ones, trying to do it in the same spirit.

Though, maybe we may want less-funny stuff (though the nostalgia is strong with this one), and more explicit sentences.

There was some code meant for multiple progress bar, like an overall progress bad and one for buildable models loading and one for player model loadings or things like that… like in Tremulous or in software installers from Windows 98 era… So I cleaned it up a bit.

Though, I have a problem, I don't get the text being displayed. I have failed to find why.


illwieckz:
@Gireen, would you notice what's my mistake?

The `CG_Rocket_DrawLoadingText()` function works, if I tell the game to call this function with `progress_value` element instead of `loadingText` one, it works.

But for some reason, the `CG_Rocket_DrawLoadingText()` is never called with `loadingText` element. By never called, I mean spamming some logs in that function never prints anything on console and even a `Sys::Error()` does not abort the game.

illwieckz:
I simply missed the comment saying `// THESE MUST BE ALPHABETIZED` 🤦‍♀️

src/cgame/cg_main.cpp
```diff
@@ -1409,6 +1394,11 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 	new(&cg) cg_t{};
 	memset( cg_entities, 0, sizeof( cg_entities ) );
 
+	cg.loadingFraction = 0.0f;
```
slipher, line 1401: This is redundant since they were just cleared with `new(&cg) cg_t{}`

src/cgame/cg_main.cpp
```diff
@@ -1337,6 +1319,9 @@ static void CG_RegisterClients()
 		CG_NewClientInfo( i );
```
slipher, line 1323: I think we eagerly load everything before this so it doesn't actually load anything new.

src/cgame/cg_main.cpp
```diff
@@ -777,74 +740,93 @@ enum {
...
-			Log::Notice("^4%%^5 Done loading everything after %is (%ims).",
-					(thisStepTime - startTime)/1000, (thisStepTime - startTime));
-			break;
-		default:
-			Log::Notice("^4%%^5 Done with Step %i after %is (%ims)… Starting Step %i",
-					step - 1, (thisStepTime - lastStepTime)/1000, (thisStepTime - lastStepTime), step );
-			break;
+	Q_strncpyz( cg.loadingText, label, sizeof( cg.loadingText ) );
+
+	Log::Notice( "Loading: %d%% %s", (int) ( 100 * cg.loadingFraction), cg.loadingText );
```
slipher, line 762: This seems too noisy for a Notice. Also `static_cast<int>`

src/cgame/cg_main.cpp
```diff
@@ -825,6 +825,10 @@ static void CG_UpdateLoadingStep( cgLoadingStep_t step )
 			CG_UpdateLoadingProgress( step, choose("Teleporting soldiers…", "Parenting dretches…", "Replicating alien DNA…", nullptr) );
 			break;
 
+		case LOAD_REMAINING:
+			CG_UpdateLoadingProgress( step, choose("Compiling GLSL shaders…", nullptr) );
```
slipher, line 829: Maybe some hint that this operation could be especially slow... like "Compiling GLSL shaders (please be patient!)"

src/cgame/cg_main.cpp
illwieckz: > This seems too noisy for a Notice.

Hmm, the purpose of funny messages is that they are expected to be read…

src/cgame/cg_main.cpp
illwieckz: Right…

src/cgame/cg_main.cpp
illwieckz: I did the `static_cast<int>` thing. For the Notice, if I set a Debug nothing is printed while I have a lot of loglevel set to debug (including `logs.logLevel.default`)…

slipher: To be read on the loading screen, not in the log file!

The level here is `logs.logLevel.cgame.default`

necessarily-equal: how about using `G_LogPrintf` to have it in the console only?

slipher: Console logs are for debugging. Funny messages don't help with that. The `Log::Notice` here should be removed

Issue #425, opened 2014-08-24
Prediction errors on edges of buildables


enneract:
Colliding with buildables often results in visible prediction errors. This is probably a consequence of using the bounding box sent by server (crapped through entityState_t's "solid" field at low precision) instead of the accurate values from config files.

slipher:
A curious feature of the bug is that it seemingly relates to closeness to the edge as a percentage of buildable size, rather than actually being on the edge. If I make a 10x scale medistation, then I get prediction errors without even being close to the edge of the bounding box. (You need to go all the way over the red part to fall down.)

![unvanquished_2021-07-17_224454_000](https://user-images.githubusercontent.com/7809431/126055608-8a73bbdd-61e0-406d-b09f-d4b8742e6a3c.jpg)

slipher:
Disregard previous comment. The bug with the giant medistation was caused by code that doesn't work with bboxes with maxs larger than 255.

Another thing I've found is that the bug cannot be reproduced via saveviewpos/setviewpos. You don't get the prediction errors after restoring the saved position. Maybe there is some sort of feedback loop with deltas to produce the bug.

illwieckz:
> You don't get the prediction errors after restoring the saved position. Maybe there is some sort of feedback loop with deltas to produce the bug.

I noticed that when restoring saved position, the spectator does not suffer from any move which seems to be some leftover for accumulated speed. Maybe the non-spectator also suffer from such accumulated speed leftover?

necessarily-equal:
@illwieckz if you want to test this, you may want to test https://github.com/necessarily-equal/Unvanquished/commit/15023cf9ef2640469111093004c5e03e004f2a17 that fixes the leftover speed. For some reason, because of a GitHub bug, I can't make a PR for it but well. You may also want to check @slipher's PR, #1471.

illwieckz:
@necessarily-equal does this link work for you:
https://github.com/Unvanquished/Unvanquished/compare/master...necessarily-equal:more-fixes

necessarily-equal:
Ah, yes it works. I think I missed the "s" at the end when copying the link. Though GitHub wasn't explicit about the error either.

slipher:
Fixed by #1471.

PR #1471, opened 2021-08-02
Fix some prediction bugs

### LINKED ISSUE #1439: Standing on the Armory lets the player bounce slightly up and down

### LINKED ISSUE #1467: Armory problems
Currently, the armory have several issues:

1. the UI is way too slow (which is currently being worked on by gireen I believe)
2. bindings allow to instantly change a full equipment set, which puts a big hole between normal players and very advanced players (and *in my opinion* this is a problem)
3. armory is usable at some distance. Menu will only pop at a *lesser* distance. This favors bind users, again.
4. bots always act as people which have everything bound (they instantly-change their whole equipment)

slipher:
Fixes #1439, #425

### DIFF
--- a/src/cgame/cg_predict.cpp
+++ b/src/cgame/cg_predict.cpp
@@ -158,15 +158,22 @@ static void CG_ClipMoveToEntities( const vec3_t start, const vec3_t mins,
 		}
 		else
 		{
-			// encoded bbox
-			x = ( ent->solid & 255 );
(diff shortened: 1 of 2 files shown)


necessarily-equal:
I've read it, and it seems fine enough. This should fix https://github.com/Unvanquished/Unvanquished/issues/1439 and #425 though, not #1467.

PR #1472, opened 2021-08-03
Center print size + small fixes

slipher:
A few small bug fixes; the most relevant one is implementing the size argument for center prints. Currently these are all rendered at the same small size, - something like 10 pt - which is barely noticeable.


src/sgame/sg_utils.cpp
```diff
@@ -295,11 +294,6 @@ void G_KillBox( gentity_t *ent )
 	{
 		hit = &g_entities[ touch[ i ] ];
 
-		if ( !hit->client )
-		{
-			continue;
```
illwieckz, line 300: is it possible to kill the world entity?

illwieckz: Ok, good.

src/sgame/sg_bot_ai.cpp
```diff
@@ -247,18 +247,19 @@ AINodeStatus_t BotSelectorNode( gentity_t *self, AIGenericNode_t *node )
 AINodeStatus_t BotSequenceNode( gentity_t *self, AIGenericNode_t *node )
 {
 	AINodeList_t *sequence = ( AINodeList_t * ) node;
-	int i = 0;
+	int start = 0;
```
bmorel, line 250: I don't get how this commit changes anything (except by having more explicit variables).
Before, 1st loop started with `i = numNodes - 1`, ended with ` i >= 0 `, and now it seems the same to me?

bmorel (commented):
lgtm

slipher:
Dropped the center print one since it's not read yet.

PR #1473, opened 2021-08-03
More fixes

necessarily-equal:


slipher:
LGTM
Not merging yet since I usually leave stuff up at least a day or two, in case someone else has a comment

illwieckz:
- _movement: drop speed to 0 when not moving_

This is greatly appreciated, this bug was already present in Tremulous 1.1 and in fact can be reproduced in Quake Ⅲ Arena itself.

LGTM as well.

PR #1474, opened 2021-08-10
sgame: fix various env_afx_gravity and func_bobbing bugs <3 Gireen

illwieckz:
sgame: remove solid flag on env_afx_gravity / trigger_gravity <3 @Gireen

The`env_afx_gravity` brushes are likely to use the `caulk` texture which is a solid texture, meaning the whole content of the `trigger_gravity` entity is solid, preventing buildables to spawn and players to move, being literally stuck in a wall.

[non solid gravity](https://dl.illwieckz.net/b/unvanquished/bugs/non-solid-gravity/20210810-125112-000.netradiant-mission-one-redux.png)

This patch always removes the solid content parm on such `env_afx_gravity`. It was verified to workaround properly the issue on map using obsolete `trigger_gravity` or recommended `env_afx_gravity`.

I tested it with maps:

- mission one redux (`trigger_gravity` replaced by `env_afx_gravity` using [Esquirel](https://github.com/DaemonEngine/Urcheon)) \
  https://betaserv.tk/pk3/pk3/2055
- mission one (`trigger_gravity`) \
  https://betaserv.tk/pk3/map/913
- hq (`trigger_gravity`) \
  https://betaserv.tk/pk3/pk3/557

[non solid gravity](https://dl.illwieckz.net/b/unvanquished/bugs/non-solid-gravity/unvanquished_2021-08-10_123950_000.jpg)

[non solid gravity](https://dl.illwieckz.net/b/unvanquished/bugs/non-solid-gravity/unvanquished_2021-08-10_124021_000.jpg)

[non solid gravity](https://dl.illwieckz.net/b/unvanquished/bugs/non-solid-gravity/unvanquished_2021-08-10_124117_000.jpg)

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -2510,11 +2510,12 @@ void SP_func_bobbing( gentity_t *self )
 
 	trap_SetBrushModel( self, self->model );
 	InitMover( self );
+	reset_moverspeed( self, 4 );
 
 	VectorCopy( self->s.origin, self->s.pos.trBase );
 	VectorCopy( self->s.origin, self->r.currentOrigin );
 
-	self->s.pos.trDuration = self->config.speed * 1000;
+	self->s.pos.trDuration = self->speed * 1000;
 	self->s.pos.trTime = self->s.pos.trDuration * phase;
 	self->s.pos.trType = trType_t::TR_SINE;
 
--- a/src/sgame/sg_spawn_afx.cpp
+++ b/src/sgame/sg_spawn_afx.cpp
@@ -283,6 +283,9 @@ void SP_env_afx_gravity( gentity_t *self )
 	self->act = env_afx_toggle;
 	self->reset = env_afx_gravity_reset;
 
+	// Remove CONTENTS_SOLID flag.
+	self->r.contents &= ~CONTENTS_SOLID;
+
 	InitEnvAFXEntity( self, true );
 }
 


illwieckz:
I add a commit to attempt to fix #1475. When `func_bobbing` `speed` property is set to `0`, I set it to `1`, because then it is used to compute a duration and then scale the model, for what I understand from the code.

I don't know if that field is always set to `0` when undefined, but at least, testing for this field to be equal to `0` and then replace it with `1` seems to fix the issue without adding regressions.

On `ethereal` map, the turret in the corridor is not destroyed by an invisible wall and projectiles are not blocked by invisible wall and player can walk it:

[broken func bobbing](https://dl.illwieckz.net/b/unvanquished/bugs/broken-func-bobbing/20210811-082521-000.unvanquished-ethereal-func-bobbing.jpg)

The moving parts are moving as expected:

[broken func bobbing](https://dl.illwieckz.net/b/unvanquished/bugs/broken-func-bobbing/20210811-082554-000.unvanquished-ethereal-func-bobbing.jpg)

And it looks like other map `mission-one-redux` that worked before is still working properly:

[broken func bobbing](https://dl.illwieckz.net/b/unvanquished/bugs/broken-func-bobbing/20210811-082656-000.unvanquished-mission-one-redux-func-bobbing.jpg)

src/sgame/sg_spawn_mover.cpp
```diff
@@ -2514,6 +2514,11 @@ void SP_func_bobbing( gentity_t *self )
 	VectorCopy( self->s.origin, self->s.pos.trBase );
 	VectorCopy( self->s.origin, self->r.currentOrigin );
 
+	if ( self->config.speed == 0 )
```
Gireen, line 2517: the default value should be 4 if not set (says netradiant)
So the question is why its 0 and where the default settings are set.

A bit above damage seems to be initiated, that could be the method needed here.
```
G_ResetIntField(&self->damage, true, self->config.damage, self->eclass->config.damage, 2);
```

illwieckz: If I do instead:

```c++
G_ResetFloatField(&self->speed, true, self->config.speed, self->eclass->config.speed, 4);
```

It fixes nothing.

illwieckz:
In `SP_func_bobbing`, if I do:

```c++
	G_ResetFloatField(&self->speed, true, self->config.speed,
		self->eclass->config.speed, 4);
```

It fixes nothing.

If I do:

```c++
	if ( self->config.speed == 0 )
	{
		self->reset( self );
	}
```

Which is calling:

```c++
	reset_moverspeed( self, 4 );
```

Which is calling (with `fallbackSpeed` being `4`):

```c++
	G_ResetFloatField(&self->speed, true, self->config.speed,
		self->eclass->config.speed, fallbackSpeed);
```

It fixes nothing.

But if I do:

```c++
	if ( self->config.speed == 0 )
	{
		self->config.speed = 4;
	}
```

The bug is fixed.

So, maybe we have a bug in `G_ResetFloatField` or we don't set properly a value used there?

----

Note: when the bug happens, the log is full of:

```
Warn: CM_AddFacetBevels... invalid bevel 
```

Gireen:
```
self->s.pos.trDuration = self->config.speed * 1000;
```
is wrong

it has to be `self->speed`
`self->config` contains all settings provided by the map and if there is none its 0

illwieckz:
> ```
> self->s.pos.trDuration = self->config.speed * 1000;
> ```
> 
> is wrong
> 
> it has to be `self->speed`
> `self->config` contains all settings provided by the map and if there is none its 0

Ah, nice find!

When I fix that, I can call `reset_moverspeed( self, 4 )` in a similar way to what other functions do, and it fix the issues.

The usage of `self->config.speed` was probably a workaround for not calling `reset_moverspeed` which is expected to set `self->speed` to `self->config.speed` if not null.

illwieckz:
I consider this ready to merge.

Gireen (approved):
lgtm

PR #1480, opened 2021-08-18
Betterai

bmorel:
this branch is a WiP, submitted to push due to real life events. It can be (maybe mostly) seen in effect on my own servers.


src/sgame/sg_bot_nav.cpp
```diff
@@ -546,6 +546,8 @@ bool BotAvoidObstacles( gentity_t *self, vec3_t dir )
 
 	if ( blocker )
 	{
+
+		self->botMind->cmdBuffer.upmove = -127;
```
slipher, line 558: I don't understand. When would this ever be a good idea?

src/sgame/sg_bot.cpp
```diff
@@ -393,6 +393,18 @@ void G_BotThink( gentity_t *self )
 	//reset command buffer
 	usercmdClearButtons( botCmdBuffer->buttons );
 
+	if ( BG_Class( self->client->ps.stats[STAT_CLASS] )->abilities & SCA_WALLCLIMBER )
+	{
+		//wall walkers really want to wall walk when possible. Naive, ugly, but if it could work... worth it.
```
slipher, line 403: > wall walkers really want to wall walk when possible

Far from true. In my experience having wallwalk on when you are trying to move around on the ground interferes with locomotion a lot. Like when you just want to jump you will stick to things when landing. Or you start to go up walls when trying to walk near them.

slipher (commented):
Looked at the 4 commits mentioned on IRC. They all could use more description in the commit message about what you are trying to do.

- "privateness improvements" basically OK
- "let bots use ladders (requires connection though)" not sure. I couldn't find what the idea is since there is so much refactoring stuff in there too.
- "implement some basic wallwalking (also requires manual connections)" left comment
- "duck when stuck" left comment

src/sgame/sg_bot_nav.cpp
bmorel: no and for 2 reasons:

* it's a bad idea to do that unconditionally, but this is something which required much work
* the most important reason, and the why I wanted your input: I remember reading code saying it was a bad idea as it can break prediction, or something like that. Also, the value is clearly something I have copied without understanding it.

I have since fixed the 1st reason, but I still don't like inserting a "random" value in there.

src/sgame/sg_bot.cpp
bmorel: The point is, this code is for bots, they won't loose orientation as humans when the world rotate like crazy. Unlike human, bots require navmesh to be generated to wallwalk, or to have manual connections added between navmeshes.
And in combat, they simply ignore navmesh stuff and thus would not be able to use the wallwalking ability.

The way I did it (which is, as my comment says, naive and ugly) allows AI to wallwalk in combat, even if this will be rare for the reasons you mentioned: it will only apply when bots pounce or jump but miss their target and end up on a wall.
I have not watched lot of games since I implemented this, and so I only witnessed this happen once.

src/sgame/sg_bot_nav.cpp
slipher: Er, that was not a yes-or-no question... why would it be a good idea to "duck when stuck"? What problem does that solve?

bmorel: When the navmesh is generated for areas in which bots needs crouching (this require a minor patch in daemonmap to use the crouched mins/maxs), the bots will get stucked/blocked when trying to follow the path.
Ducking in those conditions (and only those, but I can imagine ducking when meeting a friend so that friend might jump over, too) is the only way to allow reaching vents and the like (lot of those in parpax, for example).

slipher: > When the navmesh is generated for areas in which bots needs crouching (this require a minor patch in daemonmap to use the crouched mins/maxs), the bots will get stucked/blocked when trying to follow the path.

You really need put this kind of thing in the code comments and commit message.

Have you tested this and the modified daemonmap together successfully?

Maybe a trace with the box from crouchMaxs to maxs could be done, instead of mins to maxs, to test for this situation? Then crouching would not wrongly be triggered outside the vents.

src/sgame/sg_bot.cpp
slipher: Sounds like more testing is needed.

src/sgame/sg_bot_nav.cpp
bmorel: I have tested, and am right now in the process of writing code which checks (with trace) if the ducking is necessary (the code works, but I think it can be optimised, which is what I am doing).

You can see the results (of my current code, not the one here, I improved on the detection, but also need bigger live verifications) here: http://deadbeef.fr/random/crouch_and_climb.ogg

Issue #480, opened 2014-11-27
Not enough room to evolve!


overvolting:
When you're trying to evolve and you keep getting the message that there's not enough room because you're standing near a wall in a large open area, so you jump away and try again and now you're near a booster and so you get the message again.

It's like trying to herd a bunch of cats around in an orderly fashion after they've eaten catnip laced LSD shrooms.

Can we just agree that if there's enough room to evolve, IE, you're not stuck in a little pipe or something, then you evo and your body gently gets pushed away from the wall or obstacle?

Thanks!

Amanieu:
How about simply delaying evolution until enough room is available? A warning should be shown on the screen, and every frame the game will check if enough room is available.

overvolting:
Amanieu that might actually make a good improvement.

Veyrdite:
+1 for Amanieu's idea.

PR #1484, opened 2021-08-27
Move away from walls on evolve

necessarily-equal:
:tada:

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -2262,6 +2262,64 @@ static bool FindRoomForClassChangeVertically(
 	return !tr.startsolid && tr.fraction == 1.0f;
 }
 
+static bool FindRoomForClassChangeLaterally(
+		const gentity_t *ent,
+		const vec3_t fromMins, const vec3_t fromMaxs,
+		const vec3_t toMins, const vec3_t toMaxs,
+		vec3_t newOrigin)
+{
+	float max_x = fabs( (toMaxs[0]-toMins[0])
+	                  - (fromMaxs[0]-fromMins[0]) ) / 2.0f;
+	float max_y = fabs( (toMaxs[1]-toMins[1])
+	                  - (fromMaxs[1]-fromMins[1]) ) / 2.0f;
+	// this is an upper bound, hopefully as small as possible
+	// to avoid bugs like evolving on the other side of walls.
+	// ideally, it would be the correct minimal length for
+	// each 4 directions, but this is close enough.
+	// This is correct assuming the bbox is centered on the
+	// origin on x and y.
+	float distance = std::min(max_x, max_y);
+
+	vec3_t origin;
+	origin[0] = ent->client->ps.origin[0];
+	origin[1] = ent->client->ps.origin[1];
+	origin[2] = ent->client->ps.origin[2] + fromMins[2] - toMins[2] + 1.0f;
+
+	float best_distance = 1e37f;
+	bool found = false;
+	vec3_t deltas[] = { { 1, 0, 0 }, { 0, 1, 0 }, { -1, 0, 0 }, { 0, -1, 0 },
+	                    { 1, 1, 0 }, { -1, 1, 0 }, { -1, -1, 0 }, { 1, -1, 0 } };
+	for ( const vec3_t &delta : deltas )
+	{
+		trace_t trace;
+		vec3_t start_point;
+		VectorMA( origin, distance, delta, start_point );
+		trap_Trace( &trace, start_point, toMins, toMaxs,
+				origin, ent->s.number,
+				MASK_PLAYERSOLID, 0 );
+
+		vec3_t first_trace_end;
+		VectorCopy(trace.endpos, first_trace_end);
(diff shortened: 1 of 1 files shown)


src/sgame/sg_cmds.cpp
```diff
@@ -2252,6 +2252,56 @@ static bool FindRoomForClassChangeVertically(
 	return !tr.startsolid && tr.fraction == 1.0f;
 }
 
+static bool FindRoomForClassChangeLaterally(
+		const gentity_t *ent,
+		const vec3_t fromMins, const vec3_t fromMaxs,
+		const vec3_t toMins, const vec3_t toMaxs,
+		vec3_t newOrigin)
+{
+	float max_x = fabs(fabs(toMaxs[0]-toMins[0])
```
slipher, line 2261: maxs are always higher than mins so you can remove some `fabs`.

Since player bounding boxes are always centered on the origin, a tighter bound can be achieved by dividing by 2 relative to the current values. (Though a fudge factor beyond the tighest bound would be good since it's floating point, like add 1 at the end.) I think the tighter bound is a good idea because if the full difference between bbox sizes is used, then the new bbox at the start of the trace may not overlap with the player's current bbox. This would make it possible to emerge on the other side of a wall.

src/sgame/sg_cmds.cpp
```diff
@@ -2252,6 +2252,56 @@ static bool FindRoomForClassChangeVertically(
...
+	float distance = std::max(max_x, max_y);
+
+	vec3_t origin;
+	origin[0] = ent->client->ps.origin[0];
+	origin[1] = ent->client->ps.origin[1];
+	origin[2] = ent->client->ps.origin[2] + fromMins[2] - toMins[2] + 1.0f;
+
+	float best_distance = 1e37f;
+	bool found = false;
+	vec3_t deltas[] = { { 1, 0 }, { 0, 1 }, { -1, 0 }, { 0, -1 },
```
slipher, line 2277: This is probably going to give warnings for having less than 3 elements.

slipher (commented):
Great, fixing the inconvenience which was there since Tremulous.

src/sgame/sg_cmds.cpp
bmorel: > Since player bounding boxes are always centered on the origin

Either I misunderstand you, or it is wrong:

In https://github.com/UnvanquishedAssets/res-players_src.dpkdir/blob/master/configs/classes/human_light.model.cfg#L8-L10 we can see that, notably, this model (which can crouch) have origin not centered when standing up, but even more when crouching. IIRC, this is a general rule: height is not centered. But then, you're commenting on a X axis, and this assertion seems true for width and depth, indeed.

src/sgame/sg_cmds.cpp
```diff
@@ -2252,6 +2252,56 @@ static bool FindRoomForClassChangeVertically(
...
+		float new_distance = Distance(trace.endpos, origin);
+		if ( new_distance > best_distance )
+			continue; // not interesting
+
+		Log::Warn("Found a result! (%f, %f, %f)", newOrigin[0], newOrigin[1], newOrigin[2]);
+
+		// new best!
+		found = true;
+		best_distance = new_distance;
+		VectorCopy( (const float*)trace.endpos, newOrigin );
```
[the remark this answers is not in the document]
necessarily-equal: This cast was useful for debugging basically, ensuring it did copy to the right one.

src/sgame/sg_cmds.cpp
necessarily-equal: fixed

src/sgame/sg_cmds.cpp
necessarily-equal: I've changed the design to me more conservative in order to limit bugs a bit, but it still relies on the assumption you told a bit earlier, for the sake of simplicity.

necessarily-equal:
Ok, I found a bug.

How to break it:
* open plat23
* go in the corner with the overmind and the egg (the "innermost" part of the base)
* evolve to tyrant

src/sgame/sg_cmds.cpp
```diff
@@ -2240,22 +2236,92 @@ bool G_RoomForClassChange( gentity_t *ent, class_t pcl, vec3_t newOrigin )
...
+		Log::Warn("Trying (%f, %f, %f): %i, %i", start_point[0], start_point[1], start_point[2], trace.startsolid, trace.fraction < 1.0f);
+
+		if ( trace.startsolid /* || trace.fraction < 1.0f */)
+			continue; // collision
+
+		float new_distance = Distance(trace.endpos, origin);
+		if ( new_distance > best_distance )
+			continue; // not interesting
+
+		Log::Warn("Found a result! (%f, %f, %f)", newOrigin[0], newOrigin[1], newOrigin[2]);
```
necessarily-equal, line 2298: TODO: remove the warns

slipher:
Well it looks like the "vertically" version is supposed to make the final decision on whether evolving is allowed, but in the lateral movement case you forgot to actually call it.

slipher:
Fixes #480 !

PR #1490, opened 2021-09-05
Bot fixes

necessarily-equal:


src/sgame/sg_bot_parse.cpp
```diff
@@ -289,26 +289,30 @@ static AIValue_t randomChance( gentity_t*, const AIValue_t* )
...
+	// TODO: improve about performance when the need arises
+	std::string str = Cvar::GetValue( AIUnBoxString( params[ 0 ] ) );
 
-	if ( !c )
+	if ( str == "" )
 	{
 		return AIBoxInt( 0 );
 	}
 
-	return AIBoxInt( c->integer );
```
slipher, line 299: Use `Str::ParseInt` instead of `atoi`

src/sgame/sg_bot_parse.cpp
```diff
@@ -289,26 +289,30 @@ static AIValue_t randomChance( gentity_t*, const AIValue_t* )
...
+	std::string str = Cvar::GetValue( AIUnBoxString( params[ 0 ] ) );
 
-	if ( !c )
+	if ( str == "" )
 	{
-		return AIBoxFloat( 0 );
+		return AIBoxFloat( 0.0f );
 	}
 
-	return AIBoxFloat( c->value );
```
slipher, line 311: For floats you can use `Cvar::ParseCvarValue<float>`

src/sgame/sg_bot_parse.cpp
```diff
@@ -1276,6 +1276,7 @@ AIBehaviorTree_t *ReadBehaviorTree( const char *name, AITreeList_t *list )
 	D( UP_RADAR );
 	D( UP_JETPACK );
 	D( UP_GRENADE );
+	D( UP_FIREBOMB );
```
bmorel, line 1279: This only allows to use the symbol. Technically, they still can't buy it nor throw it.

src/sgame/sg_bot_parse.cpp
```diff
@@ -289,26 +289,30 @@ static AIValue_t randomChance( gentity_t*, const AIValue_t* )
 
 static AIValue_t cvarInt( gentity_t*, const AIValue_t *params )
 {
-	vmCvar_t *c = G_FindCvar( AIUnBoxString( params[ 0 ] ) );
+	// TODO: handle error reporting,
+	// TODO: improve about performance when the need arises
+	std::string str = Cvar::GetValue( AIUnBoxString( params[ 0 ] ) );
 
-	if ( !c )
+	if ( str == "" )
```
bmorel, line 296: why not using std::string::empty() instead of this comparison?

bmorel (commented):
LGTM

src/sgame/sg_bot_parse.cpp
necessarily-equal: I've used Cvar::ParseCvarValue for ints

src/sgame/sg_bot_parse.cpp
necessarily-equal: you mean they won't want to do it, right?

slipher:
Already committed a fixed version of that cvar thing in #1503. Needs rebase

necessarily-equal:
rebased

bmorel:
Le Tue, 28 Sep 2021 06:12:07 -0700,
necessarily-equal ***@***.***> a écrit :

> @necessarily-equal commented on this pull request.
> 
> 
> 
> > @@ -1276,6 +1276,7 @@ AIBehaviorTree_t *ReadBehaviorTree( const
> > char *name, AITreeList_t *list )
>  	D( UP_RADAR );
>  	D( UP_JETPACK );
>  	D( UP_GRENADE );
> +	D( UP_FIREBOMB );
> 
> you mean they won't want to do it, right?
> 

Actually, I was wrong. There is currently (in vanilla, and I use the
feature, how could I forget it...) possibilities to buy specific
equipment, so it makes sense to have the symbol emitted in any case.

Still, to have bots using grenades and firebombs, you'll need special
BT cases to handle both the shopping and the use, they will do neither
of those automatically.
While this make sense imo for the use (API could still be improved on
that matter, but it's doable), I think trying to handle correctly the
shopping wishes of bots in behavior trees would be an error
(unmaintainable on medium term) except in very specific cases (for
example, BT is a good place to ask bots to buy a pulse to replace their
luci when they need to defend the base).

I suppose for consumables this is doable, but I don't want to imagine
managing the remaining credits or implement bots trying to save money
for next respawn in there.

Note that all this also, of course, apply to grenade and (even
more) to jetpack (which are already in code anyway).

Issue #1373, opened 2021-04-15
nobuild content flag is the same value as the mover content flag


illwieckz:
Spectator is not affected by playerclip, it must be…

illwieckz:
I'm not sure we can fix it easily, as seen in `cgame/cg_predict.cpp`:

```c++
		cg_pmove.tracemask = MASK_DEADSOLID; // spectators can fly through bodies
```

It's expected spectator is not clipped by bodies, maybe that's the same for `playerclip` surfaces?

In `shared/bg_public.h`:

```c++
#define MASK_PLAYERSOLID ( CONTENTS_SOLID | CONTENTS_PLAYERCLIP | CONTENTS_BODY )
#define MASK_DEADSOLID   ( CONTENTS_SOLID | CONTENTS_PLAYERCLIP )
```

But `playerclip` surfaces are not bodies, and projectiles hit bodies…

illwieckz:
Well, even those spectators look to be clipped on `playerclip` given that mask…

illwieckz:
Well, if I apply `CONTENTS_BODY` to `MASK_DEADSOLID` that is expected to be used by the spectator, I can still go through `playerclip` surfaces, so maybe the bug is that sometime the spectator uses `MASK_DEADSOLID` and sometime not…

illwieckz:
When `SpectatorThink` is called, `pm.tracemask` seems to be undefined (I get different value on each game start).

illwieckz:
Spectator is affected by playerclip. The problem is elsewhere and is very bad.

The brush I was expecting to be clipped on as spectator had all those surfaceparm:

- slick, to prevent wallwalking
- nobuild, to prevent building
- playerclip, to clip player but let weapon projectile pass

The purpose of this material, named `common/outside`, was to provide mappers a very easy way to delimit the playable area: a single brush would prevent players to leave the area, missiles would still fly through, and players would not be able to build outside the playable area, assuming the `common/outside` brush is large enough to cover the surface disabled to build.

But, I discovered something very very bad. The `CONTENT_NOBUILD` flag is reusing `CONTENT_MOVER` flag. That's the same bit in the `.bsp` file. Maybe sometime around year 2000 when Tremulous was an early project, something thought that since it was a good idea to disable buildable on movers, it may also be a good idea to reuse the bit to mark content to not be buildable.

[...]

PR #1492, opened 2021-09-08
sg_buildable: also disable building on surfaces of brushes with nobuild contentflag

illwieckz:
When a brush has the `nobuild` content flag, building is forbidden inside the brush, but still allowed on the surfaces of the brush. This means even setting `surfaceparm nobuild` (which sets the `nobuild` content parm) on a sky surface or on a playerclip between player and sky does not prevent to build on it (and there is no way to build inside anyway):

Note: material syntax uses `surfaceparm` keyword for both surfaceparm and contentparm, so `surfaceparm nobuild` sets a contentparm).

[nobuild surface](https://dl.illwieckz.net/b/unvanquished/bugs/nobuild-surface/unvanquished_2021-09-08_032711_000.jpg)

***I'm looking for comments about the idea itself.***

Do we want to do that? Do you think it's a good idea?

Another solution that may look better at first glance would be to use the `nobuildsurface` surfaceparm, but in fact, while the code for it is implemented in the game and it is listed in `custinforparms.txt` read by q3map2, it was probably never used by Unvanquished (`common` textures do not set `nobuildsurface`, neither the `space` skyboxes), and it was almost unused in Tremulous (or `nobuild` was used anyway). Also, I see no situation were it would be needed to set it on content but not on surface, or on surface but not on content.

Do you see any reason to set a `nobuildsurface` on a brush without `nobuild` content?
Do you see any reason to set `nobuild` on a brush without `nobuildsurface` ?

If you see no reason, we can just rely on `nobuild` contentparm to both forbid building on content and surface.

The `nobuildsurface` surface only makes sense on solid brush (so nothing is buildable inside even if `nobuild` content is not set), and non-solid brushes to forbid building inside an area (like the `common/nobuild` material) are meant to not be solid, since it's to forbid to build where a player can go (so if it's not solid there is no surface to apply `nobuildsurface` on). And if the non-solid `common/nobuild` is used to forbid to build where a player can't go (outside of a fence or an ajar door, for example), that material is still non-solid so `nobuildsurface` would be useless anyway.

As noted [there](https://github.com/Unvanquished/Unvanquished/issues/1373#issuecomment-821862567), on 1575 Tremulous bsp, only 62 are using `nobuildsurface`, 38 use `nobuildsurface` without `nobuild`. If I deduplicate maps with different case in name and different versions, only 19 maps uses `nobuildsurface` without `nobuild`: `atcs3 atcs3a atestcs3 cage e1m2test1 gantry0 intent labyrinth loop monka1 oasys reiseb1 rtb sokolov thermal thule traps tremor volcano`. We can fix those maps when porting them if we chose to entirely drop `nobuildsurface`.

The contentparm and surfaceparm bits are very precious, and custom contentparm and surfaceparm are colliding with engines we are inheriting from.

### DIFF
--- a/src/sgame/CustomSurfaceFlags.h
+++ b/src/sgame/CustomSurfaceFlags.h
@@ -37,4 +37,4 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // custom content flags
 #define CUSTOM_CONTENTS_NOALIENBUILD    BIT( 12 ) // disallow alien building
 #define CUSTOM_CONTENTS_NOHUMANBUILD    BIT( 13 ) // disallow human building
-#define CUSTOM_CONTENTS_NOBUILD         BIT( 14 ) // disallow building,       override CONTENTS_MOVER
+#define CUSTOM_CONTENTS_NOBUILD         BIT( 14 ) // disallow building,       override CONTENTS_MOVER from RTCW and Wolf:ET we inherit through ET:XreaL, Q3 and Tremulous did not defined any flag there
--- a/src/sgame/CustomSurfaceFlags.h
+++ b/src/sgame/CustomSurfaceFlags.h
@@ -38,6 +38,3 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define CUSTOM_CONTENTS_NOALIENBUILD    BIT( 12 ) // disallow alien building
 #define CUSTOM_CONTENTS_NOHUMANBUILD    BIT( 13 ) // disallow human building
 #define CUSTOM_CONTENTS_NOBUILD         BIT( 14 ) // disallow building,       override CONTENTS_MOVER
-
-// custom surface flags
-#define CUSTOM_SURF_NOBUILD             BIT( 21 ) // disallow building,       override SURF_GLASS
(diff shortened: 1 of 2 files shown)


illwieckz:
I added two commits:

- one to drop `noalienbuildsurface` and `nohumanbuildsurface` because they were never used anyway ([see this comment](https://github.com/Unvanquished/Unvanquished/issues/1373#issuecomment-821862567)),
- one to drop `nobuildsurface` if we merge the first commit to also apply `nobuild` content parm to surfaces.

bmorel:
I think it's better to rely on volumes to alter that kind of things, which are in Q3's terminology brushes.
Having per-species brushes/surfaces also seems fragile to me, because it prevents inserting new species by mods and would require maps to be patched when the game itself evolves.

So, I'm rather in favor of those changes, but I am not really knowledgeable enough on the topic of q3 mapping to give any technical opinion.

necessarily-equal:
In principle, this PR looks reasonable if we fix the maps when porting. I'm probably not knowledgeable enough on all the legacy / Q3 / mapping stuff. I don't see the need to have a separate flag for surfaces other than some backward compat, and you noted keeping it causes some trouble with backward compat anyway.

illwieckz:
Let's do it.

necessarily-equal:
Now we need to remove the feature in netradiant too :eyes:

PR #1503, opened 2021-09-19
#1366 "CVars work by necessarily-equal" + fixes

slipher:
Fixes on top of the #1366 changes:
- Remove unneeded callbacks from BP vars
- Comma splitter: Use standard library std::string functionality instead of writing own
- Fix misspelled item names
- Don't use "allowed" as argument name for forbidden things
- Don't use vectors in 'conditions' struct in gentity_t because these things get memset
- Improve bot cvar reading

### DIFF
--- a/src/shared/bg_misc.cpp
+++ b/src/shared/bg_misc.cpp
@@ -29,6 +29,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "engine/qcommon/q_shared.h"
 #include "common/FileSystem.h"
 #include "bg_public.h"
+#include "parse.h"
 
 #define N_(x) x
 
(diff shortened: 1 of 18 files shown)


src/shared/parse.cpp
```diff
@@ -4276,3 +4276,38 @@ int Parse_SourceFileAndLine( int handle, char (&filename)[ MAX_QPATH ], int *lin
...
+see the header file for the documentation
+===============
+*/
+
+static constexpr char csv_delimiters[] = ", ";
+
+Parse_WordListSplitter::Parse_WordListSplitter(std::string str)
+	: _string(std::move(str)), _stop(0)
+{
+	++*this;
```
necessarily-equal, line 4293: neat

src/shared/bg_misc.cpp
```diff
@@ -2313,256 +2314,127 @@ int BG_UnpackEntityNumbers( entityState_t *es, int *entityNums, unsigned int cou
...
-		{
-			p++;
-			q = p;
+		if (result != PCL_NONE) {
+			results.push_back(result);
 		}
 		else
-		{
-			break;
+	       	{
```
necessarily-equal, line 2373: indentation went wrong here

necessarily-equal:
there is a regression in your version though:
While `g_disabledEquipment "lcannon"` works and `g_disabledEquipment "gren"` works, `g_disabledEquipment "lcannon,gren"` doesn't.

slipher:
> 
> 
> there is a regression in your version though:
> While `g_disabledEquipment "lcannon"` works and `g_disabledEquipment "gren"` works, `g_disabledEquipment "lcannon,gren"` doesn't.

Fixed

PR #1506, opened 2021-09-25
g_teamForceBalance 2 to only enforce balance if there is someone in another team

illwieckz:
- If no one chose to join the opposite team, everyone can join the same team.
- Once one chose the opposite team, the next players will have to balance.


illwieckz:
For example if a newcomer joined a team, and there is no one else but you around, you can chose to team up with him against bots and show him some tricks instead of giving him an unfair opposition.

slipher:
LGTM

src/sgame/sg_cmds.cpp
```diff
@@ -910,10 +910,18 @@ void Cmd_Team_f( gentity_t *ent )
...
+					if ( g_teamForceBalance.integer == 1 && G_PlayerCountForBalance( TEAM_ALIENS ) > G_PlayerCountForBalance( TEAM_HUMANS ) )
+					{
+						G_TriggerMenu( ent - g_entities, MN_A_TEAMFULL );
+						return;
+					}
+					else if ( g_teamForceBalance.integer == 2 && G_PlayerCountForBalance( TEAM_HUMANS ) > 0 && G_PlayerCountForBalance( TEAM_ALIENS ) > G_PlayerCountForBalance( TEAM_HUMANS ) )
+					{
+						G_TriggerMenu( ent - g_entities, MN_A_TEAMFULL );
+						return;
+					}
```
bmorel, line 924: You should handle the case when value is > 2.

src/sgame/sg_cmds.cpp
```diff
@@ -931,10 +939,18 @@ void Cmd_Team_f( gentity_t *ent )
...
+					if ( g_teamForceBalance.integer == 1 && G_PlayerCountForBalance( TEAM_HUMANS ) > G_PlayerCountForBalance( TEAM_ALIENS ) )
+					{
+						G_TriggerMenu( ent - g_entities, MN_H_TEAMFULL );
+						return;
+					}
+					else if ( g_teamForceBalance.integer == 2 && G_PlayerCountForBalance( TEAM_ALIENS ) > 0 && G_PlayerCountForBalance( TEAM_HUMANS ) > G_PlayerCountForBalance( TEAM_ALIENS ) )
+					{
+						G_TriggerMenu( ent - g_entities, MN_H_TEAMFULL );
+						return;
+					}
```
bmorel, line 953: You should handle the case when value is > 2.

bmorel (changes requested):
Except for the missing cases where g_teamForceBalance.integer > 2, I think this should provide at least some documentation, a minimum would be a comment near the cVar's declaration, but that would really be the minimum imo, since it would still not allow server admins to learn about the new feature (that I like, btw).

In a perfect world, the duplicated block of code would be moved into a function, but that's really a detail to me.

PR #1508, opened 2021-09-26
fix "addition of unsigned offset ..."

bmorel:
Found by UBsan.
Returning without action and emitting a warning only, if string is invalid or empty. I think the problem's root is likely  above in the call stack, but this patch will at least no longer make it happening and corrupting memory silently.

### DIFF
--- a/src/shared/parse.cpp
+++ b/src/shared/parse.cpp
@@ -1123,14 +1123,22 @@ Parse_StripDoubleQuotes
 */
 static void Parse_StripDoubleQuotes( char *string )
 {
-	if ( *string == '\"' )
+	if ( string == nullptr || *string == '\0' )
 	{
-		memmove( string, string + 1, strlen( string ) + 1 );
+		Log::Warn( "Parse_StripDoubleQuotes called with invalid or empty string" );
+		return;
 	}
 
-	if ( string[ strlen( string ) - 1 ] == '\"' )
+	size_t len = strlen( string );
+	if ( *string == '\"' )
 	{
-		string[ strlen( string ) - 1 ] = '\0';
+		memmove( string, string + 1, len );
+		--len;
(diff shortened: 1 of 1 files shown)


src/shared/parse.cpp
```diff
@@ -1123,6 +1123,12 @@ Parse_StripDoubleQuotes
...
 {
+	if ( string == nullptr || *string == '\0' )
+	{
+		Log::Warn( "Parse_StripDoubleQuotes called with invalid or empty string" );
+		return;
+	}
+
 	if ( *string == '\"' )
 	{
 		memmove( string, string + 1, strlen( string ) + 1 );
```
slipher, line 1134: Are you sure the UB is coming from the empty string case? Because it can also read out of bounds if the string begins with a quote. This line is wrong; it should be `memmove( string, string + 1, strlen( string ) );`

bmorel: The actual problem UBsan found was not from the memmove, but from the `string[ strlen( string ) - 1 ] = '\0';` line. 

The path I propose mostly prevent it from silently corrupting memory, instead it prevents said corruption and emits noise when such situation would happen.
I have not dug further since I was debugging a different, unrelated issue of my branch, but since this one could easily be ported to master, I thought it would be a good idea.
You are, indeed, right about the memmove too, let me fix it in that PR.

src/shared/parse.cpp
```diff
@@ -1123,14 +1123,22 @@ Parse_StripDoubleQuotes
 */
 static void Parse_StripDoubleQuotes( char *string )
 {
+	size_t len = strlen( string );
+	if ( string == nullptr || len == 0 )
```
slipher, line 1127: Useless null check because you already dereferenced it with strlen

PR #1510, opened 2021-09-27
Add Buildqueue

Gireen:
Allow immediate building placement but increase construction time of following buildings if build in quick succession.
Alien buildings are scaled by health percentage instead of buildtime since this now varies and isn't shared between client and server
requires  UnvanquishedAssets/unvanquished_src.dpkdir#60

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -2264,9 +2264,9 @@ void CG_Buildable( centity_t *cent )
 	{
 		// If buildable has spawned or is a human buildable, don't alter the size
 		bool  spawned = ( es->eFlags & EF_B_SPAWNED ) || ( team == TEAM_HUMANS );
-
+		health = es->generic1;
 		float adjustScale = spawned ? 1.0f :
-			sinf( static_cast<float>(cg.time - es->time) / ba->buildTime * M_PI/2.0f );
+			sinf( static_cast<float>(health) / ba->health * M_PI/2.0f );
 		ent.skeleton = bSkeleton;
 
 		if( es->modelindex == BA_H_MGTURRET || es->modelindex == BA_H_ROCKETPOD )
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -2902,6 +2902,18 @@ void CG_Rocket_DrawStaminaBolt()
 	Rocket_SetClass( "walk", !activate );
 }
 
+void CG_Rocket_DrawBuildQueueTime()
+{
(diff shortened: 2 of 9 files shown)


slipher:
I tried it briefly and it seemed like I could build an arbitrary number of buildings at once. Any increased delays were barely noticeable.

What is the motivation for this?

Gireen:
Placement is possible instantly but the construction time is increased about how much the waiting time before was.
For example if you place an armory and direct after that another beside the second one will take 6 seconds longer for construction.

The total time all constructions take is the same as before, as if a builder would stand there and wait until he can place a new building.

The motivation is to free players from being forced to wait for every building to finish its construction. The positive effect of this is that it becomes possible for a single player to between build to fight easily. Especially for rounds with few players this is helpful and makes also the overall building experience better.

bmorel:
You forgot to mention that while buildables are queued, humans can *not* repair, and I think dying or selling ckit to rebuy it does not resets the queue.

To be honest, I'm a bit mitigated as to have this in vanilla in that state: I would prefer for server owners to be able to enable the feature through a cVar (but overall I like it, especially for grangers which are very annoyingly slow and defenseless)

Gireen:
> You forgot to mention that while buildables are queued, humans can _not_ repair, and I think dying or selling ckit to rebuy it does not resets the queue.

That's not the case. You can repair with queued buildings and the additional time is supposed to not reset on death or class change.

bmorel:
> > You forgot to mention that while buildables are queued, humans can _not_ repair, and I think dying or selling ckit to rebuy it does not resets the queue.
> 
> That's not the case. You can repair with queued buildings

Just checked, I thought it was the case, I was wrong. And since I was wrong, I think this is a bug which makes human buildings even more powerful than before, since in vanilla, you can not repair while a building is processing.

necessarily-equal:
@Gireen could you rebase this? I think it's a good change overall

bmorel:
How hard would it be to have the build system controlled by a cvar? I know I'm asking for more work, but when the PR will be merged, it will be harder to get an optional thing.

bmorel (approved):

bmorel:
I'm fine with it. I'm still planning to implement to optionally use old system for a specific team, though, but that should not block this.

slipher:
I really don't like the idea of being able to build an arbitrary amount of structures instantaneously. That kind of thing makes the game more open to exploits and cheating. Like building everything and then disconnecting, or using a script to place the buildables.

Gireen:
> or using a script to place the buildables.

That's indeed a serious problem.
The other problems are existing ones but the change clearly could make them worse.
This will probably need a different approach, so i close this PR.

PR #1512, opened 2021-09-27
Query the engine to know if an extension-independent image path exists, check for multiple levelshot paths

illwieckz:
https://github.com/DaemonEngine/Daemon/pull/528 makes possible for the game to query the engine to know if an extension-independent image path exists.

A secondary commit showcase the use of it: the code checks if Unvanquished's `meta/<name>/<mapname>` levelshot exists, but if it does not exist, try to load Quake3-like `levelshot/<name>` instead.

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -1899,6 +1899,7 @@ bool   CG_ClientIsReady( int clientNum );
 void       CG_BuildSpectatorString();
 
 bool   CG_FileExists( const char *filename );
+bool       CG_ImageExists( const char *filename );
 void       CG_UpdateBuildableRangeMarkerMask();
 void       CG_RegisterGrading( int slot, const char *str );
 
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -722,6 +722,11 @@ bool CG_FileExists( const char *filename )
 	return trap_FS_FOpenFile( filename, nullptr, fsMode_t::FS_READ );
 }
 
+bool CG_ImageExists( const char *filename )
+{
+	return trap_ImageExists( filename );
+}
+
 /*
 ======================
 CG_UpdateLoadingProgress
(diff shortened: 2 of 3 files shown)


slipher:
You can implement this right now without engine changes. Attempt to load the image, and check whether 0 is returned as the handle or not. Just need to make sure it uses the same flags as librocket, namely `RSF_NOMIP` (see `DaemonRenderInterface::LoadTexture`). The engine caches the results so it won't open the image file twice.

The only caveat is that the engine prints a warning whenever no image file is found. So my suggestion is to remove the warning log from the engine, and do it in the cgame instead. Then we can avoid logging in cases where it's OK for the image not to be found.

illwieckz:
> The only caveat is that the engine prints a warning whenever no image file is found. So my suggestion is to remove the warning log from the engine, and do it in the cgame instead.

In that specific case I'm fine with the warning since I already want to print a warning that the default path is not found anyway.

Ouside of this specific case, I don't want to remove the warning in the engine, I would prefer adding testing interfaces that does not raise warning. Engine side we already replaced some explicit asset loading triggering useless warning with this kind of test prior to loading, that's why I first thought about doing it that way for this specific case.

We may want to still add this kind of interface for some other uses case like models and sounds. If I'm right we have many useless logs about some models stuff that are tried to be loaded for real instead of just checking if the optional data is there.

slipher:
> We may want to still add this kind of interface for some other uses case like models and sounds. If I'm right we have many useless logs about some models stuff that are tried to be loaded for real instead of just checking if the optional data is there.

These cases can be handled with my proposal. The idea is to have the API within the cgame look like `qhandle LoadModel(const char* name, int flags, bool warn = true)` i.e. the `warn` option is set to true by default. So it will normally warn, but you can pass `false` to deliberately suppress the warnings in cases where it is optional. 

In cases like the quoted one, it is more efficient to just try to load the model right away rather than checking first, because then there is only one cgame->engine syscall rather than two.

illwieckz:
Obsoleted by:

- https://github.com/Unvanquished/Unvanquished/pull/3214

PR #1517, opened 2021-09-30
Linux CI on Azure Pipelines

slipher:


necessarily-equal (approved):
Haven't tested it, but looks good to me otherwise

PR #1519, opened 2021-10-03
improves navcon editor

bmorel:
The global intent is to allow easier binding and edition of navmeshes

* replaced DEFAULT_CONNECTION_SIZE with the navconRadius cvar
* implement GetFeetPoint() which gives navmesh coordinates of feet
* navedit command can now also reload (quick and dirty but works)
* new putcon command placing the navcon at feet

### DIFF
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -490,6 +490,7 @@ static const struct
 	{ "notarget",         0,                       0                },
 	{ "prevframe",        CG_TestModelPrevFrame_f, 0                },
 	{ "prevskin",         CG_TestModelPrevSkin_f,  0                },
+	{ "putcon",           0,                       0                },
 	{ "reload",           0,                       0                },
 	{ "reloadHud",        CG_ReloadHud_f,          0                },
 	{ "rocket",           Rocket_Rocket_f,         0,               },
--- a/src/sgame/botlib/bot_nav_edit.cpp
+++ b/src/sgame/botlib/bot_nav_edit.cpp
@@ -266,6 +266,7 @@ void Cmd_NavEdit( gentity_t *ent )
 	}
 }
 
+//TODO: there is lot of duplicated code between Cmd_AddConnection() and Cmd_PutCon_f
 void Cmd_AddConnection( gentity_t *ent )
 {
 	if ( !CheckHost( ent ) ) return;
@@ -355,6 +356,20 @@ void Cmd_AddConnection( gentity_t *ent )
 	}
 }
 
+//TODO: there is lot of duplicated code between Cmd_AddConnection() and Cmd_PutCon_f
+// The code should be rewrote to have several improvements.
+// Actually, all those commands should be, to have a consistent API.
+// Other changes which would be *super* nice are:
+// * examples of usage
+// * user documentation
+// * storing meta-data in navmesh connections
+// * ability to select an existing connection to:
+//   * remove it
+//   * move start point
+//   * move end point
+//   * change it's meta data
+//   * change the directional info
+//   * affect it to multiple navmeshes
 void Cmd_PutCon_f( gentity_t *ent )
 {
 	if ( !CheckHost( ent ) ) return;
--- a/src/sgame/botlib/bot_nav_edit.cpp
+++ b/src/sgame/botlib/bot_nav_edit.cpp
@@ -44,8 +44,10 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 #endif
 #include "bot_navdraw.h"
 #include "nav.h"
+#include "sgame/sg_bot_util.h"
 
-static const int DEFAULT_CONNECTION_SIZE = 50;
+static Cvar::Cvar<int> navconRadius(
+	"bot_navconRadius", "radius of the navigation connection to place", Cvar::CHEAT, 15);
 
 bool GetPointPointedTo( NavData_t *nav, rVec &p )
 {
@@ -68,12 +70,16 @@ bool GetPointPointedTo( NavData_t *nav, rVec &p )
 	             CONTENTS_SOLID | CONTENTS_PLAYERCLIP, 0, traceType_t::TT_AABB );
 
 	pos = qVec( trace.endpos );
-	if ( dtStatusFailed( nav->query->findNearestPoly( pos, extents, &nav->filter, &nearRef, p ) ) )
-	{
-		return false;
-	}
+	return !dtStatusFailed( nav->query->findNearestPoly( pos, extents, &nav->filter, &nearRef, p ) );
+}
 
-	return true;
+bool GetFeetPoint( NavData_t *nav, rVec &p, gentity_t *self )
+{
+	rVec extents;
+	VectorSet( extents, 640, 96, 640 );
+	rVec pos = qVec( self->s.origin );
+	dtPolyRef nearRef;
+	return !dtStatusFailed( nav->query->findNearestPoly( pos, extents, &nav->filter, &nearRef, p ) );
 }
 
 static struct
@@ -96,13 +102,13 @@ void BotDrawNavEdit( DebugDrawQuake *dd )
 	{
 		unsigned int col = duRGBA( 255, 255, 255, 220 );
 		dd->begin( DU_DRAW_LINES, 2.0f );
-		duAppendCircle( dd, p[ 0 ], p[ 1 ], p[ 2 ], DEFAULT_CONNECTION_SIZE, col );
+		duAppendCircle( dd, p[ 0 ], p[ 1 ], p[ 2 ], navconRadius.Get(), col );
 
 		if ( cmd.offBegin )
 		{
 			duAppendArc( dd, cmd.pc.start[ 0 ], cmd.pc.start[ 1 ], cmd.pc.start[ 2 ], p[ 0 ], p[ 1 ], p[ 2 ], 0.25f,
 						0.6f, 0.6f, col );
-			duAppendCircle( dd, cmd.pc.start[ 0 ], cmd.pc.start[ 1 ], cmd.pc.start[ 2 ], DEFAULT_CONNECTION_SIZE, col );
+			duAppendCircle( dd, cmd.pc.start[ 0 ], cmd.pc.start[ 1 ], cmd.pc.start[ 2 ], navconRadius.Get(), col );
 		}
 		dd->end();
 	}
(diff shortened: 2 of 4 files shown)


src/sgame/botlib/bot_nav_edit.cpp
```diff
@@ -254,6 +260,12 @@ void Cmd_NavEdit( gentity_t *ent )
 
 		BotSaveOffMeshConnections( cmd.nav );
 	}
+	else if ( !Q_stricmp( arg, "reload" ) )
+	{
+		//TODO this is very violent, but might work for a PoC
+		G_BotNavCleanup();
+		G_BotNavInit();
```
necessarily-equal, line 267: do you know if it is it actually violent or if the surrounding code handles it correctly? (I'm fine with not resolving this tbh)

src/sgame/botlib/bot_nav_edit.cpp
```diff
@@ -68,12 +70,16 @@ bool GetPointPointedTo( NavData_t *nav, rVec &p )
...
-		return false;
-	}
+	return !dtStatusFailed( nav->query->findNearestPoly( pos, extents, &nav->filter, &nearRef, p ) );
+}
 
-	return true;
+bool GetFeetPoint( NavData_t *nav, rVec &p, gentity_t *self )
+{
+	rVec extents;
+	VectorSet( extents, 640, 96, 640 );
```
necessarily-equal, line 79: Minor enough, but why isn't Z the only coord that gets extended?

src/sgame/botlib/bot_nav_edit.cpp
bmorel: From testing, it seems fine, but I still think there's a better way to do it

src/sgame/botlib/bot_nav_edit.cpp
bmorel: This function code is `bool GetPointPointedTo( NavData_t *nav, rVec &p )` that I duplicated and adapted to my needs. I don't know where the values come from, otherwise I would have documented it.

src/sgame/botlib/bot_nav_edit.cpp
necessarily-equal: I think I got it to crash on first try :S

`/navedit reload human_naked`: `Thread 17 "daemon" received signal SIGSEGV, Segmentation fault.`

```backtrace
#0  UpdatePathCorridor (bot=bot@entry=0x7fff989fa3c8 <agents+584>, spos=..., 
    target=...) at ../src/sgame/botlib/bot_convert.h:102
#1  0x00007fff9882a172 in BotUpdateCorridor (botClientNum=1, 
    target=target@entry=0x7fff885343b0, cmd=0x7fff98974238 <g_botMind+2264>)
    at ../src/sgame/botlib/bot_nav.cpp:210
#2  0x00007fff987c3c6e in trap_BotUpdatePath (botClientNum=<optimized out>, 
    target=target@entry=0x7fff885343b0, cmd=<optimized out>)
    at ../src/sgame/sg_api.cpp:226
#3  0x00007fff987c876a in G_BotThink (self=0x7fff983e7b98)
    at ../src/sgame/sg_bot.cpp:443
#4  0x00007fff987b1318 in ClientTimerActions (ent=ent@entry=0x7fff983e7b98, 
    msec=msec@entry=25) at ../src/sgame/sg_active.cpp:892
#5  0x00007fff987b2b23 in ClientThink_real (self=self@entry=0x7fff983e7b98)
    at ../src/sgame/sg_active.cpp:2231
#6  0x00007fff987b2de0 in G_RunClient (ent=ent@entry=0x7fff983e7b98)
    at ../src/sgame/sg_active.cpp:2271
#7  0x00007fff988010a9 in G_RunFrame (levelTime=<optimized out>)
    at ../src/sgame/sg_main.cpp:2735
```

src/sgame/botlib/bot_nav_edit.cpp
```diff
@@ -276,6 +288,96 @@ void Cmd_AddConnection( gentity_t *ent )
...
+	{
+		Log::Notice( usage );
+	}
+}
+
+void Cmd_PutCon_f( gentity_t *ent )
+{
+	if ( !CheckHost( ent ) ) return;
+	const char usage[] = "Usage: putcon start <oneway|twoway> \n"
+	                     " putcon end\n"
```
necessarily-equal, line 368: It's really odd the use `start` then `end`. Maybe start/stop? or maybe rather begin/end?

src/sgame/sg_cmds.cpp
```diff
@@ -4383,6 +4383,7 @@ static const commands_t cmds[] =
 	{ "noclip",          CMD_CHEAT_TEAM,                      Cmd_Noclip_f           },
 	{ "notarget",        CMD_CHEAT | CMD_TEAM | CMD_ALIVE,    Cmd_Notarget_f         },
 	{ "pubkey_identify", CMD_INTERMISSION,                    Cmd_Pubkey_Identify_f  },
+	{ "putcon",          CMD_CHEAT,                           Cmd_PutCon_f           },
```
necessarily-equal, line 4386: « `// keep the list synchronized with the list in cg_consolecmds for completion.` »

src/sgame/botlib/bot_nav_edit.cpp
necessarily-equal: Also, I should say that this is when playing with bots.

src/sgame/botlib/bot_nav_edit.cpp
bmorel: That's using the same syntax as already existing code. It's unlikely people put bindings in their configs, so this can be changed, sure.
I don't really think this is that important, what I would like in the end is to never again have to write those damned commands, by having proper bindings.
What is important, though, is to provide the feature.

src/sgame/sg_cmds.cpp
bmorel: Sure. And how to have auto-completion enabled for the parameters? That would be almost more important to me.

src/sgame/botlib/bot_nav_edit.cpp
bmorel: I have it, too, with 4 bots. Nice catch. The fault seems related to another part of the code though. Not unrelated, but untouched, so that's about something which might have deeper roots.

bmorel: command removed, so no more segfaults.

src/sgame/sg_cmds.cpp
bmorel: Done. Without the parameters.

slipher:
Can you write a bit of documentation on how to use the addcon stuff somewhere? I think you're the only person who knows how to use it, and understanding what it's for would help in reviewing the commit.

src/cgame/cg_consolecmds.cpp
```diff
@@ -490,6 +490,7 @@ static const struct
 	{ "notarget",         0,                       0                },
 	{ "prevframe",        CG_TestModelPrevFrame_f, 0                },
 	{ "prevskin",         CG_TestModelPrevSkin_f,  0                },
+	{ "putcon",           0,                       0                },
```
slipher, line 493: command completion is the 3rd column here

src/sgame/botlib/bot_nav_edit.cpp
```diff
@@ -276,6 +282,96 @@ void Cmd_AddConnection( gentity_t *ent )
 
 	arg = Cmd_Argv( 1 );
 
+	if ( cmd.enabled )
```
slipher, line 285: Instead of copy and pasting the whole function, I suggest using the same function to implement both commands, and switching the point placement behavior based on Cmd_Argv(0).

src/cgame/cg_consolecmds.cpp
bmorel: This would require me to investigate how to write the functions, though, and that's a rabbit hole I'd rather not go in for now, especially considering the fact I have _zero_ knowledge on the cgame side.

src/sgame/botlib/bot_nav_edit.cpp
bmorel: > Can you write a bit of documentation on how to use the addcon stuff somewhere? I think you're the only person who knows how to use it, and understanding what it's for would help in reviewing the commit.

There is documentation in 2 places, currently:

* a small tutorial [https://forums.unvanquished.net/viewtopic.php?f=8&t=2244&sid=93e4d95f4a588299c789cd597e1acd12#p18577](here)
* a wiki page [https://wiki.unvanquished.net/wiki/Navigation_Meshes](there)

I agree that wiki page's quality can be improved, and that a better doc could be written, and that's even something I'd like to do at some point... *when* I will understand enough the navedit commands, as there are commands, notably, to somehow force bots to go somewhere.

With my current knowledge of the problem, I can't really write a good enough doc imo. There's also likely a «tips and tricks» database to build, and that kind of things, which even depends on what features are around.
For example, my code to allow the use of ladders and wallwalking is not ready yet, and might allow to start such a database.
At some time, I will want to implement "reach by jump or pounce" code, so that bots can access disconnected areas which are _above_ their current one. This will likely require changing the API, and thus rewrite the doc.

What can likely be done, though, is to open a ticket for next release to write/update the doc for this. I think this would be a good thing to do that for each releases: take time to update stuff which requires documentation, especially for user-facing feature. That's a different topic though.

bmorel:
> Can you write a bit of documentation on how to use the addcon stuff somewhere? I think you're the only person who knows how to use it, and understanding what it's for would help in reviewing the commit.

There is documentation in 2 places, currently:

* a small tutorial [here](https://forums.unvanquished.net/viewtopic.php?f=8&t=2244&sid=93e4d95f4a588299c789cd597e1acd12#p18577)
* a wiki page [there](https://wiki.unvanquished.net/wiki/Navigation_Meshes)

I agree that wiki page's quality can be improved, and that a better doc could be written, and that's even something I'd like to do at some point... *when* I will understand enough the navedit commands, as there are commands, notably, to somehow force bots to go somewhere.

With my current knowledge of the problem, I can't really write a good enough doc imo. There's also likely a «tips and tricks» database to build, and that kind of things, which even depends on what features are around.
For example, my code to allow the use of ladders and wallwalking is not ready yet, and might allow to start such a database.
At some time, I will want to implement "reach by jump or pounce" code, so that bots can access disconnected areas which are _above_ their current one. This will likely require changing the API, and thus rewrite the doc.

What can likely be done, though, is to open a ticket for next release to write/update the doc for this. I think this would be a good thing to do that for each releases: take time to update stuff which requires documentation, especially for user-facing feature. That's a different topic though.

src/sgame/botlib/bot_nav_edit.cpp
bmorel: Alright. Let's use the occasion to improve it further, then.

bmorel: Change of mind.
I don't want to break previous API, and handling optional parameters in this stuff is super tedious for nothing, it's even more painful than using getopt.

I will rewrite those API cleanly, yes. But this will happen *when* I'll know how to annotate the navcons with the method required to use it (ladder, WW, jump with req magnitude, pounce with magnitude, wall jump, maybe fly, etc).

For now, it's really just lot of very tedious work to do stuff correctly for no benefit. I'll still add a TODO note with what is planned, though.

src/sgame/sg_cmds.cpp
```diff
@@ -4359,6 +4359,7 @@ static const commands_t cmds[] =
...
```
necessarily-equal, line 4362: I somehow managed to add the end of a connection while being dead. Let's not allow that:
```suggestion
	{ "putcon",          CMD_CHEAT | CMD_TEAM | CMD_ALIVE,    Cmd_PutCon_f           },
```

necessarily-equal (changes requested):

src/sgame/sg_cmds.cpp
bmorel: I don' t agree that putcon must require not being spectator, as the connection start or end will just end under the feet. Well, I have not tested that, but that's the behavior I'd expect, and it makes it easier sometimes.
The alive check, though, makes sense, but is a spectator alive?
There are other problems in there, like, you can't really do some navcon commands while following, for some reason. That's annoying enough, but I don't know how to change that, yet.

necessarily-equal: If you are a spectator; how are you supposed to get the closest point on your navmesh? GetFeetPoint uses that.

I'm not sure it would be worthwhile to do the testing and integration, even if it worked.

bmorel: no, GetFeetPoint will find the navmesh position which is the closest to the entity's feet.  This is different, and there's even a message saying if things worked as expected (and now that I think about it, there's none when it does not?)

necessarily-equal: I'm not sure how that's supposed to work. What's the spectator's navmesh? Unfortunately you currently have no feedback so it's hard to know if it worked

bmorel: you don't edit the spectator's navmesh, you edit the navmesh you selected before.
There is feedback about if the putcon worked, but none for addcon, that I kept pristine.

necessarily-equal: I had no feedback at all with putcon when I tried. Is it because I didn't select a class with `navedit`? Sounds like something worth fixing anyhow

bmorel: the feedback is a Log::Notice which appear on success. See https://github.com/Unvanquished/Unvanquished/pull/1519/commits/b3f078bacd9529df7144728ad331640430a30ad4#diff-385d371b213d00191fa775c5ba68d4edd1bb55cfdfd80e3e1da2f7d3142cb4b4R453

necessarily-equal: In my case, I get no feedback at all

Screenshot from 2021-10-03 22-35-14

Step to reproduce:
* join a map, don't join a team
* `/putcon start oneway`
* `/putcon end`

bmorel: Will investigate.

bmorel:
I will not merge

illwieckz:
It looks like the only thing missing for merging is a warning telling `navedit` had not been called prior to calling `putcon` as reported by afontain. Other issues look to have been properly addressed by @bmorel (either by code change, either by deciding it's ok to do it that way).

PR #1520, opened 2021-10-03
Cleanup

bmorel:
mostly fixes crashes or UB

### DIFF
--- a/src/sgame/botlib/bot_local.cpp
+++ b/src/sgame/botlib/bot_local.cpp
@@ -49,7 +49,10 @@ make sure the vectors use the same coordinate system
 void BotCalcSteerDir( Bot_t *bot, rVec &dir )
 {
 	const int ip0 = 0;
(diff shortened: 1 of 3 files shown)


src/sgame/sg_bot_util.cpp
```diff
@@ -904,6 +904,11 @@ bool BotChangeGoal( gentity_t *self, botTarget_t target )
 
 bool BotChangeGoalEntity( gentity_t *self, gentity_t const *goal )
 {
+	if ( !goal )
+	{
```
slipher, line 904: `goal` is intentionally allowed to be null, meaning the bot has no goal. You can see that BotSetTarget handles this case.

bmorel: Then the crashes I had on that function might be due to other commits I made (which touch the goal stuff). I Will investigate more, I'll remove those commits from PR meanwhile.

slipher:
LGTM

PR #1521, opened 2021-10-03
AI: rewrite shopping code, plus rush and heal score functions

bmorel:
Those commits will require a PR in the unvanquished_dpkdir to not have AI regressions, since some stuff was moved outside of C++ (e.g. medkit usage) and the return values of the scoring functions have also been changed (to look more consistent to me).
There are also the missing cvars for evolution and shopping which require adding GUI (the [[TODO]] tag in there is to prevent forgetting that fact. I already have the code, will push when this will be reviewed and merged).


slipher:
I requested earlier to drop `fix classAttributes::cost API inconsistency`.

The cvar one should not be included until it is actually implemented.

necessarily-equal:
There's quite litterally written "[[TODO]]" in a commit title. Is that intended?

bmorel:
> I requested earlier to drop `fix classAttributes::cost API inconsistency`.
> 
> The cvar one should not be included until it is actually implemented.

You did, but the reason to change that will be more obvious in later commits (sure, one can always keep the useless different variable around, but why?).
The cost/price inconsistency is uselessly annoying, it have no real reason to exist, and I can't see any reason to keep it.
There are several other inconsistencies of that kind in config files, and at some moment I'll likely submit changes to fix them, but for now, they don't annoy me enough.

bmorel:
> The cvar one should not be included until it is actually implemented.

It is implemented. In later commits.
Alternative is, I squash all, since you don't like having too many commits? But then, I'll likely have to tweak the future commits in big chunks.

bmorel:
> I don't really like keeping "[[TODO]]" in a commit title. Can we maybe put "to be added in a later commit" or something like that?

Fixed. This was to remind me to add the submodule changes which add UI for the cvars, stuff that I totally forgot, again.

src/sgame/sg_bot_util.cpp
```diff
@@ -27,12 +27,159 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+}
+
+template <typename T>
+struct equipment_t
+{
+	int &authorized;
+	T item;
+	int price( void ) const
+	{
+		return BG_Class( item )->price;
```
slipher, line 102: How could this possibly work? BG_Class only takes a class_t as its argument, not a weapon_t or upgrade_t. Likewise, BG_Upgrade only works on an upgrade_t.

src/sgame/sg_bot_ai.cpp
```diff
@@ -1289,7 +1226,7 @@ AINodeStatus_t BotActionBuy( gentity_t *self, AIGenericNode_t *node )
 	if ( buy->nparams == 0 )
 	{
 		// equip action
-		BotGetDesiredBuy( self, &weapon, upgrades, &numUpgrades );
+		numUpgrades = BotGetDesiredBuy( self, weapon, upgrades, 4 );
```
necessarily-equal, line 1229: one less magic number here:
```suggestion
		numUpgrades = BotGetDesiredBuy( self, weapon, upgrades, ARRAY_LENGTH( upgrades ) );
```

src/sgame/sg_bot_util.cpp
```diff
@@ -27,12 +27,159 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+		case PCL_HUMAN_LIGHT:
+			return 383;
+		case PCL_HUMAN_MEDIUM:
+			return 367;
+		case PCL_HUMAN_BSUIT:
+			return 367;
+		case PCL_NONE:
+		case PCL_NUM_CLASSES:
+		default:
+			ASSERT( false && "invalid class number provided" );
```
necessarily-equal, line 91:
```suggestion
			ASSERT_UNREACHABLE();
```
?

src/sgame/sg_bot_util.cpp
```diff
@@ -303,125 +450,149 @@ int BotValueOfUpgrades( gentity_t *self )
...
+		if ( armor.canBuyNow() && usableCapital >= armor.price()
+				&& ( usedSlots & armor.slots() ) == 0 )
+		{
+			if( !currentArmor || ( armor.price() > currentArmor->price() ) )
+			{
+				currentArmor = &armor;
+			}
+			upgrades[numUpgrades] = currentArmor->item;
+			usableCapital -= currentArmor->price();
+			usedSlots |= currentArmor->slots();
```
necessarily-equal, line 536: There is a bug here: what if you have less used slots than before?

```suggestion
			usedSlots &= ~currentArmor->slots();
			usedSlots |= currentArmor->slots();
```

some day we'll have a proper abstraction for inventory support, but hey

src/sgame/sg_bot_util.cpp
```diff
@@ -303,125 +450,149 @@ int BotValueOfUpgrades( gentity_t *self )
...
-			*weapon = WP_NONE;
+			continue;
+		}
+		if ( tool.canBuyNow() && usableCapital >= tool.price()
+				&& ( usedSlots & tool.slots() ) == 0
+				&& numUpgrades < upgradesSize )
+		{
+			upgrades[numUpgrades] = tool.item;
+			usableCapital -= tool.price();
+			usedSlots |= tool.slots();
```
necessarily-equal, line 591: ditto

src/shared/bg_parse.cpp
```diff
@@ -1263,7 +1264,7 @@ void BG_ParseClassAttributeFile( const char *filename, classAttributes_t *ca )
 		else if ( !( defined & STOPSPEED ) )       { token = "stopSpeed"; }
 		else if ( !( defined & JUMPMAGNITUDE ) )   { token = "jumpMagnitude"; }
 		else if ( !( defined & ICON ) )            { token = "icon"; }
-		else if ( !( defined & COST ) )            { token = "cost"; }
+		else if ( !( defined & PRICE ) )           { token = "price"; }
```
necessarily-equal, line 1267: hmm, you didn't change the name in the parsing code but you changed it here?

necessarily-equal (changes requested):
I haven't looked in too much details, but I've found a thing or two

PR #1525, opened 2021-10-08
rocket: add “translate” element with options for enabling quake formatting and emoticons in translatable strings

illwieckz:
People can now do:

```html
<translate>Hello there!</translate>
<translate quake emoticons>^3Hello ^1there! [granger]</translate>
```


illwieckz:
Some work in progress script to produce a .pot:

```python
#! /usr/bin/env python3
#-*- coding: UTF-8 -*-

from bs4 import BeautifulSoup
import datetime
import glob

rml_file_path="ui/menu_main.rml"

translation_dict = {}

for rml_file_path in glob.glob("ui/*.rml"):
	with open(rml_file_path) as rml_file:
		rml_content = rml_file.read()
		rml_soup = BeautifulSoup(rml_content, "lxml")

		for element in rml_soup.find_all("translate"):
			translatable_string = element.contents[0]

			if translatable_string not in translation_dict.keys():
				translation_dict[translatable_string] = []

			translation_dict[translatable_string].append(rml_file_path)

datetime_string = datetime.datetime.now().astimezone().strftime("%Y-%m-%d %H:%M:%S%z")

print("""# SOME DESCRIPTIVE TITLE.
# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
# This file is distributed under the same license as the PACKAGE package.
# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
#
#, fuzzy
msgid ""
msgstr ""
"POT-Creation-Date: """ + datetime_string + """\\n"
"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
"Language: \n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\\n"
"Content-Transfer-Encoding: 8bit\\n"
""")

for translatable_string in translation_dict.keys():
	print("#: " + " ".join(translation_dict[translatable_string]))
	print("msgid \"" + translatable_string + "\"")
	print("msgstr \"\"")
	print("")
```

It produces for example:

```
# SOME DESCRIPTIVE TITLE.
# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
# This file is distributed under the same license as the PACKAGE package.
# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
#
#, fuzzy
msgid ""
msgstr ""
"POT-Creation-Date: 2021-10-08 04:38:04+0200\n"
"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE
"
"Language: 
"
"MIME-Version: 1.0
"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"

#: ui/menu_main.rml
msgid "Play"
msgstr ""

#: ui/menu_main.rml
msgid "Server listings"
msgstr ""

#: ui/menu_main.rml
msgid "Start local/LAN game"
msgstr ""

```

slipher:
My concern with this approach is that the translation string may be unpredictable in the case of extra whitespace or embedded tags. If I have a string like `<translate>  something  < b > with </ b>tags </translate>` then what is the string passed to the translation function?

illwieckz:
> My concern with this approach is that the translation string may be unpredictable in the case of extra whitespace or embedded tags. If I have a string like `<translate> something < b > with </ b>tags </translate>` then what is the string passed to the translation function?

The `< b >` and `</ b>` seems to not be considered valid by BeautifulSoup and may be translated to be kept that way (not processed as markup) or get translated using `&gt;` or `&lt,`.

But if you do `<translate> something <b > with </b >tags </translate>` the translated string is `⎵something <b> with </ b>tags⎵` with leading and trailing white space (emphasized here), I don't want to filter the content, so strings between `translate` tags must not have leading and trailing whitespace, and it is better to not do multiline strings in RML or this would add useless tabs every where. When tags are valid, the code removes the useless spaces within tags, not between tags.

The inner tags are kept, that allows to do things like `<translate>first line<br/>second line</translate>` which means the translatable string is `first line<br/>second line`, which is _wanted_.

Maybe it could also allow to do things like `<translate>Your speed is <speed/> km/h!</translate>` which would provide a translatable string being `Your speed is <speed/> km/h!`, I don't know yet if libRocket/RmlUI can process custom tags within _rewritten_ custom tags.

illwieckz:
I consider the PR ready.

slipher:
I tested and it seems like stuff outside a tag is passed through as-is. But tags themselves are reformatted - single quotes in attributes are changed to double quotes, extra spaces are removed, etc. You can't just say that these things happen as we have many extant cases of a space before the closing `>` of a tag, for example.

illwieckz:
> we have many extant cases of a space before the closing > of a tag, for example.

I fixed all of them.

illwieckz:
You can find the rewritten RML files there: https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/64/files

We just have to consider that a leading/trailing space and multiple consecutive spaces within a `translate` tag is a bug and should be reported, and then, we don't have to write extra code. We may better solve problems with conventions when possible.

illwieckz:
On thing that may be doable is to make `generate_rml_pot` return with error if there are leading/trailing space or consecutive spaces within a `translate` tag, that would even be callable by CI.

illwieckz:
For reference, our old `.pot/.po` files generated from c/c++ source code was full of leading whitespace and trailing endline, so this issue is just a global issue, we may want to check the `.pot` file in general.

I also had to fix many leading whitespace and trailing endline in our c++ code base.

illwieckz:
> // Initialization is done in OnRender instead of OnUpdate because doing it in OnUpdate somehow
> // results in a new <translate> element being created with the already-translated content?
> // But updateLanguage can't be handled in OnRender due to non-visible content.

My first implementation was saving the original text to an attribute just named `text`, so, if the attribute was missing, it would save the original text into that attribute, and then, if the attribute was there, it would translate from the attribute and save to inner element, meaning we would be able to translate an inner element that is already translated.

necessarily-equal:
So what's the way forward here? @slipher opens a pull request of its own? @illwieckz looks at it and include changes?

bmorel:
up

bmorel:
@illwieckz : will this conflict with RmlUI3 branch?

illwieckz:
@slipher I rebased your [`rmltrans2`](https://github.com/slipher/Unvanquished/tree/rmltrans2) branch on `0.53.0/sync` and ported it to RmlUi, and fixed minor things around.

It's ready.

src/utils/generate_pot.sh
```diff
@@ -1,47 +1,58 @@
...
+"${script_dir}/generate_gender_pot.pl" \
+	'src/cgame/cg_event.cpp' \
+	>> "${temp_pot_file}"
+
+find 'src/cgame' 'src/sgame' 'src/shared' \
+	-type f -name '*.c' -o -name '*.cpp' \
+| sort \
+| xgettext --from-code=UTF-8 \
+	-j -o "${temp_pot_file}" \
+	-k_ -kN_ -kP_:1,2 -k -f -
```
slipher, line 55: `G_` seems to be missing... oh well it's not a regression I guess

src/utils/generate_rml_pot.py
```diff
@@ -0,0 +1,89 @@
...
+            else:
+                content = rml_content[text_start:m.start()]
+                if content != content.strip():
+                    error("leading or trailing whitespace in translated string")
+                if "\n" in content:
+                    error("translated string should not contain newlines")
+                for empty_element in re.finditer(r"<(\w+)(\s*/\s*)>", content):
+                    if empty_element.group(2) != " /":
+                        correct = f"<{empty_element.group(1)} />"
+                        error(f"empty element should be formatted like {correct} to match libRocket's GetInnerRML()")
```
slipher: I think I got things working so that you no longer have to worry about matching the formatting.

slipher (commented):
Unfortunately I forgot to push the latest version when I was last working on this. There is another commit needed which should be squashed with the <translate> one: https://github.com/Unvanquished/Unvanquished/commit/8bed0e9e91a0a8d32e4e273ba577e1ce307a83ba. I could port it to RmlUi myself if you like.

src/utils/generate_rml_pot.py
illwieckz: > Unfortunately I forgot to push the latest version when I was last working on this. There is another commit needed which should be squashed with the one: [8bed0e9](https://github.com/Unvanquished/Unvanquished/commit/8bed0e9e91a0a8d32e4e273ba577e1ce307a83ba). I could port it to RmlUi myself if you like.

If you can do that it would be kind. You can push to this branch, and I'll squash after that.

src/utils/generate_pot.sh
illwieckz: I'm not sure to get it.

slipher: Never mind, I forgot that there is a separate script for those.

slipher:
Pushed the commits to fix <translate>.

illwieckz:
wait… this was never merged????

bmorel:
Or maybe git lies, but I doubt it.

illwieckz:
No, even `unvanquished_src.dpkdir` doesn't have any `translate` markup yet:

- https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/64

…I was so sure it was a thing of the past I wrote it was done in upcoming release post.

bmorel:
so I guess new review has to be done, now?

necessarily-equal:
It seems like the html way™ would be to use a translate attribute instead of a translate tag. https://www.w3schools.com/tags/att_translate.asp — Now I've seen that I think it makes a lot of sense to use this instead.

illwieckz:
> The attribute solution has the added advantage of being less verbose.

It is less verbose only in `h2` and `blocklink` (and things like that). We need to add a tag for almost everything else. Though we may just reuse `p` and `span` for _everything else_.

I'm afraid about how much costly doing it the attribute way would be, as we would not look for some `translate` tag but for every tag that exists. Also if I'm right the current implementation is letting RmlUi call a function everytime it sees a `translate` tag, if we do the “correct” way, we would have to look for `translate`, `emoticons` etc. on _every_ tag. If someones wants to spend that much time, it's up to him and not for 0.53.

So, well, let's keep `translate` for now, I was annoyed about this not being very pure, but it looks like real purity is too far to reach to make it a requirement anyway.

slipher:
We register `translate` as a CDATA tag so that instead of parsing markup inside it, the contents are kept as a verbatim string (during the initial document loading). This allows us to perform translation before any elements inside the translated string are translated. If we try to let the English version be instantiated first and then translate it later, stuff doesn't work correctly in certain cases. So translate has to be a tag.

illwieckz:
Thank you for the clarification. 👍️

PR #1527, opened 2021-10-09
rocket: unify emoticon class with what people do by hand

illwieckz:
Minor thing, maybe we want to delay the merge to the last minute because it needs a change on `unvanquished_src.dpkdir` side.

See https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/65

### DIFF
--- a/src/cgame/rocket/rocket.cpp
+++ b/src/cgame/rocket/rocket.cpp
@@ -542,7 +542,7 @@ Rocket::Core::String Rocket_QuakeToRML( const char *in, int parseFlags = 0 )
 					spanHasContent = true;
 					out.Append( spanstr );
 				}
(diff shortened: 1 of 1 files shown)


slipher:
Pinging this since the PRs were approved. Won't merge it myself since I don't understand the inter-repo dependencies here.

illwieckz:
I prefer to merge this at the last minute before release, maybe we should create a special tag for that?

slipher:
> I prefer to merge this at the last minute before release

Why would that be needed?

If it is really needed, then the right thing to do is put it on `0.53.0/sync` branches.

bmorel:
I do not think it's useful to delay PRs to 0.53 because of inter-repo relations. Why not just update the children repos, and push this with synced submodules?

PR #1529, opened 2021-10-10
WIP effort at rewriting strings before we start translate.

illwieckz:
WIP effort at rewriting strings before we start translate.

This PR exists so you know where I am.


slipher:
> 
> 
> For the reminder: #1566 will bring a new set of such strings (the cvar descriptions).

Nah, I don't think there is any need to translate those. Nor does the infrastructure to translate them exist.

Translating the cvar-related UI stuff is already enough work.

illwieckz:
It better be restarted from scratch.

PR #1531, opened 2021-10-16
CG_MyTeam() team rework

necessarily-equal:
See https://github.com/Unvanquished/Unvanquished/pull/1337 for the discussions about cg.snap vs cg.currentPlayerState. I've dropped the dubious use in CG_ParseTeamInfo().

I think this change is safe :)


necessarily-equal:
If we are really early in the game and predictedPlayerState isn't yet initialized, then it will be full of zeroes. Since 0 is equal to TEAM_NONE, that sounds about right.

PR #1532, opened 2021-10-17
Bot Heal Rework

necessarily-equal:


illwieckz:
> Bot heal: lessen code duplication and possibly fixes some bug
> For example, the previous code didn't check if the booster was switched off.

I confirm bots tried to heal at disabled booster (like when the overmind is dead).

PR #1533, opened 2021-10-17
Bot target refactoring

necessarily-equal:
This brings more encapsulation and much better sanity checking of the target. I think it may very well fix a bug or two.

@bmorel you may need to do some rebasing after that. Would you want some help with that?

The first two commit are from https://github.com/Unvanquished/Unvanquished/pull/1532. This PR depends on them.


slipher:
Using GentityRef

```cpp
gentity_t* someEnt;
...
GentityRef ref;
// assign nothing
ref = nullptr;
// assign an entity
ref = someEnt;

// check if the entity still exists
if (ref)
{
    // access gentity_t fields
    int modelindex = ref->s.modelindex;
}
```

necessarily-equal:
Updated to use GentityRef

PR #1546, opened 2021-10-19
Fix the armory use distance situation

necessarily-equal:
Currently, there are 4 (four!) slightly different experiences in game when trying to buy stuff

* "how far the bots see the armory" which is the same as "how far (using binds or commands) you might use the armory"
* "the tutorial says you can open the armory with the q key"
* "the hand button that hints you can open the armory appears"
* "when you press q the menu opens"

This PR makes the experience nicer and make it look and feel much more consistent. I tried to avoid making a larger refactoring to avoid breaking stuff in the process (for example maybe some maps have activatable buttons?).

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -1072,35 +1072,24 @@ public:
 	{
 		vec3_t        view, point;
 		trace_t       trace;
-		entityState_t *es;
 
 		AngleVectors( cg.refdefViewAngles, view, nullptr, nullptr );
 		VectorMA( cg.refdef.vieworg, ENTITY_USE_RANGE, view, point );
 		CG_Trace( &trace, cg.refdef.vieworg, nullptr, nullptr,
 				  point, cg.predictedPlayerState.clientNum, MASK_SHOT, 0 );
 
-		es = &cg_entities[ trace.entityNum ].currentState;
+		const entityState_t &es = cg_entities[ trace.entityNum ].currentState;
 
-		if ( es->eType == entityType_t::ET_BUILDABLE && BG_Buildable( es->modelindex )->usable &&
-			CG_IsOnMyTeam(*es) )
+		if ( es.eType == entityType_t::ET_BUILDABLE
+				&& es.modelindex == BA_H_ARMOURY
+				&& Distance(cg.predictedPlayerState.origin, es.origin) < ENTITY_USE_RANGE - 0.2f // use an epsilon to account for rounding errors
+				&& CG_IsOnMyTeam(es) )
 		{
-			//hack to prevent showing the usable buildable when you aren't carrying an energy weapon
-			if ( es->modelindex == BA_H_REACTOR &&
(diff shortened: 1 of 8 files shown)


necessarily-equal:
(by @illwieckz)
> I mean, it may make the trick work and be abused in situation it was not expected before (like through larger walls).

If it's any consolation, in such case, the previous behavior would actually let pr0 players who had binds buy gear anyway. I can change the distance back if it helps though. This is more practical though IMO.

src/cgame/cg_tutorial.cpp
```diff
@@ -140,7 +140,7 @@ static entityState_t *CG_BuildableInRange( playerState_t *ps, float *healthFract
 	trace_t       trace;
 
 	AngleVectors( cg.refdefViewAngles, view, nullptr, nullptr );
-	VectorMA( cg.refdef.vieworg, 64, view, point );
+	VectorMA( cg.refdef.vieworg, ENTITY_USE_RANGE - 0.2f, view, point );
```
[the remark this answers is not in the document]
necessarily-equal: I don't think the risk of forgetting it is too high, as I don't think someone will duplicate once more that kind of code. And if they do, they'll probably benefit from a refactor anyway :S

I'm not fond of having a
```C
#if BUILD_CGAME
#define ENTITY_USE_RANGE (64.0f - 0.2f)
#endif
#if BUILD_SGAME
#define ENTITY_USE_RANGE 64.0f
#endif
```
if that's what you had in mind

necessarily-equal:
ping @bmorel

src/cgame/cg_rocket_draw.cpp
```diff
@@ -1049,35 +1049,24 @@ class UsableBuildableElement : public HudElement
...
+		const entityState_t &es = cg_entities[ trace.entityNum ].currentState;
 
-		if ( es->eType == entityType_t::ET_BUILDABLE && BG_Buildable( es->modelindex )->usable &&
-			CG_IsOnMyTeam(*es) )
+		if ( es.eType == entityType_t::ET_BUILDABLE
+				&& es.modelindex == BA_H_ARMOURY
+				&& Distance(cg.predictedPlayerState.origin, es.origin) < ENTITY_USE_RANGE - 0.2f // use an epsilon to account for rounding errors
+				&& CG_IsOnMyTeam(es) )
 		{
-			//hack to prevent showing the usable buildable when you aren't carrying an energy weapon
```
bmorel, line 1064: to me, this checks is what advertise reloading energy weapons at reactor. I do not like the idea of removing this feature, it's handy

bmorel:
Ping. Are there still stuff to review on this?

src/cgame/cg_rocket_draw.cpp
necessarily-equal: idk, I'd say the advertizing is that the reactor is shown with a large ammo stack over it when you are out of ammo. This one is confusing IMO.

slipher:
Did anyone LGTM the range changes?

PR #1549, opened 2021-10-23
Fix the bot "leap then go back" bug

necessarily-equal:
Bots sometimes jump/leap/pounce forward, then go back on their track because their waypoint is in their back, before going forward again. This PR fixes this bug.


bmorel:
This is almost one year old. Is this still intended to be merged? I also have not seen that bug since many months (but I almost only play on betterai anyway)

PR #1550, opened 2021-10-23
Implement g_bot_{aliens,humans}AvoidStruct

necessarily-equal:
This is groundwork for a tutorial level. I guess it may be useful for a low-skill bot server too?

This implementation is quite dirty, I'm not not too happy with it either. Don't hesitate to suggest improvements.


bmorel:
Avoiding structs by hiding their navmesh won't always work, for 2 reasons:

* this relies on a bug that I intend to fix
* it depends on building's normals

Also, it means that someone on a building would be "hidden" from bots, and if that's a bot, it will be stuck.

bmorel:
ping.

Issue #1401, opened 2021-05-08
spectator's move speed depends on timescale


bmorel:
This is annoying because when looking at very accelerated games for testing (testing bot changes, trying to find the other side of the trail-bug, etc), it's hard or slow to control the spectator.
To reproduce, summon daemon with options: `+devmap plat23 +timescale 10`.

slipher:
Not a bug. If you prefer a different spectator movement speed you can set `cg_flyspeed`.

bmorel:
I see. Well, I still think it would be useful to have an option to make flyspeed scaling with timescale? Also, it could probably use at least a UI, or why not keybinds  (to increase and decrease by a step) with a tutorial info?

PR #1553, opened 2021-10-25
Make spectator move speed independant of the timescale

necessarily-equal:
made while documenting https://wiki.unvanquished.net/wiki/Modder_Documentation:_classes_.attr.cfg

I've added a fix for https://github.com/Unvanquished/Unvanquished/issues/1401.

### DIFF
--- a/src/shared/bg_pmove.cpp
+++ b/src/shared/bg_pmove.cpp
@@ -4847,17 +4847,20 @@ void PmoveSingle( pmove_t *pmove )
 
 	// determine the time
 	pml.msec = pmove->cmd.serverTime - pm->ps->commandTime;
-
-	if ( pml.msec < 1 )
-	{
-		pml.msec = 1;
-	}
-	else if ( pml.msec > 200 )
-	{
-		pml.msec = 200;
-	}
-
 	pm->ps->commandTime = pmove->cmd.serverTime;
+
+	if ( pm->ps->pm_type == PM_SPECTATOR || pm->ps->pm_type == PM_NOCLIP )
+	{
+		// HACK: ignore the effects of timescale for spectators
+		// see https://github.com/Unvanquished/Unvanquished/issues/1401
+		// we don't touch pm->ps->commandTime, because other parts of the code rely on it
+		static int last_ms;
+		auto ms = Sys::Milliseconds();
+		pml.msec = ms - last_ms;
+		last_ms = ms;
+	}
+	pml.msec = Math::Clamp( pml.msec, 1, 200 );
+
 
 	// save old org in case we get stuck
 	VectorCopy( pm->ps->origin, pml.previous_origin );


src/shared/bg_pmove.cpp
```diff
@@ -252,30 +252,24 @@ Handles both ground friction and water friction
...
 		vec[ 2 ] = 0; // ignore slope movement
 	}
+	float speed = VectorLength( vec );
 
-	speed = VectorLength( vec );
-
-	if ( speed < 0.1f )
+	if ( speed <= 0.0f )
 	{
-		VectorSet(vel, 0.0f, 0.0f, 0.0f);
```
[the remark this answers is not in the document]
illwieckz: How can speed be equal to 0.0f and spectator still move? Or was the moving bug about negative speed?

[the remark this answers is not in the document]
necessarily-equal: This if is only about preventing a negative speed to make it all explode. I've basically restored it to the previous version

necessarily-equal:
rebased

necessarily-equal:
ping @bmorel

src/shared/bg_pmove.cpp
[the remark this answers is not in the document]
slipher: Could we just drop this commit. It's consuming a lot of mental effort and in doesn't make things noticeably nicer

src/shared/bg_pmove.cpp
```diff
@@ -1737,21 +1730,23 @@ static void PM_WaterMove()
...
 
-	scale = PM_CmdScale( &pm->cmd, true );
+	float scale = PM_CmdScale( &pm->cmd, true );
 
 	//
 	// user intentions
 	//
 
-	for ( i = 0; i < 3; i++ )
+	vec3_t wishvel;
```
slipher, line 1739: I think you could get rid of one of wishvel or wishdir in all these functions

src/shared/bg_pmove.cpp
```diff
@@ -4847,17 +4847,20 @@ void PmoveSingle( pmove_t *pmove )
...
 
-	if ( pml.msec < 1 )
-	{
-		pml.msec = 1;
-	}
-	else if ( pml.msec > 200 )
+	if ( pm->ps->pm_type == PM_SPECTATOR || pm->ps->pm_type == PM_NOCLIP )
 	{
-		pml.msec = 200;
+		// HACK: ignore the effects of timescale for spectators
```
slipher, line 4854: Instead of this, find where cg_flyspeed is plugged in and do the scaling there

necessarily-equal: Well, I did that at first. Notice that this is in bg, so that you need to handle the custom code with `#ifdef` directives, but since you can't access cvars so you can't even do that easily. And for a smooth experience you should also find a way to compensate the other params that influence moving, like acceleration and friction. So the whole ends up being a much bigger mess.

slipher: It's not related to BG. You can do it like this:
```
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -632,7 +632,7 @@ void SpectatorThink( gentity_t *ent, usercmd_t *ucmd )
                        client->ps.pm_flags |= PMF_QUEUED;
                }

-               client->ps.speed = client->pers.flySpeed;
+               client->ps.speed = client->pers.flySpeed / atof(Cvar::GetValue("timescale").c_str());
                client->ps.stats[ STAT_STAMINA ] = 0;
                client->ps.stats[ STAT_FUEL ] = 0;
                client->ps.stats[ STAT_MISC ] = 0;
@@ -1994,7 +1994,7 @@ void ClientThink_real( gentity_t *self )
        // set speed
        if ( client->ps.pm_type == PM_NOCLIP )
        {
-               client->ps.speed = client->pers.flySpeed;
+               client->ps.speed = client->pers.flySpeed / atof(Cvar::GetValue("timescale").c_str());
        }
        else
        {
```
Yeah the friction is different. Who cares? BUT, I would not accept this patch in the current form, since it would trigger a lot of VM->engine syscalls to get the value.

We really need a cvar interface that lets a VM observe a cvar owned by the engine. Currently the VM can only subscribe to updates on cvars created by it.

necessarily-equal: Well, the version I'm suggesting has no such "syscall" problem :)

src/shared/bg_pmove.cpp
necessarily-equal: well, that'll be for another day

necessarily-equal:
I've merged the first 3 commits manually so this can be about nitpicking the spectator thing. I still think the current code is fine :)

src/shared/bg_pmove.cpp
bmorel: I'll give a try at the code tomorrow to see if that works. If yes, then LGTM

bmorel: That was indeed a veeeeeery long night. I checked, and it works, so as I said, I'm fine with this.

slipher:
Well, the proposed code is completely wrong as `PmoveSingle` may execute many times per frame. Measuring the time between its executions with a stopwatch does not give you the time since the last frame. It only appears to work because you tested with only one spectator at a time. 

Not particularly convinced that #1401 is a bug

PR #1554, opened 2021-10-25
Bots try to flee fast

bmorel:
* humans run when their target is friendly (repair, heal, resupply)
* mantis pounce to flee
* advanced dragoon and dragoon pounce to flee
* tyrants charge to flee

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -417,13 +417,8 @@ void G_BotThink( gentity_t *self )
 	self->botMind->cmdBuffer = self->client->pers.cmd;
 	botCmdBuffer = &self->botMind->cmdBuffer;
 
-	bool wasSprinting = usercmdButtonPressed( botCmdBuffer->buttons, BUTTON_SPRINT );
 	//reset command buffer
 	usercmdClearButtons( botCmdBuffer->buttons );
-	if ( wasSprinting )
-	{
-		usercmdPressButton( botCmdBuffer->buttons, BUTTON_SPRINT );
-	}
 
 	// for nudges, e.g. spawn blocking
 	bool hasNudge = botCmdBuffer->doubleTap != dtType_t::DT_NONE;
@@ -475,12 +470,18 @@ void G_BotThink( gentity_t *self )
 		//BotClampPos( self );
 	}
 
+	self->botMind->willSprint( false ); //let the BT decide that
 	self->botMind->behaviorTree->run( self, ( AIGenericNode_t * ) self->botMind->behaviorTree );
 
 	// if we were nudged...
 	VectorAdd( self->client->ps.velocity, nudge, self->client->ps.velocity );
 
+	// ensure we really want to sprint or not
 	self->client->pers.cmd = self->botMind->cmdBuffer;
+	self->botMind->doSprint(
+			BG_Class( self->client->ps.stats[ STAT_CLASS ] )->staminaJumpCost,
+			self->client->ps.stats[ STAT_STAMINA ],
+			self->client->pers.cmd );
 }
 
 void G_BotSpectatorThink( gentity_t *self )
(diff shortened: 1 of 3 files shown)


src/sgame/sg_bot.cpp
```diff
@@ -445,12 +445,29 @@ void G_BotThink( gentity_t *self )
...
 	// if we were nudged...
 	VectorAdd( self->client->ps.velocity, nudge, self->client->ps.velocity );
 
+	// ensure we really want to sprint or not
 	self->client->pers.cmd = self->botMind->cmdBuffer;
+	int const* stats = self->client->ps.stats;
+	int jumpCost = BG_Class( stats[ STAT_CLASS ] )->staminaJumpCost;
+	int stamina = stats[ STAT_STAMINA ];
+
+	self->botMind->m_exhausted = self->botMind->m_exhausted || ( self->botMind->botSkill.level >= 5 && stamina <= jumpCost + jumpCost / 10 );
```
necessarily-equal, line 460: I find this more readable (assuming you want to keep the current logic):
```suggestion
	if ( self->botMind->botSkill.level >= 5 && stamina <= jumpCost + jumpCost / 10 )
	{
		self->botMind->m_exhausted = true;
	}
```

But I'm wondering if you really need to set the variable twice: the current logic seems equivalent to this:
```C++
	self->botMind->m_exhausted = stamina <= jumpCost * 2 || ( self->botMind->botSkill.level >= 5 && stamina <= jumpCost + jumpCost / 10 );
```

Which is in turn equivalent to this:
```suggestion
	self->botMind->m_exhausted = stamina <= jumpCost * 2 || ( self->botMind->botSkill.level >= 5 && stamina <= jumpCost * 1.1f );
```

I think instead of jumpCost, you intended to use the skill, but that, assuming I did read right until now, would be equivalent to:
```suggestion
	self->botMind->m_exhausted = stamina <= jumpCost * 2;
```

So maybe we can just go with that for now?

src/sgame/sg_bot_nav.cpp
```diff
@@ -323,13 +323,12 @@ void BotStandStill( gentity_t *self )
 
 bool BotJump( gentity_t *self )
 {
-	int staminaJumpCost;
-
 	if ( self->client->pers.team == TEAM_HUMANS )
 	{
-		staminaJumpCost = BG_Class( self->client->ps.stats[ STAT_CLASS ] )->staminaJumpCost;
+		int const* stats = self->client->ps.stats;
```
necessarily-equal, line 328: I would use either:

```suggestion
		auto &stats = self->client->ps.stats;
```
or
```suggestion
		int (&stats)[ MAX_STATS ] = self->client->ps.stats;
```

src/sgame/sg_bot_nav.cpp
```diff
@@ -342,29 +341,38 @@ bool BotJump( gentity_t *self )
 bool BotSprint( gentity_t *self, bool enable )
 {
 	usercmd_t *botCmdBuffer = &self->botMind->cmdBuffer;
-	int       staminaJumpCost;
+	int const* stats = self->client->ps.stats;
```
necessarily-equal, line 344: ditto

src/sgame/sg_bot.h
```diff
@@ -107,14 +109,16 @@ struct botMemory_t
...
 	vec3_t      futureAim;
-	usercmd_t   cmdBuffer;
 	botNavCmd_t nav;
 
 	int lastThink;
 	int stuckTime;
 	vec3_t stuckPosition;
 
 	int spawnTime;
+	//avoid relying on buttons to remember what AI was doing
```
necessarily-equal, line 119: I think the intent was to make the fields private, so maybe something like this?

```suggestion

friend void G_BotThink( gentity_t *self );

private:
	// avoid relying on buttons to remember what AI was doing
```

src/sgame/sg_bot.cpp
```diff
@@ -445,12 +445,29 @@ void G_BotThink( gentity_t *self )
...
+	if ( !self->botMind->m_exhausted && self->botMind->m_wantSprinting )
+	{
+		usercmdPressButton( self->client->pers.cmd.buttons, BUTTON_SPRINT );
+	}
+	else
+	{
+		usercmdReleaseButton( self->client->pers.cmd.buttons, BUTTON_SPRINT );
+	}
+
+	self->botMind->m_exhausted = self->botMind->m_exhausted && stamina <= jumpCost * 2;
```
necessarily-equal, line 470: (see the previous thread)

necessarily-equal (changes requested):

src/sgame/sg_bot_nav.cpp
```diff
@@ -697,15 +703,6 @@ void BotMoveToGoal( gentity_t *self )
...
-			//dont sprint or dodge if we dont have enough stamina and are about to slow
-			if ( self->client->ps.stats[ STAT_STAMINA ] < staminaJumpCost )
-			{
-				usercmdReleaseButton( botCmdBuffer.buttons, BUTTON_SPRINT );
-
-				// walk to regain stamina
-				BotWalk( self, true );
-			}
-
 			// except if target is friendly, then let's reach it quickly
```
slipher, line 706: comment here doesn't make sense now

src/sgame/sg_bot.h
slipher: Nah no private fields in mega structs please. Also these should not use an `m_` prefix since none of the other members of this struct do that.

slipher (commented):
Could you explain what is the "moonwalk bug"? It seems that if you press both the sprint and walk keys at the same time, the walk key simply takes precedence.

src/sgame/sg_bot.h
bmorel: well, on my side of the code, I did have this private indeed, and the prefix is an heritage of that, I'll fix that.

src/sgame/sg_bot_nav.cpp
bmorel: The «moonwalk bug» is something you can only see on bots, when they are out of stamina: they constantly switch between walk/run modes, resulting in them moving slower than if they were walking.
The bug is named after how the BSuit users look like, I don't remember who named it.

It already exists in codebase, but can't be seen because currently bots only run when in fight, and the moves are then too chaotic to notice this easily.

src/sgame/sg_bot.cpp
bmorel: Have you tried your proposed changes?

slipher:
LGTM

PR #1557, opened 2021-10-25
rewrite rushScore

bmorel:


src/sgame/sg_bot_util.cpp
```diff
@@ -197,6 +208,67 @@ const gentity_t *BotGetHealTarget( gentity_t *self )
...
+					// This will break when more equipments will use multiple slots,
+					// but the game does not (yet) have easy way to handle those.
+					if ( armor.canBuyNow() && max_item_val < armor.price() )
+					{
+						max_item_val = armor.price();
+						canUseBackpack = armor.item != UP_BATTLESUIT;
+					}
+				}
+
+				if( canUseBackpack && g_bot_radar.integer && BG_UpgradeUnlocked( UP_RADAR ) )
```
slipher, line 249: You may need to check BG_UpgradeDisabled also (there are cvars to generally disable an item as well as to disable it for only bots). Actually this may be missing from the bot shopping code too.

It would make sense to add the disabled check into BG_UpgradeUnlocked.

bmorel: Yes at checking `BG_UpgradeDisabled` I guess, and yes it's missing from bot shopping too (but that's not a regression I introduced, it was already present).

bmorel: Also, this likely applies not only to upgrades, but also to weapons and classes.

bmorel: Note that I opened an issue (#1586) on that as a reminder: this problem does not belongs to this PR, or to any PR I already opened. If the problem is confirmed (which is likely) then it exists since years, and I think it requires a separated PR (which should fix more than just upgrades).

PR #1558, opened 2021-10-25
medkit usage belongs to behavior trees

bmorel:


slipher:
So maybe I just misunderstood the behavior tree the whole time? Now it seems like the behavior tree is actually executing from the start every time... the only "memory" is whether a particular node returned STATUS_RUNNING in the previous frame - some bot actions behave differently depending on if they are newly starting or were running in the previous frame.

So in that case, it makes since that healing behavior would indeed be triggered during other actions. There is still a problem though, that interrupting an action may have negative consequences, as @necessarily-equal has found.

bmorel:
> Now it seems like the behavior tree is actually executing from the start every time... 

This is how I always understood it working.
It's a tree, you start at root, explore each nodes. The action nodes (the leaves) can be running or not. They can return 3 values when tried: failure, success, running. I have not experimented with it (yet) but I think this is how the parallel (or something like that) node works: continue as long as actions return success or running.
But, well, I only read the code months ago, I could be wrong. The only thing I can say is, so far, my experiments always confirmed it.

> So in that case, it makes since that healing behavior would indeed be triggered during other actions. There is still a problem though, that interrupting an action may have negative consequences, as @necessarily-equal has found.

I'll need to verify this, but in any cases, this can be fixed rather easily.

The solution would be to place the `condition healthPercent < 0.5 { activate MEDKIT }` outside of alertedtoenemy block, and put a `decorator return success { condition healthPercent < 0.5 { activate MEDKIT } }` in the alertedtoenemy block. There may be other solutions, though.
Of course, this would also require testing, which is not exactly an easy thing to do when it come to BT.

What seems strange to me is that I never spotted such behavior, but since this change is in a lot of other changes, including in a cleaner BT (on my branch, the BT was split anew between aliens and humans, cleaned a lot, improved to avoid using big area weapons in base and to use grenades, notably) maybe the BT PR I submitted is not exactly identical (in semantic) to what I have.

PR #1562, opened 2021-10-26
duck when stuck

bmorel:
This allows bots to go in vents, if the navmesh shows them the way.
Yes, this is a one-liner, yes it works, and yes it's hackish.

This will require daemonmap to properly generate navmeshes, though, so @illwieckz might be willing to reserve this for a pre-release process.


bmorel:
Ok, there *is* a bug for now. I built a branch will all my current PRs on top of master and got it hosted on my servers, and it shows that the ducking thing have a problem. I'll investigate on it, for now I'm moving this as draft.

bmorel:
I won't merge.

Issue #1514, opened 2021-09-27
Credit accounting isn't reset properly on team change


necessarily-equal:
[EDITED] Other entities' `credits[ my_clientnum ]` isn't reset properly on team change, this leads to ~~odd credit accounting where it mixes your team and the others, so you may be penalized for damaging your team~~ odd team assist messages.

~~It may be triggered only by the juggernaut mod, though.~~

necessarily-equal:
To expand a bit, this is the kind of message displayed:
```
●[bot] Bot#2 was machinegunned by ●Antoine [bsuit]  <-- this is with a blue team color
●[bot] Bot#2 was machinegunned by ●Antoine [bsuit]  <-- this is with a blue team color
Antoine [bsuit] left the humans and joined the aliens
●[bot] Bot#4 was machinegunned by ●[bot] Bot#5; ●Antoine [bsuit] assisted  <-- this is with a blue team color
●[bot] Bot#3 was machinegunned by ●[bot] Bot#5; ●Antoine [bsuit] assisted  <-- this is with a blue team color
]/quit
```

There is some protection to avoid giving credits in such case (https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_combat.cpp#L254), but this is still confusing, and might provoke some oddities in team-switch intensive contexts (like the jugg mod).

PR #1564, opened 2021-10-27
Reset credit accounting on team leave/change

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1514.


bmorel:
Is there is a way to reproduce the bug this PR is meant to fix?

necessarily-equal:
This is indeed not the correct way to fix this issue. If ones want to clear it, one should reset the credit accounting array of the *other* people. I'll amend #1514 with more information.

PR #1569, opened 2021-10-29
cg_main: do trap_UpdateScreen after setting cgs.processedSnapshotNum, fix crash introduced in #1470

illwieckz:
I got game crash when doing `vid_restart`.

It is assumed https://github.com/DaemonEngine/Daemon/issues/540 is a symptom of that bug.


illwieckz:
The bug this PR is fixing is that the game crashes because in #1470 I wanted to add a `trap_UpdateScreen()` early to draw the loading screen as soon as possible but I added that `trap_UpdateScreen()` call too early, so this PR is making this call less early to avoid the crash.

PR #1570, opened 2021-10-30
Cleanup bot headers

necessarily-equal:
This limits the subset of bot API sgame can use. G_Bot* functions represent the "public" part of bot functions. The point is to allow for smaller compile time when touching bot data structure or when adding a function.


src/sgame/sg_struct.h
```diff
@@ -25,7 +25,7 @@ along with Unvanquished Source Code.  If not, see <http://www.gnu.org/licenses/>
 #ifndef SG_STRUCT_H_
 #define SG_STRUCT_H_
 
-#include "sg_entities.h"
+#include "sg_local.h"
```
slipher, line 28: sg_local.h includes sg_struct.h so this doesn't seem good

src/sgame/sg_bot_public.h
```diff
@@ -0,0 +1,65 @@
...
+======================
+src/sgame/sg_bot.h
+
+This file contains the public interface used by sgame to interact with bots.
+======================
+*/
+
+#include "sg_struct.h"
+
+#ifndef __BOT_PUBLIC_HEADER
```
slipher, line 34: Identifiers beginning with double underscore or underscore and capital letter are reserved for the compiler/stdlib

slipher (commented):
There's a convention in the codebase to use `*_local.h` for 'private' headers, and we already have a `bot_local.h`...

src/sgame/sg_bot_public.h
necessarily-equal: I must say I've copied that from `sg_bot.h`

src/sgame/sg_struct.h
necessarily-equal: fixed

necessarily-equal:
> There's a convention in the codebase to use *_local.h for 'private' headers, and we already have a bot_local.h...

That's true. But it's in the bot/ directory, and thus represent what used to be local to the engine. I'd like to refactor this a bit more, moving what should be so we have proper naming conventions here, but freem would hate me if I start moving files around too much :P

I can definitely rename `sg_bot.h` to `sg_bot_local.h` though.

src/sgame/sg_bot_public.h
slipher: Two wrongs don't make a right

src/sgame/sg_bot_public.h
```diff
@@ -0,0 +1,65 @@
...
+You should have received a copy of the GNU General Public License
+along with Unvanquished; if not, write to the Free Software
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+===========================================================================
+*/
+
+/*
+======================
+src/sgame/sg_bot.h
```
slipher, line 28: Remove this line? Especially since it doesn't match the actual filename :)

slipher (commented):
Yeah renaming sg_bot.h to sg_bot_local.h would work, or at least add a comment in it contrasting it with sg_bot_public.h

PR #1575, opened 2021-10-31
Allow bots to read "bool" cvars

necessarily-equal:


necessarily-equal:
Example testing method: edit `unvanquished_src.dpkdir/bots/default.bt`, and test sucessively a cvar of each type on line 3 (example: `condition cvar("g_friendlyFireHumanMultiplier") > 0.2`). If bots freeze and unfreeze, the condition worked.

src/sgame/sg_bot_parse.cpp
```diff
@@ -279,32 +279,32 @@ static AIValue_t randomChance( gentity_t*, const AIValue_t* )
...
+	}
+	if ( Cvar::ParseCvarValue( Cvar::GetValue( cvar ), floating )  )
 	{
-		Log::Warn("could not read cvar '%s' as float", cvar);
-		num = 0.0f;
+		return AIBoxFloat( floating );
 	}
 
-	return AIBoxFloat( num );
+	Log::Warn("Bot: could not read cvar '%s'", cvar);
```
slipher, line 306: I suggest appending *as number* to the message. LGTM otherwise

necessarily-equal: Fixed. Merging when pipeline is ready.

PR #1576, opened 2021-11-01
Bring back bot behavior cvars

necessarily-equal:
see https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/72

### DIFF
--- a/src/sgame/sg_extern.h
+++ b/src/sgame/sg_extern.h
@@ -203,6 +203,9 @@ extern Cvar::Cvar<bool> g_bot_level4;
 
 // misc bot cvars
 extern Cvar::Cvar<bool> g_bot_attackStruct;
+extern Cvar::Cvar<bool> g_bot_heal;
+extern Cvar::Cvar<bool> g_bot_repair;
+extern Cvar::Cvar<bool> g_bot_rush;
 extern Cvar::Cvar<float> g_bot_fov;
 extern Cvar::Cvar<int> g_bot_chasetime;
 extern Cvar::Cvar<int> g_bot_reactiontime;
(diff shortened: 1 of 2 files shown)


slipher:
I don't really like this pattern of defining cvars in the code that aren't written or read by the code at all. You can create new cvars on the fly, so it shouldn't be necessary to "declare" them here.

You could write something in the behavior tree that read potentially-unset variables and just used the resulting 0 value, except that it would print warnings. Maybe you could add an optional 2nd argument to the `cvar` BT function to provide a default value if the cvar is not defined.

Gireen:
it would still be nice to have cvars listed, with description and autocomplete.
Could we define cvars in unvanquished.dpkdir ?
Maybe with a .cfg file setting the defaults and loaded before autogen.cfg

necessarily-equal:
I think both proposals are reasonable: slipher's is simpler; and Gireen's, while more complex, brings a bit more polish to the user experience.

Noting that adding a checkbox like I did that lazy-loads cvars from the BT would not break (the ui will create the cvar on the fly), but it would lose the default value.

I personaly prefer it this way, as simple is better. Patches welcome though

Le 1 novembre 2021 23:38:07 GMT+01:00, Gireen ***@***.***> a écrit :

slipher:
Maybe it would make sense to instantiate the cvar when the BT using it is run? I see the proposed change only modifies the `default` behavior, but there are other ones that use `heal`.

And I question the assumption that every cvar should have a UI. At some point adding more options makes the UI less useful, because it's hard to find the frequently-used options among all the obscure stuff.

Completely different proposal: Implement the cvar by having BotActionHeal immediately return STATUS_FAILURE if it's set. Then there would be no need for BT changes.

bmorel:
I think introducing cvars to alter BT's actions is not a good idea, because this can make the behavior trees failing in weird ways.
Introducing cvars in the BT is a bit less bad, but then, why not simply have a different BT? BT, unlike C++ code, does not requires build time, and is rather easy to host. In fact, I consider BT as some sort of configuration, and I don't really think a cvar configuration system to change the BT configuration system is a good idea.

necessarily-equal:
So, can we take a decision about this?

bmorel:
My opinion on this has not changed, those CVars have a higher risk of bringing in bugs and hard to understand behaviors than anything positive.

PR #1581, opened 2021-11-02
cg_tutorial: display the medkit and ammo feedback in orange

illwieckz:
Not perfect, but better than what we have.

[tutorial medkit red message](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial-medkit-red-message/unvanquished_2021-11-02_235331_000.jpg)

[tutorial medkit red message](https://dl.illwieckz.net/b/unvanquished/bugs/tutorial-medkit-red-message/unvanquished_2021-11-02_235906_000.jpg)

### DIFF
--- a/src/cgame/cg_tutorial.cpp
+++ b/src/cgame/cg_tutorial.cpp
@@ -159,6 +159,8 @@ static entityState_t *CG_BuildableInRange( playerState_t *ps, float *healthFract
 	return &es;
 }
 
+#define COLOR_ALARM "^8" // Orange
+
 /*
 ===============
 CG_BuilderText
@@ -362,6 +364,7 @@ static void CG_HumanText( char *text, playerState_t *ps )
 			case WP_CHAINGUN:
 			case WP_SHOTGUN:
 			case WP_FLAMER:
+				Q_strcat( text, MAX_TUTORIAL_TEXT, COLOR_ALARM );
 				Q_strcat( text, MAX_TUTORIAL_TEXT,
 				          _( "Find an Armoury for more ammo\n" ) );
 				break;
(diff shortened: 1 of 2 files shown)


src/cgame/cg_tutorial.cpp
```diff
@@ -159,6 +159,8 @@ static entityState_t *CG_BuildableInRange( playerState_t *ps, float *healthFract
 	return &es;
 }
 
+#define COLOR_ALARM "^1" // Red
```
necessarily-equal, line 162: In orange maybe? 

```suggestion
#define COLOR_ALARM "^8" // Red
```

Otherwise, LGTM

necessarily-equal (approved):

src/cgame/cg_tutorial.cpp
bmorel: Instead of discussing colors, make it a client side cvar and call it a day? In any any cases, LGTM too.

bmorel:
Some thinking (I'm still fully ok to merge this in current state): one step further would likely be to apply color to the medkit icon too, so that even without tutorial shown, the info would be there (I know that I regularly try to use absent medkit, just in case I still have it, and basically use the warning message to know if it worked or not...).
I now know how to do that (my solution is probably not perfect though, since it implies to have one file per color variant, which is suboptimal), and will send a PR when this one will be merged, so that I can use the same color.

Also, the ammo color change should likely be done the same moment the «beacon hint» is shown, maybe with several colors, like, yellow when low on ammo, and red when there's no ammo remaining at all (current behavior).
This would not help when tutorial is not shown, though, for that, ideally the ammo and clip bars should change color or, maybe even start blinking?

bmorel:
Any reason to not merge this?

slipher:
LGTM

src/cgame/cg_tutorial.cpp
illwieckz: > In orange maybe?

Good idea, it may be more readable, done.

illwieckz:
> Also, the ammo color change should likely be done the same moment the «beacon hint» is shown, maybe with several colors, like, yellow when low on ammo, and red when there's no ammo remaining at all (current behavior).

I have looked at it at the time I submitted the PR and it looks like detecting a weapon is low on ammo is tricky (especially, it's per-weapon specific), while “empty” is universal. Of course more can be done later.

PR #1583, opened 2021-11-03
Drop bot dodge support

necessarily-equal:
The "dodge" mechanism was removed years ago.

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -631,12 +631,6 @@ AINodeStatus_t BotActionAlternateStrafe( gentity_t *self, AIGenericNode_t* )
 	return STATUS_SUCCESS;
 }
 
-AINodeStatus_t BotActionClassDodge( gentity_t *self, AIGenericNode_t* )
-{
-	BotClassMovement( self, BotTargetInAttackRange( self, self->botMind->goal ) );
-	return STATUS_SUCCESS;
-}
-
 AINodeStatus_t BotActionChangeGoal( gentity_t *self, AIGenericNode_t *node )
 {
 	AIActionNode_t *a = ( AIActionNode_t * ) node;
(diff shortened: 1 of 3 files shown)


slipher:
This doesn't actually seem to have any relation to double-tap dodge?

necessarily-equal:
I would assume the "dodge" mechanism was implemented as a special class movement (just like the tyrant class movement is running). So that would be why. In any case, it currently is dead code.

slipher:
I checked the history and the code was not changed since it was introduced. BotClassDodge never did anything for humans.

It's not really dead code since it's exposed to the behavior tree scripting language.

necessarily-equal:
I would say that it never was changed, because nobody noticed that it 
wasn't useful anymore. I think when people removed dodge they just 
didn't notice this was left.

slipher:
Nothing was removed from the code this action uses.

> I checked the history and the code was not changed since it was introduced. BotClassDodge never did anything for humans.

Gireen:
Its not useless 😐
Look at what the action does. It lets aliens attack the goal. 
It would probably be good to change the name to something more fitting.

necessarily-equal:
Well, it's not quite useless for aliens, but then it's a bad interface. Goons and tyrants need to charge, and they have no way of doing that. That's why aliens do it in other functions. Like, this action really was designed for the human dodge 😅

Also, I still think it can be considered dead code, as it is never ever called, and I'd be pretty surprised if there is a mod out there calling it

Le 4 novembre 2021 23:16:19 GMT+01:00, Gireen ***@***.***> a écrit :

slipher:
Mind showing your evidence that this is somehow connected to double-tap dodge?

necessarily-equal:
I still don't think this function has any usefulness, but shrug

necessarily-equal:
(It is indeed unrelated to double tab dodge though)

PR #1589, opened 2021-11-07
Refactor things a bit to drop code

necessarily-equal:


src/sgame/sg_bot_util.cpp
```diff
@@ -309,10 +309,9 @@ float BotGetBaseRushScore( gentity_t *ent )
 	// return average value on which other parameters can weight
 	float rush_score = 0.5;
 
-	float self_value = static_cast<float>( ent->client->ps.persistant[PERS_CREDIT]
-		+ BotValueOfWeapons( ent )
-		+ BotValueOfUpgrades( ent )
-		+ BG_Class( ent->client->ps.stats[ STAT_CLASS ] )->price );
+	float self_value = static_cast<float>(
+		ent->client->ps.persistant[PERS_CREDIT]
+			+ BG_GetPlayerPrice( ent->client->ps ) );
```
bmorel, line 314: Does not work. the class is missing.

necessarily-equal: I don't think it is

bmorel: right, it's done with that function. Still not convinced, though, since current code is team agnostic, and the proposed patch is not.

bmorel: LGTM anyway

bmorel: LGTM

necessarily-equal:
merging because good enough and all that, but I'll keep it in mind should we have an inventory rework

PR #1590, opened 2021-11-07
Revert ai heal

### LINKED ISSUE #1588: AI try to heal with buildables on ceilings
since #1532, alien AI try to heal 1st with ceiling leeches.
While the idea seems good at first, considering the fact AI don't know the heights, it makes aliens run toward walls endlessly, which is a considerable regression.

bmorel:
fixes #1588

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -213,15 +213,10 @@ const gentity_t *BotGetHealTarget( gentity_t *self )
 		// powered booster
 		return self->botMind->closestBuildings[BA_A_BOOSTER].ent;
 	}
-	else
-	{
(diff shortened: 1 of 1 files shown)


slipher:
LGTM

bmorel:
> LGTM

Actually it's not: some other commits sneaked in there. The only commit I wanted to get merged is the one which only lets barricade and overmind to be considered worth checking, for the obvious reason that those can only be built on ground.
Let me rebase.

bmorel:
Even with just barricades and overmind, the change is still insufficient. A real revert is likely required.

slipher:
> Even with just barricades and overmind, the change is still insufficient.

How so? They still run into walls or something?

bmorel:
they keep doors open

PR #1593, opened 2021-11-11
Fix crash

necessarily-equal:
Dereferencing a null pointer is not such a great idea :S

Includes a submodule sync, because why not I guess


necessarily-equal:
merging as freem confirmed it fixes the bug

PR #1594, opened 2021-11-11
Increase radius of granger spit fire extinguishing

necessarily-equal:


necessarily-equal:
how to test:

1. give human bots flamer
2. spawn as granger
3. find fire on the ground
4. spit on it

slipher:
It turns out the feature actually already exists and works on master. Try this patch to get a debug log:

```
diff --git a/src/sgame/sg_missile.cpp b/src/sgame/sg_missile.cpp
index 87f87cdac..f1719a3ab 100644
--- a/src/sgame/sg_missile.cpp
+++ b/src/sgame/sg_missile.cpp
@@ -253,7 +253,8 @@ static int ImpactSlowblob( gentity_t *ent, trace_t *trace, gentity_t *hitEnt )
        {
                if ( neighbor != hitEnt )
                {
-                       neighbor->entity->Extinguish( ABUILDER_BLOB_FIRE_IMMUNITY );
+                       if (neighbor->entity->Extinguish(ABUILDER_BLOB_FIRE_IMMUNITY))
+                               Log::Notice("extinguished something");
                }
        }
```

necessarily-equal:
@slipher: fixed

For the record, the default size of 100 is the visual size of the pseudo-creep the blob creates on ground.

slipher:
What do you mean "fixed"? What I'm saying is the code actually already works, so no pull request is needed.

necessarily-equal:
I've updated the PR to use a non-almost-zero radius for the "put out" 
zone, and make that configurable in a CVar.

slipher:
100 seems too large of a range (I made #1603 to visualize the radius of the fire damage zone which is 60). IMO 40 would be good.

sgame cvars conventionally begin with `g_`, not `w_`.

necessarily-equal:
I've set it to 100 because this is the size of the blob on the ground. So it seems coherent to whatever is displayed.

Another angle of approach is to see how much would be appreciable in-game, and 40 seems a too low here. 20 wasn't even noticed as doing something even by players who explicitely tried to do it. I'd like the granger to feel like it is actually able to fight fire :P If you think 100 is too powerful, maybe 75 can be a better option?

bmorel:
considering it's now a cvar, I'd be okay with any value.

Except for the cvar's naming, LGTM

slipher:
> I've set it to 100 because this is the size of the blob on the ground. 

No way, that's much larger. The decal has is of a similar or a bit smaller size to the flame radius (which is 60).

Gireen:
I think the x5 increase is to much and i m not sure if a increase is needed at all.
If know that fire on the ground can be extinguished and where the center of flames is it is relative easy to do.
But players don't know its possible or where to aim for, at least i didn't.

It could be useful to display some goals/beacons in the fire to all grangers as assists.

slipher:
> If know that fire on the ground can be extinguished and where the center of flames is it is relative easy to do.

It may be difficult to tell where the center of the flames is, since the flame sprites (something like 14 of them per flame) are positioned randomly, up to a considerable distance away. When there are multiple flame objects, it is hard to discern the locations of individual ones. So I think the original value of 20 is too small.

bmorel:
The current value of 20 is ridiculous indeed, I even didn't knew it was doable at all!
If 100 is too big a value, and 20 too small, maybe use 64? This value have 2 niceties: it's a power of 2, and I love those, and it's close to ( 100 + 20 ) / 2.

bmorel:
LGTM but I suppose better to wait for some other approvals on this

slipher:
Range is reasonable now, just the cvar prefixes need to be fixed

bmorel:
@slipher : I did the `s/w_/g_/` changes. To squash now.

slipher:
LGTM

bmorel:
ah shit... wanted to squash... damn that github editor... but otherwise can't move on...

PR #1595, opened 2021-11-13
Fix corpse team: ps.clientNum is actually reused for the corpse model

necessarily-equal:
The function is not supposed to be used on a corpse anywhere AFAIK, but let's avoid a potential bug.


slipher:
https://github.com/Unvanquished/Unvanquished/pull/1337#discussion_r676034376 :stuck_out_tongue: 
(LGTM)

necessarily-equal:
I remerber that message :S

PR #1596, opened 2021-11-13
Drop custom Com_Memset and Com_Memcpy

necessarily-equal:
They are only hiding compiler warnings, which makes debugging harder


slipher:
LGTM

PR #1597, opened 2021-11-14
Drop typedefs

necessarily-equal:
Remove the typedef where they are not used for anything more than forward declaration, and turn them into `using` directives otherwise.


slipher:
LGTM, thanks. These `_s` were always annoying when I clicked "Go to definition" in Visual Studio :stuck_out_tongue:

PR #1598, opened 2021-11-14
Make bot healing fail when medistation is busy

necessarily-equal:
Backport a fix by @bmorel

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1083,19 +1083,32 @@ static AINodeStatus_t BotActionReachHealH( gentity_t *self )
 {
 	vec3_t targetPos;
 	vec3_t myPos;
+	VectorCopy( self->s.origin, myPos );
 
 	if ( G_Team( self ) != TEAM_HUMANS )
 	{
 		return STATUS_FAILURE;
 	}
 
-	self->botMind->goal.getPos( targetPos );
-	VectorCopy( self->s.origin, myPos );
+	auto const& goal = self->botMind->goal;
+	const gentity_t * medistation = goal.getTargetedEntity();
+
+	goal.getPos( targetPos );
 	targetPos[2] += BG_BuildableModelConfig( BA_H_MEDISTAT )->maxs[2];
 	myPos[2] += self->r.mins[2]; //mins is negative
+	float distanceSquared = DistanceSquared( myPos, targetPos );
+
+	// If medistation is busy, do something else until can go on it anew.
+	// See https://github.com/Unvanquished/Unvanquished/pull/1598
+	// (It would be nice to allow the BT to check for the failure cause.
+	//  How? That's a good question)
+	if ( medistation->target && medistation->target.get() != self )
+	{
+		return STATUS_FAILURE;
+	}
 
 	//keep moving to the medi until we are on top of it
-	if ( DistanceSquared( myPos, targetPos ) > Square( BG_BuildableModelConfig( BA_H_MEDISTAT )->mins[1] ) )
+	if ( distanceSquared > Square( BG_BuildableModelConfig( BA_H_MEDISTAT )->mins[1] ) )
 	{
 		BotMoveToGoal( self );
 	}


slipher:
This could only make sense if the bot is already close to the medistation. If it is far away, then the medistation's current status says little about whether it will be open when the bot arrives.

necessarily-equal:
For the record, this kind of things happen without this, where you can witness most bots waiting to heal while they, or their base, get attacked by aliens.

image
image


Needless to say humans tend to loose fairly quickly when that happens.

slipher:
How about failing the action if the medistation is busy AND the bot's distance from it is less than 200?

necessarily-equal:
> How about failing the action if the medistation is busy AND the bot's distance from it is less than 200?

Should be nice enough. Done

slipher:
LGTM

slipher:
What's wrong with it?

bmorel:
No idea what's wrong with this, except maybe the fact it does not fail if bot is too far :)

For info, the way I use the original feature is to make bot roam in circle around the closer medistat, so the distance check is not really useful.

bmorel:
@necessarily-equal , @slipher I intend to merge this soon (next week) since this PR being a non-merged draft for no reason blocks improvements.
Anything against that?

necessarily-equal:
I would like to keep the BT thing 
(https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/86) 
unmerged until I can check it actually works, but I'd say that we could 
merge this one *with* the "don't fail if too close" reverted — which 
is what I just pushed.

I really think there is a bug lurking in 
https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/86 
though, so I'll have to test.

bmorel:
> "don't fail if too close" reverted

I don't remember having seen such thing. Don't fail if too close would indeed be a bug, since it's definitely possible to have 2 players on a medistation.
Only the check on medistation's target being someone else can mean the medistation is busy, other stuff can only be about it being unpowered (or the distance thing slipher asked for, but that defines a status_running until close enough, so as long as close enough is away enough, things are fine).

> I really think there is a bug lurking in UnvanquishedAssets/unvanquished_src.dpkdir#86 though, so I'll have to test.

I would be curious to see which bug you suspect. The only thing I can see there might be because of changing goals too often, and if it's proven, the fix will be trivial: put the roam in a `decorator delay 2000` block.

bmorel:
Marking this as ready for review, and will merge it on sunday except if there is something more than "suspicion of bug".

bmorel:
@necessarily-equal said "LGTM" on irc.

bmorel:
This has been waiting for more than enough time and received approval from the author anyway.

slipher:
The part about the distance being greater than 200 disappeared. This should be reverted

slipher:
https://github.com/Unvanquished/Unvanquished/commit/48a132e9315145eb44c0aafc7973071df9f3972d is the version I LGTMed

bmorel:
Then I can just add back that patch I guess

bmorel:
I can revert, or #1723 can be merged.

necessarily-equal:
I've removed the distance check from the C++, and said I was ok merging this without the 
distance check because 
https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/86 makes 
it useless anyway. I realize this didn't make it clear I had updated the branch, so sorry for that. However, I believe that if you checkout every branch on master, there is no regression, so there is no need to revert ?

Le 30 janvier 2022 00:09:42 GMT+01:00, "Morel Bérenger" ***@***.***> a écrit :

slipher:
https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/86 does not make the distance check useless.

necessarily-equal:
Could you explain how so?

Le 30 janvier 2022 22:34:18 GMT+01:00, slipher ***@***.***> a écrit :

slipher:
If the distance thing is removed then the healing action fails arbitrarily far from the medistation.

bmorel:
I agree that this is like bots cheating, but the distance suffers the same problems as usual in any cases: don't check walls. Should better do a IsVisible() now that I think about it.
In any cases, the PR I do to sync the submodule do the requested distance change.

PR #1599, opened 2021-11-14
Draw emoticons and colors in player's usernames

necessarily-equal:


illwieckz (approved):
LGTM

necessarily-equal:
Fixes this bug:
image


I couldn't test this, because connecting to a custom server with a dll is hard.

PR #1600, opened 2021-11-14
Eradicate old-style cvars (vmCvar_t)

slipher:


src/sgame/sg_client.cpp
```diff
@@ -1126,8 +1126,7 @@ const char *ClientConnect( int clientNum, bool firstTime )
 	// check for a password
 	value = Info_ValueForKey( userinfo, "password" );
 
-	if ( g_password.string[ 0 ] && Q_stricmp( g_password.string, "none" ) &&
-	     strcmp( g_password.string, value ) != 0 )
+	if ( g_needpass.Get() && !Str::IsIEqual( g_password.Get(), value ) )
```
necessarily-equal, line 1129: Do we really want to check passwords in a case-insensitive way?

```suggestion
	if ( g_needpass.Get() && g_password.Get() != value )
```

src/cgame/cg_main.cpp
```diff
@@ -417,21 +372,9 @@ CG_UpdateCvars
...
-		{
-			trap_Cvar_Update( cv->vmCvar );
-		}
-	}
-
-	// check for modifications here
 	CG_SetPVars();
 	CG_UpdateBuildableRangeMarkerMask();
-}
+ }
```
necessarily-equal, line 377: :eyes:
```suggestion
}
```

necessarily-equal (approved):
LGTM

src/sgame/sg_client.cpp
slipher: Nope!

necessarily-equal (approved):
looked fairly rapidly, LGTM

PR #1602, opened 2021-11-16
Introduce a new way to iterate entities

necessarily-equal:
The point of this PR is to indroduce a new way to iterate entities, that will be more convenient and more extensible, and hopefully has a lower running cost (this really would need to be mesured though).

Here is an example of how it could be used:

```C++
gentity_t *FindOvermind()
{
    for ( gentity_t *ent : iterate_buildable_entities )
    {
        if ( ent->s.modelindex == BA_A_OVERMIND )
            return ent;
    }
    return nullptr;
}
```

Each individual commit in the history compiles, in case this would be to be merged in several steps. I think this needs some extensive testing. From a quick check everything seems to work.

Further work that could be done in the following PRs:
* add more `iterate_*_entities`
* use std::function to allow for any function-like predicates, and consider providing functions to things such as `for ( gentity_t *ent :iterate_client_entities_in_team(TEAM_ALIENS) )`
* maybe introduce a `iterate_entity_with_component<>` and see if we can replace `ForEntities`
* replace the `G_IterateEntities{,OfClass,WithinRadius,WithField}` functions

### DIFF
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -26,6 +26,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 #include "sg_local.h"
 #include "shared/parse.h"
+#include "sg_entities_iterator.h"
 #include "Entities.h"
 #include "CBSE.h"
 #include "backend/CBSEBackend.h"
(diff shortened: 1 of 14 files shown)


slipher:
We already have e.g. `MedipadComponent::GetAll()` if you want to iterate through every `MedipadComponent*`. Or if you want to iterate over `Entity`s instead of components, a wrapper could be made around those iterators. There is no need to linearly search through all entities, or to write custom code for specific component types.

necessarily-equal:
> There is no need to linearly search through all entities, or to write custom code for specific component types.

I mean, yes, but on the other side, it's the simplest and the most versatile solution, as it allows checking for everything, and not just testing for the presence of one component. This means that this (or a variation of this) will allow rewriting every entity iterating code, like the `G_IterateEntities` and the `ForEntities<SmthComponent>`. **This means less code that solves the same problem**, in a way that is more readable and more concise than the alternatives.

In case you are concerned about performance, I did collect runtime performance data using a profiler (the one used in Intel vTune). The profiling did last 20s while the game was started. With the 34 instances of `iterate_*` present here in this PR, several of which have been called several times a frame, the profiler did stop only a handful of time in some function of the new files, making the run time barely mesurable, and accounting for (very roughly) "1ms" out of 20s, which is less than 0.01% of the run time. **The performance impact is thus negligible**.

In a way, there is no need to iterate linearly through all the entities, but on the other hand, there is no need to keep a list of every component an entity may have.

slipher:
iterate_entities, iterate_client_entities, and iterate_buildable_entities are all *exact duplicates* of functionality provided by ForEntities<Something>. We do not need yet another API that does the same thing as existing ones.

necessarily-equal:
The point is to replace and supersede the other API in the next PR though. There is currently 3 way of iterating entities in the game, `G_Interate*Entities`, `ForEntities` and the good old `for()` loop. This would be able to supersede them all, in subsequent PR 

For example,
1. using `std::function` instead of function pointers to take some extra params. This which would allow replacing functions like `G_IterateEntitiesInRadius`. 
2. and templates to replace ForEntities).

The first could look like this: `for ( gentity_t *ent : iterate_entities_in_radius(origin, ARMORY_USE_DISTANCE) )`, and the second like this: `for ( gentity_t *ent : iterate_entities_with_component<SmthComponent>() )`.

src/sgame/sg_main.cpp
```diff
@@ -1135,17 +1135,7 @@ void CalculateRanks()
 
 		if ( level.clients[ clientNum ].pers.connected != CON_DISCONNECTED )
 		{
```
bmorel, line 1138: Dropping this loop is obviously a good idea.

bmorel (commented):
The 1st 2 commits are fine, you could likely cherry pick them.

src/sgame/sg_main.cpp
necessarily-equal: Just cherry-picked them as 55254c94e61609a3fb7aec1ec1bdbe8d7ded5903 and f765dd7966da0c63a76d556ae7997806c8f4ae8e

DolceTriade:
I think having a newer unified API to replace ForEntities would be good (ie, the last commit in this chain should delete ForEntities).

DolceTriade (approved):

PR #1603, opened 2021-11-17
cg_drawBBOX - show ground fires

slipher:
A sphere is drawn around ET_FIRE entities showing the damage zone.


necessarily-equal (approved):

necessarily-equal:
I would make it a bit more transparent, like `0.2`. Otherwise, LGTM.

slipher:
> I would make it a bit more transparent, like `0.2`. Otherwise, LGTM.

Done.

PR #1611, opened 2021-11-24
Introduce BoundedVector and use it in some places

necessarily-equal:
The implementation is a bit primitive, for example it doesn't provide a .erase() method. Otherwise I think this is useful due to the large number of variable-size arrays that can't really use std::vector because other parts of the code rely on their capped size. I guess the absence of malloc can also be a nice touch for performance-constrained code, but this wasn't really optimized for that usage.

More parts of the code could use it, but this PR is large enough as is

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -217,8 +217,7 @@ void G_BotChangeBehavior( int clientNum, const char* behavior )
 
 bool G_BotSetBehavior( botMemory_t *botMind, const char* behavior )
 {
-	memset( botMind->runningNodes, 0, sizeof( botMind->runningNodes ) );
-	botMind->numRunningNodes = 0;
+	botMind->runningNodes.clear();
 	botMind->currentNode = nullptr;
 	memset( &botMind->nav, 0, sizeof( botMind->nav ) );
 	BotResetEnemyQueue( &botMind->enemyQueue );
@@ -517,8 +516,7 @@ void G_BotSpectatorThink( gentity_t *self )
 	memset( &self->botMind->nav, 0, sizeof( self->botMind->nav ) );
 	self->botMind->futureAimTime = 0;
 	self->botMind->futureAimTimeInterval = 0;
-	self->botMind->numRunningNodes = 0;
-	memset( self->botMind->runningNodes, 0, sizeof( self->botMind->runningNodes ) );
+	self->botMind->runningNodes.clear();
 
 	if ( self->client->sess.restartTeam == TEAM_NONE )
 	{
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -35,28 +35,22 @@ static AITreeList_t treeList;
 Bot management functions
 =======================
 */
-static struct
-{
-	int count;
-	struct
-	{
-		char *name;
-		bool inUse;
-	} name[MAX_CLIENTS];
-} botNames[NUM_TEAMS];
+struct nameInfo_t {
+	std::string name;
+	bool inUse;
+};
+static BoundedVector<nameInfo_t, MAX_CLIENTS> botNames[NUM_TEAMS];
 
 static void G_BotListTeamNames( gentity_t *ent, const char *heading, team_t team, const char *marker )
 {
-	int i;
-
-	if ( botNames[team].count )
+	if ( !botNames[team].empty() )
 	{
 		ADMP( heading );
 		ADMBP_begin();
 
-		for ( i = 0; i < botNames[team].count; ++i )
+		for ( nameInfo_t &nameInfo : botNames[team] )
 		{
-			ADMBP( va( "  %s^* %s", botNames[team].name[i].inUse ? marker : " ", botNames[team].name[i].name ) );
+			ADMBP( va( "  %s^* %s", nameInfo.inUse ? marker : " ", nameInfo.name.c_str() ) );
 		}
 
 		ADMBP_end();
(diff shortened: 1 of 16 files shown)


src/sgame/sg_bot.cpp
```diff
@@ -71,68 +65,47 @@ void G_BotListNames( gentity_t *ent )
...
 					goto next;
 				}
 			}
 		}
 
-		botNames[team].name[i].name = ( char * )BG_Alloc( strlen( name ) + 1 );
-		strcpy( botNames[team].name[i].name, name );
-
-		botNames[team].count = ++i;
+		botNames[team].append({ .name = name, .inUse = false });
```
slipher, line 108: You can't use C++20 features

src/sgame/sg_maprotation.cpp
```diff
@@ -124,11 +118,11 @@ G_MapExists
 Check if a map exists
 ===============
 */
-bool G_MapExists( const char *name )
+bool G_MapExists( const std::string &name )
```
slipher, line 121: `Str::StringRef` would be better

src/shared/bg_public.h
```diff
@@ -1525,6 +1525,159 @@ meansOfDeath_t            BG_MeansOfDeathByName( const char *name );
...
+	void clear() {
+		while (pop_back()) {}
+	}
+	bool append(T elem) {
+		if (count_ == maximum_size_)
+			return false;
+		array[count_++] = std::move(elem);
+		return true;
+	}
+	size_t count() const { return count_; }
```
slipher, line 1675: This should be called `size`

src/shared/bg_public.h
```diff
@@ -1525,6 +1525,159 @@ meansOfDeath_t            BG_MeansOfDeathByName( const char *name );
...
+		return *this;
+	}
+	~BoundedVector() {
+		while (pop_back());
+	}
+
+	class iterator
+	{
+	private:
+		BoundedVector<T, maximum_size_>* vector_;
```
slipher, line 1568: You can do without an iterator class, by having the iterator be a `T*` pointing to the element.

slipher:
Oh I think you said somewhere that you introduced a new class-memaccess warning. Please don't do that!

src/shared/bg_public.h
necessarily-equal: I'm not fond of calling it size, as the size of the vector is constant (in that it always takes as much memory space).

src/shared/bg_public.h
necessarily-equal: Neat!

src/sgame/sg_struct.h
```diff
@@ -74,10 +74,10 @@ struct gentityConditions_t
 	team_t   team;
 	int      stage;
 
-	class_t     classes[ PCL_NUM_CLASSES ];
-	weapon_t    weapons[ WP_NUM_WEAPONS ];
-	upgrade_t   upgrades[ UP_NUM_UPGRADES ];
-	buildable_t buildables[ BA_NUM_BUILDABLES ];
+	BoundedVector<class_t,     PCL_NUM_CLASSES>   classes;
```
slipher, line 77: To expand on why introducing a class-memaccess warning is bad: the warning doesn't tell you which field of the struct generated the error, so you have to manually inspect all the fields to see which one it is and check whether it is a true positive.

src/shared/bg_public.h
```diff
@@ -1525,6 +1525,69 @@ meansOfDeath_t            BG_MeansOfDeathByName( const char *name );
...
+		}
+	}
+	BoundedVector<T, maximum_size_>& operator=(const BoundedVector<T, maximum_size_>& other) {
+		for (size_t i = 0; i < other.count_; i++) {
+			array[i] = other.array[i];
+		}
+		count_ = other.count_;
+		return *this;
+	}
+	~BoundedVector() {
```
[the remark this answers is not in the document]
necessarily-equal: I think this is not true for C-style arrays. You would need to use `std::array` for this

src/shared/bg_public.h
necessarily-equal: updated

src/sgame/sg_struct.h
necessarily-equal: Fixed

src/sgame/sg_main.cpp
```diff
@@ -524,7 +524,10 @@ void G_InitGame( int levelTime, int randomSeed, bool inClient )
 	G_admin_readconfig( nullptr );
 
 	// initialize all entities for this game
-	memset( g_entities, 0, MAX_GENTITIES * sizeof( g_entities[ 0 ] ) );
+	for ( int i=0; i < MAX_GENTITIES; i++ )
+	{
+		new(&g_entities[i])gentity_t{};
```
slipher, line 529: So the placement new seems to crash nacl-clang. You can try parentheses instead of curly braces there - I think that worked to stop MSVC crashes in the past...

src/shared/bg_public.h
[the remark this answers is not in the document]
slipher: Yes it is true. `Foo a[3];` is no different from `Foo a0; Foo a1; Foo a2;` in terms of destruction.

src/sgame/sg_bot.cpp
```diff
@@ -35,28 +35,22 @@ static AITreeList_t treeList;
...
-	{
-		char *name;
-		bool inUse;
-	} name[MAX_CLIENTS];
-} botNames[NUM_TEAMS];
+struct nameInfo_t {
+	std::string name;
+	bool inUse;
+};
+static BoundedVector<nameInfo_t, MAX_CLIENTS> botNames[NUM_TEAMS];
```
slipher, line 42: I don't think there's any reason for this to be fixed-size. It's no worse than the current code though, so I'm not demanding to rewrite.

src/sgame/sg_main.cpp
necessarily-equal: src/cgame/cg_main.cpp line 1118 contains `new(&cgs) cgs_t{}; // Using {} instead of () to work around MSVC bug`. Which is why I've used `{}` here

slipher: Well, that was a long-obsolete MSVC version so we shouldn't have to worry about it anymore...
Anyway gentity_t is only 3K in size, so we don't need to worry about it overflowing the stack. `g_entities[i] = {}` would work

src/shared/bg_public.h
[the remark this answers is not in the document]
necessarily-equal: Apparently, it is also equivalent in terms of construction, at least from testing with my implementation (GNU GCC + GNU libstdc++)

src/sgame/sg_bot.cpp
necessarily-equal: The initial reason was that the gentity_t was being memset. A reason that is not valid anymore indeed.

PR #1615, opened 2021-11-28
makes unvanquished more UBsan friendly

bmorel:
1st commit just remove empty lines.
2nd commit replaces a hack to get addresses offsets with a standard compiler feature doing the same: `offsetof`


slipher:
LGTM

PR #1617, opened 2021-12-04
yellow blueprints when place is valid but reqs are not met

bmorel:
Requires a PR into unvanquished_src.dpkdir before merging.
This allows to have 3 ghostly colours for blueprints:

* green: place is ok, pre-requisites are ok
* yellow: place is ok, pre-requisites are not (BP, momentum, RC/OM absent, etc etc)
* red: place is not ok

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -1129,7 +1129,8 @@ void CG_GhostBuildable( int buildableInfo )
 	vec3_t        mins, maxs;
 	trace_t       tr;
 	float         scale;
-	buildable_t   buildable = (buildable_t)( buildableInfo & SB_BUILDABLE_MASK ); // assumed not BA_NONE
+	buildable_t   buildable = (buildable_t)( buildableInfo & SB_BUILDABLE_MASK );
+	ASSERT( BA_NONE < buildable && buildable < BA_NUM_BUILDABLES );
 	const buildableModelConfig_t *bmc = BG_BuildableModelConfig( buildable );
 
 	ps = &cg.predictedPlayerState;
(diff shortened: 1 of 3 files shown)


src/cgame/cg_buildable.cpp
```diff
@@ -1129,7 +1129,8 @@ void CG_GhostBuildable( int buildableInfo )
 	vec3_t        mins, maxs;
 	trace_t       tr;
 	float         scale;
-	buildable_t   buildable = (buildable_t)( buildableInfo & SB_BUILDABLE_MASK ); // assumed not BA_NONE
```
slipher, line 1132: The assertion should be on `buildable` not `buildableInfo`, right? Also `ASSERT_NQ`

src/cgame/cg_buildable.cpp
```diff
@@ -1156,8 +1157,19 @@ void CG_GhostBuildable( int buildableInfo )
 
 	ent.hModel = cg_buildables[ buildable ].models[ 0 ];
 
-	ent.customShader = ( SB_BUILDABLE_TO_IBE( buildableInfo ) == IBE_NONE )
-	                     ? cgs.media.greenBuildShader : cgs.media.redBuildShader;
+	int mask = SB_BUILDABLE_TO_IBE( buildableInfo );
```
slipher, line 1160: Why is it called `mask`? It's not like a set of bit flags

src/cgame/cg_buildable.cpp
bmorel: Now that I look anew at it, yes, this would make more sense. Looks like buildableInfo is really a "struct" with both the buildable_t and some flags in it...

src/cgame/cg_buildable.cpp
bmorel: I'm not saying I understand what IBE thingy is, but I probably grepped and found this line: ` #define SB_BUILDABLE_TO_IBE(x)   ( ( ( x ) & SB_BUILDABLE_STATE_MASK ) >> SB_BUILDABLE_STATE_SHIFT )` which likely made me think it really is a mask (yes: probably and likely, I don't remember, I worked on other stuff since then).

src/cgame/cg_buildable.cpp
slipher: assert not equal

slipher:
LGTM except the submodule should be updated

src/cgame/cg_buildable.cpp
bmorel: Ok. Well, I replaced it with a range anyway, which makes more sense (as long as NUM_BUILDABLE is really the last one). Or there's also a specific assert for that? (that seems a bit overkill to have specific asserts for those kind of stuff though)

Issue #1202, opened 2020-05-24
Move `BUTTON_` definitions into gamelogic


slipher:
The constants like `BUTTON_ATTACK`, `BUTTON_ACTIVATE` etc. should be moved from the engine into the gamelogic.
A few of them are referenced in the engine for some reason but we can at least move all the other ones.

DolceTriade:
Yeah, it seems probably that a lot of the engine logic that requires these buttons could be moved to the gamelogic as well. The main thing that seems concerning is something like VoIP...

bmorel:
The only ones engine actually use are TALK/ANY/WALKING. It could very likely be fixed, I'm not sure.
Functions which uses those symbols only writes the values into `cmd->buttons`. They are:

* void CL_KeyMove( usercmd_t *cmd )
* void CL_JoystickMove( usercmd_t *cmd )
* void CL_GameControllerMove( usercmd_t *cmd )
* void CL_CmdButtons( usercmd_t *cmd )

One can see comments such as:

```
  // adjust for speed key / running
  // the walking flag is to keep animations consistent
  // even during acceleration and deceleration
```

Code in CL_JoystickMove() and CL_GameControllerMove() seems mostly duplicated, and likely could be factored to be put in a common block with CL_KeyMove(), and I still doubt setting the flag is really required to be done in engine.

BUTTON_ANY is *never* read, and BUTTON_TALK is only checked in PmoveSingle( pmove_t *pmove ), but seems actually useful.

bmorel:
Considering the fact that:

> A few of them are referenced in the engine for some reason but we can at least move all the other ones.

has been done, I'm closing this, despite being likely that things could be definitely moved out of engine.

PR #1618, opened 2021-12-04
make gamelogic responsible for more BUTTON/btn definitions

bmorel:
This improves situation on #1202
Only (useful) remaining BUTTON_xxx occurrences in daemon are
BUTTON_WALKING, BUTTON_TALK and BUTTON_ANY, which are used to define
values for their btn_xxx equivalents.
Renaming from BUTTON_xxx to btn_xxx was done to avoid having any symbol
conflict between daemon and unvanquished.

### DIFF
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -584,7 +584,6 @@ void CG_InitConsoleCommands()
 		trap_AddCommand( commands[ i ].cmd );
 	}
 
-	// Defined in src/engine/qcommon/q_shared.h, see BUTTON_ATTACK etc.
 	trap_RegisterButtonCommands(
 	    // 0      123     45      6        78       9       ABCD           E      <- bit nos.
 	      "attack,,,taunt,,sprint,activate,,attack2,attack3,,,,deconstruct,rally"
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -584,6 +584,7 @@ void CG_InitConsoleCommands()
 		trap_AddCommand( commands[ i ].cmd );
 	}
 
+	// defined in enum buttonNumber_t in file src/shared/bg_public.h
 	trap_RegisterButtonCommands(
 	    // 0      123     45      6        78       9       ABCD           E      <- bit nos.
 	      "attack,,,taunt,,sprint,activate,,attack2,attack3,,,,deconstruct,rally"
--- a/src/shared/bg_public.h
+++ b/src/shared/bg_public.h
@@ -1127,9 +1127,23 @@ enum meansOfDeath_t
 // TODO: move other button definitions into gamelogic
 enum buttonNumber_t
 {
-  BUTTON_ATTACK3 = 9,
-  BUTTON_DECONSTRUCT = 13,
+	btn_attack       = 0,
+	// defined by daemon
+	btn_talk         = BUTTON_TALK, // = 1, // disables actions
+	btn_use_holdable = 2,
+	btn_gesture      = 3,
+	// defined by daemon
+	btn_walking      = BUTTON_WALKING, // = 4, // walking can't just be inferred from MOVE_RUN
+	btn_sprint       = 5,
+	btn_activate     = 6,
+	// defined by daemon
+	btn_any          = BUTTON_ANY, // = 7, // if any key is pressed
+	btn_attack2      = 8,
+	btn_attack3      = 9,
+	btn_deconstruct  = 13,
+	btn_rally        = 14 // also named "taunt" in some places
 };
+
 
 #define DEVOLVE_RETURN_FRACTION 0.9f
 
--- a/src/shared/bg_public.h
+++ b/src/shared/bg_public.h
@@ -1127,21 +1127,21 @@ enum meansOfDeath_t
 // TODO: move other button definitions into gamelogic
 enum buttonNumber_t
 {
-	btn_attack       = 0,
+	BTN_ATTACK       = 0,
 	// defined by daemon
-	btn_talk         = BUTTON_TALK, // = 1, // disables actions
-	btn_use_holdable = 2,
-	btn_gesture      = 3,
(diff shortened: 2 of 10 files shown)


slipher:
I was thinking, for ones that Daemon doesn't define a specific purpose for, there could be definitions like `BUTTON_USR5` etc. which the gamelogic could define its constants based on. WDYT?

bmorel:
I had a quick read at the "purpose" of those in engine. To me, only BUTTON_TALK seems to be useful. BUTTON_ANY is *never* read at all, and BUTTON_WALK seems to simply be able to switch animation, which could be inferred from the actual speed.

So, I think it's just asking for troubles to have the engine defining them, except for the BUTTON_TALK, and... well, even this one, I'd require more read of the code: maybe it can be moved in gamelogic as well.

src/shared/bg_public.h
```diff
@@ -1127,10 +1127,24 @@ enum meansOfDeath_t
 // TODO: move other button definitions into gamelogic
 enum buttonNumber_t
 {
-  BUTTON_ATTACK3 = 9,
-  BUTTON_DECONSTRUCT = 13,
+	BTN_ATTACK       = 0,
+	// defined by daemon
+	BTN_TALK         = BUTTON_TALK, // = 1, // disables actions
+	BTN_USE_holdable = 2,
```
necessarily-equal, line 1133: This is still partly lowercase

src/shared/bg_public.h
```diff
@@ -1127,10 +1127,24 @@ enum meansOfDeath_t
...
+	// defined by daemon
+	BTN_WALKING      = BUTTON_WALKING, // = 4, // walking can't just be inferred from MOVE_RUN
+	BTN_SPRINT       = 5,
+	BTN_ACTIVATE     = 6,
+	// defined by daemon
+	BTN_ANY          = BUTTON_ANY, // = 7, // if any key is pressed
+	BTN_ATTACK2      = 8,
+	BTN_ATTACK3      = 9,
+	BTN_DECONSTRUCT  = 13,
+	BTN_RALLY        = 14 // also named "taunt" in some places
```
necessarily-equal, line 1144: (nitpick) Typically one adds a comma to help limiting future diffs
```suggestion
	BTN_RALLY        = 14, // also named "taunt" in some places
```

src/shared/bg_public.h
```diff
@@ -1127,10 +1127,24 @@ enum meansOfDeath_t
 // TODO: move other button definitions into gamelogic
 enum buttonNumber_t
 {
-  BUTTON_ATTACK3 = 9,
-  BUTTON_DECONSTRUCT = 13,
+	BTN_ATTACK       = 0,
+	// defined by daemon
+	BTN_TALK         = BUTTON_TALK, // = 1, // disables actions
```
necessarily-equal, line 1132: I find these confusing, how about changing the two lines to

```suggestion
	BTN_TALK         = BUTTON_TALK, // = 1 (redefined from daemon), disables actions
```

src/shared/bg_public.h
```diff
@@ -1127,10 +1127,24 @@ enum meansOfDeath_t
...
+	BTN_USE_holdable = 2,
+	BTN_GESTURE      = 3,
+	// defined by daemon
+	BTN_WALKING      = BUTTON_WALKING, // = 4, // walking can't just be inferred from MOVE_RUN
+	BTN_SPRINT       = 5,
+	BTN_ACTIVATE     = 6,
+	// defined by daemon
+	BTN_ANY          = BUTTON_ANY, // = 7, // if any key is pressed
```
necessarily-equal, line 1140: (ditto)

src/shared/bg_public.h
```diff
@@ -1127,10 +1127,24 @@ enum meansOfDeath_t
...
 {
-  BUTTON_ATTACK3 = 9,
-  BUTTON_DECONSTRUCT = 13,
+	BTN_ATTACK       = 0,
+	// defined by daemon
+	BTN_TALK         = BUTTON_TALK, // = 1, // disables actions
+	BTN_USE_HOLDABLE = 2,
+	BTN_GESTURE      = 3,
+	// defined by daemon
+	BTN_WALKING      = BUTTON_WALKING, // = 4, // walking can't just be inferred from MOVE_RUN
```
slipher, line 1136: Actually BUTTON_WALKING is used directly by pmove to control speed. And there isn't anything anywhere called MOVE_RUN. So better to remove this comment.

src/cgame/cg_consolecmds.cpp
```diff
@@ -584,7 +584,6 @@ void CG_InitConsoleCommands()
 		trap_AddCommand( commands[ i ].cmd );
 	}
 
-	// Defined in src/engine/qcommon/q_shared.h, see BUTTON_ATTACK etc.
```
slipher, line 587: Mention the new location?

bmorel: I'm not even sure what this comment is really meant to explain.

slipher: The nth item in the list is the name of the "plus" command for the nth button. So it explains where the buttons are defined.

src/shared/bg_public.h
```diff
@@ -1125,12 +1125,24 @@ enum meansOfDeath_t
 };
 
 // TODO: move other button definitions into gamelogic
```
slipher, line 1127: You did the TODO so delete it :)

bmorel: Technically, there are still some though?

src/cgame/cg_consolecmds.cpp
bmorel: Right, I'll try to make a better comment on that then.

src/shared/bg_public.h
slipher: You got all the ones that aren't used by the engine so there's nothing left that can be easily done.

bmorel:
I think all comments have been addressed.

slipher:
LGTM

PR #1619, opened 2021-12-05
Bot BT: add hysteresis node support

necessarily-equal:
Bots have terrible remembrance of the past. One way we do currently add some persistance is with a `sequence { condition (x>300); action }`, but it has the terrible downside of continuing the action until the bot dies, or the action succeed or fail. This will allow for finer grained control where one can tell the bot to do an action *until* the score value goes down, or the action succeeds.

Tested with the following BT (designed for human bots only):
```javascript
selector
{
        hysteresis 100 300 distanceTo( E_H_MEDISTAT )
        {
                action moveTo( E_H_MEDISTAT )
        }

        action moveTo( E_A_OVERMIND )
}

```

@bmorel hopefully you'll like it a lot

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -202,8 +202,9 @@ void G_BotChangeBehavior( int clientNum, const char* behavior )
 
 bool G_BotSetBehavior( botMemory_t *botMind, const char* behavior )
 {
-	botMind->runningNodes.clear();
 	botMind->currentNode = nullptr;
+	BotClearRunningNodesHysteresisState( &g_entities[clientNum] );
+	botMind->runningNodes.clear();
 	memset( &botMind->nav, 0, sizeof( botMind->nav ) );
 	BotResetEnemyQueue( &botMind->enemyQueue );
 
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -444,6 +444,22 @@ bool EvalConditionExpression( gentity_t *self, AIExpType_t *exp )
 	return false;
 }
 
+void BotClearRunningNodesHysteresisState( gentity_t *self )
+{
+	int clientNum = self - g_entities;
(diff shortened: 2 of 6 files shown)


Gireen:
Whats the difference to a normal condition?
```
selector
{
        condition ( distanceTo( E_H_MEDISTAT ) > 100 && distanceTo( E_H_MEDISTAT ) < 300 )
        {
                action moveTo( E_H_MEDISTAT )
        }

        action moveTo( E_A_OVERMIND )
}
```

necessarily-equal:
One question left to answer is: should we reset the hysteresis for a bot when the action failed or succeeded? I would tend to say yes, but this is harder than you would think.

necessarily-equal:
@Gireen: in your example, if the distance is greater than 300, the bot will go to the overmind, while in mine, the bot will go to the medipad, so it is different.

To explain better how it works:
* if a bot is further than 300, it will always go back
* if a bot is closer than 100, it will always go forward
* if a bot is at a distance between 100 and 300, it will continue doing what it was doing (going back if it was going back, going forward if it was going forward.

slipher:
I don't understand the intended semantics of the example.


>    if a bot is further than 300, it will always go back
>    if a bot is closer than 100, it will always go forward
>    if a bot is at a distance between 100 and 300, it will continue doing what it was doing (going back if it was going back, going forward if it was going forward.

Which way is "back" and which way is "forward"?
What happens if the distance is between 100 and 300 at the beginning?

bmorel:
Is there is any use case for that? I'm failing to see the point, I'm afraid.

necessarily-equal:
I've thought about this for some time, and I think this is quite similar to

```js
sequence {
    condition (distanceTo(E_A_OVERMIND) > 700)
    condition (distanceTo(E_A_OVERMIND) > 400) {
        action moveTo(E_A_OVERMIND)
    }
}
```

So this makes this branch not justified enough to stay open I think

PR #1621, opened 2021-12-06
Distinguish between VFS and homepath file access

slipher:
Currently there is only a single syscall, `FSFOpenFileMsg`, for reading files both in the pakpath and in the homepath (it checks the pakpath first). This is not a good situation; often you only want one or the other, and if you want to check both, you want the reverse order.

After this PR it will be like this:
- `FS::PathPak::ReadFile` or `trap_FS_OpenPakFile` (requires https://github.com/DaemonEngine/Daemon/pull/549) for pakpath-only reading
- `trap_FS_FOpenFile` for homepath-only access
- `G_FOpenGameOrPakPath` for homepath first and pathpak second reading

Pakpath-only access can already be emulated by checking the file's existence beforehand. Homepath-only read access will have to wait for 0.53.

Inspired by @bmorel 's report about the navmesh files in https://github.com/DaemonEngine/Daemon/issues/547.

### DIFF
--- a/src/cgame/cg_gameinfo.cpp
+++ b/src/cgame/cg_gameinfo.cpp
@@ -28,6 +28,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 // gameinfo.c
 //
 
+#include "common/FileSystem.h"
 #include "cg_local.h"
 
 //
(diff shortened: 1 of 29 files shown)


src/cgame/cg_gameinfo.cpp
```diff
@@ -120,31 +121,21 @@ CG_LoadArenasFromFile
...
+	std::string text = FS::PakPath::ReadFile( filename, err );
+	if ( err )
 	{
-		Log::Warn( "%sfile not found: %s", Color::ToString( Color::Red ), filename );
+		Log::Warn( "file not found: %s", filename );
 		return;
 	}
-
-	if ( len >= MAX_ARENAS_TEXT )
+	else if ( text.size() >= MAX_ARENAS_TEXT )
```
bmorel, line 131: no need for an else when the if block returns anyway

src/cgame/cg_rocket.cpp
```diff
@@ -77,23 +76,10 @@ void CG_Rocket_Init( glconfig_t gl )
 	Rocket_RegisterProperty( "locked-marker-color", "red", false, false, "color" );
 
 	// Preload all the menu files...
-	len = trap_FS_FOpenFile( rocket_menuFile.Get().c_str(), &f, fsMode_t::FS_READ );
+	std::error_code err;
+	std::string content = FS::PakPath::ReadFile( rocket_menuFile.Get() );
```
bmorel, line 80: error checking disappeared

src/cgame/cg_rocket.cpp
```diff
@@ -265,29 +251,12 @@ void CG_Rocket_Init( glconfig_t gl )
...
-	{
-		trap_FS_FCloseFile( f );
-		Sys::Drop( "File %s too long.", rocket_hudFile.Get() );
-	}
-
-	trap_FS_Read( text, len, f );
-	text[ len ] = 0;
-	text_p = text;
-	trap_FS_FCloseFile( f );
+	std::string text = FS::PakPath::ReadFile( rocket_hudFile.Get() );
```
bmorel, line 258: error checking disappeared

slipher:
Removal of error checking was intentional. The dependee Daemon pull request edits the filesystem code to make it so that an exception message including the filename will be automatically generated.

bmorel:
So there are exceptions to handle, too?

slipher:
The exceptions are allowed to be generated only in cases where the error is fatal. For environments with no exceptions support, we can if necessary adjust the SetErrorCode* stuff in filesystem to use Sys::Error instead of throwing.

bmorel:
Alright, LGTLM  then.

src/cgame/cg_rocket.cpp
```diff
@@ -77,23 +76,10 @@ void CG_Rocket_Init( glconfig_t gl )
 	Rocket_RegisterProperty( "locked-marker-color", "red", false, false, "color" );
 
 	// Preload all the menu files...
-	len = trap_FS_FOpenFile( rocket_menuFile.Get().c_str(), &f, fsMode_t::FS_READ );
+	std::error_code err;
```
necessarily-equal, line 79: Is this line really useful? (and if it is, why isn't it here in the next function?)

necessarily-equal (approved):
I haven't tested. From reading, I spotted one unused variable, otherwise LGTM.

src/cgame/cg_rocket.cpp
slipher: Removed the line.

PR #1622, opened 2021-12-08
Exec map configs before map_restart when possible

slipher:
@bmorel This makes it possible to load "mods" on a per-map basis. You can configure the server with `set g_mapConfigs map` and write `set fs_extrapaks <whatever>` in `config/map/<mapname>.cfg`. And then if you want the mods to be off for other maps, `set fs_extrapaks ""` in `config/map/default.cfg`.


Gireen:
I'm not sure if that is a good idea.
What happens with commands that only apply to the current map?
Like /bot or /g_disabledBuildables

slipher:
Um, right, it doesn't make sense!

PR #1626, opened 2021-12-11
make bot default skill in Cvars

bmorel:
This allows for more generic configurations, notably, if the rotation makes a different number of bots depending on the map, a single `bot fill X` is enough, while currently it requires one line per team.
The 2 other ones are because why not. I see no con at having things less hard-coded.

Originally, all were in a single commit, I've split it just in case.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -5716,7 +5716,7 @@ static bool BotAddCmd( gentity_t* ent, const Cmd::Args& args )
 	}
 	else
 	{
-		skill = BOT_DEFAULT_SKILL;
+		skill = g_bot_default_skill.Get();
 	}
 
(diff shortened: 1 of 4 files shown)


bmorel:
This PR (at least 1st commit) allows to define default bot skill level in config/server.cfg and to specify the number of bots per map in game/maprotation.cfg with a single line, instead or requiring duplication of the maprotation.cfg file.

slipher:
So in bot commands that take a name, you can put the magic string `*` which means the name should be taken from the stored name list, rather than what's specified in the command. It appears that you have made this magic string *configurable*. So if I do `bot names aliens alice bob` and `g_bot_default_namelist bmorel` and then `bot add bmorel aliens`, then a bot named either `alice` or `bob` would be added to the aliens... :thinking:

bmorel:
Tbh, I was just doing this one since I was near it, without thinking much about it. I can easily remove that commit (which includes the default behavior, but this does not makes much sense neither in the end)

slipher:
LGTM

PR #1630, opened 2021-12-12
make bot say where they suicide

bmorel:
This is a bit hackish, but should be safe, and could be handy when there will be a way to know a player's position.
Multiple improvements on this can be imagined though, such as placing beacons, which would be more handy, so that teammate would be able to  check the cause, in case theres a building preventing the moves.

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1047,6 +1047,10 @@ AINodeStatus_t BotActionHeal( gentity_t *self, AIGenericNode_t *node )
 
 AINodeStatus_t BotActionSuicide( gentity_t *self, AIGenericNode_t* )
 {
+	//hopefully this is enough digits for all maps.
+	char str[] = "I'm stuck at x=0123456789 y=0123456789 z=0123456789 !?!";
+	snprintf( str, sizeof( str ), "I'm stuck at x=%-.0f y=%-.0f z=%-.0f !?!", self->s.origin[0], self->s.origin[1], self->s.origin[2] );
+	G_Say( self, SAY_TEAM, str );
 	Entities::Kill( self, MOD_SUICIDE );
 	return AINodeStatus_t::STATUS_SUCCESS;
 }


slipher:
I think it would be better as a log rather than a say. It seems mainly interesting for developers and would be annoying in chat

bmorel:
One of the points is that, like this, players can move problematic buildings. In a log, they won't be able to do so.

slipher:
The coordinates are not human-readable. It takes dev-level knowledge to use them. Anyway, if someone is actually building they can simply see the bots with their own eyes, which would work better.

src/sgame/sg_bot_ai.cpp
```diff
@@ -1047,6 +1047,10 @@ AINodeStatus_t BotActionHeal( gentity_t *self, AIGenericNode_t *node )
 
 AINodeStatus_t BotActionSuicide( gentity_t *self, AIGenericNode_t* )
 {
+	//hopefully this is enough digits for all maps.
+	char str[] = "I'm stuck at x=0123456789 y=0123456789 z=0123456789 !?!";
+	snprintf( str, sizeof( str ), "I'm stuck at x=%-.0f y=%-.0f z=%-.0f !?!", self->s.origin[0], self->s.origin[1], self->s.origin[2] );
+	G_Say( self, SAY_TEAM, str );
```
necessarily-equal, line 1053: I'd use `G_LogPrintf` or `Log::Debug` here.

Generally speaking, `snprintf` seems fairly unidiomatic in Unvanquished and daemon. The usual way to do that is to use `va` or the newer `Str::Format`. Of course, with `G_LogPrintf` or `Log::Debug` you don't have this problem in the first place.

---

By the way, I think that you would benefit from defining a logger in the bot code. You can define this:

```C++
Log::Logger botLog = Log::Logger("bot", "", Log::Level::NOTICE);
```
Then use it like you would usually, except with `botLog.` instead of `Log::`. For example  `botLog.Debug("%i enemies", nbEnemies)`, `botLog.Warn("Unexepected something")`, `botLog.Notice("Some message")`

So that the day you want to debug stuff, you can change it to this:
```C++
Log::Logger botLog = Log::Logger("bot", "", Log::Level::DEBUG).WithoutSuppression();
```

(the `.WithoutSuppression()` is to disable the log spam rate-limiting.)

bmorel: coords are hardly readable, yes, and I would have preferred indication of the room, but I don't know how to get it. This, or to place a tag.

@necessarily-equal: as I said, I want players to move buildings which can cause such suicides, thus i think that, unlike for most messages which disappear way too fast anyway, this information can be useful.
Could be eggs that bots don't know how to move from, could be turrets or barricades placed so that they prevent bots to jump above them or steer around, etc. And it might not happen when someone is paying attention, too.

Gireen:
I see this only useful for debugging, so it should be optional and disabled by default.
Normal players and server admins are not interested in such informations and would be unnecessarily spammed with it.

bmorel:
Since the consensus is more or less against it, I guess I should close?

As I said, the intent it both to be able to debug, *and* to allow real players to move buildings which either block AIs or make them spawn on stuck places. Sure, raw coordinates are rather inefficient for that, beacons would likely be more useful, but I thought it was a start.

Team chat about suicide do give players a useful info though: since they are team chat, the area name where the suicide happens is also indicated, unlike what currently happens with the "[bot] Bot#2 was killed by noname".

PR #1631, opened 2021-12-12
start rotation immediately by default

bmorel:
Workaround to improve on https://github.com/DaemonEngine/Daemon/issues/552 (servers should just run maprotation instead of forcing a specific map).

The intent is to have an example configuration file which is less surprising, that is, starting immediately with the maprotation.


illwieckz:
I wonder if it would be possible to load the next map without having loaded a map first (so there would be no `map` command in `server.cfg` but `nextmap` one instead).

bmorel:
If this PR is not considered useful, I'm ok to close it. I'd like to try to only have one page of PRs...

slipher:
LGTM

PR #1635, opened 2021-12-17
GoalInRange automatically adds self's radius to dist

### LINKED ISSUE #1577: BotActionFlee is broken
steps to reproduce:
* make a BT that contains `action flee` (for example replacing the `action heal` in the `default.bt`)
* `/bot add * aliens 5 my_bt_name`
* notice how the bot will rush to the overmind, and will never consider this action done: it's stuck.

how to fix:
* copy or rather reuse code from BotActionMoveTo, so that the action succeeds if the bot is "close enough" from the buildable.

bmorel:


bmorel:
After further thinking, this code is still wrong for various reasons.

1. it iterates of entities to see if they're both the goal *and* close enough, which seems useless to me
2. it does not checks the entity is on the same plane, which is not ok in various cases, like, when the target is bellow, through a floor yet in range
3. it does not checks if there are walls or alike between self and target, which can be ok in some cases (like booster or armory, and I'm not even certain it is)

PR #1637, opened 2021-12-19
UI: allow different styles to show sprint status

### LINKED ISSUE #1390: Give "sprint" a visual clue
The only way to know you are sprinting is that your stamina reduces when you move. I think this could be improved, maybe by changing bar's colour, the same way as wall-walking have visual clue.

bmorel:
This requires a UI PR to be useful, but does not breaks anything without it.

Fixes #1390

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -1005,13 +1005,15 @@ class WallwalkElement : public HudElement
 {
 public:
 	WallwalkElement( const Rocket::Core::String& tag ) :
-			HudElement( tag, ELEMENT_ALIENS ),
-			isActive( false ) {}
+			HudElement( tag, ELEMENT_ALIENS )
+	{
+		SetActive( false );
+	}
 
-	void DoOnUpdate()
+	void DoOnUpdate() override
 	{
 		bool wallwalking = cg.snap->ps.stats[ STAT_STATE ] & SS_WALLCLIMBING;
-		if ( ( wallwalking && !isActive ) || ( !wallwalking && isActive ) )
+		if ( wallwalking != isActive )
 		{
 			SetActive( wallwalking );
 		}
@@ -1033,13 +1035,15 @@ class SprintElement : public HudElement
 {
 public:
 	SprintElement( const Rocket::Core::String& tag ) :
-			HudElement( tag, ELEMENT_HUMANS ),
-			isActive( false ) {}
+			HudElement( tag, ELEMENT_HUMANS )
+	{
+		SetActive( false );
+	}
 
-	void DoOnUpdate()
+	void DoOnUpdate() override
 	{
 		bool sprinting = cg.snap->ps.stats[ STAT_STATE ] & SS_SPEEDBOOST;
-		if ( ( sprinting && !isActive ) || ( !sprinting && isActive ) )
+		if ( sprinting != isActive )
 		{
 			SetActive( sprinting );
 		}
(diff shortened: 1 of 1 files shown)


slipher:
How would this be used?

bmorel:
like this: https://github.com/bmorel/unvanquished_src.dpkdir/commit/0e0f0fc5f942faccd5a5f42701643e52a5f8fd8e

src/cgame/cg_rocket_draw.cpp
```diff
@@ -1029,6 +1029,34 @@ class WallwalkElement : public HudElement
...
+		{
+			SetActive( sprinting );
+		}
+	}
+
+private:
+	void SetActive( bool active )
+	{
+		isActive = active;
+		SetClass( "active", active );
```
slipher, line 1052: The class should be set from the beginning instead of only after the first time the user sprints.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -1029,6 +1029,34 @@ class WallwalkElement : public HudElement
...
 };
 
+class SprintElement : public HudElement
+{
+public:
+	SprintElement( const Rocket::Core::String& tag ) :
+			HudElement( tag, ELEMENT_HUMANS ),
+			isActive( false ) {}
+
+	void DoOnUpdate()
```
slipher, line 1039: `override`

src/cgame/cg_rocket_draw.cpp
```diff
@@ -1029,6 +1029,34 @@ class WallwalkElement : public HudElement
...
+	{
+		bool sprinting = cg.snap->ps.stats[ STAT_STATE ] & SS_SPEEDBOOST;
+		if ( ( sprinting && !isActive ) || ( !sprinting && isActive ) )
```
slipher, line 1042: `sprinting != isActive`

bmorel: I admit I shamelessly copy-pasted code from wallwalking. Thus I applied your remarks to both (there's probably a better way to implement that, though).

src/cgame/cg_rocket_draw.cpp
```diff
@@ -1029,6 +1031,36 @@ class WallwalkElement : public HudElement
...
+			SetActive( sprinting );
+		}
+	}
+
+private:
+	void SetActive( bool active )
+	{
+		isActive = active;
+		SetClass( "active", active );
+		SetClass( "inactive", !active );
```
necessarily-equal, line 1057: Can you change the name to something more understandable? Maybe something like `running` and `sprinting`?

necessarily-equal (approved):
Some clarity thing, otherwise wfm.

src/cgame/cg_rocket_draw.cpp
bmorel: Ideally, I would prefer to merge this class and the wallwalking one, to avoid code duplication. To some extent, it could also be used for no-clip weapons (to disable showing the clip) and other elements like that, and renaming the "vars" would not help at that.

slipher: You could have a virtual function that returns a string for what CSS class to use. Store that instead of a bool, and then you could use any names and it generalizes to more than two states.

bmorel: yes, that would likely be the best solution, but I'm not really fond of the idea of playing too much with librocket. The current situation seems buggy to me (see #1651) but I am clueless about how to really fix it.

bmorel:
ping @slipher is it ok if I merge it in current state?

slipher:
LGTM

PR #1642, opened 2021-12-21
cleanup

bmorel:
The only change this does is to remove the number of clips for painsaw, which have very little use considering it has infinite ammo (and thus, no clips).

Should probably just *not* show the hud element for infinite ammo weapons or weapon with a single clip (flamer, luci, chaingun, lasgun).

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -330,30 +330,21 @@ public:
 		int           value;
 		playerState_t *ps = &cg.snap->ps;
 
-		switch ( BG_PrimaryWeapon( ps->stats ) )
+		if ( BG_Weapon( BG_PrimaryWeapon( ps->stats ) )->infiniteAmmo )
 		{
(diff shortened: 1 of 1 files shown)


necessarily-equal (commented):
How about keeping the switch, but showing an infinity text instead for non-build weapons with infinite ammo?
```
			case WP_BLASTER:
 			case WP_PAIN_SAW:
				if ( clips != -1 )
				{
					SetText( "∞" );
				}
				clips = -1;
				return;

			case WP_ABUILD:
			case WP_ABUILD2:
			case WP_HBUILD:
				if ( clips != -1 )
				{
					SetText( "" );
				}
				clips = -1;
				return;
```

Otherwise, you are probably right that the amount of clip shouldn't be displayed for weapons without clips. (Not sure that belongs to this PR, how you see fit.)

bmorel:
Even with infinity symbol, the switch is still useless, and still would broke when weapons properties are changed.

slipher:
LGTM

Issue #538, opened 2015-01-26
`listmaps` command allowed to be called by registered user 0  or 1 but not listed by `adminhelp`


illwieckz:
Hi, registered users level 0 and level 1 can call the command `listmaps` on a server, but `adminhelp` does not list this command.

* calling `adminhelp` command as registered user level 1 lists `listplayers` and `listlayouts` commands and both are allowed, and the hidden `listmaps` command is allowed too.
* calling `adminhelp` command as registered user level 0 lists `listplayers` command only and `listplayers` command is allowed but `listlayouts` command is not allowed, but the hidden `listmaps` command is allowed.

I don’t know if the bug is “_they must no be able to list maps_” or “_adminhelp must list this command_”, but there is a bug for sure. :wink:

DolceTriade:
listmaps is not an admin command.

illwieckz:
Hmm, it's strange to think we can forbid/allow some people to list layouts, but not to list maps…

DolceTriade:
Indeed. Team, what do you think about abandoning different command systems for admin commands and game commands? Is there any benefit to this separation?

Kangz:
+1

Viech:
Sounds like a good idea to me.

PR #1643, opened 2021-12-22
allow giving momentum to specific team without joining

bmorel:
This is not the cleaner implementation, but it works. I think it's not *much* of a problem if a cheat command is not cleanly implemented, and overall this whole file should probably be done the same way sg_cmds.cpp is done, as said in #538.
Still, this patch is useful as it reduces a lot the pain when doing tests (checking if events bound to momentum or if a specific change in AI works correctly, notably, I don't really see other reasons to use that).

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -593,7 +593,7 @@ void Cmd_Give_f( gentity_t *ent )
 	char     *name;
 	bool give_all = false;
 	float    amount;
-	team_t team = static_cast<team_t>( ent->client->pers.team );
+	team_t team = G_Team( ent );
 
 	if ( trap_Argc() < 2 )
 	{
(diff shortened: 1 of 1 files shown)


Gireen:
what happens if the `give` other commands are used as spectator?

bmorel:
Good question. I just tried a `give all`, and it simply behave as I would expect: makes credits at their maximum value.

src/sgame/sg_cmds.cpp
```diff
@@ -593,12 +593,13 @@ void Cmd_Give_f( gentity_t *ent )
 	char     *name;
 	bool give_all = false;
 	float    amount;
+	team_t team = static_cast<team_t>( ent->client->pers.team );
```
necessarily-equal, line 596:
```suggestion
	team_t team = G_Team(ent);
```

necessarily-equal (commented):
I would replace all the references to `TEAM_NONE` using `BG_IsPlayableTeam`, which is coherent with `BG_PlayableTeamFromString`.

The error reporting is bad, but to be honest, that's a problem of this function in the first place, and it can be done in another PR as this would require a refactor.

bmorel:
This function, and others in that file, do not just have bad error reporting actually. This stuff likely requires real work to get cleaned.

bmorel:
Are there still problems with this PR?

bmorel:
Any reason to not merge it?

necessarily-equal:
There is still this weird thing where once you use `G_IsPlayableTeam` 
and once `team != TEAM_NONE`. You can merge it anyway though

bmorel:
Fixed.

PR #1644, opened 2021-12-22
merge HUMAN_MAX_CREDITS and ALIEN_MAX_CREDITS since they're the same

bmorel:


src/shared/bg_gameplay.h
```diff
@@ -35,6 +35,7 @@ along with Unvanquished Source Code.  If not, see <http://www.gnu.org/licenses/>
 
 #define BUILDER_DECONSTRUCT_RANGE 100
 
+constexpr int MAX_CREDITS = 2000;
```
necessarily-equal, line 38: How about keeping the comment? You are the first one to complain about lack of doc :P
```suggestion
constexpr int MAX_CREDITS = 2000; // for aliens, CREDITS_PER_EVO converts this to evos for display
```

necessarily-equal (approved):
A nitpick, otherwise, LGTM

Gireen:
I don't think this values should be merged.
Its brings no benefit but makes it more work if we later want to have different values per team again.

Also merging a PR in less than 24 hours is not ok imo.

bmorel:
I'll revert then.

necessarily-equal:
If you want, I can revert it except making it TEAM_* agnostic.

PR #1654, opened 2021-12-30
BotBuy failure + different AI cheating handling

bmorel:
This PR updates the BotBuy functions so that they can fail if buying failed (forbidden equipment, not enough money, locked item), allowing the BT to handle the situation correctly.
It also changes how cheating cvar acts, by not giving any credit to the bot, but instead avoids removing their funds. This makes the code more resilient if someday ALIEN_MAX_CREDITS != HUMAN_MAX_CREDITS (or in case other team is added).


bmorel:
This likely is buggy, plus, if reworking AI cheating code, there's other things which could be done, like make them ignore momentum (in a separate cvar). I'll extract the Buy failure code in a separate PR though

PR #1664, opened 2022-01-03
Fix crash on cg_shadows modification

slipher:
Fix cvar-related crashes on changing cg_shadows.

This doesn't fix crashes related to uninitialized shaders with certain cg_shadows values.

I have a WIP branch for allowing multiple cvar proxies on one variable (so that multiple modules can observe changes), so I will make sure to fix the underlying cvar system bug with that branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,6 @@ if (BUILD_SGAME)
     # Fastlz
     add_library(srclibs-fastlz EXCLUDE_FROM_ALL ${FASTLZLIST})
     set_target_properties(srclibs-fastlz PROPERTIES POSITION_INDEPENDENT_CODE 1 FOLDER "libs")
-    set(LIBS_ENGINE ${LIBS_ENGINE} srclibs-fastlz)
 
     # Detour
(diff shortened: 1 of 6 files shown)


slipher:
Apparently remove the latch flag in the cgame breaks the latching somehow, even if it's latched in the renderer. So this is not ready yet.

slipher:
Switched to a different approach, now it works.

illwieckz:
Yay!

On a side topic, I wonder if we should rename it to `r_shadows`, and it's already named this way in engine code.

This one is very different to `cg_playerShadows` or `cg_buildableShadows`, those is about gameplay stuff : how must be displayed shadows of gameplay elements according to their gameplay kind. But `cg_shadows` is about the way the renderer renders shadow whatever the meaning of what is casting shadow in game.

The `cg_` prefix is probably inherited from 20 years old code, and we are at a time no one we complain if we rename it (we currently only use one value which is the one by default) so it's the perfect time to do a rename. I'm OK to do all the renaming in every repository, and this can be done _after_ this one is merged.

slipher:
OK, rename to r_shadows on 0.53 branches SGTM

illwieckz (approved):
It works! Thank you very much.

PR #1666, opened 2022-01-03
Bot behaviors: add closest human/alien building

slipher:


bmorel:
Nice. I would probably just use ally/enemy building instead of explicitly naming the team though, but that's a minor detail.

src/sgame/sg_bot_ai.cpp
```diff
@@ -175,6 +196,14 @@ botEntityAndDistance_t AIEntityToGentity( gentity_t *self, AIEntity_t e )
 	{
 		return self->botMind->closestDamagedBuilding;
 	}
+	else if ( e == E_HUMANBUILDING )
```
bmorel, line 199: while you're there, maybe we'd rather see a switch, here? The only case in which it would not work nicely is when a specific buildable is required for, but the check could go in a default: statement, I guess?

bmorel (approved):
What is needed to make it a normal PR?

slipher:
Applied suggestions from comments

Issue #1586, opened 2021-11-06
bots try to buy forbidden items despite global restrictions on equipment


bmorel:
As mentioned by Slipher on https://github.com/Unvanquished/Unvanquished/pull/1557:

> You may need to check BG_UpgradeDisabled also (there are cvars to generally disable an item as well as to disable it for only bots). Actually this may be missing from the bot shopping code too.
>
It would make sense to add the disabled check into BG_UpgradeUnlocked.

And thus, from myself:

> this likely applies not only to upgrades, but also to weapons and classes

Gireen:
It should be optional for bots to follow this. 
We can also want to have things disabled only for players but available for bots.

bmorel:
I'm not sure. To me, having different values for players and bots make sense, and the player system... well, from what I remember, it's hard to use, compared to AI one (one cvar per equipment).
I'm ok to fix that, that can be done quickly, I personally don't care for one way or another. Up to you and @slipher here (and others, I guess).

necessarily-equal:
So, to describe this issue better: bots actually *can't* buy these, but they will desire to do so because BotDesiredBuy tell them to. So bots will stay in front of the armory trying to buy some piece of equipment forever, while not being able to:

![Screenshot from 2021-11-14 13-22-50](https://user-images.githubusercontent.com/59283660/141680975-246dc45f-f274-47c0-881f-a19e38ea023b.png)

This screenshot is bots with `/give momentum 300; set g_bot_infinite_funds 1; set g_disabledEquipment lcannon`

necessarily-equal:
To give my opinion on the question, I don't think bots should be allowed to override this, because it's hard to tell what they should be allowed to override, and what they shouldn't. For example, should be they allowed to buy anyway when they are too far from armory? This kind of special favor would probably introduce bugs.

If we want to have a way for bots to have this, I'd like to have some "magic methods" that give you this on that piece of equipment, like this `BotActionTeleport` thing.

Also note that this restriction wrongly checked may apply to classes, upgrades, weapons and buildables. So we would need to check all the ones a bot can buy, and ensure that if we update `BG_*Unlocked` to check `BG_*Disabled`, we do so for all 4. I personally prefer to keep  `BG_*Unlocked` and `BG_*Disabled` separate.

bmorel:
>  I'd like to have some "magic methods" that give you this or that piece of equipment, like this BotActionTeleport thing.

[...]

PR #1674, opened 2022-01-04
Bot buy can fail

bmorel:
Allow BT to handle if buying a weapon or upgrade failed. This should never happen if nothing specific was requested (`action equip` or `action evolve`) but can happen if a specific equipment or class is asked for which is not unlocked yet, forbidden to buy (by either bot specific cvars or general disabling of equipment) or bot does not have the ¤ for, resulting for humans in possibly selling the weapon and roaming around with a blaster (no ckit nor smg).

This PR does not fix any situation, it only allows things to be saved.


bmorel:
Code is extracted from #1654

src/sgame/sg_bot_util.cpp
```diff
@@ -1986,111 +1986,112 @@ bool BotEvolveToClass( gentity_t *ent, class_t newClass )
...
 
-		//can afford this?
-		if ( BG_Weapon( weapon )->price > ( short )self->client->pers.credit )
-		{
-			return;
-		}
+	//already got this?
+	if ( BG_InventoryContainsWeapon( weapon, self->client->ps.stats ) )
+	{
+		return false;
```
slipher, line 1999: I wouldn't call that a failure

src/sgame/sg_bot_util.cpp
```diff
@@ -1986,111 +1986,112 @@ bool BotEvolveToClass( gentity_t *ent, class_t newClass )
...
 	{
-		return;
+		return false;
 	}
 
 	//already got this?
 	if ( BG_InventoryContainsUpgrade( upgrade, self->client->ps.stats ) )
 	{
-		return;
+		return false;
```
slipher, line 2064: I wouldn't call this a failure

necessarily-equal:
> This should never happen if nothing specific was requested (action equip or action evolve)

Well, it can :P — that's among the bugs I'll have to fix before I can make a tutorial release. See https://github.com/Unvanquished/Unvanquished/issues/1586#issuecomment-968281421 for a description and a reproducer.

bmorel:
> Well, it can :P — that's among the bugs I'll have to fix before I can make a tutorial release. See #1586 (comment) for a description and a reproducer.

One more reason to make BT able to handle the situation, then.

src/sgame/sg_bot_util.cpp
bmorel: Agreed.

src/sgame/sg_bot_util.cpp
bmorel: Agreed.

necessarily-equal (approved):
From a quick look, LGTM

PR #1677, opened 2022-01-04
spectators should not zoom (at least not like this)

### LINKED ISSUE #1399: using right click when observing a sniper user triggers the zooming shader
To reproduce, with RC6's default bindings: run with ` /devmap plat23; bot fill 4`, forbids human bots to wear any weapon which is not sniper, give them infinite funds. Spec a human bot, and try to "unspec".

bmorel:
Fix #1399

### DIFF
--- a/src/cgame/cg_draw.cpp
+++ b/src/cgame/cg_draw.cpp
@@ -321,7 +321,7 @@ static void CG_Draw2D()
 
 	CombatFeedback::DrawDamageIndicators();
 
-	if ( cg.zoomed )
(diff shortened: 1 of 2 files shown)


necessarily-equal:
This shows some kind of scope anyway before unattaching, which makes it look weird. I'm don't think this one is ready for merging

bmorel:
It does not fixes the whole problem, but it improves the situation a lot, as there's no more artefact when the spectator is indeed moving around.

bmorel:
You got a better fix, so let's wait for it to land there.

PR #1678, opened 2022-01-04
show balloon over head of chatting players

bmorel:
The idea is to use a state bit to communicate about the chat "state" of players, and for players which are chatting, to display an icon over their head, same way as when they seems disconnected.

[This](https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/blob/9a7bc181205bc2c4470f531dd7c233213dc800bf/gfx/feedback/chatballoon.png) is the icon which will be used, if all works correctly.

This is a draft because I could never test it, it seems that I'd need building complete dpk and all the stuff: I'm not in the mood right now.

### DIFF
--- a/src/cgame/cg_players.cpp
+++ b/src/cgame/cg_players.cpp
@@ -2571,10 +2571,13 @@ Float sprites over the player's head
 */
 static void CG_PlayerSprites( centity_t *cent )
 {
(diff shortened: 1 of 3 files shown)


bmorel:
This does not work

PR #1679, opened 2022-01-05
workaround bots not able to reach armory in some cases

bmorel:


slipher:
Can you give more detail how to reproduce the bug? I added some bots (from a recent master) with infinite funds on both Chasm and Urban, and they managed to buy weapons in both cases.

bmorel:
Easy: just go on chasm with current default layout. There's the same on urbanp. Give human bots money and momentum, and see how they're still only using smg without armor.

slipher:
I found an even quicker repro which is to do `bot fill 2 h` on Chasm. One of the spawning bots will immediately get stuck on the reactor. The patch fixes it, nice!

I wonder if this stuff should be in G_BotAddObstacle (where the other fudge factors are applied). But the only other use cases (besides buildings) are movers, which might be hard to test.

bmorel:
This is simply a workaround and is destined to be remove, but the intent is to improve the situation until we get the knowledge to do that properly. Movers are currently unusable for bots anyway, so as you say, it's hard to test them. Also, I suspect the mover's code to be pretty buggy.

slipher:
LGTM. Is there an issue about this?

bmorel:
Not that I'm aware of.

PR #1686, opened 2022-01-07
Pmove refactoring: part 1

bmorel:
There's still some more or less easy work to do and documentation to write on things which are either long or not obvious.

Outside of the cleanup, there are (unreported AFAIK) candidate bug fixes in there, too:

* water level management was based on a constant mins[2] (aka ZMINS or something like that)
* when crashing on the ground, previously there was emitted sound only for living entities if distance is medium or far, not for close or no-fall, I made it so that the noise is emitted only if the entity is living (I think this is arguable tbh, but I have not dug into the ins and outs of that noisy story)

There are several code removal and simplifications that I refrained myself to do, but I'm pretty sure I'll see them pop in comments.


src/shared/bg_pmove.cpp
```diff
@@ -4516,6 +4516,21 @@ PmoveSingle
 
 ================
 */
+// set the firing flag for continuous beam weapons
+static void setFireBeam( buttonNumber_t btn )
+{
+	if ( !( pm->ps->pm_flags & PMF_RESPAWNED ) && pm->ps->pm_type != PM_INTERMISSION &&
+			usercmdButtonPressed( pm->cmd.buttons, BTN_ATTACK ) &&
```
necessarily-equal, line 4483: This function doesn't use its argument
```suggestion
			usercmdButtonPressed( pm->cmd.buttons, btn ) &&
```

src/shared/bg_pmove.cpp
```diff
@@ -4516,6 +4516,21 @@ PmoveSingle
...
+			( ( pm->ps->ammo > 0 || pm->ps->clips > 0 ) || BG_Weapon( pm->ps->weapon )->infiniteAmmo ) )
+	{
+		pm->ps->eFlags |= EF_FIRING;
+	}
+	else
+	{
+		pm->ps->eFlags &= ~EF_FIRING;
```
necessarily-equal, line 4490: the previous code did set `EF_FIRING2` or `EF_FIRING3` in some cases, too. I would use that as second argument.

src/shared/bg_pmove.cpp
```diff
@@ -4605,15 +4588,7 @@ void PmoveSingle( pmove_t *pmove )
...
-
-	if ( pml.msec < 1 )
-	{
-		pml.msec = 1;
-	}
-	else if ( pml.msec > 200 )
-	{
-		pml.msec = 200;
-	}
+	Math::Clamp( pml.msec, 1, 200 );
```
necessarily-equal, line 4551: This does nothing. use this instead:
```suggestion
 	pml.msec = Math::Clamp( pml.msec, 1, 200 );
```

src/shared/bg_pmove.cpp
```diff
@@ -4667,14 +4667,7 @@ static void PM_HumanStaminaEffects()
...
 	// Check stamina limits
-	if ( stats[ STAT_STAMINA ] > STAMINA_MAX )
-	{
-		stats[ STAT_STAMINA ] = STAMINA_MAX;
-	}
-	else if ( stats[ STAT_STAMINA ] < 0 )
-	{
-		stats[ STAT_STAMINA ] = 0;
-	}
+	Math::Clamp( stats[ STAT_STAMINA ], 0, STAMINA_MAX );
```
necessarily-equal, line 4470: This doesn't do anything. Use this instead:
```suggestion
	stats[ STAT_STAMINA ] = Math::Clamp( stats[ STAT_STAMINA ], 0, STAMINA_MAX );
```

src/shared/bg_pmove.cpp
```diff
@@ -51,11 +51,13 @@ struct pml_t
 	vec3_t   previous_origin;
 	vec3_t   previous_velocity;
 	int      previous_waterlevel;
+
```
necessarily-equal, line 54: (negligible, but probs a typo)

necessarily-equal (changes requested):
I still have to review 37038de05b77bf113f5680f13b0c5ba771de2d08 and the last commit, but here's some stuff

src/shared/bg_pmove.cpp
```diff
@@ -85,6 +85,11 @@ inline bool PM_Paralyzed( int pmt )
 	return PM_Paralyzed( static_cast<pmtype_t>( pmt ) );
 }
 
+inline bool segmentedModel( playerState_t const* ps )
```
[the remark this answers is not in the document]
bmorel: Done.

src/shared/bg_pmove.cpp
bmorel: good catch!

src/shared/bg_pmove.cpp
bmorel: always confused by that math lib... thanks

src/shared/bg_pmove.cpp
bmorel: not a typo: it "splits" the struct between properties and methods

src/shared/bg_pmove.cpp
```diff
@@ -4431,75 +4431,41 @@ PM_Animate
...
-			if ( pm->ps->torsoTimer == 0 )
-			{
-				PM_ForceLegsAnim( NSPA_GESTURE );
-				pm->ps->torsoTimer = TIMER_GESTURE;
-				pm->ps->tauntTimer = TIMER_GESTURE;
-
-				PM_AddEvent( EV_TAUNT );
-			}
-		}
+		PM_ForceLegsAnim( NSPA_GESTURE );
```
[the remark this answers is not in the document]
bmorel: no, because RALLY/TAUNT can not, AFAIK, be used at all. Those things should likely be removed from codebase.

src/shared/bg_pmove.cpp
bmorel: good catch, again.

bmorel:
@necessarily-equal I think all the remarks were fixed or answered.

src/shared/bg_pmove.cpp
```diff
@@ -4431,75 +4431,41 @@ PM_Animate
...
 
-	if ( usercmdButtonPressed( pm->cmd.buttons, BTN_GESTURE ) )
+	bool doit = false;
+	if ( IsSegmentedModel( pm->ps ) )
 	{
-		if ( pm->ps->tauntTimer > 0 )
+		if ( usercmdButtonPressed( pm->cmd.buttons, BTN_GESTURE ) )
 		{
-			return;
+			int wp_anim = TORSO_GESTURE_BLASTER + ( pm->ps->weapon - WP_BLASTER );
```
slipher, line 4210: Nit: our style for variable names is `wpAnim`

src/shared/bg_pmove.cpp
```diff
@@ -3446,58 +3399,39 @@ static void PM_Footsteps()
...
-					}
+					PM_ContinueLegsAnim( NSPA_WALK );
 				}
 			}
 		}
 	}
 
-	bobmove *= BG_Class( pm->ps->stats[ STAT_CLASS ] )->bobCycle;
+	classAttributes_t const* pcl = BG_Class( pm->ps->stats[ STAT_CLASS ] );
+	ASSERT( pcl );
```
slipher, line 3344: Useless assertion as BG_Class always returns something even if given an out-of-range value.

src/shared/bg_pmove.cpp
```diff
@@ -4755,8 +4711,23 @@ void PmoveSingle( pmove_t *pmove )
 	}
 	else if ( pml.walking )
 	{
-		if ( BG_ClassHasAbility( pm->ps->stats[ STAT_CLASS ], SCA_WALLCLIMBER ) &&
-		     ( pm->ps->stats[ STAT_STATE ] & SS_WALLCLIMBING ) )
+		if ( pm->waterlevel > 2 && DotProduct( pml.forward, pml.groundTrace.plane.normal ) > 0 )
```
slipher, line 4637: This one `PM_ClimbMove PM_WalkMove dedup (to PmoveSingle)` is not correct because there were early returns from PM_WalkMove, but now the rest of the function will always execute.

slipher (commented):
In general I like the -300 lines. Didn't get through all commits; next one for me is `PmoveSingle dedup + clean`

src/shared/bg_pmove.cpp
```diff
@@ -168,31 +168,29 @@ static void PM_StartWeaponAnim( int anim )
 PM_StartLegsAnim
 ===================
 */
+// checks for higher priority animation
+// return false if a higher priority animation is running,
+// there was a comment for the segmentedModel case:
+// "legsTimer is clamped too tightly for nonsegmented models"
+// keeping it here in case it makes sense to someone later.
+// FYI: this code was originally 4 branchs.
+static bool PM_LegsPriority( playerState_t const* ps )
```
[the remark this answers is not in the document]
bmorel: I think you'll change your mind after some commits, since code surrounding `PM_ContinueLegsAnim` and `PM_StartLegsAnim` gets moved to be organised in a more logical way (that is, according if the model is segmented or not). At this point, `PM_LegsPriority` only gets called twice, but I agree that a better name could be found. At least, to have the code in a single place means that any required change would only need to happen at a single place.

src/sgame/sg_struct.h
```diff
@@ -364,6 +364,11 @@ struct gentity_t
 	int       splashMethodOfDeath;
 
 	int       watertype;
+	// from -inf to 3, apparently.
+	// * 0 and less: air/ground
+	// * 1: touch water I guess
```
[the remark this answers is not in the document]
bmorel: Thanks, will update the comment.

src/shared/bg_pmove.cpp
bmorel: Good catch, needs an additional "else" block there

src/shared/bg_pmove.cpp
bmorel: When I get a pointer (and not a reference) from a function, I always consider things can go wrong and thus always put an assert after if I do not have full control on input parameters, just to be sure.
I could remove the assert, but it's likely the best fix here would be to return a reference (which by definition can not be null without being a bug).

src/shared/bg_pmove.cpp
bmorel: exact.

src/shared/bg_pmove.cpp
```diff
@@ -89,6 +89,14 @@ inline bool IsSegmentedModel( playerState_t const* ps )
 	return !( ps->persistant[ PERS_STATE ] & PS_NONSEGMODEL );
 }
 
+// well... this only checks for sky and slick, so it might
+// need changes. At least now, there's a single place.
+inline bool hitGrippingSurface( trace_t const& tr )
```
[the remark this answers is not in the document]
bmorel: good catch

src/shared/bg_pmove.cpp
```diff
@@ -143,13 +142,13 @@ PM_StartTorsoAnim
 */
 void PM_StartTorsoAnim( int anim )
 {
-	if ( PM_Paralyzed( pm->ps->pm_type ) )
+	playerState_t * ps = pm->ps;
```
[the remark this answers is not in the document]
bmorel: true enough, will change that. It's just kind of an habit I've taken while reading bot code :)

src/shared/bg_pmove.cpp
```diff
@@ -2418,37 +2418,33 @@ PM_CorrectAllSolid
...
-	vec3_t point;
-
 	if ( pm->debugLevel > 1 )
 	{
 		Log::Notice( "%i:allsolid\n", c_pmove );
 	}
 
 	// jitter around
-	for ( i = -1; i <= 1; i++ )
+	vec3_t point;
```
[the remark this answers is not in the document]
bmorel: Understandable, but considering this is done a lot, I wanted to reduce the number of calculations. This commit have zero impact on the API though, and thus is pretty easy to remove from the stack. I also don't think I touched that area of code later, so that probably means no conflict if you prefer it to be removed.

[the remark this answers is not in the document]
slipher: Moving the code that sets `point` around won't make the code more efficient, and this is the part that is hard to read, so I'd rather have this change dropped.

Removing the extra copy into pml.groundTrace is good.

src/shared/bg_pmove.cpp
bmorel: I suppose doing that (returning a ref instead of a pointer) should be addressed in a different PR (as this one is enormous enough as it is, and this would be kind of out of scope).
Maybe this could be done more generally, though: I think I'll do a cleanup on that matter too, if that's ok.

bmorel:
@slipher : I addressed the comments, but to keep things easier to track, since have not merged them, instead I pushed more commits on top.
I intend to squash the PR before merging, so this won't show anyway, or at  least will compact that if that's preferred.

slipher:
Personally I find it easier to follow when the original commit that had mistakes or improvements is modified to include the changes, rather than having more commits at the end. That way, there is a single correct commit to look at, rather than two, possibly individually nonsensical, commits that I have to mentally piece together. 

A good example is the commit undoing part of "PM_CorrectAllSolid cleanup". The undo commit missed some bits; this fact is hard to see now but would be easy if those two were squashed together.

bmorel:
Will do that then.

bmorel:
I believe all comments have been addressed properly now, except for «This commit (PM_ContinueLegsAnim PM_StartLegsAnim dedup ) makes the code harder to read IMO» which is necessary for later refactoring.

src/shared/bg_pmove.cpp
```diff
@@ -5229,18 +5228,18 @@ void PM_StepEvent( const vec3_t from, const vec3_t to, const vec3_t normal )
 	}
 }
 
-void slide( vec3_t wishdir, float wishspeed, playerState_t &ps )
+void pml_t::slide( vec3_t wishdir, float wishspeed, playerState_t &ps )
```
slipher, line 5202: This is not a good thing to make a member function. pml_t is a hodgepodge of variables which doesn't correspond to any coherent concept, so it is not a good fit for OOP techniques (consider e.g. the "single-responsibility principle").

src/shared/bg_pmove.cpp
```diff
@@ -2125,13 +2124,13 @@ PM_CheckLadder
 Check to see if the player is on a ladder or not
 =============
 */
-static void PM_CheckLadder()
+static void PM_CheckLadder( const pmove_t * pm_, pml_t & pml )
```
slipher, line 2127: `pm` should be global either everywhere or nowhere.

src/shared/bg_pmove.cpp
```diff
@@ -2011,12 +2011,12 @@ static void PM_WalkMove( pml_t & pml )
...
 
 	// set the movementDir so clients can rotate the legs for strafing
-	PM_SetMovementDir();
+	PM_SetMovementDir(); // }
 
 	// project moves down to flat plane
 	pml.forward[ 2 ] = 0;
 	pml.right[ 2 ] = 0;
-
+	// dup {
```
slipher, line 2019: Huh?

src/shared/bg_pmove.cpp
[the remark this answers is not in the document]
slipher: Yeah, part of my objection is the semantics are unclear from the name. But also it returns a different result from the original code, in the case of a segmented model with torsoTimer > 0.

How about writing the code more concisely in its original spot like

```
int timer = IsSegmentedModel(pm->ps) ? pm->ps->legsTimer : pm->ps->torsoTimer;
if(timer > 0) {
  return; // high preority animation is running
}
```

slipher (commented):
Reviewed the rest except "PM_Weapon Split big one"

src/shared/bg_pmove.cpp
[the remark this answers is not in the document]
bmorel: I just reverted the change this commit did, and used the IsSegmentedModel() in the conditions. I think this solves the issue (maybe some more cleanup will come around later, still so much work to do in there...)

src/shared/bg_pmove.cpp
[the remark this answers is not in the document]
bmorel: fixed.

src/shared/bg_pmove.cpp
bmorel: Fixed.

src/shared/bg_pmove.cpp
bmorel: I think I fixed it, but have a doubt.

src/shared/bg_pmove.cpp
bmorel: was a helper to allow me to better identify the stuff still to dedup, I moved it in another commit on top, so that it won't appear in code anymore (think: construction marks, nothing to do in real code indeed)

bmorel:
I think I addressed all comments in there.

src/shared/bg_pmove.cpp
```diff
@@ -3603,9 +3598,65 @@ PM_Weapon
 Generates weapon events and modifies the weapon counter
 ==============
 */
+
+// for weapons with long, noticeable cooldown (mantis pounce)
+static void cooldown( pml_t & pml )
```
slipher, line 3600: This could be a single line instead of a separate function. `weaponCharge = std::min(0, weaponCharge - pml.msec)`

src/shared/bg_pmove.cpp
```diff
@@ -3824,7 +3796,9 @@ static void PM_Weapon()
 			if ( pm->ps->weapon != WP_NONE )
 			{
 				// drop the current weapon
-				PM_BeginWeaponChange( pm->ps->persistant[ PERS_NEWWEAPON ] );
+				int weapon = pm->ps->persistant[ PERS_NEWWEAPON ];
+				ASSERT( weapon >= WP_NONE || weapon < WP_NUM_WEAPONS );
```
slipher, line 3797: &&

src/shared/bg_pmove.cpp
```diff
@@ -3603,9 +3598,65 @@ PM_Weapon
...
+		pm->ps->weaponCharge = 0;
+	}
+}
+
+// charge the skill up to max by amount.
+// If the skill was at maximum and trigger is true, multiply weaponCharge
+// but mult.
+// Triggers a warning if warning is reached
+// Return true if skill was already at maximum value
+static bool chargeSkill( int max, int amount, int warning, bool trigger, float mult )
```
slipher, line 3614: The idea of this function is hard to understand - a real head-scratcher. I think it would be better to just apply min/max/clamp more in the original code. Same goes for dischargeSkill

src/shared/bg_pmove.cpp
bmorel: Part of the intent there is to have weapons properties as isolated as possible, so that one day it would be possible to no longer have per-weapon code.
But right, maybe it's too soon for that kind of things.

src/shared/bg_pmove.cpp
bmorel: fixed

src/shared/bg_pmove.cpp
bmorel: To be honest, the way the charging of weapons is done in code is not intuitive neither, and I refrained myself to make it more generic, since it would be out of the scope of this PR (it would explicitly affect gameplay).

I agree that this is suboptimal, though. Maybe it would be better to just drop this commit for now, until I (or someone else) can come with a better solution?

slipher: Dropping the commit works for me

bmorel: Done.

slipher:
LGTM

PR #1687, opened 2022-01-07
re-do proper stamina handling

bmorel:
looks like rebase dance introduced bugs when I did the previous PR. Here's the fix.


slipher:
LGTM

PR #1688, opened 2022-01-07
use the g_bot_radarRatio cvar

bmorel:
looks like rebase dance introduced bugs when I did the previous (related) PR. Here's the fix.


necessarily-equal (approved):
sure

PR #1690, opened 2022-01-07
CalcAimPitch aims at coord + fleeing refactoring

bmorel:
Only use class to decide what action to do, instead of useless team switch.
Make the switch generate the flee parameters, and take action only after.
Try to aim at next corner to flee faster, instead of toward a random position.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -628,6 +628,7 @@ void BotMoveToGoal( gentity_t *self )
 {
 	vec3_t dir;
 	VectorCopy( self->botMind->nav.dir, dir );
+	const playerState_t& ps  = self->client->ps;
 
 	if ( dir[ 2 ] < 0 )
 	{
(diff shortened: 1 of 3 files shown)


src/sgame/sg_bot_util.cpp
```diff
@@ -1803,6 +1803,8 @@ void BotFireWeaponAI( gentity_t *self )
 	trap_Trace( &trace, muzzle, nullptr, nullptr, targetPos, ENTITYNUM_NONE, MASK_SHOT, 0 );
 	distance = Distance( muzzle, trace.endpos );
 	bool readyFire = self->client->ps.IsWeaponReady();
+	vec3_t dir;
```
slipher, line 1806: a position should not be called "dir"

src/sgame/sg_bot_nav.cpp
```diff
@@ -689,56 +690,69 @@ void BotMoveToGoal( gentity_t *self )
...
+	if ( G_Team( self ) != targetTeam )
 	{
-		case TEAM_HUMANS:
-			// if target is friendly, reach it quickly
-			if ( G_Team( self ) == targetTeam )
+		return;
+	}
+
+	usercmd_t &botCmdBuffer = self->botMind->cmdBuffer;
+	vec3_t dest;
```
slipher, line 701: This variable is unnecessary

slipher (commented):
> improvements being: aim at next angle when pouncing for flee.

What are the improvements? AFAICT it's a pure refactoring commit which does not change behavior.

src/sgame/sg_bot_nav.cpp
bmorel: It is, because vec3_t can not be easily copied on stack (passing byval) and CalcAimPitch modifies it (it would be better with a byval).

src/sgame/sg_bot_util.cpp
bmorel: Good point, I guess when I did that stuff I didn't had everything clear enough in my mind.

bmorel (commented):
This is indeed just a refactoring, I don't remember why I wrote there would be improvements in behavior.
I have searched in my other related commits and can't find anything in there, probably something I did at a point and removed later (because was broken or not working as expected, likely).

src/sgame/sg_bot_nav.cpp
bmorel: In any cases, I moved the declaration closer to where it's used.

bmorel:
@slipher : commit name and descriptions were updated accordingly.

src/sgame/sg_bot_nav.cpp
```diff
@@ -628,6 +628,7 @@ void BotMoveToGoal( gentity_t *self )
 {
 	vec3_t dir;
 	VectorCopy( self->botMind->nav.dir, dir );
+	const auto& state = self->client->ps;
```
necessarily-equal, line 631: The rest of the code uses this convention:
```suggestion
	const playerState_t& ps = self->client->ps;
```

necessarily-equal (approved):

necessarily-equal (approved):

Issue #1493, opened 2021-09-08
bots obstacle avoidance is broken


bmorel:
Bots have a code to find path around obstacle when they are stuck ( https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_bot_nav.cpp#L471). This code probably does not work, as can be seen by:

* loading perseus
* giving aliens max momentums
* allow bots to cheat
* write a BT for aliens which makes them 1: evolve with 1.5s and rush (nothing else)

This will trigger the following behavior:

Bot will evolve to tyrants, try to go to "Crusher" area by the alien base door, but will be stuck in the right hand side of the door.

I do not know the exact causes. This bug was somehow revealed by my rewrite of BotGetRushScore and my BTs.
Note that this is likely the most obvious failure.
Normally, stuck bots should, after some time, find a new target (https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/blob/c1dbbf36ec480c637a84c546c0fe6d345c0baedb/bots/unstick.bt#L34). This, also, seems to not work as expected.

Desired behaviors here are:

* for a start, to never be stuck against walls. A correct use of navmesh should prevent this, at all times.
* if stuck against a wall or a building and no steering path was found, define new target as a close, backpedaling direction
* if stuck against another friendly player, backpedal and wait a random amount of time. Randomness here is important, because otherwise the block will happen infinitely when other side will also resume it's moves (that's what humans do in streets when such situation happens)

To check: are all entities placed on the navmesh? If no, there is no way the navmesh handling code can find correct pathfinding.
It might also be desirable to find a way to use DetourCrowd, but considering the use of various navmeshes this seems unlikely to be doable (amongst other problems that crowd would bring)

[edit]
french word sneaking in text + some precisions

bmorel:
The original description happens because of several bugs:

1. navmesh is not generated at door between "cooling" and "cell block bottom". Still no idea why, it might be either a bug in the navmesh generator, or a bug in Perseus' geometry, which I think is more probable, but I do not know enough on the topic, maybe you would know how to identify something blocking only bots there, @illwieckz , @Viech ?
2. stair platform near door in "storage" is too tight, no navmesh is generated because of margins. Navmesh generator related, clearly. 2 possible fixes: enlarge the platforms, or find a way to reduce the margins.
[...]

Issue #1448, opened 2021-07-05
Proof that navmesh can go through walls


illwieckz:
This looks OK right?

[![bots and fake door](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_230758_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_230758_000.jpg)

[![bots and fake door](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_231237_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_231237_000.jpg)

No!

[![bots and fake door](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_230817_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_230817_000.jpg)

[![bots and fake door](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_231044_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_231044_000.jpg)

[![bots and fake door](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_231104_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_231104_000.jpg)

[![bots and fake door](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_231119_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/maps/station15/bots-and-fake-door/unvanquished_2021-07-05_231119_000.jpg)

[...]

Issue #310, opened 2014-06-10
Bots have navigation problems in complex parts of maps


Veyrdite:
# Problem

Bots sometimes either freeze or move around in tight circles/spasm.  They stay like this until provoked (sometimes they do not react at all) or pushed out of their little zone (commonly difficult, as they will want to move back).
- when most of a team becomes stuck, team balance suffers
- bots can camp on spawns, preventing them from working.  Bots are not 'kicked' away from spawns like players are.
# Steps to reproduce
- play on non bot-friendly maps like chasm or spacetracks
- wait for a bot to get stuck, or build a spawn up somewhere where a bot can't take a simple path to escape from.
# Examples

Dawdling on a high crate, spazzing in circles
![multiple](https://cloud.githubusercontent.com/assets/6168874/3225469/788e3a4c-f051-11e3-9586-db143abc8081.jpeg)
Stationary and non-reactive to everything, including attacks.  Blocks spawning
![overview](https://cloud.githubusercontent.com/assets/6168874/3225470/788f9770-f051-11e3-8414-2cff3613432e.jpeg)

Other known places:
- chasm: marauders circle on-top of the weather-sensor-thing in the outside area
- karith: crates on the mezzanine of the elevator room, near the vent
- spacetracks: opposite side of the wall the bar sits against in the cafeteria.  Bots like treadmilling this.

illwieckz:
Bots now know how to escape crates, see #984, but we can't guarantee it would not happen on special case. Can current implementation be considered as acceptable to close this issue?

illwieckz:
May be better to close this one, if needed anyone can open a new issue with documented  complex case instead of keeping this one open "in case of".

t4im:
We could try dynamically connecting entities with surroundings to specifically address disconnected player spawns. Not sure how much comes out of it, but might be worth looking into.

bmorel:
This bug is NOT closed, the issue is still there.

There are 3 possible fixes:

* either have mappers place AI-clip or alike on surfaces which are reachable by accident by bots, and have no navmesh generated for those (resulting in bot code implementing a "find nearest navmesh" and using it)
* or have them place one-way connections, this would be long, tedious and not doable in practice
* or have the daemonmap to do that. This is the best option, but also the most complex one to implement. It would require to stop considering daemonmap as a tool meant to disappear, but as the valuable and essential tool it currently really is, and give it some love.

illwieckz:
[...]

PR #1691, opened 2022-01-09
improve blocker situation

bmorel:
This basically allows "big bots" (humans, dragoons, tyrants) to navigate better in the map if the proper navmesh is generated, *and* to not be stopped when a human jumps over their defenses.

Proper navmesh means, the climbHeight value should be `std::min( STEPSIZE + ( JumpMagnitude * JumpMagnitude ) / ( gravity * 2 ), maxs[2] - mins[2] - 1)`.

My previous (not published because buggy) attempt at doing that was not capped, and thus had navigation problems when maps had bridges.

[edit]
I merged this with #1701 since they really participate to the same thing: make bots less stuck.
I'm copying here that other PR's text for convenience:

> This improves on #1493 by allowing bots to move on flat or on downward
tight corridors.
The code is a quick workaround, as explained in the commit, a correct
fix would require a complete rewrite of the obstacle management code,
and would even require better engine API to give bots a usable "vision".
>
The "if" block was kept as close to original on purpose, to make it more
obvious that this improves a situation while not fixing all problems.

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -439,7 +439,10 @@ void G_BotThink( gentity_t *self )
 		//BotClampPos( self );
 	}
 
-	self->botMind->willSprint( false ); //let the BT decide that
+	// let the BT decide what kind of moves to do
+	self->botMind->willSprint( false );
+	self->botMind->willCrouch( false );
+
 	self->botMind->behaviorTree->run( self, ( AIGenericNode_t * ) self->botMind->behaviorTree );
 
 	// if we were nudged...
@@ -632,4 +635,17 @@ void botMemory_t::doSprint( int jumpCost, int stamina, usercmd_t& cmd )
 	}
 
 	exhausted = exhausted && stamina <= jumpCost * 2;
+}
+
+void botMemory_t::willCrouch( bool enable )
+{
+	wantCrouch = enable;
+}
+
+void botMemory_t::doCrouch( usercmd_t& cmd )
+{
+	if ( wantCrouch )
+	{
+		cmd.upmove = -127;
+	}
 }
--- a/src/sgame/sg_bot_local.h
+++ b/src/sgame/sg_bot_local.h
@@ -101,6 +101,9 @@ struct botMemory_t
 	botTarget_t goal;
 	void willSprint( bool enable );
 	void doSprint( int jumpCost, int stamina, usercmd_t& cmd );
+	void willCrouch( bool enable );
+	void doCrouch( usercmd_t& cmd );
+
 	usercmd_t   cmdBuffer;
 
 	botSkill_t botSkill;
@@ -126,6 +129,7 @@ struct botMemory_t
 	//avoid relying on buttons to remember what AI was doing
 	bool wantSprinting = false;
 	bool exhausted = false;
+	bool wantCrouch = false;
 };
 
 bool G_BotSetupNav( const botClass_t *botClass, qhandle_t *navHandle );
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -357,19 +357,45 @@ gentity_t* BotGetPathBlocker( gentity_t *self, const vec3_t dir )
 	}
 
 	BG_ClassBoundingBox( ( class_t ) self->client->ps.stats[STAT_CLASS], playerMins, playerMaxs, nullptr, nullptr, nullptr );
+	playerMins[2] -= STEPSIZE;
+	playerMaxs[2] -= STEPSIZE;
 
+	gentity_t *blockers[3] = {};
 	//account for how large we can step
-	playerMins[2] += STEPSIZE;
-	playerMaxs[2] += STEPSIZE;
-
-	VectorMA( self->s.origin, TRACE_LENGTH, dir, end );
-
-	trap_Trace( &trace, self->s.origin, playerMins, playerMaxs, end, self->s.number, MASK_SHOT, 0 );
-	if ( ( trace.fraction < 1.0f && trace.plane.normal[ 2 ] < 0.7f ) || g_entities[ trace.entityNum ].s.eType == entityType_t::ET_BUILDABLE )
+	for ( int i = -1; i != 1; ++i )
 	{
-		return &g_entities[trace.entityNum];
+		playerMins[2] += STEPSIZE;
+		playerMaxs[2] += STEPSIZE;
+
+		VectorMA( self->s.origin, TRACE_LENGTH, dir, end );
+
+		trap_Trace( &trace, self->s.origin, playerMins, playerMaxs, end, self->s.number, MASK_SHOT, 0 );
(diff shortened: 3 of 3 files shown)


bmorel:
Note that this auto-jump implementation is very trivial, but effective enough to allow this:
2022-01-09-171016_1366x768_scrot

In this screenshot, the blue shield was a medistation. A goon went nearby, jumped over the obstacle, and handled the building.

Optimal values to use for climbing in current status are, for reference, those ones (extracted from the logs of my patched daemonmap):

```
class:  builder, radius: 20.0, height: 40.0, stepsize: 18.0, jumpMag: 195.0, jumpHeight: 41.8
Agent have too much jump height (41.8), caping to 38.9 with a magnitude of 183.0
class:  builderupg, radius: 20.0, height: 40.0, stepsize: 18.0, jumpMag: 270.0, jumpHeight: 63.6
Agent have too much jump height (63.6), caping to 38.9 with a magnitude of 183.0
class:  level0, radius: 15.0, height: 30.0, stepsize: 18.0, jumpMag: 250.0, jumpHeight: 57.1
Agent have too much jump height (57.1), caping to 28.9 with a magnitude of 132.0
class:  level1, radius: 15.0, height: 30.0, stepsize: 18.0, jumpMag: 310.0, jumpHeight: 78.1
Agent have too much jump height (78.1), caping to 28.9 with a magnitude of 132.0
class:  level2, radius: 23.0, height: 36.0, stepsize: 18.0, jumpMag: 340.0, jumpHeight: 90.2
Agent have too much jump height (90.2), caping to 34.8 with a magnitude of 164.0
class:  level2upg, radius: 25.0, height: 40.0, stepsize: 18.0, jumpMag: 340.0, jumpHeight: 90.2
Agent have too much jump height (90.2), caping to 38.9 with a magnitude of 183.0
class:  level3, radius: 26.0, height: 55.0, stepsize: 18.0, jumpMag: 270.0, jumpHeight: 63.6
Agent have too much jump height (63.6), caping to 54.0 with a magnitude of 240.0
```

bmorel:
This is one improvement over #310, while I am at it. Specifically, it does the gamelogic side of the 5th solution in a theoretically reliable way (even if that's still not efficient enough for bigger vaults): "generate navmeshes which consider that places one can reach by jump are connected. When bots hit geometry, they can jump... "

bmorel:
I just backported the change to allow human bots to crawl in vents, so this is rather close in features to my most advanced code.
Since that's backports, though, the architecture was not cleaned, but that should allow to use the lesser auto-jump and crouching code when there are corresponding navmeshes, plus tyrants will have less troubles to move out of perseus' alien base (still probably won't, because navmesh is broken... although it might do the job, since the lesser autojump might help there... we'll see when I'll have my AI testing servers updated I guess).

bmorel:
Actual testing seems to show that the code works fine.
About navmeshes, though, there is new data. To generate navmeshes which takes the most out of this PR, the contraint is to cap the climbHeight at `( crouchMaxs[2] - mins[2] ) * 0.625`. Don't ask me why this value.

For info, this is what says my patched daemonmap:

>class:  human_naked, radius: 15.0, height: 40.0, stepsize: 18.0, jumpMag: 220.0, jumpHeight: 48.2
Agent have too much jump height (48.2), caping to 24.9 with a magnitude of 105.0
class:  human_bsuit, radius: 15.0, height: 57.0, stepsize: 18.0, jumpMag: 220.0, jumpHeight: 48.2
class:  builder, radius: 20.0, height: 40.0, stepsize: 18.0, jumpMag: 195.0, jumpHeight: 41.8
Agent have too much jump height (41.8), caping to 24.9 with a magnitude of 105.0
class:  builderupg, radius: 20.0, height: 40.0, stepsize: 18.0, jumpMag: 270.0, jumpHeight: 63.6
Agent have too much jump height (63.6), caping to 24.9 with a magnitude of 105.0
class:  level0, radius: 15.0, height: 30.0, stepsize: 18.0, jumpMag: 250.0, jumpHeight: 57.1
Agent have too much jump height (57.1), caping to 18.7 with a magnitude of 34.0
class:  level1, radius: 15.0, height: 30.0, stepsize: 18.0, jumpMag: 310.0, jumpHeight: 78.1
Agent have too much jump height (78.1), caping to 18.7 with a magnitude of 34.0
class:  level2, radius: 23.0, height: 36.0, stepsize: 18.0, jumpMag: 340.0, jumpHeight: 90.2
Agent have too much jump height (90.2), caping to 22.4 with a magnitude of 84.0
class:  level2upg, radius: 25.0, height: 40.0, stepsize: 18.0, jumpMag: 340.0, jumpHeight: 90.2
Agent have too much jump height (90.2), caping to 24.9 with a magnitude of 105.0
class:  level3, radius: 26.0, height: 55.0, stepsize: 18.0, jumpMag: 270.0, jumpHeight: 63.6
Agent have too much jump height (63.6), caping to 34.2 with a magnitude of 161.0
class:  level3upg, radius: 29.0, height: 66.0, stepsize: 18.0, jumpMag: 270.0, jumpHeight: 63.6
class:  level4, radius: 32.0, height: 92.0, stepsize: 18.0, jumpMag: 170.0, jumpHeight: 36.1

What this means is that the navmesh generated this way have optimal jump values for the following classes:

* tyrant
* advanced dragoon
* battlesuit

Notably, this fixes the problem of tyrants being unable to go out of their base on perseus. As a (probably) final note, there can be no regressions (and only minor improvements) with the currently generated navmeshes.

slipher:
So this requires navgen changes? In that case the navgen changes should be in the same PR, as navgen now lives in the same repository.

Navgen changes might need to be on the 0.53 branch though, since we can't rely on the navmeshes being automatically regenerated yet.

bmorel:
>  So this requires navgen changes? 

Yes and no.
Of the 3 changes, only the crouching requires navmesh change. The lesser-autojump brings improvements even without that, since it allows aliens to not be stuck on drills when chasing someone (but navgen brings further improvements), and the one which improves the handling of slopes is unrelated to navmesh.

bmorel:
From what I've seen, generating navmeshes accounting for the jump's height poses lot of problems on parpax (but again, with current navmesh generation, there's only improvements) around the grids.

I remember there were troubles with pathfinding on them on @Viech 's first iterations (of the current version) but I do not know much about how those grids are done or if it's possible at all to remove navmesh in the empty areas (which seems to be the problem, somehow they end up connected and that messes up with bot's pathfinding) other than filling it with some clipping brush (out of question, would take a lot of time to apply on all maps, and would break on maps where there's no source code).
To put as many information as possible here, there's already navmesh generated there, tweaking the climbheights only seems to connect them.
This might be related to the erosion done by recast: those grids are fairly thin, and even disconnected, so I can imagine tweaking the erosion could solve this, and others, bug (maybe #1448 is caused by the same problem).

bmorel:
Won't be merged since requires changes in the navmesh generator, which are currently quite experimental, plus I think the original navmesh generator from daemonmap (which has been imported mostly as-is) needs a real cleanup (and doc) so this will wait.

PR #1692, opened 2022-01-09
Hide scope for spectators

### LINKED ISSUE #1399: using right click when observing a sniper user triggers the zooming shader

necessarily-equal:
Fixes Unvanquished#1399 and fixes Unvanquished#1505.


bmorel (approved):
Tested and confirmed the problem is fix. Also, LGTM.

Issue #1613, opened 2021-11-25
Disabled buildings not always displayed as such


Gireen:
Sometimes buildings disabled with g_disabledBuildables are displayed possible to build for players. 

With this setting `g_disabledBuildables "drill,leech"`
there also sometimes is this error show on mapstart
```
^3Warn: unknown buildable drill
^3Warn: unknown buildable leech
```

slipher:
> Sometimes buildings disabled with g_disabledBuildables are displayed possible to build for players.

The value of the cvar is reset after each game restart. So if you want to set this consistently, you have to put it in a map config. Maybe the buildables being enabled is just the effect of the cvar resetting to the default value?

The warning is harmless, but I'm working on a commit to fix it.

Gireen:
it was set in default.cfg for maps so it should have worked.
And it also did if one tries to build a message appears that its not possible. 
So on the server side it is enforced but sometimes not displayed as such for the client.

necessarily-equal:
Has this been fixed by the slipher PR?

necessarily-equal:
So, this wasn't fixed. Here's an hacky patch to fix it. @slipher, if I make a PR with it, is it considered "good enough" to be merged? It removes all the warnings and it technically works, but it technically calls BG_SetForbiddenEquipment with a wrong value before it's ready early on.

```diff
commit 362bdfff89f3e014a5947e4cacbf7fc0987b1f7c
Author: Antoine Fontaine <antoine.fontaine@epfl.ch>
Date:   Mon Jan 10 00:33:34 2022 +0100

    HACK: fixes https://github.com/Unvanquished/Unvanquished/issues/1613 (disabled equipment not shown disabled in cgame)

diff --git a/src/cgame/cg_main.cpp b/src/cgame/cg_main.cpp
index 475bae599..ea9e0b04a 100644
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1187,6 +1187,7 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 	// load configs after initializing particles and trails since it registers some
 	CG_UpdateLoadingStep( LOAD_CONFIGS );
 	BG_InitAllConfigs();
+	CG_ParseServerinfo(); // g_disabledEquipment couldn't be parsed correctly before we loaded the configs
 
 	CG_UpdateLoadingStep( LOAD_SOUNDS );
 	CG_RegisterSounds();
[...]

PR #1694, opened 2022-01-10
Fix disabled equipment not shown disabled in cgame

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1613.


slipher:
LGTM

PR #1696, opened 2022-01-11
Misc fixes

necessarily-equal:


src/sgame/sg_bot.cpp
```diff
@@ -202,21 +202,11 @@ const char * G_BotGetBehavior( int clientNum )
 	return bot->botMind->behaviorTree->name;
 }
 
-void G_BotChangeBehavior( int clientNum, const char* behavior )
+bool G_BotSetBehavior( int clientNum, const char* behavior )
 {
-	gentity_t *bot = &g_entities[clientNum];
-
-	if ( !( bot->r.svFlags & SVF_BOT ) || !bot->botMind )
```
slipher, line 209: The check is needed because the `bot behavior` subcommand can be invoked on random client nums.

src/sgame/sg_momentum.cpp
```diff
@@ -353,9 +353,7 @@ void G_DecreaseMomentum()
  */
 float G_AddMomentumGeneric( team_t team, float amount )
```
slipher, line 354: These could just return void, it seems.

src/sgame/sg_client.cpp
```diff
@@ -1377,11 +1377,17 @@ void ClientBegin( int clientNum )
...
 		case 0:
-			if (0)
-		default:
-			if ( !client->pers.admin )
+			break;
 		case 1:
 			G_TriggerMenu( client->ps.clientNum, MN_WELCOME );
+			break;
+		case 2:
+		default:
```
slipher, line 1385: We shouldn't need a default since there are no other possible values.

bmorel:
I'm not fond of what `Drop G_BotChangeBehavior and make G_BotSetBehavior the real public interface` does, because if (and I intend to) an action to switch behavior is done, it will require searching in a global array until one find an entry which is equal, and do a difference to get one's "clientnum" value.
As to why have bots switching between behaviors by themselves, if this need to be explained, then:

1. an ally give instructions to a bot (follow, stand, defend, rush...): the bot should switch behavior according to instructions (and does it itself, it's not swapped by the other player)
2. rush should, imo, be a behavior, instead of an action. If it was a behavior, it would be easier to make it a real rush. According to wikipedia: «The rush is often a suicidal attack (for the units involved); ».

I'm fine with the 2 other commits, though (well, maybe the switch would be better written as a simple if, not sure, but I can hardly see anything being worst that the current shape).

src/sgame/sg_bot.cpp
necessarily-equal: Fair enough, dropped that commit

src/sgame/sg_momentum.cpp
necessarily-equal: It could yeah. I've kept the return value because it could be useful, but this is true.

necessarily-equal:
I've dropped the commit that removed that function that @bmorel liked, so this should be good for review again

src/sgame/sg_client.cpp
bmorel: I wonder if keeping a switch here is really a good idea.
From what I understand, all that code means:

```
if ( ! ( g_showHelpOnConnection.Get() == 2 && client->pers.admin ) )
{
    G_TriggerMenu( client->ps.clientNum, MN_WELCOME );
}
```

which I consider more readable.

src/shared/bg_public.h
```diff
@@ -750,7 +750,6 @@ enum dynMenu_t
 {
   MN_NONE,
 
-  MN_WELCOME,
```
slipher, line 753: Deleting this from the list changes the client/server ABI (by renumbering the other ones), so it should be done on a release branch per https://wiki.unvanquished.net/wiki/Compatibility

bmorel: solutions to that point:

* make it clear it's candidate for removal, and remove it for 0.53
* just do the PR against 0.53/sync

About that one, note that I think @Gireen was against removal on IRC (I may be wrong though).

bmorel:
At least 1st commit seems (more than) legit to  me, I have no opinion on 2nd one.

bmorel:
I wonder if this PR is still valid.

src/shared/bg_public.h
necessarily-equal: I note that from IRC discussions, Gireen don't seem to think resurrecting this is the correct way to proceed.

necessarily-equal: About the ABI change, I've changed the base branch to 0.53.0/sync

Issue #1669, opened 2022-01-04
give resupply API to BTs


bmorel:
Current system is a rather hackish solution combining luck (lot of), headScore and rushScore, that is, they will resupply ammo because the medistat is traditionally near armory and they need heal, or because they buy new equipment.
This only concerns humans, of course.
They can know how much ammo they have, but that's all (and probably move to closest armory if weapon is empty?).

Proper resupply API should include:

* a resupply score function, working the same way as the healing score
* handling for poison
* possibly handling for medkit
* handling for reloading
* handling for barbs regeneration (might be trickier since it might make adv goons too cowardly)

The bot's skill should be considered while implement all that.

bmorel:
This actually works in my mod. There's just an agreement to be found for merging.

bmorel:
Is improved by #2504

bmorel:
Done.

PR #1668, opened 2022-01-03
Use creep for heal

bmorel:

PR #1698, opened 2022-01-12
Ai resupply

bmorel:
Provides BT means to handle reloading of (human) primary weapon:

* action "reload"
* condition "resupplyScore"
* condition "percentAmmoClip"
* condition "percentClips"

Removes condition "percentAmmo"

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1210,6 +1210,27 @@ AINodeStatus_t BotActionHealH( gentity_t *self, AIGenericNode_t *node )
 	return STATUS_RUNNING;
 }
 
+// triggers a reload.
+// Fails if there is no remaining clips.
+// TODO: return STATUS_RUNNING as long as reload is not
+//   finished (this would allow entering a more "dancing"
+//   stance for the bot, to be more careful, etc)
+// TODO: make Cmd_Reload_f able to report status (that's
+//   true for all player commands actually, to allow bots
+//   to use them more easily)
+AINodeStatus_t BotActionReload( gentity_t *self, AIGenericNode_t *node )
+{
+	ASSERT( self && self->client );
+
+	if ( self->client->ps.clips == 0 )
+	{
+		return STATUS_FAILURE;
+	}
+
+	Cmd_Reload_f( self );
+	return STATUS_SUCCESS;
+}
+
 AINodeStatus_t BotActionRepair( gentity_t *self, AIGenericNode_t *node )
 {
 	vec3_t forward;
@@ -1258,6 +1279,7 @@ AINodeStatus_t BotActionRepair( gentity_t *self, AIGenericNode_t *node )
 	}
 	return STATUS_RUNNING;
 }
+
 AINodeStatus_t BotActionBuy( gentity_t *self, AIGenericNode_t *node )
 {
 	AIActionNode_t *buy = ( AIActionNode_t * ) node;
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1062,6 +1062,104 @@ AINodeStatus_t BotActionResetStuckTime( gentity_t *self, AIGenericNode_t* )
 	return AINodeStatus_t::STATUS_SUCCESS;
 }
 
+// resupply action does 2 things:
+// * it makes the bot move at range of a resupply building
+// * it makes the bot to actively reload when at range to
+//   prevent them to go back to fight with one less clip...
+//   or to be stuck in the action.
+// TODO: handle aliens hugs to regen ammo (that's true for
+//   the BotActionHeal, mind you). This likely requires support
+//   for crowds.
+AINodeStatus_t BotActionResupply( gentity_t *self, AIGenericNode_t *node )
+{
+	team_t team = G_Team(self);
+	playerState_t const& ps = self->client->ps;
+	weapon_t const wp = BG_PrimaryWeapon( ps.stats );
+	weaponAttributes_t const* wpa = BG_Weapon( wp );
+	ASSERT( self && self->botMind );
+	botMemory_t & mind = *self->botMind;
+	bool needsPoison = team == TEAM_ALIENS
+		&& !( ps.stats[ STAT_STATE ] & SS_POISONED )
+		&& mind.closestBuildings[ BA_A_BOOSTER ].ent;
+	bool needsAmmo = ps.ammo != wpa->maxAmmo || ps.clips != wpa->maxClips;
+
+	bool needsSupply = needsAmmo || needsPoison;
+
+	float range = 0.f; // poison requires touching the booster, hence 0.f val
+	if ( mind.currentNode != node )
+	{
+		// not sure if it's really logical to fail when the action
+		// is not needed, but imitating other actions (heal) here.
+		if ( !needsSupply )
+		{
+			return STATUS_FAILURE;
+		}
+
+		buildable_t dest;
+		switch( team )
+		{
+			case TEAM_ALIENS:
+				dest = BA_A_BOOSTER;
+				break;
+			case TEAM_HUMANS:
+				range = ENTITY_USE_RANGE;
+				if ( wpa->usesEnergy &&
+						mind.closestBuildings[ BA_H_REACTOR ].distance <= mind.closestBuildings[ BA_H_ARMOURY ].distance )
+				{
+					dest = BA_H_REACTOR;
+				}
+				else
+				{
+					dest = BA_H_ARMOURY;
+				}
+				break;
+			case TEAM_NONE:
+			case TEAM_ALL:
+			case NUM_TEAMS:
+				dest = BA_NONE;
+		}
+
+		if ( !BotChangeGoalEntity( self, mind.closestBuildings[ dest ].ent )
+				|| dest == BA_NONE )
+		{
+			return STATUS_FAILURE;
+		}
+		mind.currentNode = node;
+	}
+
+	if ( !needsSupply )
+	{
+		return STATUS_SUCCESS;
+	}
+
+	if ( !mind.goal.targetsValidEntity() )
+	{
+		return STATUS_FAILURE;
+	}
+
+	// Can't resupply at powered off buildables
+	if ( !mind.goal.getTargetedEntity()->powered )
+	{
+		return STATUS_FAILURE;
+	}
+
+	if ( !GoalInRange( self, range ) )
+	{
+		BotMoveToGoal( self );
+		return STATUS_RUNNING;
+	}
+	// we are in range, but maybe current clip is not fully loaded
+	// resupply of last clip will be done automatically since we're
+	// already in range
+	else if( ps.ammo != wpa->maxAmmo )
+	{
+		Cmd_Reload_f( self );
+		return STATUS_RUNNING;
+	}
+
+	return STATUS_SUCCESS;
+}
+
 AINodeStatus_t BotActionGesture( gentity_t *self, AIGenericNode_t* )
 {
 	usercmd_t *botCmdBuffer = &self->botMind->cmdBuffer;
--- a/src/sgame/sg_bot_parse.cpp
+++ b/src/sgame/sg_bot_parse.cpp
@@ -141,9 +141,48 @@ static AIValue_t haveUpgrade( gentity_t *self, const AIValue_t *params )
 		&& BG_InventoryContainsUpgrade( upgrade, self->client->ps.stats ) );
 }
 
-static AIValue_t percentAmmo( gentity_t *self, const AIValue_t* )
+// This function returns the ratio of 2 ammo values (ammo/clips)
+// for a specific weapon.
+// Only used to reduce duplicated code.
+// TODO: make this a weapon_t or weaponAttributes_t method?
+static AIValue_t Ratio( int current, int max, bool infiniteAmmo )
 {
-	return AIBoxFloat( PercentAmmoRemaining( BG_PrimaryWeapon( self->client->ps.stats ), &self->client->ps ) );
+	if ( max == 0 || infiniteAmmo )
+	{
+		return AIBoxFloat( 1.f );
+	}
+	return AIBoxFloat( current / static_cast<float>( max ) );
+}
+
+// Return emptyness ammo ratio of currently loaded clip
+// TODO: implement for non-primary weapon (such as barbs or future secondary weapons)
+static AIValue_t percentAmmoClip( gentity_t *self, const AIValue_t* )
+{
+	ASSERT( self && self->client );
+	playerState_t const& ps = self->client->ps;
+	weaponAttributes_t const* wpa = BG_Weapon( BG_PrimaryWeapon( ps.stats ) );
+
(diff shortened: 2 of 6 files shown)


bmorel:
added action "resupply", since only asking bots to reach armoury could make them stuck in certain cases

bmorel:
Some of the TODOs in this require a commit from #1668, namely: 6e33a1848d (which makes a function available from all `sg_bot*.cpp` files)

This PR will *also* allow to fix some of the TODOs in #1668, as:

```
//TODO: remove the booster exception, which is made so that
//bots can have poison. Instead, a real resupply API should be
￼//done, so that bots would not go grab poison by sheer luck.
```

will no longer be true.

bmorel:
Fix #1669

bmorel:
rebased to share code with #1668

bmorel:
conflicts from the review changes in #1668 solved.

bmorel:
Removed useless commit, implemented poison booster handling and energy weapon

src/sgame/sg_bot_util.cpp
```diff
@@ -245,6 +245,63 @@ static int GetMaxEquipmentCost( gentity_t const* self )
...
+// resupply action would be desirable.
+// The returned value depends on:
+// * distance between the caller and the closest supply building
+// * caller's skill
+//
+// NOTE: the distance modifier is the same as for BotGetHealScore
+//
+// TODO: use a navigation distance
+// TODO: add poison support
+// TODO: in some cases (goon's barbs) the weapon have infiniteAmmo
```
slipher, line 258: Barb support seems useless to add because you can't replenish them by going to a specific spot.

src/sgame/sg_bot_util.cpp
```diff
@@ -245,6 +245,63 @@ static int GetMaxEquipmentCost( gentity_t const* self )
...
+{
+	float dist = 0;
+	auto const& closestBuildings = self->botMind->closestBuildings;
+	playerState_t const& ps = self->client->ps;
+	team_t team = G_Team( self );
+	weapon_t wp = BG_PrimaryWeapon( ps.stats );
+	weaponAttributes_t const* weapon = BG_Weapon( wp );
+
+	// why check team? Because of goons.
+	if ( weapon->infiniteAmmo && team == TEAM_HUMANS )
```
slipher, line 272: Just return here if the team is not humans? It doesn't seem to do anything useful for aliens of any type.

src/sgame/sg_bot_parse.cpp
```diff
@@ -142,9 +142,38 @@ static AIValue_t haveUpgrade( gentity_t *self, const AIValue_t *params )
...
+// specific weapon. Only used to reduce duplicated code.
+// TODO: make this a weapon_t or weaponAttributes_t method?
+static AIValue_t ratio( int current, int max, weapon_t wp )
 {
-	return AIBoxFloat( PercentAmmoRemaining( BG_PrimaryWeapon( self->client->ps.stats ), &self->client->ps ) );
+	if ( max == 0.f || BG_Weapon( wp )->infiniteAmmo )
+	{
+		return AIBoxFloat( 1.f );
+	}
+	return AIBoxFloat( static_cast<float>( current ) / static_cast<float>( max ) ); //TODO: only keep the useful one (I never remember which one it must be)
```
slipher, line 154: What does this comment mean?

src/sgame/sg_bot_parse.cpp
```diff
@@ -142,9 +142,38 @@ static AIValue_t haveUpgrade( gentity_t *self, const AIValue_t *params )
 	return AIBoxInt( !BG_UpgradeIsActive( upgrade, self->client->ps.stats ) && BG_InventoryContainsUpgrade( upgrade, self->client->ps.stats ) );
 }
 
-static AIValue_t percentAmmo( gentity_t *self, const AIValue_t* )
+// This function returns the ratio of 2 ammo values for a
+// specific weapon. Only used to reduce duplicated code.
+// TODO: make this a weapon_t or weaponAttributes_t method?
+static AIValue_t ratio( int current, int max, weapon_t wp )
```
slipher, line 148: Function name should begin with capital letter

If it's an infinite ammo weapon, `max` should be 0 right? So maybe `wp` is not needed.

src/sgame/sg_bot_ai.cpp
```diff
@@ -1050,6 +1050,103 @@ AINodeStatus_t BotActionResetStuckTime( gentity_t *self, AIGenericNode_t* )
...
+	playerState_t const& ps = self->client->ps;
+	weaponAttributes_t const* wpa = BG_Weapon( BG_PrimaryWeapon( ps.stats ) );
+	ASSERT( self && self->botMind );
+	botMemory_t & mind = *self->botMind;
+
+	bool needsSupply = ps.ammo != wpa->maxAmmo
+		&& ps.clips != wpa->maxClips
+		&& wpa->maxAmmo != 0
+		&& !wpa->infiniteAmmo
+		&& ( team == TEAM_ALIENS && ( ps.stats[ STAT_STATE ] & SS_POISONED ) );
```
slipher, line 1073: You can only "need supply" when on aliens?

src/sgame/sg_bot_ai.cpp
```diff
@@ -1296,23 +1296,23 @@ AINodeStatus_t BotActionHealH( gentity_t *self, AIGenericNode_t *node )
...
-	if ( self->client->ps.clips == 0 )
+	if ( ps.clips == 0 )
 	{
 		return STATUS_FAILURE;
 	}
 
 	Cmd_Reload_f( self );
-	return STATUS_SUCCESS;
+
+	return ps.IsWeaponReady() ? STATUS_SUCCESS : STATUS_RUNNING;
```
slipher, line 1315: As discussed on IRC, I don't think it makes since for the reload action to have a "running" state, since any handling of this couldn't be applied to naturally occurring reloads when a clip runs out.

And dodging behavior would only be useful while fighting, but it would be stupid to reload in a fight with a non-empty clip.

src/sgame/sg_bot_util.cpp
bmorel: Except that one could be willing to go near a very close booster to regen them much faster

src/sgame/sg_bot_util.cpp
bmorel: adv goons and barbs. I want to get them able to use that routine at some point.

src/sgame/sg_bot_parse.cpp
bmorel: it means that I never remember which damned side of a divide must be a float to not loose precision, and that I forgot to check and fix it before doing the PR :/

src/sgame/sg_bot_parse.cpp
bmorel: barbs are infinite, yet they have a max value.

src/sgame/sg_bot_ai.cpp
bmorel: barbs. Also, I'm trying to make the code as generic as possible to anticipate future changes.

src/sgame/sg_bot_ai.cpp
bmorel: Example of use: you just killed an alien close to a door. If you reload instantly, you'll just continue to move on, opening the door (which are neat places to put defenses) while reloading.
You should more likely reload at safe distance of door and, when the weapon is ready, go open the door fully prepared.
Ofc I'm talking about bots being able to detect doors, which is currently not possible, since the navmesh generator just mostly ignore them, instead of creating offmesh links with the info as it should, but I can't address all problems at the same time, and I'd rather have things done once, or at least prepare for later improvements as much as possible.

Now, I agree that a better solution would likely be to check if the weapon (which one, actually? Humans can have up to 3 weapons, and aliens have only one, but some of it's modes might be ready or not at some time, effectively acting as if there were 3 weapons) is ready.
In the meantime, the code this PR does seems to work rather nicely. Seems, because since recently I'm starting to smell a bug in it, that I suspected in another part of my AI changes. Need more investigations, as this is pretty difficult (as always with AIs) to reproduce.

src/sgame/sg_bot_parse.cpp
slipher: It's a floating-point division if either side is floating point.

src/sgame/sg_bot_parse.cpp
slipher: Barbs are not infinite, you can only have 3. And the maxAmmo value is 3. So everything works out by just ignoring the infiniteAmmo field.

src/sgame/sg_bot_ai.cpp
slipher: *Only* barbs? Never any human weapons?

src/sgame/sg_bot_parse.cpp
bmorel: barbs are infinite, but they have a max amount. This is obviously super weird, but this is that kind of super weird things which makes the code so hard to get right or generic enough to prevent code duplication everywhere.
I do not remember all the details on that, there's stuff to clean obviously. The whole equipment/class/upgrade system is in a very bad state there.

See [this](https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/blob/08a511f95cfc7bff735109127d27fcdfc5ba88f9/configs/weapon/level3upg.attr.cfg#L11) as a proof I'm not spouting random noise through keyboard.

src/sgame/sg_bot_ai.cpp
bmorel: huh... oh, right, misunderstood your point. Indeed, bug spotted. Thanks.

src/sgame/sg_bot_ai.cpp
slipher: It seems better to have a condition function for WeaponIsReady. Then you could detect any reload, not only a BT-requested one. If BT users wanted to reload without waiting, they could do that. And if you wanted this behavior of waiting until reload is finished, you could make a `sequence` with `reload` and then `condition !weaponIsReady { <something>} `

src/sgame/sg_bot_parse.cpp
slipher: Yes I understand that. I'm saying ignore infiniteAmmo and base the decision only on whether maxAmmo is 0 - give a resupply score of 0 whenever there is 0 maxAmmo. Hypothetically speaking, if there were a gun that has 0 ammo and non-infinite ammo (thus unfirable), you would also want to return 0, and need this to avoid dividing by 0.

src/sgame/sg_bot_util.cpp
bmorel: The main differences between barbs and most human weapons are:

* barbs slowly regenerate without help
* some buildings can drastically accelerate barb regeneration, which is something players use a lot

src/sgame/sg_bot_parse.cpp
bmorel: > if there were a gun that has 0 ammo and non-infinite ammo (thus unfirable)

Good point that it's technically possible to get that... I'll add a Log::Warn() too.

src/sgame/sg_bot_ai.cpp
bmorel: > and then condition `!weaponIsReady { <something>}`

Ok, but doing that is quite fragile, because `playerState_t::IsWeaponReady()` do not apply to aliens. This is not too much important currently, since aliens won't reload anyway (none of them have clips), but I remember seeing cases in which aliens have only some of their "modes" not ready. IIRC this applies to `mantis` which can pounce between attacks and advanced dragoons with their barbs.
Of course, here, the correct solution would be to allow aliens to hold multiple weapons instead of considering that everything is the same weapon.
This remark also applies to the barbs being both infinite but with limited number of ammo.

src/sgame/sg_bot_ai.cpp
```diff
@@ -1050,6 +1050,104 @@ AINodeStatus_t BotActionResetStuckTime( gentity_t *self, AIGenericNode_t* )
...
+				}
+				else
+				{
+					dest = BA_H_ARMOURY;
+				}
+				break;
+			case TEAM_NONE:
+			case TEAM_ALL:
+			case NUM_TEAMS:
+				dest = BA_NONE;
```
slipher, line 1107: Why this code for an impossible case? Should be an assertion at most.

src/sgame/sg_bot_parse.cpp
```diff
@@ -136,6 +136,18 @@ static AIValue_t currentWeapon( gentity_t *self, const AIValue_t* )
 	return AIBoxInt( BG_GetPlayerWeapon( &self->client->ps ) );
 }
 
+// Return true if currently selected weapon is ready
+// This won't work for all cases since some alien "modes"
+// do not necessarily block all other modes.
+// It might also be a problem for mass driver's scope, but
```
slipher, line 142: Mass driver scope is not a weapon. It's a client-side shader thing changing the view.

src/sgame/sg_bot_util.cpp
```diff
@@ -245,6 +245,63 @@ static int GetMaxEquipmentCost( gentity_t const* self )
...
+				if ( rcDist < dist && weapon->usesEnergy )
+				{
+					dist = rcDist;
+				}
+			break;
+			}
+		case TEAM_NONE:
+		case TEAM_ALL:
+			return 0.f;
+		case NUM_TEAMS:
```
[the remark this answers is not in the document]
bmorel: Because unvanquished is *not* safe. Enums in here are *not* reliable, sadly.

src/sgame/sg_bot_ai.cpp
bmorel: I'm growing wary about assertions in unvanquished's case, seriously.

src/sgame/sg_bot_parse.cpp
bmorel: does not means that the zoom in there is handled like a weapon. BUT I agree with your comment.

src/sgame/sg_bot_util.cpp
illwieckz: I always considered booster as a kind of armoury for the barb. The in-game behaviour is the same: return to armoury, reload.

src/sgame/sg_bot_util.cpp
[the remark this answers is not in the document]
slipher: There at least shouldn't be different handling for different illegal values.

src/sgame/sg_bot_ai.cpp
slipher: If the bot is "playing" on spectators then something is severely broken and the assumptions of the program are violated everywhere.

If you are worried that this might happen, it would be better to add a check at the entry point of behavior tree evaluation, instead of in all the leaf nodes.

src/sgame/sg_bot_parse.cpp
slipher: I know for a fact that zoom is implemented in the cgame. So please remove the inaccurate comment.

bmorel:
This code is old and actually does not work for poison. I made it working in the meantime, but the fixes are implemented in a branch  with more than just that. They shuold be easy to cherry-pick if people are interested, anyway.

PR #1699, opened 2022-01-12
Make bots adhere to equipment restrictions

bmorel:
I am not sure it's such a great idea (see my previous opinions about g_bot_roam and g_bot_rush as to why), since this can be simply removed from the BT instead, but as long as there is a UI to do that, there should be code to make the UI actually doing something.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -65,11 +65,20 @@ struct equipment_t
 	T item;
 	int price( void ) const;
 	bool unlocked( void ) const;
+	bool allowed( void ) const;
 	bool canBuyNow( void ) const
 	{
-		return authorized.Get() && unlocked();
+		return allowed() && unlocked();
 	}
 	int slots( void ) const;
+	bool operator==( T other ) const
+	{
+		return item == other;
+	}
+	bool operator!=( T other ) const
+	{
+		return item != other;
+	}
 };
 
 /// let's write some duplicated code because of nice API
@@ -111,6 +120,25 @@ template <>
 bool equipment_t<weapon_t>::unlocked( void ) const
 {
 	return BG_WeaponUnlocked( item );
+}
+
+// allowed
+template <>
+bool equipment_t<class_t>::allowed( void ) const
+{
+	return authorized.Get() && !BG_ClassDisabled( item );
+}
+
+template <>
+bool equipment_t<upgrade_t>::allowed( void ) const
+{
+	return authorized.Get() && !BG_UpgradeDisabled( item );
+}
+
+template <>
+bool equipment_t<weapon_t>::allowed( void ) const
+{
+	return authorized.Get() && !BG_WeaponDisabled( item );
 }
 
 // slots
@@ -176,6 +204,7 @@ equipment_t<weapon_t> weapons[] =
 	{ g_bot_shotgun , WP_SHOTGUN },
 	{ g_bot_painsaw , WP_PAIN_SAW },
 	{ g_bot_rifle   , WP_MACHINEGUN },
+	{ g_bot_ckit    , WP_HBUILD },
 };
 
 /*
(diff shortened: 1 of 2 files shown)


slipher:
Just delete the UI then?

slipher:
That's a good observation that it just duplicates the functionality of g_bot_ckit. What if you just rewrote the text for g_bot_ckit and got rid of the other UI option? Like "Bots use ckit (for repairing)"

bmorel:
Technically, the ckit can be used for building, too, but yes, of course, bots don't know how to do that (yet).
I'll just do a PR to remove that dumb GUI, and alter ckit description to reflect it's the currently only reason for a bot to buy ckit (and trust the user's brain).

bmorel:
to squash.

bmorel:
Actually, I'm not sure...
It is possible, and easy, to remove the repair code from the BT and, as I said in #1576, this would likely be better because it would keep BTs simpler to understand (by making certain no other config can mess with BTs which really don't need that), but then, the ckit is the only "primary weapon" which do not have option to be disabled for bots, and in practice, current code can lead to bots buying it (by forbidding all primary weapons), so what should happen there?
Of course, the ckit is *not* a primary weapon, so it should probably be excluded from the list of weapons...

slipher:
I agree with your logic in #1576, and it has some sense here. But in this case I'd prefer to think of it as a forbidden game element that only applies to bots. So by that logic we can have the cvar and actions which require buying the ckit should fail.

bmorel:
Now the ckit is just a weapon like other ones. Took the occasion to make the testing of allowance coherent, and since it was requested (in #1586), added a cvar to allow bypassing the "normal" cvars (in a different commit, in case it's asked to remove this (I don't really like this feature tbh).

src/sgame/sg_bot_util.cpp
```diff
@@ -504,7 +533,7 @@ AINodeStatus_t BotActionEvolve ( gentity_t *self, AIGenericNode_t* )
 
 	for ( auto const& cl : classes )
 	{
-		if ( cl.authorized.Get() && BotCanEvolveToClass( self, cl.item ) && BotEvolveToClass( self, cl.item ) )
+		if ( BotEvolveToClass( self, cl.item ) )
```
slipher, line 536: The checks that the item is unlocked, enabled, and doesn't cost too much should not be removed.

bmorel: The evolution is done, in practice, by `BotEvolveToClass` and thus I think the checks belong there, instead of in the code deciding which class to select when none is specified.

slipher: OK so you added enabled and unlocked checking there, but the price checks and also the prohibition on devolving was removed.

bmorel: This is done by `BotCanEvolveToClass` which is called by `BotEvolveToClass`.

In the original code, only the bot specific cvar was not checked by `BotEvolveToClass`, the `BotCanEvolveToClass` was (and still is) already called there.
I added the canBuyNow() (and a check that the class actually exists, I guess this was because it can be called from BT directly, I don't remember all details, that's old code to me) there, which only, in practice, adds the cvar check, since the unlocked check is one of the several checks done by BotCanEvolveToClass:

* unlocked (manual way)
* no devolve (which is probably because originally aliens could not devolve, no longer true, I'll have to implement this for fire extinguishing grangers, some day)
* enough credits

IMO this could use some cleanup too, ofc, but that's not the point of this PR (which is only to implement a way to allow admins to disable `bot repairs`).

slipher: Oh I didn't see that BotCanEvolveToClass is called in that function. LGTM

bmorel: Understandable, the code repeats itself a lot.

PR #1701, opened 2022-01-13
Improve blocker handling

bmorel:
This improves on #1493 by allowing bots to move on flat or on downward
tight corridors.
The code is a quick workaround, as explained in the commit, a correct
fix would require a complete rewrite of the obstacle management code,
and would even require better engine API to give bots a usable "vision".

The "if" block was kept as close to original on purpose, to make it more
obvious that this improves a situation while not fixing all problems.


bmorel:
merged into #1691

PR #1702, opened 2022-01-13
armouries do not work when destroyed

### LINKED ISSUE #1632: armory menu can be triggered when buildable has been destroyed
To reproduce: start a map with not bot, spawn as human, move near armory, shoot at it until destruction, try to use it: menu will spawn.

What would be expected: menu should not appear, but a text saying one can't use a powered down or destroyed building should appear.

bmorel:
Fix #1632


bmorel:
approved by @necessarily-equal  on IRC

PR #1703, opened 2022-01-13
Partial revert of #1532

### LINKED ISSUE #1588: AI try to heal with buildables on ceilings

bmorel:
Fix #1588

This fix 2 problems introduced in said PR:

* alien bots trying to heal with buildings on ceiling happening much more often (because overmind is no longer the highest priority target after booster, and booster is usually placed in easy to access locations)
* alien bots try to heal on very stupid places, like between barricades and enemies, keeping doors blocked by barricades open, or on open grounds because there's a leech. Eggs and boosters are usually placed in rather easy to defend spots, or even defended by actual defenses, which is obviously untrue for almost all other buildings.

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1018,51 +1018,18 @@ AINodeStatus_t BotActionRush( gentity_t *self, AIGenericNode_t *node )
 	return STATUS_RUNNING;
 }
 
-static AINodeStatus_t BotActionReachHealA( gentity_t *self );
-static AINodeStatus_t BotActionReachHealH( gentity_t *self );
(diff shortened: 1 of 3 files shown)


bmorel:
I'm doing this PR knowing that it will bring back the old behavior, and not fix anything.
The point is that the regression lasted 3 months already, and it means anyone using master will suffer it. On my side, I maintained that patch since long, waiting for a proper fix, which seems won't come.

slipher:
LGTM

bmorel:
Good. I'll merge it tomorrow then if there are no more concerns.

Issue #1714, opened 2022-01-16
use cmake to generate unvanquished.dpk


bmorel:
Currently, to mod unvanquished, one needs several things:

* to modify and build the C++ side, with $EDITOR, cmake, ninja or make or an IDE, a C++ compiler, and python2 (because of nacl, that will change someday)
* to modify the data side
* to use Urcheon to take the C++ build, mix it with the data side, and generate the final .dpk with those

The reason the data side is in a different module is, IIRC, because it makes things easier for Urcheon, which itself depends on at least 7 programs unrelated with building the central mod (they are related to build maps, though).

The problems this brings are:

* many changes requires at least 2 PRs (not counting changes which touch to BBoxes...)
* the toolchain requires unorthodox tooling, sometimes so unorthodox that they are maintained only or mostly by some unvanquished or daemon-engine devs
* requires Urcheon's user to RTFM :D

The suggestions I have is to make the build-chain of the gamelogic "mod" simpler. For that:

*  stop using crn in there, which requires a very specific program only maintained by the unvanquished project AFAIK (use webp instead, which is packaged by distributions)
* merge the 2 repos in a single one (would ease maintenance by reducing number of PRs to do when touching to UI or bots, notably)
* finally use cmake to build the .dpk which, in unvanquished's case, only requires a zip tool (7z or another, could be made configurable) and tools to convert the images to the desired format (for now, I'd suggest webp which have better compression rate and is supported by many tools, including simple image viewers like sxiv)

slipher:
Currently there are a couple of options:
- Use Urcheon for a fancy build with crn compression, delta paks and so on
- Idiot method - copy nexe's into unvanquished_src.dpkdir and `zip *` (OK I don't really know what the command is)

So this is proposing yet another method with intermediate fanciness between those two, which would use webp (but why?) instead of crn or the source (typically png). It doesn't seem to me worth the developer resources to build such a 3rd method. 

Of course I agree with merging unvanquished_src.dpkdir into the Unvanquished repo, but that's a separate question.

bmorel:
Well, I had explanations since then about why crn is better than webp (uses less GPU memory, for reference), so I guess such a cmake could use crn (and thus depend on one tool, that's still less) to get the same quality than Urcheon's dpk

PR #1715, opened 2022-01-17
Migrate from libRocket to RmlUi

### LINKED ISSUE #1200: RmlUi issues
Global Issue tracker for RMLUI issues
- [x] ColorInput doesn't work
- [x] In game focus is borked
- [x] Fix deprecated decorator syntax aka broken checkbox and others
- [x] Rocket progress bar needs to be refactored
- [x] Human hud has color issue (was probably using [foreground-color bug](https://github.com/Unvanquished/libRocket/issues/7) as a feature)
- [x] Cursor scaling was lost in translation
- [x] UI scaling was lost in translation
- [x] Add support for transforms.
- [x] Circle menus are broken
- [x] human health blinks when it goes down
- [x] Backup font face
- [x] Cursor only visible over document

### LINKED ISSUE #1551: Circle menu loses keyevent after multiple mouse clicks
To reproduce buy sell Light armor a few times with mouse.
After that the circlemenu doesn't react to key inputs anymore in this case 0-9 and ESC

This doesn't happen if keys are uses to buy and sell.

slipher:
Note that this is based on a March 2020 version of RmlUi, so there is still a bit of migration to get up to the latest version.

Companion https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/95

Fixes #1200.
Fixes #1551.

TODO: consider squashing commits up to the point where it at least compiles and runs without crashing.

---

_Edit by illwieckz._

Needs:

- https://github.com/DaemonEngine/Daemon/pull/572


bmorel:
I do not know how CI passes, but here I can't compile at all, something unsets the -std=c++14 that I put in the CXXFLAGS, somewhere, somehow, and since this is required for RMLUI, it does not builds.

slipher:
You can check out the `rmlui3/sync` branch on Daemon to get C++14 activated.

The 1st commit I believe just contains the trivial find-and-replace stuff like Rocket -> Rml and CString -> c_str. I guess we have to replace that stuff at some point. I want it to be reviewable but I don't understand how putting it in a later commit would help, can you explain more?

bmorel:
> The 1st commit I believe just contains the trivial find-and-replace stuff like Rocket -> Rml and CString -> c_str.

Nope, there's more than just that, I've seen stuff removed, notably.

> I want it to be reviewable but I don't understand how putting it in a later commit would help, can you explain more?

Well, when the code have things like: `Rocket::Core::String...` which are replaced by rml::Core::String...`, that's one diff for each change.
If the code was using `using namespace Rocket;` before, then there would only be `Core::String...` which would remove all those changes from the diff.
Of course, those changes would have to be in a rocket/rml unrelated commit before, but it would bee much more trivial to review since the only diff in there would be either the removal of the `Rocket::` string _or_ the addition of the `using namespace Rocket;` at start of file.

slipher:
@bmorel I've moved the most common and trivial renaming parts into a separate commit (the 2nd one).

The 1st commit now consists of the original first 20 or so commits squashed together. The minimal stuff to get it to build and launch (though mostly without displaying anything). I used a define hack to delay the renaming stuff.

slipher:
Just realized this contradicts our [compatibility guidelines](https://wiki.unvanquished.net/wiki/Compatibility) by making it impossible to use master gamelogic with the latest release's DPKs. Thoughts anyone? Should it go on the 0.53 branch?

illwieckz:
It's fine to go on 0.53.

bmorel:
How does it conflicts? The unvanquished.dpk also contains the GUI configuration files. If the RML changes are to be considered breaking compat, then some of the changes I made to bots also are, since I changed interface of BotGetHealScore (and probably other stuff) and the BTs are not in the nexe.
It is possible to have a package built from only sgame or only cgame, without the data, and to rely on unvanquished dpk to provide those, but that's not how people do it so far (except me, and that's a very recent change, done to make it faster for all to upload/download only the data or only the binary patches), and doing things in a different way than upstream should imply some understanding on the consequences (in this case, to provide the GUI file at the correct format).

What is likely required though, is to have this PR accompanied with the `.rml` files themselves, which are in a submodule (see #1714 about ending the split of inter-dependent files into 2 different repos, without much answers so far) when necessary (I do not know how much rml and rocket are compatible data-wise).

slipher:
I want master gamelogic to be in a more-or-less usable state without any -pakpath flags, if you have the latest release installed. The bot changes were small enough that they still mostly work with 0.52 behavior trees. Similarly the yellow build shader is broken without new assets, but it's a minor point that doesn't prevent testing most things.

DolceTriade:
Thanks for taking this over!

DolceTriade:
It builds for me. I'll go test it some more later tonight.

src/cgame/rocket/rocketProgressBar.h
```diff
@@ -149,27 +149,14 @@ class RocketProgressBar : public Rocket::Core::Element
...
 		Element::OnPropertyChange( changed_properties );
 
-		if ( changed_properties.find( "color" ) != changed_properties.end() )
+		if ( changed_properties.Contains( Rocket::Core::PropertyId::Color ) )
 		{
-			color = Color::Adapt( GetProperty( "color" )->Get<Rocket::Core::Colourb>() );
+			color = Color::Adapt( GetProperty( Rocket::Core::PropertyId::Color )->Get<Rocket::Core::Colourb>() );
 		}
 
-		if ( changed_properties.find( "image" ) != changed_properties.end() )
```
bmorel, line 157: just checking that this is not accidental change and that I understand (could be useful some day): image is no longer a property, but an attribute?

src/cgame/rocket/rocket_documents.cpp
```diff
@@ -59,19 +58,7 @@ void Rocket_LoadDocument( const char *path )
...
-		// This gets 12px on 1920×1080 screen, which is libRocket default for 1em
-		int fontSize = std::min(cgs.glconfig.vidWidth, cgs.glconfig.vidHeight) / 90;
-
-		// 1.6×2.3em ≈ 20×28px on 1920×1080 screen
-		document->SetProperty( "width", va( "%fpx", 1.6 * fontSize ) );
-		document->SetProperty( "height", va( "%fpx", 2.3 * fontSize ) );
-
-		document->RemoveReference();
-	}
+	// TODO: Remove.
```
bmorel: I guess this TODO is done in later commits? Can't really assert anything here, but maybe it would be possible to use [deprecated](https://en.cppreference.com/w/cpp/language/attributes/deprecated) attribute to avoid forgetting (would have a compiler warning, which can be checked or turned as error)?
Since C++14 is required for RMLUI anyway, let's use it?

bmorel (commented):
Only reviewed the 1st commit, which LGTM despite some very minor details.
I will continue reviewing the other commits later.

src/cgame/rocket/rocket.cpp
```diff
@@ -679,7 +679,8 @@ class EngineCursor
         }
         else if ( show_cursor )
         {
-            mode = MouseMode::CustomCursor;
```
bmorel, line 647: maybe get some Log::Notice for this?

src/cgame/rocket/rocketKeyBinder.h
```diff
@@ -55,7 +55,7 @@ class RocketKeyBinder : public Rml::Core::Element, public Rml::Core::EventListen
 	{
 	}
 
-	void OnAttributeChange( const Rml::Core::ElementAttributes &changed_attributes )
+	void OnAttributeChange( const Rml::Core::ElementAttributes &changed_attributes ) override
```
bmorel, line 58: Most of those changes could have been done in the rocket code (would have fixed a bunch of warnings, too, and possibly could enable treating the lack of override as error).

src/cgame/cg_rocket.cpp
```diff
@@ -99,13 +99,7 @@ void CG_Rocket_Init( glconfig_t gl )
 		{
 			token = COM_Parse2( &text_p );
 
-			rocketInfo.cursor = trap_R_RegisterShader( token, (RegisterShaderFlags_t) RSF_DEFAULT );
```
bmorel: This looks like it could be squashed with  18adf9a7a ?

src/cgame/cg_rocket.cpp
```diff
@@ -642,3 +639,29 @@ bool CG_Rocket_IsCommandAllowed( rocketElementType_t type )
...
+	}
+	rocketInfo.cursor = trap_R_RegisterShader( cursorPath.c_str(), (RegisterShaderFlags_t) RSF_DEFAULT );
+	if ( rocketInfo.cursor == 0 )
+	{
+		return false;
+	}
+	// Scale cursor with resolution while maintaining the original aspect ratio.
+	int x, y;
+	trap_R_GetTextureSize( rocketInfo.cursor, &x, &y );
+	float ratio = static_cast<float>( x ) / static_cast<float>( y );
```
bmorel, line 655: This reminds me a TODO I kept in another PR :) a single cast would suffice (not important at all though).

src/cgame/cg_rocket.cpp
```diff
@@ -642,3 +639,29 @@ bool CG_Rocket_IsCommandAllowed( rocketElementType_t type )
...
+	rocketInfo.cursor_pos.h = cgs.glconfig.vidHeight * 0.025f;
```
bmorel, line 656: `rocketInfo.cursor_pos.h = cgs.glconfig.vidHeight / 40.f;` would be more readable, to me. Ideally, would have liked some info about that divide, too, but not sure the info is still around?

src/cgame/rocket/rocketProgressBar.h
```diff
@@ -36,289 +36,50 @@ Maryland 20850 USA.
...
-			{
-				orientation = DOWN;
-			}
-			else
-			{
-				orientation = RIGHT;
-			}
-		}
-	}
+	virtual ~RocketProgressBar() {}
```
bmorel, line 53: override?
Also, there is the ` = default` thing. It's a bit longer, though, but makes it easy to notice with syntax coloration. Dunno about it's use.

src/cgame/rocket/rocket.h
```diff
@@ -88,6 +88,10 @@ struct UnvPropertyId {
 	static Rml::Core::PropertyId Orientation;
 };
 
+#define REGISTER_ELEMENT( tag, clazz ) static std::unique_ptr<Rml::Core::ElementInstancerGeneric< clazz >> instancer_##clazz; \
```
bmorel, line 92: I was not very fond of seeing this macro in the C++ side, but in C++ was okayish considering the scope of the change (and how much it's desired).
Now it lands in a header and is meant to be used by other stuff, so I'd rather see a template function.

Attached patch *compiles* but I have not checked that it *works* (use with care!).
[0001-suggested-patch.patch.txt](https://github.com/Unvanquished/Unvanquished/files/8078310/0001-suggested-patch.patch.txt)

src/cgame/rocket/rocket_documents.cpp
```diff
@@ -56,6 +57,15 @@ void Rocket_LoadDocument( const char *path )
 
 }
 
+// Scale the UI proportional to the screen size
+void Rocket_SetDocumentScale( Rml::Core::ElementDocument& document )
+{
+	// This gets 12pt on 1920�1080 screen, which is libRocket default for 1em
```
bmorel, line 63: Nitpick: not sure it's worth using the UTF8 character for `×` here, looks like just asking for problems to me (even if should not do problems to most).

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2130,17 +2130,20 @@ class PredictedMineEfficiencyElement : public HudElement
 			{
 				if        ( deltaBudget < 0 ) {
 					color = Color::Red;
-					msg = va( "<span class='material-icon error'>&#xE000;</span> You are losing build points!"
+					// Icon U+E000
```
bmorel, line 2131: I think I've seen in other commits things to convert to/from UTF-8, maybe it would be possible to use them instead of manually made UTF-8 codes? (I might be wrong though)

src/cgame/cg_rocket_draw.cpp
```diff
@@ -3526,46 +3531,50 @@ static void CG_Rocket_HaveJetpck()
 struct elementRenderCmd_t
 {
 	const char *name;
-	void ( *exec )();
+	void ( *update )(); // Modifies the RML or element properties
+	void ( *render )(); // Does rendering outside of the HTML engine
 	rocketElementType_t type;
 };
 
+// This kind of element is a hack left over from when VMs could only use C code
```
bmorel, line 3539: just asking, maybe this could use the `deprecate` thingy of C++14? Or a TODO, since this looks like a pretty good candidate of things to get rid of soon (those manually sorted list are quite annoying, but at least here it's documented...)

src/cgame/rocket/rocket.cpp
```diff
@@ -140,6 +140,14 @@ class DaemonSystemInterface : public Rml::Core::SystemInterface
 
 	void SetMouseCursor(const Rml::Core::String& cursor_name) override
 	{
+		if ( cursor_name.empty() )
```
bmorel, line 143: Maybe it's somehow possible to prevent the cursor's name to get empty, instead?
Also, this commit looks like it could be squashed with e61b3a38 ?

src/cgame/rocket/rocket.cpp
```diff
@@ -701,22 +704,8 @@ static EngineCursor engineCursor;
...
-		default:
-			if ( !( catcher & KEYCATCH_CONSOLE ) )
-			{
-				engineCursor.Show( false );
-			}
-
-			drawMenu = false;
-			break;
-	}
+	drawMenu = !!( catcher & KEYCATCH_UI );
```
bmorel, line 707: If this is a cast to boolean, would not it be better to use something easier to read than `!!`? Maybe `( catcher & KEYCATCH_UI ) != 0`?

bmorel:
I have reviewed each commits one by one, still have to read the companion PR though.

Overall things seems ok, but I really think the `override` stuff could go in separate commit which would apply to current state of code. This is not important though, as long as we get rid of those... also, this means we could be able to enable the related warnings (including in CI), which is a good thing.
Some commits could likely be squashed to make reviewing easier, too, but in my case it's too late. Of course, the PR needs a merge so I guess it will have to be reviewed again...

The thing I would like to see changed the most is the addition of a macro where a template could apply. Since templates are instantiated once per type and since there's exactly one instance of each instanciers, this would work. What would probably be needed in my patch is some added documentation in the template or even better, some kind of actual security, no idea how (verifying that the function is called exactly once, maybe?).

src/cgame/rocket/rocketChatField.h
```diff
@@ -294,6 +295,8 @@ class RocketChatField : public Rocket::Core::Element, Rocket::Core::EventListene
 protected:
 	void GenerateCursor()
 	{
+		Rocket::Core::FontFaceHandle font = text_element->GetFontFaceHandle();
+		if ( font == 0) return;
```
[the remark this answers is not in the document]
slipher: Fixed

src/cgame/rocket/rocket.h
slipher: Done.

src/cgame/rocket/rocket.cpp
slipher: I was paranoid about compiler warnings, but no compiler is actually warning about int to bool conversion so I just removed the `!!`.

slipher (commented):
Replied to most comments, but I'm going to beg off of fixing any of that override specifier stuff.

slipher:
Also sorry about commits which contained bugs / should have been squashed; I'm just reluctant to squash together commits with different authors.

src/cgame/rocket/rocket.cpp
```diff
@@ -246,21 +246,26 @@ class DaemonRenderInterface : public Rocket::Core::RenderInterface
...
 
-	void SetScissorRegion( int x, int y, int width, int height )
+	void SetScissorRegion( int x, int y, int width, int height ) override
 	{
 		trap_R_ScissorSet( x, cgs.glconfig.vidHeight - ( y + height ), width, height );
 	}
+
+	void SetTransform( const Rocket::Core::Matrix4f* transform ) override
+	{
+		// TODO: implement.
```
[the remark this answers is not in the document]
bmorel: My comment was done when reading 1st commit, the warn was added in "RMLUI: warn if it calls unimplemented transform"

src/cgame/rocket/rocketChatField.h
```diff
@@ -317,35 +320,31 @@ class RocketChatField : public Rocket::Core::Element, Rocket::Core::EventListene
 
 	void UpdateCursorPosition()
 	{
-		if ( text_element->GetFontFaceHandle() == nullptr )
+		if ( text_element->GetFontFaceHandle() == 0 )
```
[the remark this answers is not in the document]
bmorel: This is what I've seen:
```
-		if ( text_element->GetFontFaceHandle() == nullptr )
￼+		if ( text_element->GetFontFaceHandle() == 0 )
```

I do not get what you mean with uintptr_t?

src/cgame/rocket/rocketProgressBar.h
bmorel: Possibly, I didn't reviewed the squashed PR but each commits one by one

src/cgame/cg_rocket.cpp
bmorel: I guess. The factor/div is not that important, the most important thing would have been to know why the value, but since knowledge is lost, it's can't be helped.

src/cgame/cg_rocket_draw.cpp
bmorel: Ok then, but I would not expect anyone to be able to guess that. Is there something to do to help on that? (adding comment everywhere might be counter-productive, if this happens in other places, but I do not think I've seen other ones?)

src/cgame/cg_rocket_draw.cpp
bmorel: True, but it might help encourage us to fix the situation if we see them :)

src/cgame/rocket/rocket.cpp
bmorel: This clearly would feel like dirty hack. Maybe in later rmlui version it's possible to get a default cursor... we'll see I guess.

src/cgame/rocket/rocketKeyBinder.h
bmorel: Note that I was referring to the *override* changes, in the commit they were introduced.

src/cgame/rocket/rocketProgressBar.h
bmorel: I meant, since the class inherits, the DTor could use override instead.

src/cgame/rocket/rocket_documents.cpp
bmorel: true

bmorel (approved):
I think everything is fine, that is, I don't think this can make things worst that they are. There's some nitpicking on stuff, but really don't care much considering the size and importance of this PR.

src/cgame/rocket/rocketChatField.h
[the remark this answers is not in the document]
necessarily-equal: 0 is an integer, and if you have `using FontFaceHandle = uintptr_t;`, then you should test for equality against the same type like slipher did, so ==0.

necessarily-equal:
I'm not sure how this is supposed to compile, but with current master, when doing this:
```
$ cmake -D'BUILD_GAME_NACL=OFF' -D'BUILD_GAME_NACL_NEXE=OFF' -D'BUILD_GAME_NATIVE_EXE=OFF' -D'BUILD_GAME_NATIVE_DLL=ON' -GNinja -Bbuild-nixos-rmlui
$ ninja -j1 -C build-nixos-rmlui
```
I get https://paste.gnugen.ch/paste/5aAK

bmorel:
it's a sync branch, so it won't work with normal daemon branch, but with a sync one

necessarily-equal:
So I tried again, this time with the daemon rmlui3/sync, and I got this on map load:

```gdb
Thread 10 "daemon" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffbd7fc640 (LWP 250217)]
0x00007fffa2483cab in Rml::Core::PropertyDefinition::ParseValue(Rml::Core::Property&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
(gdb) backtrace
#0  0x00007fffa2483cab in Rml::Core::PropertyDefinition::ParseValue(Rml::Core::Property&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#1  0x00007fffa248a39a in Rml::Core::PropertySpecification::ParseShorthandDeclaration(Rml::Core::PropertyDictionary&, Rml::Core::ShorthandId, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#2  0x00007fffa24bcb0e in Rml::Core::SpritesheetPropertyParser::Parse(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#3  0x00007fffa24b484b in Rml::Core::StyleSheetParser::ReadProperties(Rml::Core::AbstractPropertyParser&) ()
   from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#4  0x00007fffa24bad06 in Rml::Core::StyleSheetParser::Parse(Rml::Core::StyleSheetNode*, Rml::Core::Stream*, Rml::Core::StyleSheet const&, robin_hood::detail::Table<true, 80ul, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, Rml::Core::Keyframes, robin_hood::hash<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::equal_to<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, robin_hood::detail::Table<true, 80ul, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, Rml::Core::DecoratorSpecification, robin_hood::hash<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::equal_to<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, Rml::Core::SpritesheetList&, int) () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#5  0x00007fffa2496b34 in Rml::Core::StyleSheet::LoadStyleSheet(Rml::Core::Stream*, int) () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#6  0x00007fffa24a7aac in Rml::Core::StyleSheetFactory::LoadStyleSheet(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
   from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#7  0x00007fffa24a84d0 in Rml::Core::StyleSheetFactory::GetStyleSheet(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
   from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#8  0x00007fffa24a89cd in Rml::Core::StyleSheetFactory::GetStyleSheet(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) ()
   from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#9  0x00007fffa24592cc in Rml::Core::ElementDocument::ProcessHeader(Rml::Core::DocumentHeader const*) ()
   from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#10 0x00007fffa24e81b1 in Rml::Core::XMLNodeHandlerHead::ElementEnd(Rml::Core::XMLParser*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#11 0x00007fffa24eaf91 in Rml::Core::XMLParser::HandleElementEnd(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
   from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#12 0x00007fffa2560af9 in Rml::Core::BaseXMLParser::ReadCloseTag() () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#13 0x00007fffa25625d3 in Rml::Core::BaseXMLParser::ReadBody() () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#14 0x00007fffa2562751 in Rml::Core::BaseXMLParser::Parse(Rml::Core::Stream*) () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#15 0x00007fffa24763c6 in Rml::Core::Factory::InstanceDocumentStream(Rml::Core::Context*, Rml::Core::Stream*) ()
   from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#16 0x00007fffa243a0c9 in Rml::Core::Context::LoadDocument(Rml::Core::Stream*) () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#17 0x00007fffa243a23a in Rml::Core::Context::LoadDocument(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
   from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#18 0x00007fffa2391724 in Rocket_LoadDocument(char const*) () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
#19 0x00007fffa23354fc in CG_Rocket_Init(glconfig_t) () from /home/antoine/unv/Unvanquished/build-nixos-rmlui/cgame-native-dll.so
```

illwieckz:
With https://github.com/Unvanquished/RmlUi/pull/2 this code builds and run on my end.

illwieckz:
The first commit of the branch had a wrong submodule because of https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/95#issuecomment-1046003362

If you have pulled Unvanquished `0.53.0/sync`, you have to reset to b8dd98d290d0a74931a61ed7d2285213ad7e2650 and pull again.

PR #1716, opened 2022-01-17
Miscellaneous commits from RmlUi branch

slipher:
Ones which don't directly relate to RmlUi migration.

### DIFF
--- a/src/cgame/cg_snapshot.cpp
+++ b/src/cgame/cg_snapshot.cpp
@@ -124,7 +124,7 @@ void CG_SetInitialSnapshot( snapshot_t *snap )
 	}
 
 	// Need this check because the initial weapon for spec isn't always WP_NONE
-	if ( CG_MyTeam() == TEAM_NONE )
+	if ( snap->ps.persistant[ PERS_TEAM ] == TEAM_NONE )
 	{
 		Rocket_ShowHud( WP_NONE );
 	}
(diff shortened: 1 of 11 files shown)


src/cgame/cg_snapshot.cpp
```diff
@@ -124,7 +124,7 @@ void CG_SetInitialSnapshot( snapshot_t *snap )
 	}
 
 	// Need this check because the initial weapon for spec isn't always WP_NONE
-	if ( CG_MyTeam() == TEAM_NONE )
```
bmorel, line 127: I checked what CG_MyTeam() does, and it checks `cg.predictedPlayerState.persistant[PERS_TEAM]`.
Why the change? (I don't say this is wrong, I focused my efforts on sgame so far)

BTW we should change `G_Team( gentity_t* )` so that it becomes `BG_Team( playerState_t const& )` to avoid that kind of things... I thought this already existed, but after checking, nope, it does not.

bmorel (approved):
LGTM

src/cgame/cg_snapshot.cpp
slipher: It doesn't work with predictedPlayerState - it's all zero the first time this executes

bmorel: I see. Thanks for explanations.

bmorel:
Concerning the revert, it appears that it's the reason why packages does not work at all, even with custom maprotation file.
I do know know the reason, but here, neither the engine nor the sgame module gives any hint about what happened.

I think this commit should be reverted quickly, possibly individually from this PR, since it breaks master.

slipher:
@necessarily-equal  @ing in case you want to investigate the reverted one

PR #1719, opened 2022-01-20
Deduplicate navmesh file path building

slipher:
@bmorel Have a look at this. When this is merged, you could make a 1-line change in your mod in the `NavmeshFilename` function to use a different path scheme for the navmesh files. Then the ones in map paks will no longer be used and you'll be free to use navmeshes stored in the homepath.

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1108,8 +1108,7 @@ static void GenerateNavmeshes()
 	for ( class_t species : RequiredNavmeshes() )
 	{
 		fileHandle_t f;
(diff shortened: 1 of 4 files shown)


bmorel:
Dedup code always LGTM.

> you could make a 1-line change in your mod in the NavmeshFilename function to use a different path scheme for the navmesh files

The path looks like it's related to pak's root, though?

slipher:
> The path looks like it's related to pak's root, though?

It also works relative to <homepath>/game/. The problem in 0.52 is that pak files have a higher priority than homepath ones.

PR #1725, opened 2022-01-30
prevent cg_bounceParticles without cg_depthSortParticles

bmorel:
Partially fix https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/issues/20 ideally the GUI would make it impossible, but then it would still be possible to get things wrong when people do not go through GUI (console or manual edition of configuration file).


slipher:
What's the point of the "prevent cg_bounceParticles without cg_depthSortParticles" if you're deleting the latter?

bmorel:
I just added a commit on top, is all. Should be squashed.

bmorel:
@necessarily-equal said LGTM on irc

Issue #395, opened 2014-08-04
Music volume-slider does nothing


Veyrdite:

slipher:
Fixed by DaemonEngine/Daemon#578.

PR #1726, opened 2022-01-30
add a cvar to disable background music

### LINKED ISSUE #1667: Allow to disable music in main and map-loading screens
Currently, there is no way to disable those music themes except to disable all sound (see #395), and they are quite annoying notably when debugging.

bmorel:
Fix #1667

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -1842,6 +1842,7 @@ extern Cvar::Cvar<bool> cg_spawnEffects;
 extern Cvar::Cvar<std::string> cg_sayCommand;
 
 extern Cvar::Cvar<bool> cg_lazyLoadModels;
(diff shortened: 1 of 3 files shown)


slipher:
How about implementing the music volume cvar instead: https://github.com/DaemonEngine/Daemon/pull/578

bmorel:
That would be better, indeed, but I was sick of this music and gave it a shot. I honestly don't like at all how music is handled in unvanquished, so I'll just put it to 0.
Now, I know that in some of the (rare) games I keep the music on, sometimes I just want to shut the music, and for that, the ideal situation is to have both.

Last but not least, this PR is compatible with 0.52, while the daemon PR (which is welcome and even fixes 2 issues: add #395 to the list) will have to wait for the next release (which I hope will come soon, but still).

PR #1730, opened 2022-02-02
fix regression in 8461a64c83

bmorel:
Regression was not checking if rally or gesture button was pressed for
non segmented entities, resulting in flooding them.

### DIFF
--- a/src/shared/bg_pmove.cpp
+++ b/src/shared/bg_pmove.cpp
@@ -4249,22 +4249,26 @@ static void PM_Animate()
 	}
 
 	bool doit = false;
+	bool gesture = usercmdButtonPressed( pm->cmd.buttons, BTN_GESTURE );
+	bool rally  = usercmdButtonPressed( pm->cmd.buttons, BTN_RALLY );
 	if ( IsSegmentedModel( pm->ps ) )
 	{
-		if ( usercmdButtonPressed( pm->cmd.buttons, BTN_GESTURE ) )
+		if ( gesture )
 		{
 			int wpAnim = TORSO_GESTURE_BLASTER + ( pm->ps->weapon - WP_BLASTER );
 			//and now I know why build stuff must be last in the weapon list...
 			PM_StartTorsoAnim( wpAnim > WP_LUCIFER_CANNON ?  TORSO_GESTURE_CKIT : wpAnim );
 			doit = true;
 		}
-		else if( usercmdButtonPressed( pm->cmd.buttons, BTN_RALLY ) )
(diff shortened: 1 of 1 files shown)


bmorel:
Note that I took the occasion to make things a tiny bit more readable by using temp variables instead of those long function calls. Can't hurt.

src/shared/bg_pmove.cpp
```diff
@@ -4249,22 +4249,30 @@ static void PM_Animate()
...
 	{
-		if ( usercmdButtonPressed( pm->cmd.buttons, BTN_GESTURE ) )
+		if ( gesture )
 		{
 			int wpAnim = TORSO_GESTURE_BLASTER + ( pm->ps->weapon - WP_BLASTER );
 			//and now I know why build stuff must be last in the weapon list...
 			PM_StartTorsoAnim( wpAnim > WP_LUCIFER_CANNON ?  TORSO_GESTURE_CKIT : wpAnim );
 			doit = true;
 		}
-		else if( usercmdButtonPressed( pm->cmd.buttons, BTN_RALLY ) )
```
slipher, line 4261: The else if makes more sense. You can't play two torso animations at once.

slipher (commented):
I don't understand, what's the regression?

src/shared/bg_pmove.cpp
bmorel: Agree, but now the code does the same as old, "just in case" again (see my note above it: I think all mentions to rallly could be removed anyway).

bmorel:
> I don't understand, what's the regression?

WHen this fix is not there, aliens will taunt non-stop, because the test for the button to be pressed is no longer there (reintroduced in this commit).

src/shared/bg_pmove.cpp
slipher: It's obviously useless to do both at once, so why revert to the old behavior? Even the change to not produce falling sounds for dead players at a certain distance or whatever seemed far more bold than this one :stuck_out_tongue:

bmorel: "so why revert to the old behavior?" because the intent of the commit was *not* to change behaviors but only to make code more readable.
Otherwise, I would have boldly removed every mentions of BTN_RALLY since AFAIK it does not really exists.

bmorel: OTOH since this thing is basically just a no-OP I guess I can keep the else anyway

slipher:
LGTM

PR #1733, opened 2022-02-05
make creepslow modifier a class attribute

bmorel:
This includes default values matching the ones in 0.52.1 if nothing is found in configuration, to avoid breaking compat.

Note that alien classes are also checked, because I thought that it might be interesting to give aliens a speed boost when in their base. Can easily be removed and the option made humans-only.

### DIFF
--- a/src/shared/bg_parse.cpp
+++ b/src/shared/bg_parse.cpp
@@ -997,7 +997,8 @@ void BG_ParseClassAttributeFile( const char *filename, classAttributes_t *ca )
 		STAMINASPRINTCOST  = 1 << 21,
 		STAMINAJOGRESTORE  = 1 << 22,
 		STAMINAWALKRESTORE = 1 << 23,
-		STAMINASTOPRESTORE = 1 << 24
+		STAMINASTOPRESTORE = 1 << 24,
+		CREEPMODIFIER      = 1 << 25,
 	};
 
 	if( !BG_ReadWholeFile( filename, text_buffer, sizeof(text_buffer) ) )
@@ -1230,6 +1231,12 @@ void BG_ParseClassAttributeFile( const char *filename, classAttributes_t *ca )
 			ca->staminaStopRestore = atoi( token );
 			defined |= STAMINASTOPRESTORE;
 		}
+		else if ( !Q_stricmp( token, "creepModifier" ) )
+		{
+			PARSE( text, token );
+			ca->creepModifier = atof( token );
+			defined |= CREEPMODIFIER;
+		}
 		else
 		{
 			Log::Warn( "%s: unknown token '%s'", filename, token );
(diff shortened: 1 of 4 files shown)


slipher:
What is the point of this?

bmorel:
Several ones.

1. this fix a TODO in code
2. this makes things less hard-coded
3. with the removal of the 0.52 compat, lines are removed
4. with minor code changes, it would be possible to give aliens a speed bonus in their base, depending on their class
5. it makes it possible to have different maluses depending on the armour ratio of the human

bmorel:
Guess it won't be merged.

necessarily-equal:
Actually I think this needs a companion PR on unvanquished.dpkdir ?

bmorel:
Technically it does not since it will adopt default values in absence of the right config keys (also, since I closed it, I will not maintain it, so you'll have to do the changes one might request)

src/shared/bg_parse.cpp
```diff
@@ -1279,6 +1286,37 @@ void BG_ParseClassAttributeFile( const char *filename, classAttributes_t *ca )
...
+				break;
+			case PCL_HUMAN_NAKED:
+				ca->creepModifier = 0.5f;
+				break;
+			case PCL_HUMAN_LIGHT:
+			case PCL_HUMAN_MEDIUM:
+			case PCL_HUMAN_BSUIT:
+				ca->creepModifier = 0.75f;
+				break;
+			case PCL_NONE:
```
necessarily-equal, line 1315:
```suggestion
```

src/shared/bg_parse.cpp
```diff
@@ -1279,6 +1286,37 @@ void BG_ParseClassAttributeFile( const char *filename, classAttributes_t *ca )
 			            token, filename );
 		}
 	}
+
+	// TODO(0.53) move the values into config files
+	if ( !( defined & CREEPMODIFIER ) )
+	{
+		Log::Warn( "creepModifier not defined in %s (using previously hard-coded values)", filename );
+		switch( ca->number )
+		{
```
necessarily-equal, line 1295:
```suggestion
		{
			case PCL_NONE:
```

necessarily-equal (commented):
Can you just apply that?

necessarily-equal (approved):
approved for squash&merge

necessarily-equal:
@slipher are you ok with merging this?

slipher:
1. It's kind of perverse that the creep modifier attribute is required for alien configs, when it is impossible for them to have SS_CREEPSLOWED.
2. I don't want to have warning spam when using 0.52 assets. Why not put this on the 0.53 branch? Then you could do everything in 1 step without all the default values.
3. The companion PR should be opened before submitting this.

bmorel:
> It's kind of perverse that the creep modifier attribute is required for alien configs, when it is impossible for them to have SS_CREEPSLOWED.

This can be fixed easily by making it a human-only thing. I'm not fond of that idea though, but you are right that the attribute SS_CREEPSLOWED should be applied to aliens if the configuration key works for them.
I think I'll just implement this, it's likely just replacing `ForEntities<HumanClassComponent>([&] (Entity& other, HumanClassComponent&) {` with... dunno what exactly, but I'll find out.

Before I try though, do you agree with that plan?
That is, to have aliens possibly affected by creep too (considering that it might be an actual speed increase, why not)?

> I don't want to have warning spam when using 0.52 assets. Why not put this on the 0.53 branch? Then you could do everything in 1 step without all the default values.

Fine with me.

> The companion PR should be opened before submitting this.

Ok.

necessarily-equal:
> > It's kind of perverse that the creep modifier attribute is required for alien configs, when it is impossible for them to have SS_CREEPSLOWED.
>
> This can be fixed easily by making it a human-only thing. 

Or we can just call that a creepSpeedMultiplier, and apply it to everyone ?


Le 19 février 2022 14:34:51 GMT+01:00, "Morel Bérenger" ***@***.***> a écrit :

bmorel:
That's what it would become in practice yes.

slipher:
It wastes network bandwidth if you set a useless state flag that doesn't do anything.

bmorel:
Well, I did closed that PR because I thought it would not be merged without endless discussions, too much for what the PR is worth.
The problem is, I regularly browse all PRs I have pending, just in case, and github does not give a shit that I'm not the person which opened this in last, so I'm more or less tricked in staying active in this.
@necessarily-equal I would really like it if you could clone than branch and do another PR which would supersede this one, so that this one can be closed and that I no longer see it in the PRs I'm the author of.

PR #1738, opened 2022-02-09
Vec3 removal

bmorel:
The removal is not complete, CBSE and botlib still use Vector<> (Vec3 and Vec2) but I do not see yet how to finish the work on those.

### DIFF
--- a/src/cgame/CombatFeedback.cpp
+++ b/src/cgame/CombatFeedback.cpp
@@ -71,7 +71,7 @@ struct DamageIndicator {
 
 	// the following fields are computed every frame and can be ignored
 	// in all init/clustering code
-	Vec2 pos2d;
+	vec2_t pos2d;
 	float dist, scale;
 	Color::Color color;
 	float alpha;
@@ -264,7 +264,8 @@ static bool EvaluateDamageIndicator(DamageIndicator *di, bool *draw)
 	}
 
 	// CG_WorldToScreen returns a point on the virtual 640x480 screen
-	di->pos2d = Vec2::Load(tmp) * Vec2(cgs.screenXScale, cgs.screenYScale);
+	di->pos2d[ 0 ] = tmp[ 0 ] * cgs.screenXScale;
+	di->pos2d[ 1 ] = tmp[ 1 ] * cgs.screenYScale;
 	*draw = true;
 
 	t_fade = (float)(cg.time - di->ctime) / 1000;
@@ -294,24 +295,26 @@ static void DrawDamageIndicator(DamageIndicator *di)
 		text = "!" + text;
 	total_width = width * text.length();
 
-	x = di->pos2d.x() - total_width / 2;
-	y = di->pos2d.y() - height / 2;
+	x = di->pos2d[ 0 ] - total_width / 2;
+	y = di->pos2d[ 1 ] - height / 2;
 
 	for (size_t i = 0; i < text.length(); i++) {
 		int glyph;
-		Vec2 st0, st1;
+		vec2_t st0, st1;
 
 		if (text[i] >= '0' && text[i] <= '9')
 			glyph = text[i] - '0';
 		else
 			glyph = 10;
 
-		st0 = Vec2((glyph % 4) * 0.25f, (glyph / 4) * 0.25f);
-		st1 = st0 + Vec2(0.25f, 0.25f);
+		st0[ 0 ] = ( glyph % 4 ) / 4.f;
+		st0[ 1 ] = ( glyph / 4 ) / 4.f;
+		st1[ 0 ] = st0[ 0 ] + 0.25f;
+		st1[ 1 ] = st0[ 1 ] + 0.25f;
 
 		trap_R_SetColor( di->color );
 		trap_R_DrawStretchPic(x, y, width, height,
-		                      st0.x(), st0.y(), st1.x(), st1.y(),
+		                      st0[ 0 ], st0[ 1 ], st1[ 0 ], st1[ 1 ],
 		                      cgs.media.damageIndicatorFont);
 
 		x += width;
--- a/src/cgame/CombatFeedback.cpp
+++ b/src/cgame/CombatFeedback.cpp
@@ -55,7 +55,7 @@ enum damageIndicatorLayer_t {
 	DAMAGE_INDICATOR_LAYERS
 };
 
-float damageIndicatorColors[DAMAGE_INDICATOR_LAYERS][3] = {
+Color::Color damageIndicatorColors[DAMAGE_INDICATOR_LAYERS] = {
 	{1, 1, 1},  // enemy
 	{1, 0.5, 0}, // friendly
 	{0.54, 1, 0.68}, // alien building
@@ -73,7 +73,7 @@ struct DamageIndicator {
 	// in all init/clustering code
 	Vec2 pos2d;
 	float dist, scale;
-	Vec3 color;
+	Color::Color color;
 	float alpha;
 };
 
@@ -153,7 +153,7 @@ static void AddDamageIndicator(DamageIndicator di)
 			i++;
 	}
 
-	di.color = Vec3::Load(damageIndicatorColors[di.layer]);
+	di.color = damageIndicatorColors[di.layer];
 	di.velocity = Vec3(crandom() * 20, crandom() * 20, 100);
 
 	damageIndicators.push_back(di);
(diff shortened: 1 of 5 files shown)


src/sgame/CombatFeedback.cpp
```diff
@@ -77,15 +77,25 @@ void HitNotify(gentity_t *attacker, gentity_t *victim,
...
+	vec3_t point;
+	if ( !point_opt || indirect )
+	{
+		VectorCopy( victim->r.currentOrigin, point );
+		// those asserts compare floats, but said floats are
+		// read from config files. Original code just did the
+		// calculation for each dimensions, but I strongly doubt
+		// the engine supports assymetric BBoxes for horizontal
+		// plane. Putting asserts to make the assumption explicit.
+		ASSERT( fabs( victim->r.mins[ 0 ] ) == fabs( victim->r.maxs[ 0 ] ) );
```
slipher, line 89: mins == -maxs would be a nicer way to write the assertion

src/sgame/CombatFeedback.cpp
```diff
@@ -31,7 +31,7 @@ Log::Logger logger("sgame.cf");
  * @brief Notify a client of a hit
  */
 void HitNotify(gentity_t *attacker, gentity_t *victim,
-               Util::optional<Vec3> point_opt, float damage,
+               vec3_t const& point_opt, float damage,
```
slipher: This loses the optional-ness of point_opt. I imagine you have a compile error on `!point_opt` below.

slipher: OK so there was a 2nd commit later. I request that broken commits be squashed with their fixes, so that I don't have to waste time commenting on them.

You'd have to have some code to pass nullptr in the missing vector case... maybe it's easier to leave these optionals until glm is added.

src/sgame/CombatFeedback.cpp
bmorel: Seriously, I think it would be better to just have a single value instead of those 4 ones (X is always same as Y, mins is always -maxs...) this would make a lot of things easier, and use the memory space for more useful stuff like squared radius and radius...
But yes, what you propose seems a bit more readable.

src/sgame/CombatFeedback.cpp
bmorel: fixed

src/sgame/CombatFeedback.cpp
bmorel: fixed

bmorel:
About the remaining Vec3 (and Vec2) instances, I'll remove them in the glm branch in one go, to avoid having multiple libs.
The GLM branch will likely need huge rework before that though, and I'd rather have the other refactorings I've made merged before, will be easier to manage conflicts...

src/shared/bg_pmove.cpp
```diff
@@ -1289,12 +1285,12 @@ static bool PM_CheckWallRun()
...
 
-	normal = Vec3::Load( trace.plane.normal );
-	velocity = Vec3::Load( pm->ps->velocity );
-
-	velocity += Math::Normalize( normal + Vec3( 0, 0, 0.9 ) ) * jumpMag;
+	vec3_t tmp;
+	VectorSet( tmp, 0.f, 0.f, 0.9f );
+	VectorAdd( tmp, trace.plane.normal, tmp );
+	VectorNormalize( tmp );
+	VectorMA( pm->ps->velocity, jumpMag, tmp, pm->ps->velocity );
```
necessarily-equal, line 1292: I would write
```suggestion
	vec3_t tmp;
	VectorCopy( trace.plane.normal, tmp );
	tmp[2] += 0.9f;
	VectorNormalize( tmp );
	VectorMA( pm->ps->velocity, jumpMag, tmp, pm->ps->velocity );
```

src/cgame/cg_weapons.cpp
```diff
@@ -1281,85 +1281,53 @@ CG_CalculateWeaponPosition
...
 	//weaponInfo_t *weapon = cg_weapons + cg.predictedPlayerState.weapon;
 	Filter<WeaponOffsets> &filter = cg.weaponOffsetsFilter;
 	WeaponOffsets offsets;
 
-	origin = Vec3::Load( cg.refdef.vieworg );
-	angles = Vec3::Load( cg.refdefViewAngles );
-	right = Vec3::Load( cg.refdef.viewaxis[ 1 ] );
-	up = Vec3::Load( cg.refdef.viewaxis[ 2 ] );
-
 	// on odd legs, invert some angles
```
necessarily-equal, line 1288: you forgot this comment here

src/cgame/cg_weapons.cpp
```diff
@@ -1292,23 +1292,26 @@ static void CG_CalculateWeaponPosition( vec3_t out_origin, vec3_t out_angles )
...
 
 	if( !filter.IsEmpty( ) )
 	{
 		auto last = filter.Last( );
 		float dt = ( cg.time - last.first ) * 0.001f;
 
-		offsets.angvel = angles.Apply2( AngleDelta, last.second.angles ) / dt;
+		offsets.angvel[ 0 ] = AngleNormalize180( angles[ 0 ] - last.second.angles[ 0 ] ) / dt;
+		offsets.angvel[ 1 ] = AngleNormalize180( angles[ 1 ] - last.second.angles[ 1 ] ) / dt;
+		offsets.angvel[ 1 ] = AngleNormalize180( angles[ 2 ] - last.second.angles[ 2 ] ) / dt;
```
necessarily-equal, line 1314: you want `angvel[ 2 ]` here

src/cgame/cg_weapons.cpp
```diff
@@ -1328,10 +1325,9 @@ static void CG_CalculateWeaponPosition( vec3_t out_origin, vec3_t out_angles )
 	const float scaleY = 2e-3;
 
 	VectorAdd( angles, offsets.bob, angles );
-	origin += up * atanf( offsets.angvel[ 0 ] * scaleY ) * limitY;
-	origin += right * atanf( offsets.angvel[ 1 ] * scaleX ) * limitX;
+	VectorMA( cg.refdef.vieworg, atanf( offsets.angvel[ 0 ] * scaleY ) * limitY, cg.refdef.viewaxis[ 2 ], out_origin );
+	VectorMA( out_origin, atanf( offsets.angvel[ 1 ] * scaleX ) * limitX, cg.refdef.viewaxis[ 1 ], out_origin );
```
necessarily-equal, line 1329: maybe it would be a good idea to use some `vec3_t& up, right;` here to avoid obfuscating the code more than it already is. (I don't think it's required either though)

src/cgame/CombatFeedback.cpp
```diff
@@ -294,24 +295,26 @@ static void DrawDamageIndicator(DamageIndicator *di)
...
 
 		if (text[i] >= '0' && text[i] <= '9')
 			glyph = text[i] - '0';
 		else
 			glyph = 10;
 
-		st0 = Vec2((glyph % 4) * 0.25f, (glyph / 4) * 0.25f);
-		st1 = st0 + Vec2(0.25f, 0.25f);
+		st0[ 0 ] = ( glyph % 4 ) / 4.f;
+		st0[ 1 ] = ( glyph % 4 ) / 4.f;
```
necessarily-equal, line 311: the original code did that:
```suggestion
		st0[ 1 ] = ( glyph / 4 ) / 4.f;
```

necessarily-equal (changes requested):

src/shared/bg_pmove.cpp
slipher: If we're nitpicking initialization, better is `vec3_t tmp = {0, 0, 9}`

src/sgame/components/HealthComponent.cpp
```diff
@@ -238,7 +238,9 @@ Util::optional<Vec3> direction, int flags, meansOfDeath_t meansOfDeath) {
 
 	if (entity.oldEnt != source && source->client) {
 		bool lethal = (health <= 0);
-		CombatFeedback::HitNotify(source, entity.oldEnt, location, take, meansOfDeath, lethal);
+		vec3_t tmp;
+		location.value().Store( tmp );
```
slipher, line 242: This will throw an exception if location is not present.

Hint hint, you could just leave this for direct replacement with GLM...

src/shared/bg_pmove.cpp
bmorel: Considering the reason of this PR, I prefer slipher's suggestion (no need for later changes). Also, still considering the reasons for this, I would prefer to avoid nitpicks, since the idea is to replace vec3_t *soon*.

src/sgame/CombatFeedback.cpp
```diff
@@ -85,6 +77,14 @@ void HitNotify(gentity_t *attacker, gentity_t *victim,
 	event->s.angles2[0] = damage;
 	event->s.otherEntityNum = victim - g_entities;
 
+	Vec3 point;
```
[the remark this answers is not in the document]
bmorel: I think the point was to group it to ease the removal or finding of code duplication. I'll fix that.

src/cgame/cg_weapons.cpp
bmorel: All this will soon be rewritten to use glm anyway. This PR is "just" to make the glm migration one easier to do (because of the power of sed) and review.

slipher:
LGTM

necessarily-equal (approved):

PR #1741, opened 2022-02-12
reset stuck time on long actions

bmorel:
Bots reset stuck time on long actions which are still running, which
are:

* heal
* repair
* fight

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -729,6 +729,7 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 
 		self->botMind->currentNode = node;
 		self->botMind->enemyLastSeen = level.time;
+		BotResetStuckTime( self );
 		return STATUS_RUNNING;
 	}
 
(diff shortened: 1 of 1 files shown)


bmorel:
Note that the BotActionResetStuckTime should no longer be necessary in BT after this is merged.

The problem this fixes is, when fighting and having to stay put for some reason (destroying buildings, healing, blocked by wall) bots otherwise go in the unstuck routine, which is a workaround, and which makes them freeze or other bad things to do while trying to accomplish something.

slipher:
This is just stopping the stuck time from working. If the bot thinks it has a path to some goal but isn't making any progress, the stuck timer must be allowed to count in order to detect this. If the goal is a medistation or a damaged buildable, the bot could continue the action indefinitely without progress. *Maybe* this wouldn't happen for `fight`; I don't know how it works.

I guess you're seeing problems because you made the stuck timer too short (5 s) in your server.

bmorel:
> This is just stopping the stuck time from working. 

Yes, sorry, I should have placed comments here about what I talked on IRC afterward.
That is, this PR is very fragile and should not be merged. Ishould have even put it in draft, I apologizes again.

> I guess you're seeing problems because you made the stuck timer too short (5 s) in your server.

This is obviously the reason.
But the current value is also too high and makes bots stuck for too long. I think we should discuss about a more robust workaround, as this issue is not easy to fix.

Issue #1727, opened 2022-01-31
Behavior tree API cleanup


bmorel:
Currently, the BT API is inconsistent (example: bots can know the ammo they have, but can't reload nor go resupply, aliens can know if they can evolve in something, but humans can't know if they can buy specific stuff), as well as it's code quality (duplicated code, partial checks, see #15586, code split in files at random, absence of documentation and I suspect conflicts between some stuff, like the straffing things and the C++).

I want to clean this up, preferably for 0.53, but for that I need input from the other people (@Gireen at least, probably @necessarily-equal , I'm not aware of anyone else besides me?) which actually do use the BT, as I do not want to break used things in the process.

So, which `action` and `condition` have you used?
Which ones would you like to see?

As a reminder (comments in parens):

Actions as of 0.52:

* `activateUpgrade`
* `aimAtGoal`
* `alternateStrafe` (done in C++ anyway)
* `buy`
* `changeGoal`
* `classDodge` (likely conflicts with actual actions)
* `deactivateUpgrade` (what is it even for?)
* `equip`
* `evolve`
* `evolveTo` (could use same semantics as `equip`, that is: no argument, let bot decide)
* `fight`
* `fireWeapon` (handled by `fight` anyway, and can't select which weapon to fire)
* `flee` (mostly same as heal, reported buggy in #1577)
* `heal`
* `jump` (there is no way to know if stamina would allow jumping though)
* `moveInDir`
* `moveTo`
*` moveToGoal` (could be merged with `moveTo`, plus, this won't detect breaks when navmeshes evolve, i.e. when activable doors opens/close or when elevators move around)
* `repair` (can not extinguish burning)
* `resetStuckTime`
* `roam`
* `roamInRadius`
* `rush`
* `say`
* `strafeDodge`
* `suicide`

Condtions as of 0.52:

* `alienMomentum`
* `humanMomentum` (has been merged with alienMomentum for obvious reasons)
* `alertedToEnemy` (no distinction between detection and vision, being shot does not makes a bot alerted)
* `baseRushScore`
* `buildingIsDamaged` (does not detect fires)
* `canEvolveTo` (humans do not have access to that kind of feature)
* `class`
* `cvarFloat`
* `cvarInt`
* `directPathTo`
* `distanceTo`
* `goalBuildingType`
* `goalIsDead`
* `goalTeam`
* `goalType`
* `haveUpgrade`
* `haveWeapon` (does not checks for the bot cvars)
* `healScore`
[...]

PR #1750, opened 2022-02-15
in BT: toggling of upgrades can fail

bmorel:
This PR fixes a "regression" introduced in [this commit](https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/commit/b3107f469986c77dd490f8d7b96ecbf72b557025) which made human bots act weird (freezing, usually) in presence of alien buildings.

This could likely be further improved by making the API to upgrade/disable upgrades return a boolean instead of nothing, but I think the fix I'm proposing is the least intrusive one while still fixing the actual problem.

How to reproduce the faulty behavior is rather simple:

* /devmap plat23
* /bot add * h 9
* allow bots to cheat
* go in alien base and look at main entrance: bot will destroy a 1st tube then freeze.

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -577,11 +577,17 @@ AINodeStatus_t BotActionActivateUpgrade( gentity_t *self, AIGenericNode_t *node 
 	AIActionNode_t *action = ( AIActionNode_t * ) node;
 	upgrade_t u = ( upgrade_t ) AIUnBoxInt( action->params[ 0 ] );
 
-	if ( !BG_UpgradeIsActive( u, self->client->ps.stats ) &&
-		BG_InventoryContainsUpgrade( u, self->client->ps.stats ) )
+	if ( !BG_InventoryContainsUpgrade( u, self->client->ps.stats ) )
 	{
-		BG_ActivateUpgrade( u, self->client->ps.stats );
+		return STATUS_FAILURE;
 	}
+
+	if ( BG_UpgradeIsActive( u, self->client->ps.stats ) )
+	{
+		return STATUS_FAILURE;
+	}
+
+	BG_ActivateUpgrade( u, self->client->ps.stats );
 	return STATUS_SUCCESS;
 }
 
@@ -590,11 +596,17 @@ AINodeStatus_t BotActionDeactivateUpgrade( gentity_t *self, AIGenericNode_t *nod
 	AIActionNode_t *action = ( AIActionNode_t * ) node;
 	upgrade_t u = ( upgrade_t ) AIUnBoxInt( action->params[ 0 ] );
 
-	if ( BG_UpgradeIsActive( u, self->client->ps.stats ) &&
-		BG_InventoryContainsUpgrade( u, self->client->ps.stats ) )
+	if ( !BG_InventoryContainsUpgrade( u, self->client->ps.stats ) )
 	{
-		BG_DeactivateUpgrade( u, self->client->ps.stats );
+		return STATUS_FAILURE;
 	}
+
+	if ( !BG_UpgradeIsActive( u, self->client->ps.stats ) )
+	{
+		return STATUS_FAILURE;
+	}
+
+	BG_DeactivateUpgrade( u, self->client->ps.stats );
 	return STATUS_SUCCESS;
 }
 


necessarily-equal (approved):
I would rather return `STATUS_SUCCESS` if the state is already in the requested state (trying to activate while already active or trying to deactivate while already inactive), but it doesn't matter too much.

bmorel:
> I would rather return STATUS_SUCCESS if the state is already in the requested state (trying to activate while already active or trying to deactivate while already inactive), but it doesn't matter too much.

Why not, but I prefer to be stricter, as this can allow easier debugging (debugging BT is painful enough as it is). Also, from quick glance at other parts of the code, it seems stuff like grenades can be enabled while still being in inventory, returning SUCCESS in that case might trigger new bugs.
This is something which needs to be investigated more actually...

src/sgame/sg_bot_ai.cpp
```diff
@@ -590,11 +596,17 @@ AINodeStatus_t BotActionDeactivateUpgrade( gentity_t *self, AIGenericNode_t *nod
 	AIActionNode_t *action = ( AIActionNode_t * ) node;
```
slipher, line 596: It seems to me this action should just be deleted, because there is no item in the game that can be meaningfully deactivated. Upgrade activate/deactivation is only used as an implementation detail of grenades and medkits which are activated for 1 frame when you use them.

src/sgame/sg_bot_ai.cpp
```diff
@@ -577,11 +577,17 @@ AINodeStatus_t BotActionActivateUpgrade( gentity_t *self, AIGenericNode_t *node
...
 
-	if ( !BG_UpgradeIsActive( u, self->client->ps.stats ) &&
-		BG_InventoryContainsUpgrade( u, self->client->ps.stats ) )
+	if ( !BG_InventoryContainsUpgrade( u, self->client->ps.stats ) )
 	{
-		BG_ActivateUpgrade( u, self->client->ps.stats );
+		return STATUS_FAILURE;
+	}
+
+	if ( BG_UpgradeIsActive( u, self->client->ps.stats ) )
```
slipher, line 585: I suppose the only way this could happen is if you tried to use medkit/grenade/firebomb more than once in a single frame. (This is not a criticism, just an observation.)

Medkits can fail to activate if you have full health, but the behavior tree wouldn't detect this.

bmorel: Honestly, I don't know enough that part of the code for now, so I'm trying to get the problem fixed once and for once. The symptoms really don't make any sense when the BT shows a bug, and I would rather have a pedantic API that makes it obvious what should happen if an action explicitly failed that to continue with that Schrödinger cat API...

There are, AFAIK, 4 upgrades:

1. grenade
2. firebomb
3. medkit
4. jetpack

Of those, grenade and firebomb have specific code, but do not really make sense as upgrades anyway. IMO they should be put in one's hands, and then thrown, possibly adding some better time and direction control, which would possibly be more fitting as weapon (don't know, really). Those are the ones which actually triggered the bug.
You covered medkit's case.
Jetpack is not yet usable by bots, but this could happen. When it becomes, I don't think it will belong to BT anyway, but rather to pathfinding, and is the only case in which one would want to enable an already enabled upgrade.

src/sgame/sg_bot_ai.cpp
bmorel: Jetpack?
But as I said earlier, jetpack's usage would not belong to BT anyway, since it directly and only affects the path finding.

slipher: No. Jetpack activation has nothing to do with this stuff - it's handled in pmove, triggered by the jump button.

src/sgame/sg_bot_ai.cpp
bmorel: Was going to edit, but changes in between, so separate post:

You covered medkit's case, but it is worth mentioning that it could use the same BT code reduction as the grenades received, that is (not tested, might trigger other problems):

```
decorator return( STATUS_FAILURE )
{
	condition ( percentHealth( E_SELF ) <= 0.5  )
	{
		action activateUpgrade( UP_MEDKIT )
	}
}
```

Instead of:

```
decorator return( STATUS_FAILURE )
{
	condition ( percentHealth( E_SELF ) <= 0.5 && haveUpgrade( UP_MEDKIT ) )
	{
		action activateUpgrade( UP_MEDKIT )
	}
}
```

src/sgame/sg_bot_ai.cpp
bmorel: I'm ok to delete it, but honestly I think #1727 is what should be done instead of random API changes. I kind of doubt it, but maybe @Gireen uses it for example (the ticket I've linked was created to get that kind of infos before doing wasteful efforts).

bmorel:
@slipher: just checking, do you see a blocker here?

src/sgame/sg_bot_ai.cpp
slipher: I think the BotActionDeactivateUpgrade changes don't make sense, but the action is useless so it doesn't matter. LGTM

PR #1756, opened 2022-02-17
use real mass to shove, not health

bmorel:
Removes old code using the actual health to know someone's mass in order to shove them out of the way.
This might improve situation of aliens being blocked by smaller one when fleeing (because hurt, hence less life)

### DIFF
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -246,7 +246,7 @@ static void ClientShove( gentity_t *ent, gentity_t *victim )
 	}
 
 	force = g_shove.Get() * entMass / vicMass;
-	force = Math::Clamp( force, 0.f, 150.f );
+	force = Math::Clamp( force, 0.f, g_shoveMaxForce.Get() );
 
 	// Give the victim some shove velocity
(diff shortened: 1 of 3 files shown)


bmorel:
Thinking about this function even more:

```
  // Cannot push enemy players unless they are walking on the player
  if ( !G_OnSameTeam( ent, victim ) && 
       victim->client->ps.groundEntityNum != ent - g_entities )
  {
    return;
  }
```

I think there is no reason to not being able to push enemies, especially for a charging tyrant or a dragoon thrown at full speed.
Not sure yet what's the 2nd part of the check though

bmorel:
Ok, so, now, I have verified in-game that this works. The default behavior is *not* changed, instead I've added CVar values to tweak the limitations (max force of 150 and only shove allies).
I think those should allow one to experiment with settings to prevent big aliens being blocked by lesser beings.

An easy way to see what it does is:

* /devmap plat23
* /god
* join humans (because human bots won't go melee)
* add a bunch of aliens
* set g_shove to 150
* set g_shoveEnemies to true
* move to someplace aliens can surround you
* enjoy the free moving

bmorel:
This does not really works nor feel good in game, so I'm closing it.

PR #1757, opened 2022-02-17
Simpler bg class

bmorel:
This basically allows to repeat oneself less. That is, when I do `BG_Class(...)` I do not need to say that I want STAT_CLASS.
I'm not sure how useful this is, considering the amount of changes it triggers.

This was made out of curiosity, but I think it is a decent candidate for merge. "Requires" #1756 because it touches things I changed in there. This is by doing this curiosity check that I found that health is used as mass to shove...


bmorel:
I'm converting this to draft, not because it does not work or might bug, but because I am thinking this change alone really does not makes much sense.
I think overall it would be better to have more [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) and less exposure of internals, but having this for only BG_Class() makes little sense.

PR #1759, opened 2022-02-18
double ENTITY_USE_RANGE

bmorel:
I think it's too much and situation could be improved in a better way, but let's have something which works before something which is nice.

See this to get an idea about the actual distance it allows one to buy:

2022-02-18-160021_1362x748_scrot

### DIFF
--- a/src/shared/bg_gameplay.h
+++ b/src/shared/bg_gameplay.h
@@ -307,7 +307,7 @@ extern int   MEDKIT_STARTUP_SPEED;
 
 #define QU_TO_METER                        0.03125 // in m/qu
 
-#define ENTITY_USE_RANGE                   64.0f
+#define ENTITY_USE_RANGE                   128.0f
 
 // fire
 #define FIRE_MIN_DISTANCE                  20.0f


necessarily-equal:
Doubling feels large IMO, but the current value is indeed too small.

Just to say I did propose 96.0f as a middle ground in my first PR, and encountered quite some resistance because it was « too large » :P

If anything, I still feels like 128.0f is better than 64.0f.

bmorel:
> and encountered quite some resistance because it was « too large » :P

I still think 128 is way too large, and I think this PR is just a workaround, but I lack the time to find a real, clean, solution. Why I do this PR is to avoid having to remember to apply/cancel a random number of patches when building my own mod.

A better solution would have be to make this value a configuration value for each "usable" building, and maybe (ok, likely!) it would be worth it to refactor code a bit in the process to have booster using such value too.

> Just to say I did propose 96.0f as a middle ground in my first PR, and encountered quite some resistance because it was « too large » :P

That's why I tend to avoid changes that bring non-configurable gameplay changes :)

slipher:
Should we just revert #1546? Apparently it used to be 128 but only for buying.

necessarily-equal:
The distance for buying always used to be 64, but what is new is that 
it does the check including if the trace succeeded when you are looking 
at something.

Maybe we should use 64 + buildableSide/2 instead?

slipher:
> The distance for buying always used to be 64

What was `#define ENTITY_BUY_RANGE 128.0f` then?

necessarily-equal:
It should have been called `ENTITY_RESUPPLY_RANGE` because ot was about refuelling and getting munitions, although it allowed bots and people with binds to buy from further away

Le 21 février 2022 08:46:42 GMT+01:00, slipher ***@***.***> a écrit :

bmorel:
Was it not the range at which one could buy with commands and resupply, which what what gave the feeling the bots could do instant buy or whatever (they were simply not facing armory, as any player with keybinds would do, coupled with the insane range)

necessarily-equal:
Yes, it is also that range

Le 21 février 2022 09:24:25 GMT+01:00, "Morel Bérenger" ***@***.***> a écrit :

bmorel:
> Yes, it is also that range

And I think that was the problem (and likely still is).

necessarily-equal:
Ok, it _was_ that range, it has been unified since

Le 21 février 2022 09:30:08 GMT+01:00, "Morel Bérenger" ***@***.***> a écrit :

bmorel:
The main difference between after and before the #1546 PR is: there's a single value for everyone, bots, players with binds, and players without binds.
I do not think it's a good idea to revert that improvement.

Now, it introduced a regression, which is that it makes things very tedious for human players now, having to hug the armoury.

Using `BG_BuildableBoundingBox( buildable, mins, nullptr ); 64 + mins[0]` would possibly be good, but I think this belongs to a different PR, as to be done cleanly, this would require modifying many more places in the code, which is out of the scope of current PR. The scope of this PR is to make the game enjoyable, not (for once considering my PRs!) to refactor things.

illwieckz:
We can adjust in the future if we feel the need to.

PR #1760, opened 2022-02-18
fix FireBullet types: no need for float

bmorel:


src/sgame/sg_weapon.cpp
```diff
@@ -483,14 +483,14 @@ MACHINEGUN
...
-static void FireBullet( gentity_t *self, float spread, float damage, meansOfDeath_t mod, int other )
+static void FireBullet( gentity_t *self, int spread, int damage, meansOfDeath_t mod, int other )
 {
 	trace_t   tr;
 	vec3_t    end;
 	gentity_t *target;
 
 	VectorMA( muzzle, 8192 * 16, forward, end );
-	if ( spread > 0.f )
+	if ( spread )
```
necessarily-equal, line 493:
```suggestion
	if ( spread > 0 )
```
maybe?

necessarily-equal (approved):
I'm not fond of the change for spread, and I'd say if you actually wanted less casts, you would have to go and change that in the parsing code too.

This still has the advantage of using one type (almost) all along the way, and doesn't change the behavior of the current code, so LGTM anyway.

bmorel:
I'm closing this because I will supersede it soon, with a sounder API, that is, just pass the damn weapon number to the function instead of passing each elements one by one.

PR #1762, opened 2022-02-18
debian: update packaging, also rewrite unvanquished script

illwieckz:
The unvanquished-helper script is now renamed to unvanquished
naming it (or naming a symlink to it) to unvanquished-server
or unvanquished-tty will launch daemonded or daemon-tty instead.
It now ignores unvanquished-migrate-profile script if not there.

The dpk archives are now provided within unvanquished-data package
instead of providing a download script with (now removed)
unvanquished-data-downloader package.

The unvanquished .desktop file is updated and a separate protocol
handler one is added.


debian/copyright
```diff
@@ -3,11 +3,12 @@ Upstream-Name: unvanquished
 Source: http://github.com/Unvanquished/Unvanquished
 
 Files: *
-Copyright: 2011-2013 Unvanquished Development (and others)
+Copyright: 2011-2021 Unvanquished Development (and others)
```
bmorel, line 6: 2021 or 2022?

bmorel (commented):
Would have been nice to have this in multiple commits, 38 files modified/deleted in one go is pretty huge, considering the changes are not all trivial.
Still, from the stuff I understand, LGTML.

What I do not understand much is the bloated^W systemd and the .desktop stuff.

debian/copyright
illwieckz: Welp, I have set 2021 because the last release in changelog is from 2021, but on the other hand it also builds commits from 2022 so, maybe 2022.

illwieckz:
> Would have been nice to have this in multiple commits, 38 files modified/deleted in one go is pretty huge, considering the changes are not all trivial. Still, from the stuff I understand, LGTML (considering the fact I may have missed stuff with all the deleted files and the likes, also I have _not_ tested it).

Most of those changes are just noise, like file renaming noise. For example package unvanquished-data-downloader has been renamed to unvanquished-data, so even before editing anything it makes a lot of files being moved (added and deleted), git can detect moved files but it's wrong to rename them and not edit them to implement their new behavior, so git can't track them as renames… In the same way, the unvanquished-helper script was edited to be fully autonomous and remove the need for specific unvanquished, unvanquished-server, unvanquished-tty callers, but then once unvanquished, unvanquished-server, unvanquished-tty are deleted, unvanquished-helper can be renamed to unvanquished, so once again git sees a bit edit to unvanquished while it's unvanquished-helper being renamed to unvanquished and all other variants being deleted.

Also some files are just one liners because that's the way Debian does, so editing multiple lines may edit multiple files because everything is split in tiny files.

It makes no sense to make it in multiple commits because no one intermediary commit would work. Also this is not really Unvanquished, this is the packaging script for Debian, so either it works at packaging Unvanquished, either it does not.

illwieckz:
Some things that are not noise and may receive comments:

Not Debian specific:
- the idea to make a single unvanquished script, only symlinking it (or copying it) as unvanquished-server or unvanquished-tty;
- the idea to ignore the unvanquished-migrate-profile script if not found, a packager for a specific distro can just chose to ignore it;
- the idea to make possible to just ignore `/etc/unvanquished/paths.conf` if not found (the config file purposed for specific distro path specialization so one script can be used on many distro as soon as the distro-specific paths are set there), so the script may also be used in non-standard installations (like a portable installation).

Debian specific:

- the idea of packaging data instead of packaging a data downloader.

bmorel:
> Also some files are just one liners because that's the way Debian does

And considering that they even use tarballs which have a file just to store version, and how that seems so neat to me now, I'd say I don't blame them for that :)

Overall, I didn't knew the situation before, so I can't judge all. The size of the commit is quite overwhelming, and if splitting is not possible, then so be it. The stuff I've looked at really felt ok to me, only the 2021 thing made me questioning the content.
I only warn, again, that I do not know Freedesktop's standards (systemd and .desktop) since I never use them (using runit and standalone window manager makes me quite far from those stuff) but otherwise it seems good for me.
You could probably wait for someone else's opinion though.

illwieckz:
This does not affect the game itself neither the way we distribute it officially anyway, and this fixes something that was broken.

PR #1763, opened 2022-02-18
Weapon refactoring

bmorel:
The intent of this branch is the reduction of the number of named global variables like LEVEL0_BITE_RANGE (and others)

### DIFF
--- a/src/cgame/cg_weapons.cpp
+++ b/src/cgame/cg_weapons.cpp
@@ -2566,7 +2566,7 @@ static void ShotgunPattern( vec3_t origin, vec3_t origin2, int seed, int attacke
 	for ( i = 0; i < SHOTGUN_PELLETS; i++ )
 	{
 		r = Q_crandom( &seed ) * M_PI;
-		a = Q_random( &seed ) * SHOTGUN_SPREAD * 16;
+		a = Q_random( &seed ) * BG_Weapon( WP_SHOTGUN )->spread * 16;
 
 		u = sinf( r ) * a;
 		r = cosf( r ) * a;
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1151,7 +1151,6 @@ bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target )
 			secondaryRange = 0;
 			break;
 		case WP_PAIN_SAW:
-			range = PAINSAW_RANGE;
 			secondaryRange = 0;
 			break;
 		case WP_FLAMER:
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1088,14 +1088,16 @@ bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target )
 {
 	ASSERT( target.targetsValidEntity() );
 
-	float range = BG_Weapon( BG_PrimaryWeapon( self->client->ps.stats ) )->range;
+	weaponAttributes_t const* wpa = BG_Weapon( BG_PrimaryWeapon( self->client->ps.stats ) );
+	float range = wpa->range;
+	float width = wpa->width;
+	float height = wpa->height;
 	float secondaryRange;
 	vec3_t forward, right, up;
 	vec3_t muzzle;
 	vec3_t maxs, mins;
 	vec3_t targetPos;
 	trace_t trace;
-	float width = 0, height = 0;
 
 	AngleVectors( self->client->ps.viewangles, forward, right, up );
 	G_CalcMuzzlePoint( self, forward, right, up , muzzle );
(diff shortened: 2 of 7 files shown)


bmorel:
This is a draft because while it could likely be merged as is (and it would be a gain already!) further developments will likely make it more useful.
I am trying to do one commit per variable in order to make things easier to review, but this imply there will be a lot of commits in there.

bmorel:
This removes most of the melee specific variables, favoring weaponAttributes_t struct. Unfortunately, means of deaths are hard-coded and can not be put in the struct easily, preventing to get rid of the huge switch case for melee weapons.

I will likely do the same work for ranged weapons later.
This takes a modified version of #1764 (that is, the check for mantis is done by the FireMelee instead of the calling switch)

slipher:
The point of the `bg_configVars` thing seems to be precise accounting of which attributes are needed for which classes. Thanks to this, you will get a warning if you put a range attribute for a non-ranged weapon, or if you leave off the range attribute for a ranged weapon. With the current refactoring that feature is lost.

Maybe this would allow eventual removal of duplicate code paths in some cases, but from this PR on its own we don't seem to get any benefits.

Note that the `defined |= ...` stuff is for warning if there are missing attributes. I see that there are some new values for those enums defined, but they are never actually checked. (And checking them would usually be wrong since the values are not always required.)

illwieckz:
I really appreciate the fact that after this PR there is less boilerplate in the code. @slipher would you know if there can be alternate ways to check for needed attributes? For example maybe we can set some bools for required features?

illwieckz:
There is also this point of view: let's imagine the fact if it has a range attribute it defines it's a ranged weapon or not. There would be no check to do. At some point the code would just read the config file and run the code related to the attributes. So to create a new weapon, one would just have to create a new config file. New code would only be required when adding an new mechanism never implemented before. Then a weapon would be ranged for the sole reason it has a ranged attribute. There would be no meaning of a ranged weapon without range attribute because the range attribute would define it is ranged or not.

To me this is the direction to go. This is also the direction where modding can be done by adding config files (or pushing buttons in an editor) and where new code would only be required for new mechanism.

slipher:
> So to create a new weapon, one would just have to create a new config file. New code would only be required when adding an new mechanism never implemented before. Then a weapon would be ranged for the sole reason it has a ranged attribute. There would be no meaning of a ranged weapon without range attribute because the range attribute would define it is ranged or not.

No, this vision does not appeal to me. I don't see any benefit in try in to create some sort of building blocks for weapons like Unity or something. Unvanquished is a game, not a game engine. Modding is great, but hopefully modders would add something genuinely new, rather than just changing some stat numbers and enabling the mass driver scope on the lasgun. For interesting contributions, you need to write code.

illwieckz:
> For interesting contributions, you need to write code.

Even us (as Unvanquished) have failed at writing such interesting code in years. First beneficiaries of making gameplay editing easier would be us (Unvanquished).

This is not about a game engine, it's about having a gamecode architecture we can rely on without having to reinvent the wheel and copy-paste boiler plate every time we have to implement a gameplay change.

slipher:
No one added new weapons simply because the team did not decide to add a new weapon. Weapon code is not a bottleneck.

FWIW, I was working on a "total conversion" mod a few months ago and added a new weapon; I didn't experience particular difficulties from the current architecture. The ability to mix and match properties of previously existing weapons wouldn't have helped as the new weapon was different from existing ones (a thrown projectile like a grenade, but I wanted to have it as a weapon in order to have it displayed with a "gun" model).

In my modding experience, everything in config files was a giant pain in the ass and a worse experience than working with code. (Other than model stuff, where I support config files, although binary searching for the right model bounding boxes/scaling factors is still too tedious with vid_restarting; what I actually do is temporarily hack in a cvar that offsets the bounds.)

illwieckz:
> The ability to mix and match properties of previously existing weapons wouldn't have helped

Does this PR would have been an obstacle to do such mod?

We currently have contributors to Unvanquished gamecode, something we didn't have for years. My concern is to make their life easy. I'm astonished by the patience (or stubbornness) to still be there and try pushing forward while getting so few results at getting their improvements merged. If someone like @bmorel considers such PR as a foundation for more to come, I would be afraid to not see this merged if not merging it would prevent other changes to come.

How much this PR is an obstacle to improve Unvanquished?

slipher:
My particular mod is probably unaffected the changes in this PR.

The regression is that it no longer tracks which variables are required and which ones are not used.

The current situation with bg_configVars is ugly, yes. But this PR hardly improves it. Having a weapon configuration god struct with 100 options is not a nice thing, especially when there are many attribute combinations that would be broken if you tried to turn them on simultaneously.

I think refactoring PRs should only be accepted if they are a clear improvement.

illwieckz:
From: https://github.com/Unvanquished/gameplay/issues/25

> This is a good example of how it's bad to allow config options that don't do anything.

This argument is invalid because if those config files were the only places to do things, they would not be dead variables there.

slipher:
> This argument is invalid because if those config files were the only places to do things, there would not be dead variables there. 

This PR doesn't even change anything with what's in the config file. It just stores values from the config file in weapon attribute structs instead of individually wrapped global variables. 

> It just means things are currently inconsistent, not that a solution is better than the other.

If all of the bg_configVars variables were migrated to the config structs following the approach of this PR, the only behavioral difference would be removing the warnings for missing required variables and useless variables.

bmorel:
There's actually a bug in that PR, because of a typo. I didn't updated since it seems pointless.

illwieckz:
I just notice this proves we can also have plenty of dead values in C++ files without warnings, just like when doing them in config files:

- #1779

bmorel:
> I just notice this proves we can also have plenty of dead values in C++ files without warnings, just like when we do them in config files:

Yes. As long are things are public, they may be used as external symbols.
For private members, (some, including clang and gcc) compilers will emit warnings, but it has been decided I think that partially private structs is a bad thing.

PR #1764, opened 2022-02-19
Clarify usage of DAMAGE_xxx flags

slipher:


bmorel:
So, this makes mantis' slowness back as a special case... I can't say I understand the motivation behind it, but the code is okay at least.
I suppose the special case could be handled by the CBSE stuff, but considering how hard it is for me to just drop it's Vec3 usage, I'd rather having touching it myself when possible and thus won't ask for others to do that neither.

bmorel:
I took the freedom to adapt your commit and merge it in #1763 along with more cleanup. Main difference is, the check for mantis attack in done in FireMelee, which imo makes more sense as there is already a specific case for the painsaw anyway.

necessarily-equal:
idk what's the current situation about https://github.com/Unvanquished/Unvanquished/pull/1763, but LGTM

PR #1766, opened 2022-02-19
partial revert of 44b29a6 to fix a pathfinding regression

### LINKED ISSUE #1765: pathfinding regression found in 44b29a6c4
There's a bug which has been been plaguing my mod since several weeks. I thought it was a change in my own code since I play a lot with obstacle detection and handling, but today I decided to find and fix it, and after several bissects, I identified the commit 44b29a6c4 to be the root cause, while not seeing any obvious cause for that.

Here is a chasm layout which makes it easy to reproduce:
[chasm-prison.dat.tar.gz](https://github.com/Unvanquished/Unvanquished/files/8103099/chasm-prison.dat.tar.gz)

How to reproduce:

* /devmap chasm chasm-prison
* /bot fill 5 a 9
* /bot fill 5 h 9
* move as spec so that you can see the blue arrow on following screenshot
* when bots will spawn on the nearby telenode, they will be stuck

2022-02-19-232819_1362x748_scrot

I'm still trying to figure out more, and hopefully find a patch to fix the situation.

bmorel:
fix #1765

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -5727,12 +5727,6 @@ static bool BotAddCmd( gentity_t* ent, const Cmd::Args& args )
 
 	const char* behavior = args.Argc() >= 6 ? args[5].data() : BOT_DEFAULT_BEHAVIOR;
 
-	if ( !G_BotInit() )
-	{
-		ADMP( QQ( N_( "Navigation mesh files unavailable for this map" ) ) );
-		return false;
-	}
-
 	bool result = G_BotAdd( name, team, skill, behavior );
 	if ( !result )
 	{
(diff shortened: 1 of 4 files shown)


bmorel:
I applied this patch on top of my last betterai release, it works and seems to bring no more bugs.

bmorel:
While describing the issue to someone, I said "it's like they have no navmesh" and actually, I think this might be the real issue there, would explain why it only seems to happen after spawn.
In any cases, I'll also need to double check something about performances: I had the feeling somehow this PR made game starting faster. That would be surprising, ofc, so I need to do actual measurements. I'm curious to see if other problems arise, since I've noticed huge performance problems those days, and put them on my experiments, but I'm no longer certain about that.

slipher:
LGTM

PR #1769, opened 2022-02-20
migrate generated code to CBSE

bmorel:
This PR aim at removing the Vec3 dependency in the code generated by the CBSE. It does *not* compile, though, and I need help on that.
It is required to touch the CBSE stuff in order to fully migrate to glm in any cases.


slipher:
A double-migration Vec3 -> vec3_t -> glm::vec3 for this kind of stuff requires dealing with a lot of semantic differences between the types, which is a big waste of your and reviewers' time. Directly Vec3 -> glm::vec3 would be so much easier since it is a nearly drop-in replacement.

bmorel:
well, it seems it passed with glm instead indeed, so that's how things will go: a 1st shot to convert remaining Vec3/Vec2 (others?) to glm stuff, and then progressive migration to vec3_t to glm::vec3.

PR #1724, opened 2022-01-30
Glm

bmorel:

PR #1770, opened 2022-02-20
Glm2

bmorel:
This is a step by step PR to replace Vec3 in the CBSE subsystem by glm::vec3.
I think it would be better to squash all commits working on TurretComponnent, as they are even smaller steps than others and include some code-style changes I needed to help me navigate the code, but I think it will be easier to review in the split state.
One important thing though: the middle steps ARE UGLY. Do not nitpick on them, as there is a "cleanup" commit to fix that and match the usual unvanquished code style at the end.

The remaining instances after this are in botlib, which already have a pending PR, and in cgame/CombatFeedback.cpp. I'll try to migrate the cgame ones to vec3_t, but in case it's impossible this will have to wait for this PR to merged.
After those steps, I'll progressively merge the changes I have pending in #1724 , that should be atomic enough in the end, but I need to review them to 1st provide some compatibility layer.


bmorel:
Just added the commit to remove Vec3 from cgame.

.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "pkg/unvanquished_src.dpkdir"]
 	path = pkg/unvanquished_src.dpkdir
 	url = https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir.git
+[submodule "libs/glm"]
+	path = libs/glm
+	url = https://github.com/g-truc/glm
```
illwieckz, line 18: This should not be the cause of your problem BUT it would be better if you add the trailing `.git`.

bmorel: let's check then

bmorel: It seems it was because I had not added `libs/glm` to the commit.

bmorel:
Note that the [2 things](https://github.com/Unvanquished/Unvanquished/pull/1724#issuecomment-1033538725) issued by @slipher on #1724 are fixed, in this one, that is:

> Noting a couple of things discussed on IRC:
>
Define GLM_FORCE_EXPLICIT_CTOR, to avoid the unwanted type conversions discussed above
Get rid of all instances of the Math::Vec3 (etc.) library

The plan is to replay partially #1724 on top of this, adapting it on the fly with the things I've learned since then. I'm keeping #1724 open for reference, but that's all: new PRs will be issued one after the other to progressively move from vec3_t to glm::vec3 (and other quaternions, vec2_t and all).
The status of the draft is: all the AI code is migrated to glm::vec3, which means at the end of the draft merge, there will be considerably more code migrated to glm than there ever were with Vec3.

bmorel:
Converting to draft because I'll rebase this over #1755 and #1763 and am planning to continue working on this PR (on top of the other two)

src/sgame/components/TurretComponent.cpp
```diff
@@ -1,19 +1,19 @@
 #include "TurretComponent.h"
 #include "../Entities.h"
```
slipher, line 2: This file is in the "new style" as described at https://wiki.unvanquished.net/wiki/Coding_convention. Please remove the `for readability's sake!` commit.

slipher (commented):
I looked at things up through the commit `glm: convert SpikerComponnent`.

src/sgame/components/TurretComponent.cpp
illwieckz: I'm OK to look away from this if it makes easier to merge this PR. I know how it can be hard to rebase some work _after_ such change. This can be a mistake but I prefer time and effort invested on making the game better. Or maybe freem can just revert the style as an ending commit, without rebasing everything.

We have to reduce contribution cost and review cost.

PR #1773, opened 2022-02-21
adds cg_drawPosition: position indicator for the HUD

bmorel:
Requires a change in the UI files to be effective. One problem is, when merged, this will trigger a conflict with RMLUI, hence targeting future 0.54 for now (will need patches, but people will be able to apply this on their local repos if willing to do so).

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -1780,6 +1780,7 @@ extern Cvar::Cvar<bool> cg_thirdPersonPitchFollow;
 extern Cvar::Cvar<float> cg_thirdPersonRange;
 extern Cvar::Cvar<bool> cg_lagometer;
 extern Cvar::Range<Cvar::Cvar<int>> cg_drawSpeed;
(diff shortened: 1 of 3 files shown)


slipher:
You can target UI changes to the 0.53 branch, where RmlUi has been merged.

bmorel:
yes, but I want to benefit from it now :)
I still maintain my mod, it have some bugs sometimes, but the features I ask and bring are always related to it, and I'm kind of afraid that 0.53 with the rmlui change will block us for nothing (it's really just about C++14 for what I understand, which is a *bugfix*)

slipher:
Pull requests to the libRocket code (i.e. ones that will create merge conflicts with the RmlUi changes) will not be accepted.

bmorel:
I know, I'm keeping this open for the day we'll be in 0.53. See the target.

necessarily-equal:
Reviving this would need a cleanup and a port to RmlUI

PR #1774, opened 2022-02-21
bots do not lag, so do not call G_UnlaggedOn

bmorel:
According to a comment found, G_UnlaggedOn might be rather costly and is
only useful for clients, not stuff running in server.
Bots do run in server, so this patch prevents them to enable Unlagged
mode *in sg_weapon*.
In my opinion, the best thing to do would be to do the check in the
G_UnlaggedOn() and G_UnlaggedOff() calls, but this would take a bit more
time and I'm not even certain it is a good idea to do so, hence doing it
the quick'n dirty way

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -290,6 +290,7 @@ bool G_BotAdd( const char *name, team_t team, int skill, const char *behavior, b
 
 	// register user information
 	userinfo[0] = '\0';
+	Info_SetValueForKey( userinfo, "cg_unlagged", "0", false ); // bots do not lag
 	Info_SetValueForKey( userinfo, "name", name ? name : "", false ); // allow defaulting
 	Info_SetValueForKey( userinfo, "rate", "25000", false );
 	Info_SetValueForKey( userinfo, "snaps", "20", false );


bmorel:
@slipher I'm afraid only you of the active contribs knows enough on that topic to tell if that's a good, or a bad idea.
Basically, I inserted a check that the entity doing the action is not a bot, that is, checked that there's a client, but no botmind.

slipher:
Good idea. It could be accomplished more cleanly by setting `cg_unlagged` to 0 in the bot's userinfo. See the userinfo stuff in G_BotAdd

bmorel:
I did that, with the:

```
  // DEPRECATED: Use InfoMap
  void       Info_SetValueForKey( char *s, const char *key, const char *value , bool big );
```

Since I have no clue what "big" is and it's set to false around, I used false too. I would have used the non-deprecated way, but then I have no clue how it is meant to be be used.
I would suppose something along the lines of `InfoMap[ "cg_unlagged" ] = "false";`, but no clue.

src/sgame/sg_bot.cpp
```diff
@@ -286,6 +286,7 @@ bool G_BotAdd( const char *name, team_t team, int skill, const char *behavior, b
 
 	// register user information
 	userinfo[0] = '\0';
+	Info_SetValueForKey( userinfo, "cg_unlagged", "true", false ); // bots do not lag
```
slipher, line 289: It's an integer and you need to turn it off, not on.

src/sgame/sg_bot.cpp
```diff
@@ -290,6 +290,7 @@ bool G_BotAdd( const char *name, team_t team, int skill, const char *behavior, b
 
 	// register user information
 	userinfo[0] = '\0';
+	Info_SetValueForKey( userinfo, "cg_unlagged", 0, false ); // bots do not lag
```
slipher, line 293: The value has to be a string. This is interpreted as nullptr

PR #1781, opened 2022-03-07
cg_rocket_datasource: sort resolutions by size (larger first)

illwieckz:
[sort resolutions](https://dl.illwieckz.net/b/unvanquished/bugs/sort-resolutions/unvanquished_2022-03-07_042646_000.png)

I don't think `CG_Rocket_SortResolutionList` is used anyway.

### DIFF
--- a/src/cgame/cg_rocket_datasource.cpp
+++ b/src/cgame/cg_rocket_datasource.cpp
@@ -547,6 +547,9 @@ void CG_Rocket_BuildResolutionList( const char* )
 
 	buf[ 0 ] = '\0';
 
+	// Sort resolutions by size by default (larger first).
+	std::sort(rocketInfo.data.resolutions,rocketInfo.data.resolutions+rocketInfo.data.resolutionCount,[](const resolution_t&a, const resolution_t& b){return a.width*a.height > b.width*b.height;});
+
 	Rocket_DSClearTable( "resolutions", "default" );
 
 	for ( i = 0; i < rocketInfo.data.resolutionCount; ++i )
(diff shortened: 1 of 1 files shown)


src/cgame/cg_rocket_datasource.cpp
```diff
@@ -547,6 +569,9 @@ void CG_Rocket_BuildResolutionList( const char* )
 
 	buf[ 0 ] = '\0';
 
+	// Sort resolutions by size by default (larger first).
+	qsort( rocketInfo.data.resolutions, rocketInfo.data.resolutionCount, sizeof( resolution_t ), &ResolutionListCmpBySize );
```
slipher, line 573: Use std::sort. `std::sort(rocketInfo.data.resolutions,rocketInfo.data.resolutions+rocketInfo.data.resolutionCount,[](const resolution_t&a, const resolution_t& b){return a.width*a.height < b.width*b.height;});`

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -569,31 +594,16 @@ void CG_Rocket_BuildResolutionList( const char* )
...
-	if ( a->width > b->width ) return -1;
-
-	if ( b->width > a->width ) return 1;
-
-	if ( a->width == b->width )  return 0;
-
-	return 0; // silence compiler
 }
 
 void CG_Rocket_SortResolutionList( const char*, const char *sortBy )
```
slipher, line 599: CG_Rocket_SortResolutionList could indeed be deleted - it's only useful to have a sort function if the table has more than one column.

src/cgame/cg_rocket_datasource.cpp
illwieckz: Done (your suggestion listed smaller first, turned it to larger first).

illwieckz:
Did the suggested `std::sort` thing and also deleted `CG_Rocket_SortResolutionList()` and `ResolutionListCmpByWidth()`.

slipher:
LGTM

PR #1782, opened 2022-03-08
Map list: lazily load levelshots; ignore .arena

slipher:
Companion: https://github.com/DaemonEngine/Daemon/pull/601

See https://github.com/DaemonEngine/Daemon/issues/557.

Before: When the map list is opened, all paks with the map- prefix are
opened and files with the meta/ prefix are opened. Arena files are
parsed. The map's name is displayed in the list according to the
'longname' key in the .arena.

After: The map list is built based on pak names that begin with "map-".
The meta/ prefix is loaded for a given pak only when you click on it in
the list. The pak name (excluding the prefix) is used as the name of the
map in the list. Arena files are not used.

### DIFF
--- a/src/cgame/cg_gameinfo.cpp
+++ b/src/cgame/cg_gameinfo.cpp
@@ -117,9 +117,11 @@ int CG_ParseInfos( const char *buf, int max, char *infos[] )
 /*
 ===============
 CG_LoadArenasFromFile
+
+TODO: consider deleting, unless we want to display 'longname' somewhere
 ===============
 */
-static void CG_LoadArenasFromFile( char *filename )
+void CG_LoadArenasFromFile( char *filename )
 {
 	std::error_code err;
 	std::string text = FS::PakPath::ReadFile( filename, err );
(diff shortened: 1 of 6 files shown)


illwieckz:
I tested with ~200 maps and it is instantaneous! while before it lagged for some seconds!

I was a bit wondering about the list only displaying the short name, but in fact it's good this way (also it helps people to know them by names, so they can mapvote them), and it ensures the list is clean (not wonky colors in the map list).

illwieckz:
About:

```c
// TODO: consider deleting, unless we want to display 'longname' somewhere
```

On of my old idea is that in the future we display the long name and the author name(s), and maybe the version in the map load screen, instead of baking them in levelshot (and to avoid submitting 4K images to repo anytime we want to change the text to be displayed at map load).

illwieckz:
In fact, it would be good to display map longname when selecting the map, for example above or under the level shot. But that can be done later.

illwieckz (approved):
LGTM.

illwieckz:
Now that https://github.com/DaemonEngine/Daemon/pull/601 is merged we need to merge this one or build will break.

Issue #1368, opened 2021-04-13
Telenodes and eggs can spawn a player overlapping another


slipher:
When a human is standing at a certain spot on the telenode, another human may spawn with an overlapping bounding box with the first one. This renders both of them unable to move. It is pretty easy to reproduce like this:
- Have 1 telenode
- Stand in the middle of the telenode
- Add a human bot
- Walk slowly toward the edge of the node until the bot spawns

This happened to me twice while testing the 0.52 release candidate. But I was able to reproduce using my own 0.52/sync gamelogic with 0.51 assets.

Below are programmatic repro instructions, which I used with `map-forlorn_0.15.dpk`. You can conveniently run them by saving in a file `<homepath>/config/hspawnbug.cfg`, and typing `/exec hspawnbug.cfg` in the console.

```
devmap forlorn
delay 10f team h
delay 20f class rifle
delay 30f setviewpos 1879.062134 1602.956055 341.411255 -49.762573 8.412258
delay 40f bot fill 3 h
```

necessarily-equal:
This issue is also reproducible with aliens

illwieckz:
I believe there is a cvar to enable in order to push out people standing on a telenode, to prevent people blocking the spawn.
It probably does not pushes players from the other team though.

It would be better to fix the bug anyway.

slipher:
It also happens in 0.51

Gireen:
Duplicate of #1160 
Seems like the spawnblock area is neither high nor wide enough.
So far i couldn't find out how this area is defined.

necessarily-equal:
Some patch making the denial zone large enough for most cases:

```diff
diff --git a/src/sgame/components/EggComponent.cpp b/src/sgame/components/EggComponent.cpp
index d016879f9..9a0d59d81 100644
--- a/src/sgame/components/EggComponent.cpp
+++ b/src/sgame/components/EggComponent.cpp
@@ -27,11 +27,12 @@ bool EggComponent::CheckSpawnPoint(
 
 	// Method 1.
 	// This describes the worst case for any angle.
-	/*{
+	{
 		displacement = std::max(VectorLength(spawnerMins), VectorLength(spawnerMaxs)) +
 		               std::max(VectorLength(clientMins),  VectorLength(clientMaxs))  + 1.0f;
-	}*/
+	}
 
+#if 0
 	// Method 2
 	// For normals with equal X and Y component these results are optimal. Otherwise they are still
[...]

Issue #1768, opened 2022-02-19
spawn buildings are not blocked when allied player is standing nearby


bmorel:
How to reproduce: remove all but 1 spawn, move on that spawn, add a bot.

What the game should do: warn the player that they're blocking the spawn, and *not* spawn the next player until the place is blocked.
Other actions like nudging the blocker or killing it can be nice, killing it being already implemented.

slipher:
Try grepping (case-insensitively) for `cvar.*block` and see if that's helpful

necessarily-equal:
I don't think this is about the same thing. It's that the the spawn distance check is borked. Here is a patch that kinda fixes it:

```diff
diff --git a/src/sgame/components/EggComponent.cpp b/src/sgame/components/EggComponent.cpp
index d016879f9..9a0d59d81 100644
--- a/src/sgame/components/EggComponent.cpp
+++ b/src/sgame/components/EggComponent.cpp
@@ -27,11 +27,12 @@ bool EggComponent::CheckSpawnPoint(
 
 	// Method 1.
 	// This describes the worst case for any angle.
-	/*{
+	{
 		displacement = std::max(VectorLength(spawnerMins), VectorLength(spawnerMaxs)) +
 		               std::max(VectorLength(clientMins),  VectorLength(clientMaxs))  + 1.0f;
-	}*/
+	}
 
+#if 0
 	// Method 2
 	// For normals with equal X and Y component these results are optimal. Otherwise they are still
 	// better than naive worst-case calculations.
@@ -63,6 +64,7 @@ bool EggComponent::CheckSpawnPoint(
 		// optimal (closest possible) distance.
 		displacement = (1.0f - frac) * minDistance + frac * maxDistance + 1.0f;
 	}
+#endif
 
 	spawnPoint = spawnerOrigin + displacement * spawnerNormal;
 
```

Maybe I misunderstood, but I think this is a duplicate of https://github.com/Unvanquished/Unvanquished/issues/1368.

bmorel:
Yes, duplicate of that one indeed. I tried to found related issue before opening but looks like I failed to use the correct search words.

Issue #1160, opened 2020-02-24
Spawnblock area of telenodes to small


Gireen:
The area above telenodes that prevents spawning is smaller then the space occupied by a spawned human. 

reproduce: jump on a telenode and wait for a bot to spawn results in 2 stuck players.

slipher:
Closing in favor of #1368 since it has more details.

PR #1783, opened 2022-03-10
Simple fix for the bounding box problem

necessarily-equal:
It's not optimal, but the "optimal" version we currently have has many
issues.

Fix https://github.com/Unvanquished/Unvanquished/issues/1368 and its
duplicates https://github.com/Unvanquished/Unvanquished/issues/1160 and
https://github.com/Unvanquished/Unvanquished/issues/1768.

Note by trying hard enough, you can still get stuck, but that requires
jumping towards the egg/telenode at an unfortunate time. Fixing that
would require more design.

### DIFF
--- a/src/sgame/components/EggComponent.cpp
+++ b/src/sgame/components/EggComponent.cpp
@@ -25,13 +25,15 @@ bool EggComponent::CheckSpawnPoint(
 	// TODO: Merge both methods below.
 	float displacement;
 
+#if 1
 	// Method 1.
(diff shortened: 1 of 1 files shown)


slipher:
Can you check if #1785 fixes the problems with egg spawning? It fixes the bug with telenodes but I'm not sure how to reproduce it with eggs.

illwieckz:
I assume related issues were fixed by #1785 that superseded this.

If someone can reproduce with egg, please reopen #1368.

PR #1788, opened 2022-03-17
rework flamer damage

illwieckz:
@bmorel

See also: https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/107

### DIFF
--- a/src/shared/bg_gameplay.h
+++ b/src/shared/bg_gameplay.h
@@ -215,7 +215,7 @@ extern float FLAMER_IGNITE_CHANCE;
 extern float FLAMER_IGNITE_SPLCHANCE;
 #define FLAMER_DAMAGE_MAXDST_MOD 0.5f    // damage decreases linearly from full damage to this during missile lifetime
 #define FLAMER_SPLASH_MINDST_MOD 0.5f    // splash damage increases linearly from this to full damage during lifetime
-#define FLAMER_LEAVE_FIRE_CHANCE 0.3f
+#define FLAMER_LEAVE_FIRE_CHANCE 0.1f
 
 extern int   PRIFLE_DMG;
 extern int   PRIFLE_SPEED;


illwieckz:
The idea is that the flamer is meant to set things on fire, not to be a short-range pulse-rifle neither a fire paint tool.

Drastically reduce direct flamer hit, it's a flame, not a bullet! Also the rate is already very high anyway.

```diff
-damage             10
+damage             2
```

Require more than 3 hit to be sure to set things on fire:

```diff
-igniteChance 0.5
+igniteChance 0.2
```

Do not be able to paint fire on the ground while running backward, you need to have the time to:

```diff
-#define FLAMER_LEAVE_FIRE_CHANCE 0.3f
+#define FLAMER_LEAVE_FIRE_CHANCE 0.1f
```

illwieckz:
I discovered an interesting mechanics while testing this with bots in a 10v10 game: the ability to denial an area even if not killing the aliens: hurting them enough to make them flee.

I would be interested in seeing how human brains playing alien react to this. Where an human carrying flamer can just run at them to hurt them, while they can stare at each other without hurting each other if they are far, knowing that maybe neither one or the other can manage to kill the other one while both have high chance to _hurt_, not necessarily kill.

illwieckz:
On my end I played a 10v10 game with bots and managed to rip out bases, but this was a **challenge**.

I had to buy a medium armour to be able to set plat23 tubes on fire while dancing with goons defending their bases.

If I was teaming up with human brains, it would have been easier to coordinate an attack, with one flamer and someone else carrying a rifle.

I reduced direct damage hit from 10 to 2 because I found 5 ridiculously powerful. And I find the mechanism of “the flamer is to set on fire” very interesting, this also makes it an alternative to the painsaw.

illwieckz:
Thanks to the new nightly test server, we can actually merge this and test it online, allowing us to adjust with future commits if needed.

bmorel:
It's interesting that nobody answered on this yet. LGTM. I don't even see why approval is needed on such a thing, since the flamer is a well known problem, the almost inexistant risks of this change, and the presence of the "new" nightly server.

illwieckz:
Let's do it, we can make it stronger if needed.

PR #1791, opened 2022-03-22
dist/scripts: add record-benchmark-demo

illwieckz:
_Note:_ the script assumes Linux.

So, someone can do something like that:

```bash
./record-benchmark-demo
```

And get a file named `benchmark.dm_86`.

This is to make us able to produce a compatible benchmark demo for every release, without having to care about backward and forward compatibility.

The purpose is to be used by things like https://openbenchmarking.org/test/pts/unvanquished-1.6.0 making life easier for third-party benchmark maintainers by providing them up-to-date demo when releasing the game.

The demo is expected to be used this way:

```
daemon -set in_grab off -set demo.shutdownOnDrop on -set demo.timedemo on +demo_play benchmark
```

The `+attack` trick to build a buildable in the script is currently unreliable because of:

- https://github.com/DaemonEngine/Daemon/issues/613

We may extend the demo to actually buy and display every weapons, evolve to and display every class, and build and display every buildable, etc. so this may also be used to troubleshoot the game.

Anyway, demos produced by this script are good enough for benchmarkers.


illwieckz:
I added a shell script automatizing the demo generation (currently assumes Linux), I updated first post.

dist/scripts/record-benchmark-demo/record-benchmark-demo
```diff
@@ -0,0 +1,64 @@
...
+	-set in_joystick 0 \
+	-set r_fullscreen 0 \
+	-set r_noBorder 1 \
+	-set r_mode -1 \
+	-set r_customwidth 640 \
+	-set r_customheight 480 \
+	-set g_synchronousClients 1 \
+	"${@}" \
+	+preset 'presets/graphics/lowest.cfg' \
+	+minimize \
```
slipher, line 58: If you don't want to see the screen why not use `daemon-tty`?

dist/scripts/record-benchmark-demo/record-benchmark-demo
```diff
@@ -0,0 +1,64 @@
...
+
+mkdir -p "${config_dir}"
+cp -a "${script_dir}/record-benchmark-demo.cfg" "${config_dir}"
+
+"${daemon_path}" \
+	-homepath "${temp_home_path}" \
+	-set name 'benchmark' \
+	-set common.shutdownOnDrop on \
+	-set common.framerate.max 0 \
+	-set common.framerate.maxMinimized 0 \
```
slipher, line 46: Is there some reason to max out the frame rate instead of limiting it to 100 or something?

illwieckz: Hmm, this is not deeply thought, I just wanted to be sure nothing was an obstacle for anything.

Also I was concerned for minimized/unfocused things because of:

- https://github.com/DaemonEngine/Daemon/issues/613

But I'm still trying to figure out things, there may be useless things. I even don't know if we need to set an high framerate. I have also another question: does it have an impact when running `daemon-tty`?

dist/scripts/record-benchmark-demo/record-benchmark-demo
illwieckz: Ah right, I had some trouble when I first experimented with this, but now that I figured it out it seems to work.

illwieckz: Also `minimize` was an attempt to workaround this but it doesn't work it around:

- https://github.com/DaemonEngine/Daemon/issues/613

slipher (commented):
LGTM

PR #1793, opened 2022-04-06
Implement center print font size

slipher:
unvanquished_2022-04-05_223248_000

### DIFF
--- a/src/cgame/cg_draw.cpp
+++ b/src/cgame/cg_draw.cpp
@@ -162,10 +162,11 @@ Called for important messages that should stay in the center of the screen
 for a few moments
 ==============
 */
-void CG_CenterPrint( const char *str, int, int )
+void CG_CenterPrint( const char *str, int, int charWidth )
 {
 	Q_strncpyz( cg.centerPrint, str, sizeof( cg.centerPrint ) );
 	cg.centerPrintTime = cg.time;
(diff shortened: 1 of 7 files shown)


necessarily-equal (commented):
So I'm having trouble seeing how this should fix `g_mapStartupMessage`. That variable is used in `src/sgame/sg_client.cpp` to send a message using `cpd`, which passes the message to `CG_CenterPrint_Delay_f`. This function will then use `/delay` to execute it later.

The problem in that chain, if I'm reading it correctly is that when /delay will execute the command as a regular command and not a server command.

Do you mean that you're fixing it for all but `g_mapStartupMessage`?

slipher:
The problem with `cpd` is that delay commands are executed in a later part of the frame after the UI is updated. The code was written such that CG_CenterPrint only works if it is called earlier in the frame than the UI updating step.

necessarily-equal:
I've tried it out and I couldn't reproduce the startup message working. I tried `+set g_mapStartupMessage hi` and `+devmap chasm +set g_mapStartupMessage hi +set g_mapStartupMessageDelay 20`.

Also I find the font quite large for centerprint, but that's something that can be fixed later. Also, `#define`s for sizes would be nice.

Otherwise, LGTM

slipher:
Probably 20 is too short of a delay

PR #1794, opened 2022-04-18
sg_admin.cpp: more default permissions for lvl4 admin (written into file admin.dat)

Sswweet:
Level 4 admins are the highest level below what is called "server operator". By default, they should be able to pause the game and revert the buildables to a previous state. As far as I know, this is the best response to evil players vandalizing a game.


slipher:
Makes sense.

PR #1795, opened 2022-04-23
allow per-layout mapconfigs

cu-kai:
Please note that default behaviour has changed and... 
### The sgame will attempt to exec configs with the following priority:
```
<g_mapConfigs>/default.cfg
<g_mapConfigs>/mapname/default.cfg
<g_mapConfigs>/mapname/layoutname.cfg
```

### DIFF
--- a/src/sgame/sg_spawn_shared.cpp
+++ b/src/sgame/sg_spawn_shared.cpp
@@ -69,6 +69,10 @@ void think_aimAtTarget( gentity_t *self )
 	VectorAdd( self->r.absmin, self->r.absmax, origin );
 	VectorScale( origin, 0.5f, origin );
 
+	if ( self->s.origin2[0] != 0.f || self->s.origin2[1] != 0.f || self->s.origin2[2] != 0.f )
+	{
+		return;
+	}
 	pickedTarget = G_PickRandomTargetFor( self );
 
 	if ( !pickedTarget )


Gireen:
Isn't this already the current behavior?

slipher:
Reading `G_InitGame`, `G_MapConfigs` is called about 60 lines *before* G_LayoutSelect, so the code as written cannot work. The layout is not known yet when G_MapConfigs executes.

Also I think you are supposed to be able to select the layout(s) by setting `g_layouts` in a map config. This would not be possible if the configs execute after layout selection. However, this is already broken: due to command buffering, map configs only execute after the first frame, and therefore fail to influence layout selection of the *current* map.

As an alternative to layout-specific configs, you use an `if` command. For example `if $layout$ eq foo "/set abc 123"` This takes advantage of the same command buffering - the layout is known by the time the config file is executed.

cu-kai:
> Isn't this already the current behavior?

No, now mapconfigs must go in a subfolder for that map, and there is an extra `default.cfg` for that map always loaded before any layout specific one (if required)



> Reading `G_InitGame`, `G_MapConfigs` is called about 60 lines _before_ G_LayoutSelect, so the code as written cannot work. The layout is not known yet when G_MapConfigs executes.
> 
> Also I think you are supposed to be able to select the layout(s) by setting `g_layouts` in a map config. This would not be possible if the configs execute after layout selection. However, this is already broken: due to command buffering, map configs only execute after the first frame, and therefore fail to influence layout selection of the _current_ map.
> 
> As an alternative to layout-specific configs, you use an `if` command. For example `if $layout$ eq foo "/set abc 123"` This takes advantage of the same command buffering - the layout is known by the time the config file is executed.

I mentioned this on IRC last night, I have to admit it was a quick and dirty patch that worked at initial testing. Perhaps if #1796 is resolved this could be merged. I'll leave this PR open for now unless anyone objects.

Gireen:
I could have sworn that this was already the case. Maybe i mix it up with something that was there in tremfusion.

Regarding file structure 
```
<g_mapConfigs>/default.cfg
<g_mapConfigs>/mapname.cfg
<g_mapConfigs>/mapname/layoutname.cfg
```
I think just adding a per map folder for layout specific config would be better.
Its less work if you want to only add default configs and we also have a layout that's named default, it overwrites *BUILTIN* if no layout is defined.

bmorel:
Note that the PR was done after me asking how to change config depending on the layout, for a PvE layouts. Slipher's answer allowed me to implement that, but this lacks documentation AFAICT.
The only concern I have with having a single map config file with a lot of `if $layout$ eq bla { ... }` is that it might become a mess to maintain, especially if there's several server instances to keep in sync. But then, there's very little use for that currently, so I'm not sure if it's worth patching C++, in terms of efforts (would still be a nice to have though).

slipher:
+1 to Gireen's directory layout suggestion.

Actually as long as #1796 is *not* fixed, I believe the PR can quite easily be fixed by moving G_MapConfigs after layout selection. (If #1796 were fixed it seems more complicated because then you might want some configs to run before layout selection and some after.) I doubt #1796 will be fixed any time soon so I'd be fine the suggestion above.

bmorel:
+1 too for that (it's how I had understood the PR even...). Only question is, is "default.cfg" run before the per-map configs, and then is that per-map config run before the per-layout config?

cu-kai:
> +1 too for that (it's how I had understood the PR even...). Only question is, is "default.cfg" run before the per-map configs, and then is that per-map config run before the per-layout config?

Yes that's exactly how it's intended to work, though `<g_mapConfigs>/mapname.cfg` is a better way to do this, like @Gireen said on IRC.

PR #1799, opened 2022-05-01
Give bot permission to server admins

necessarily-equal:


necessarily-equal:
Note I haven't tested this works

slipher:
LGTM

PR #1800, opened 2022-05-01
Use the new team balancing cvar in the recommended server config

necessarily-equal:


illwieckz:
Can you also update it there?

https://github.com/Unvanquished/Unvanquished/blob/e87791f1f1e69032fe74ec6e0f7d0a241f92bd87/src/sgame/sg_main.cpp#L85

Also it would be useful to update the comment in the sample file, maybe listing the other option too.

slipher:
> Can you also update it there?

That would be a bad idea because the default values should be something suitable for a LAN game. In a local game the players should be able to do whatever they want. The point of the sample configs is that good defaults for a server are different than for a local game.

illwieckz:
Oups, I missread the line I quoted, the default value is “no restriction at all”, so yes, only the sample config file has to be edited.

dist/configs/config/server.cfg
```diff
@@ -58,7 +58,7 @@ set timelimit 0
 set g_emptyTeamsSkipMapTime 15
 
 // force each team to have a similar number of players
-set g_teamForceBalance 1
+set g_teamForceBalance 2
```
illwieckz, line 61: Wording suggestion:

```c
// team enforcement
// 0 - no restriction
// 1 - force each team to have a similar number of players
// 2 - force each team to have a similar number of players once there are players in both team, make "players versus bots" possible if no player joins the other team
set g_teamForceBalance 2
```

necessarily-equal: I've pushed another formulation, that is
```diff
 // 1 - force each team to have a similar number of players
+// 2 - if players start going in both teams, force each team to have a similar number of players; otherwise players can all join the same team for a human vs bots game
 set g_teamForceBalance 2
```

Tell me which you prefer

slipher:
LGTM

PR #1805, opened 2022-05-23
Fix g_teamForceBalance and multiline command output

slipher:


bmorel:
I guess the fix is only the 1st commit. I'm failing to understand the point of the 4th commit though, could you provide (in the new `Quote()`'s comment possibly) an example where it is useful to have multiline commands?

Other than that, LGTM

slipher:
So in response to a client command e.g. `listplayers` the server may send back a command like ```print_tr "<multiline string>"```

PR #1810, opened 2022-05-28
remove outdated archlinux folder

bmorel:
This folder's content seems superseeded by stuff in the dist folder.
The content is basically files to run the daemon on a server.
Stuff in the `debian` dir does not looks like the same, rather, it looks like it's how  to build a .deb, so not removed, but it could likely be cleaned up too: I've not checked.


necessarily-equal:
The AUR package uses it https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=unvanquished#n136

I guess you would need to ask @Viech to check what's still useful

bmorel:
> The AUR package uses it

Then it's broken, since there's no more "server.lanOnly" option.

Viech:
> I guess you would need to ask @Viech to check what's still useful

You can remove it, I will then update my PKGBUILDs to point to `dist`.

bmorel:
Let's wait a bit more and merge if nobody is against then. @Viech: I think you can already update the PKGBUILDs, though, since from what I've seen, stuff in the `dist` folder is way more recent (and that's the way arch's stuff should be, right? ;)) notably the .desktop file is for version 1.5 instead of 1, the server.cfg does not contains stuff which are no longer in code, etc.

necessarily-equal:
I've checked the NixOS package and Faltpak and they don't use it either. So no problem on that side either

necessarily-equal (approved):

Viech:
> @Viech: I think you can already update the PKGBUILDs

Yes but the pressure is not there yet. :wink:

bmorel:
Nobody argued against in a week, so let's merge.

PR #1815, opened 2022-05-28
warn about bug 1811

bmorel:
I do not like the idea of releasing 0.53 with that bug without warning the unwarry admin or to delay the release date longer.
This PR does not solves the problem, but at least gives a warning, which might be enough considering the fact this file has to be installed manually.

### DIFF
--- a/dist/configs/config/map/default.cfg
+++ b/dist/configs/config/map/default.cfg
@@ -1,5 +1,9 @@
 // define some bot names
 bot names clear
+
+// Note than defining names will make the game crash if one
+// adds more bots than there are names.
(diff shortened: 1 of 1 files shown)


necessarily-equal (approved):

slipher:
This bug sounds very easy to fix so it seems silly to add a warning instead of just fixing it.

bmorel:
I agree with that tbh, but fixing stuff requires people to actually do it. Yet, it's a bug that will rather rarely happen, but that I would consider critical, while I'm more than a bit reticent to delay 0.53 *more*.

slipher:
#1818

Issue #1812, opened 2022-05-28
"Warn: no more slots for bot" is spammed all the time


bmorel:
Instead, this should be only printed *once* when the command is run.
And the warning should be more visible, like what happens when one does a `/warn` in console, since the actual chat might be flooded already, and this means that teams will be imbalanced, so it's important to not miss it. And one can't scroll the scoreboard AFAIK so only the log can be used to notice that fact.
I also suspect this can have a huge performance cost.

necessarily-equal:
This doesn't seem to be run every frame, so I'd say this is fine performance-wise. Does this happen only with `bot fill 32`? I could reproduce with `bot fill 32` but not `30`.

bmorel:
Just do `/bot fill 127`, observe the output of `top`, and then do `bot fill FOO` replacing FOO with half the maximum number of clients on your server.
On my cheap VPS, there is a real performance impact, maybe because I log unvanquished's stdout and stderr by piping them to another process (which only adds a timestamp, saves them and does rotation).

slipher:
Bot fill quotas are checked every 2 seconds IIRC.

necessarily-equal:
Workaround: don't try to add more bots than the maximum number of allowed client minus one.

bmorel:
> Bot fill quotas are checked every 2 seconds IIRC.

Exact. One warning per full team, every 2 seconds. The problem is less important than I thought then, but I think a correct behavior would be to warn only when the command is ran, with number of effectively filled bots set to maximum value.

bmorel:
I no longer think this can be fixed, as there is no way to know what will happen in the future, that is, players or spectators arriving or leaving the game. This would change the number of clients, and hence affect the number of slots.

xXRaptorSc0pezXx:
The “Warn: no more slots for bot” message really needs to get fixed before the next release it is driving me crazy having it flash up in the corner constantly. Like I'll just start a local game and spawn in and it automatically fills the teams up and then that message just keeps flashing over and over and over and over it's incredibly annoying. It literally makes the game unplayable with how distracting it is for me.

slipher:
@xXRaptorSc0pezXx This is likely a result of using the `Server setup > Bots: General > Minimum bots per team` option. You can fix your personal issue by sliding it to 0 (or any number less than 10).

xXRaptorSc0pezXx:
[...]

PR #1817, opened 2022-05-28
Add an upper bound for the number of bots added by `/bot fill`

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1812.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -5735,6 +5735,33 @@ static bool BotAddCmd( gentity_t* ent, const Cmd::Args& args )
 	return result;
 }
 
+static int BotFillCalculateMax(const std::vector<team_t> &teams)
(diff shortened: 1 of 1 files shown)


slipher:
I don't really see the point of this. The error can still happen e.g. if there are too many spectators. And it might interfere with legitimate commands, let's say you were experimenting with max alien bots like `bot fill 99 a` and then decided to get rid of the alien bots and do max human bots and typed `bot fill 99 h; bot fill 0 a` but then the game would refuse to add any human bots. If the message is that annoying we could change it to a Verbose log level instead.

necessarily-equal:
In the meantime, I found another way to break it: adding more than half the available bot count with `/bot add`, then asking it to add as many bots as possible `/bot fill 50`. This would need some more work. Marking as draft.

PR #1818, opened 2022-05-28
Fix crash when bot name list exhausted

### LINKED ISSUE #1811: game crash when names are exhausted
When running a server using bot names, adding more bots that there are names lead to a crash.

slipher:
Fixes #1811.


bmorel:
LGTM

Issue #1802, opened 2022-05-15
Implement levelTime condition for BTs


bmorel:
This would allow bots to change behavior based on how old the game is, to become more aggressive, for example.
See #1801

necessarily-equal:
"LGTM" by @bmorel on IRC

bmorel:
This actually already exists:

* https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_bot_parse.cpp#L351
* https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_bot_parse.cpp#L223-L226

PR #1820, opened 2022-05-29
Add levelTime BT function

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1802


necessarily-equal:
"LGTM" by @bmorel on IRC

slipher:
There is already `matchTime` which gives `level.matchTime`, which seems a bit more appropriate for behavior trees than `level.time`.

bmorel:
Agree that matchTime is likely more suited. I simply didn't not noticed this BT condition when I opened the isssue. Mea culpa!

PR #1823, opened 2022-05-31
dead code and macro removal

bmorel:
Remove unused function.

### DIFF
--- a/src/sgame/sg_client.cpp
+++ b/src/sgame/sg_client.cpp
@@ -1376,24 +1376,19 @@ void ClientBegin( int clientNum )
 	}
 }
 
-/** Sets shared client entity parameters. */
-#define CLIENT_ENTITY_SET_PARAMS(params)\
-	params.oldEnt = ent;\
-	params.Client_clientData = client;\
-	params.Health_maxHealth = BG_Class(pcl)->health;
-
-/** Copies component state for components that all client entities share. */
-#define CLIENT_ENTITY_COPY_STATE()\
-	if (evolving) {\
-		*ent->entity->Get<HealthComponent>() = *oldEntity->Get<HealthComponent>();\
-	}
-
 /** Creates basic client entity of specific type, copying state from an old instance. */
-#define CLIENT_ENTITY_CREATE(entityType)\
-	entityType::Params params;\
-	CLIENT_ENTITY_SET_PARAMS(params);\
-	ent->entity = new entityType(params);\
-	CLIENT_ENTITY_COPY_STATE();
+template <typename entityType>
+void ClientEntityCreate( gentity_t* ent, Entity *oldEntity, gclient_t *client, class_t pcl, bool evolving )
+{
+	typename entityType::Params params;
+	params.oldEnt = ent;
+	params.Client_clientData = client;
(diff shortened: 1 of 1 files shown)


src/sgame/sg_buildable.cpp
```diff
@@ -1702,110 +1702,97 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
...
+	ent->entity = new entityType(params);
+}
 
-#define BUILDABLE_ENTITY_START(entityType)\
-	entityType::Params params;\
-	BUILDABLE_ENTITY_SET_PARAMS(params);
-
-#define BUILDABLE_ENTITY_END(entityType)\
+template < typename entityType, typename std::enable_if<
+	std::is_same<entityType,DrillEntity>::value ||
```
slipher, line 1715: This will never compile as written (`void = 0`)

src/sgame/sg_buildable.cpp
```diff
@@ -1702,110 +1702,97 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
...
+	std::is_same<entityType,LeechEntity>::value
+	>::type = 0>
+void BUILDABLE_ENTITY_CREATE( gentity_t *ent, buildable_t buildable )
```
slipher, line 1718: This seems overall less readable than what we have now. The specialization requires some verbose code duplication and is far from its point of use.

If you insist on replacing the macros with something else, it seems better to make template specializations for only the part that's different (i.e. nothing for most of them and the one line for mining structures). 

Personally if I were to simplify the original code somehow, I would just get rid of BUILDABLE_ENTITY_CREATE and write out BUILDABLE_ENTITY_START and BUILDABLE_ENTITY_END for each structure so that it's obvious which parts are in common.

src/sgame/sg_buildable.cpp
bmorel: Wdym? This actually passes my local build and the CI, so I guess you're not talking about compiler itself?

src/sgame/sg_buildable.cpp
bmorel: > it seems better to make template specializations for only the part that's different (i.e. nothing for most of them and the one line for mining structures).

I don't see how this would work. To have 2 template functions, as with before but instead of macros?

I've split the commit in 2, so that it would be easy to remove it since it's possibly how this will end.

src/sgame/sg_buildable.cpp
slipher: > This actually passes my local build and the CI, so I guess you're not talking about compiler itself?

Yes I expect the build to pass, but *this template specialization* will never compile. With the `= 0` or `= true` type examples you need to specify a 2nd argument for `enable_if` so that `type` is something that can be 0.

src/sgame/sg_buildable.cpp
bmorel: I'm not too sure how to do that, nor when i'll have time and motivation to do that, so I'll just remove this part from the PR. Easier for everyone like this (and it's not important at all anyway).

slipher:
LGTM

PR #1825, opened 2022-06-04
Memleak fixes

bmorel:
The new version of BG_strdup is used only once, but it is possible other places could use it, I've not checked for the calls in rocket stuff.

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -1366,11 +1366,23 @@ static void CG_BuildableParticleEffects( centity_t *cent )
 CG_BuildableStatusParse
 ==================
 */
+static bool CG_RegisterShaderFromConf( int handle, qhandle_t& qhandle )
+{
+	pc_token_t token;
+
+	if ( !Parse_ReadTokenHandle( handle, &token ) )
+	{
+		return false;
+	}
+
+	qhandle = trap_R_RegisterShader( token.string, static_cast<RegisterShaderFlags_t>( RSF_NOMIP ) );
+	return true;
+}
+
 void CG_BuildableStatusParse( const char *filename, buildStat_t *bs )
 {
 	pc_token_t token;
 	int        handle;
-	const char *s;
 	int        i;
 	float      f;
 	Color::Color c;
@@ -1391,38 +1403,22 @@ void CG_BuildableStatusParse( const char *filename, buildStat_t *bs )
 
 		if ( !Q_stricmp( token.string, "frameShader" ) )
 		{
-			if ( PC_String_Parse( handle, &s ) )
-			{
-				bs->frameShader = trap_R_RegisterShader( s, (RegisterShaderFlags_t) ( RSF_NOMIP ) );
-			}
-
+			CG_RegisterShaderFromConf( handle, bs->frameShader );
 			continue;
 		}
 		else if ( !Q_stricmp( token.string, "overlayShader" ) )
 		{
-			if ( PC_String_Parse( handle, &s ) )
-			{
-				bs->overlayShader = trap_R_RegisterShader( s, (RegisterShaderFlags_t) ( RSF_NOMIP ) );
-			}
-
+			CG_RegisterShaderFromConf( handle, bs->overlayShader );
 			continue;
 		}
 		else if ( !Q_stricmp( token.string, "noPowerShader" ) )
 		{
-			if ( PC_String_Parse( handle, &s ) )
-			{
-				bs->noPowerShader = trap_R_RegisterShader( s, (RegisterShaderFlags_t) ( RSF_NOMIP ) );
-			}
-
+			CG_RegisterShaderFromConf( handle, bs->noPowerShader );
 			continue;
 		}
 		else if ( !Q_stricmp( token.string, "markedShader" ) )
 		{
-			if ( PC_String_Parse( handle, &s ) )
-			{
-				bs->markedShader = trap_R_RegisterShader( s, (RegisterShaderFlags_t) ( RSF_NOMIP ) );
-			}
-
+			CG_RegisterShaderFromConf( handle, bs->markedShader );
 			continue;
 		}
 		else if ( !Q_stricmp( token.string, "healthSevereColor" ) )
--- a/src/shared/bot_nav_shared.h
+++ b/src/shared/bot_nav_shared.h
@@ -202,12 +202,12 @@ struct BasicMeshProcess : public dtTileCacheMeshProcess
 
 struct LinearAllocator : public dtTileCacheAlloc
 {
-	unsigned char* buffer;
-	size_t capacity;
-	size_t top;
-	size_t high;
+	unsigned char* buffer = nullptr;
+	size_t capacity = 0;
+	size_t top = 0;
+	size_t high = 0;
 
-	LinearAllocator( const size_t cap ) : buffer(nullptr), capacity(0), top(0), high(0)
+	LinearAllocator( const size_t cap )
 	{
 		resize(cap);
 	}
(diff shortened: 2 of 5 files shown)


src/cgame/cg_buildable.cpp
```diff
@@ -1366,11 +1366,23 @@ static void CG_BuildableParticleEffects( centity_t *cent )
...
+static bool PC_String_Parse( int handle, qhandle_t& qhandle )
+{
+	pc_token_t token;
+
+	if ( !Parse_ReadTokenHandle( handle, &token ) )
+	{
+		return false;
+	}
+
+	qhandle = trap_R_RegisterShader( token.string, static_cast<RegisterShaderFlags_t>( RSF_NOMIP ) );
```
slipher, line 1378: Why is there something about registering shaders in a function called "string parse"?

src/shared/bg_public.h
```diff
@@ -1777,6 +1777,8 @@ int cmdcmp( const void *a, const void *b );
 char *Quote( const char *str );
 char *Substring( const char *in, int start, int count );
 char *BG_strdup( const char *string );
+// This BG_strdup version does *not* allocates memory
+bool BG_strdup( const char *src, char *dst, size_t dst_sz );
```
slipher, line 1781: that's just the same as `Q_strncpyz`

src/sgame/sg_bot_parse.cpp
```diff
@@ -1403,7 +1403,10 @@ pc_token_list *CreateTokenList( int handle )
 		current->token.intvalue = token.intvalue;
 		current->token.subtype = token.subtype;
 		current->token.type = token.type;
-		current->token.string = BG_strdup( token.string );
+		if ( !BG_strdup( token.string, current->token.string, sizeof( current->token.string ) ) )
```
slipher, line 1406: This does not make sense. Size of a pointer is being used as a buffer size

src/shared/parse.cpp
```diff
@@ -1351,19 +1351,14 @@ Parse_CopyToken
 */
 static token_t *Parse_CopyToken( const token_t *token )
 {
-	token_t *t;
-
-//  t = (token_t *) malloc(sizeof(token_t));
-	t = ( token_t * ) Z_Malloc( sizeof( token_t ) );
+	token_t *t = new (std::nothrow) token_t( *token );
```
slipher, line 1345: > By default, when the new operator is used to attempt to allocate memory and the handling function is unable to do so, a [bad_alloc](https://www.cplusplus.com/bad_alloc) exception is thrown. But when nothrow is used as argument for new, it returns a null pointer instead.

This is not what we want. We want the game to immediately crash when it runs out of memory, which an unhandled exception accomplishes. Returning a null pointer means the game crashes at some later point which may make it hard to trace where the failed allocation occurs.

src/shared/parse.cpp
```diff
@@ -337,7 +328,7 @@ static void Parse_CreatePunctuationTable( script_t *script, punctuation_t *punct
 	if ( !script->punctuationtable )
 	{
 		script->punctuationtable = ( punctuation_t ** )
-		                           Z_Malloc( 256 * sizeof( punctuation_t * ) );
+		                           BG_Alloc( 256 * sizeof( punctuation_t * ) );
```
slipher, line 331: The commit description could be more precise e.g. "Replace Z_Malloc with BG_Alloc in parse.cpp"

src/shared/bot_nav_shared.h
```diff
@@ -219,12 +219,15 @@ struct LinearAllocator : public dtTileCacheAlloc
 
 	void resize( const size_t cap )
 	{
-		if ( buffer )
+		auto * tmp_buffer = static_cast<unsigned char*>( realloc( buffer, cap ) );
+		if ( !tmp_buffer )
 		{
-			free( buffer );
+			//TODO: I doubt a Log::Warn is enough for that kind of situation,
```
slipher, line 225: `Sys::Drop` would be good

src/shared/parse.cpp
bmorel: Immediate crashes are fine to me, except when the only error message is "Pipe broken", which happens fairly often to me.
I think using `Sys::Drop()` would be better than to rely on exceptions being replaced by "abort()" here.
Ideally, [[nodiscard]] would have been nice (not only for those ones) but it's a C++17 feature, and I guess we can't use it, right?

src/shared/bot_nav_shared.h
bmorel: So be it, then!
Didn't knew about this one, I was naively searching for "Log::Error" or something along this line...

src/shared/parse.cpp
slipher: > I think using Sys::Drop() would be better than to rely on exceptions being replaced by "abort()" here.

Actually we already do this via our `operator new` override in daemon/src/common/System.cpp.

> Ideally, [[nodiscard]] would have been nice (not only for those ones) but it's a C++17 feature, and I guess we can't use it, right?

You can't use the spelling `[[nodiscard]]`, but vendor-specific annotations with equivalent functionality are available in earlier versions. If you want that you can add the appropriate defines for each supported compiler in daemon/src/common/Compiler.h.

slipher: Actually we already have one, `WARN_UNUSED_RESULT`

src/sgame/sg_bot_parse.cpp
bmorel: You gave me a doubt, so I did a quick try locally:

```
#include <stdio.h>

struct bla { char foo[1024]; };
int main() {
	bla hello;
	printf( "%zu\n", sizeof( hello.foo ) );
}
```

Then: `clang++ test.cpp && ./a.out` prints 1024, as expected.

src/shared/parse.cpp
bmorel: For now I'll just do the crash thing though. I dream of getting rid of that parsing code anyway, considering how unreliable it is, so just fixing the memleak bugs shall be enough, especially since this is the scope of this PR.

src/shared/parse.cpp
```diff
@@ -1723,8 +1713,8 @@ static void Parse_FreeDefine( define_t *define )
 	}
 
 	//free the define
-	free( define->name );
-	Z_Free( define );
+	BG_Free( define->name );
```
necessarily-equal, line 1716: This string is allocated with strdup, so I'd say for consistency it'd be better to keep `free()`. (For now this would work fine, but this relies on the current impl of `BG_Free`)

necessarily-equal:
Cherry-picked two commits

src/shared/parse.cpp
bmorel: That's a weird definition of consistency, considering BG_Free() is just free() anyway

src/shared/parse.cpp
```diff
@@ -1351,19 +1351,7 @@ Parse_CopyToken
...
-
-//  t = freetokens;
-	if ( !t )
-	{
-		Sys::Error( "out of token space" );
-	}
-
-//  freetokens = freetokens->next;
-	memcpy( t, token, sizeof( token_t ) );
+	token_t *t = new token_t( *token );
```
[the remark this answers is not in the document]
slipher: I don't understand your comment, but it's illegal to mix `new` and `free`.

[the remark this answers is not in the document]
slipher: There is a `Parse_FreeToken` with numerous call sites.

src/cgame/cg_buildable.cpp
```diff
@@ -1366,11 +1366,23 @@ static void CG_BuildableParticleEffects( centity_t *cent )
...
+static bool CG_RegisterShaderFromConf( int handle, qhandle_t& qhandle )
+{
+	pc_token_t token;
+
+	if ( !Parse_ReadTokenHandle( handle, &token ) )
+	{
+		return false;
+	}
+
+	qhandle = trap_R_RegisterShader( token.string, static_cast<RegisterShaderFlags_t>( RSF_NOMIP ) );
```
slipher, line 1378: Nit: there is no need for `static_cast<RegisterShaderFlags_t>`.

LGTM

bmorel:
Recent changes were simply about renaming a Z_Malloc to BG_Alloc, which was preventing the game to compile. Those functions did the same thing though.

PR #1827, opened 2022-06-04
0.53/glm complete

bmorel:
supersedes #1770. I'm doing branch cleanup, hence the deletion. I'm not particularly interested at having this merged, though. I use it for my own comfort and for my own mod, is all.


bmorel:
With the level of integration I did there, there is only negligible impact to expect. The main reason behind this move from my point of view are: documentation and readability.
The impacts comes from the fact there's a lot of conversions, since daemon itself does not exposes an interface with glm. There's also places in code which still uses vec3_t, vec4_t, col3_t or others, so, conversions.
OTOH vec3_t for example is literally the size of 3 pointers, which are twice as big as float and same size of double (doubles are used by glm internally IIRC, but I think it can be changed).
Add to this the indirections, and there might be performance increase in that area.
GLM is also widely used, and thus maintained by other peoples, which might have added optimisations, however, as they say themselves, they do not have performance in their primary scope.

The reasons I switched from Vec3 first and then started moving even the vec3_t to glm are:

* to have only one lib used across the code
* to have documentation about the math functions (that one is *very* important)
* readability (return by value, passing by value sometimes, `end = start + dir * length;` vs `VectorMA( start, dir, length, end );`, even if those are possible with Vec3 API, this is also a very important reason)
* to use a widely used lib will make it easier for new contributors, since it will be more likely they already used it (glm is used in most if not all the opengl or vulkan tutorials I've seen those last years)

Note that I am already using this branch since several months (more than 6, easily) and my servers are currently the most populated (since gireen stopped his).
I host 3 instances on the cheapest VPS I could find, and do regular debug.
I could find no performance impact (neither negative nor positive) linked to this changed. I have not  seen any regression neither (related to this work, of course: I can name a few regressions in my betterai code).

I have several months of development on top of this branch, so it's hard to evaluate precisely the impact of this work.

I hope this answers the questions.

illwieckz:
I doubt it can be slower that what we have. Maybe things like [realtime math](https://github.com/nfrechette/rtm) can get more performance as such other library is explicitly tailored for this, but if GLM was slower than what people usually expects from such library it would not have its success. There is a lot of SIMD code in GLM.

bmorel:
What I am saying is, performance was not the reason which made me do the work, and that I have seen no performance change.
The reason there is pretty obvious though: heavy calculations are likely done by the engine. There's only minor things in gamelogic.

illwieckz:
One thing is that, even if I know this PR is only about gamelogic and your only GLM effort is on gamelogic, I would not be against a similar migration in engine, so if we already use GLM in gamelogic we would better chose GLM as a potential replacement in engine to avoid deps proliferation, which means the plus and cons of GLM or whatever library are important to me.

illwieckz:
@bmorel said:

> it's a draft because I don't intend to do more efforts on this
> I did more than enough already

@DolceTriade would you look at it to merge this, eventually doing the minor changes that are required because of desync?

illwieckz:
@DolceTriade I just forked https://github.com/Unvanquished/glm so we can actually use this in submodule and tag the commits we use at time of release with tags like `unvanquished/0.53.2`. There is no purpose to have custom code there, just pull it from time to time and tag commits for release.

bmorel:
To be exact, since the opening of this:

> supersedes https://github.com/Unvanquished/Unvanquished/pull/1770. I'm doing branch cleanup, hence the deletion. **I'm not particularly interested at having this merged, though**. I use it for my own comfort and for my own mod, is all.

illwieckz:
Yes but I want this.

DolceTriade:
Yes I will.

PR #1837, opened 2022-06-07
improve use_entity_range mecanism for armory

### LINKED ISSUE #1720: buildings range is not checked properly
Happens on master (b31952f), probably since ENTITY_USE_RANGE was merged with ENTITY_BUY_RANGE, and in consequence reduce to 64qu (with other stuff).

How to reproduce:

* go on armory roof center
* aim anywhere
* try to buy stuff: it works
* try to get the "usable" icon: it won't appear (even if crouching)
* move a bit further
* try to buy stuff: it does not work

For information, armoury is 66qu height (above origin), humans in bsuit is 38qu (under origin), other humans are 24qu (still under), for totals of 104qu and 90qu distances.
Old USE range was 128, and is now 64.

From the behavior in game, one could guess that multiple algorithms are used for the same thing, and after a check in code, indeed, code is duplicated around, the most obvious thing being `CG_BuildableInRange` vs `G_BuildableInRange`, but grepping gave me hints that in some places it implemented outside of function (duplicated code).

In my opinion, the correct thing which should be done is checking distance between BBoxes, and not between origins. Or maybe between the player's origin and the armory's BBox, eventually with a radius bonus.
This problem also brings the question of: "what about the reactor?". Well, the reactor have a maxs height of 100, which is quite bigger compared to armory.

About the "half-sides" (very bad approx of radius, used in some places in code), reactor is 41, armory 31,  humans are 15 (yes, bsuit is exactly as wide as naked humans).

bmorel:
fix #1720

The way it works is by applying the player's sizes before searching for armories in range on sgame.
On cgame, both player's and armory's sizes are considered. There, Reactor is not handled, not sure if that's because I could not find where it is handled though or if it's handled a different way.

Note that this means the current value for ENTITY_USE_RANGE is likely way too big, and that it should probably be back at 64, but this should be done *after* testing.

### DIFF
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -2186,8 +2186,7 @@ void ClientThink_real( gentity_t *self )
 
 		if ( ent && ent->use &&
 		     ( !ent->buildableTeam   || ent->buildableTeam   == client->pers.team ) &&
-		     ( !ent->conditions.team || ent->conditions.team == client->pers.team ) &&
-		     Distance( self->s.origin, ent->s.origin ) < ENTITY_USE_RANGE )
+		     ( !ent->conditions.team || ent->conditions.team == client->pers.team ) )
 		{
 			if ( g_debugEntities.Get() > 1 )
 			{
@@ -2196,28 +2195,9 @@ void ClientThink_real( gentity_t *self )
 
 			ent->use( ent, self, self ); // other and activator are the same in this context
 		}
-		else
+		else if ( client->pers.team == TEAM_ALIENS )
 		{
-			// no entity in front of player - do a small area search
-			for ( ent = nullptr; ( ent = G_IterateEntitiesWithinRadius( ent, client->ps.origin, ENTITY_USE_RANGE ) ); )
-			{
-				if ( ent && ent->use && ent->buildableTeam == client->pers.team)
-				{
-					if ( g_debugEntities.Get() > 1 )
-					{
-						Log::Debug("Calling entity->use after an area-search for %s", etos(ent));
-					}
-
-					ent->use( ent, self, self ); // other and activator are the same in this context
-
-					break;
-				}
-			}
-
-			if ( !ent && client->pers.team == TEAM_ALIENS )
-			{
-				G_TriggerMenu( client->ps.clientNum, MN_A_INFEST );
-			}
+			G_TriggerMenu( client->ps.clientNum, MN_A_INFEST );
 		}
 	}
 
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -1075,9 +1075,23 @@ public:
 
 		const entityState_t &es = cg_entities[ trace.entityNum ].currentState;
 
+		vec3_t originMins, originMaxs;
+		BG_ClassBoundingBox( cg.predictedPlayerState.stats[STAT_CLASS]
+				, originMins, originMaxs, nullptr, nullptr, nullptr );
+
+		//TODO: do not hard-code ARMORY here.
+		//While we're listing stuff to do: check that target is
+		//a usable *before* doing any maths.
+		vec3_t targetMins, targetMaxs;
+		BG_BuildableBoundingBox( BA_H_ARMOURY, originMins, originMaxs );
+
+		float distance = Distance( cg.predictedPlayerState.origin, es.origin )
+			- RadiusFromBounds( originMins, originMaxs )
+			- RadiusFromBounds( targetMins, targetMaxs );
+
 		if ( es.eType == entityType_t::ET_BUILDABLE
 				&& es.modelindex == BA_H_ARMOURY
-				&& Distance(cg.predictedPlayerState.origin, es.origin) < ENTITY_USE_RANGE - 0.2f // use an epsilon to account for rounding errors
+				&& distance < ENTITY_USE_RANGE - 0.2f // use an epsilon to account for rounding errors
 				&& CG_IsOnMyTeam(es) )
 		{
 			if ( !IsVisible() )
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -2153,7 +2153,7 @@ void BotSellWeapons( gentity_t *self )
 	int i;
 
 	//no armoury nearby
-	if ( !G_BuildableInRange( self->client->ps.origin, ENTITY_USE_RANGE, BA_H_ARMOURY ) )
+	if ( !G_InUseRange( self->client->ps, BA_H_ARMOURY ) )
 	{
 		return;
 	}
@@ -2193,7 +2193,7 @@ void BotSellUpgrades( gentity_t *self )
 	int i;
 
 	//no armoury nearby
-	if ( !G_BuildableInRange( self->client->ps.origin, ENTITY_USE_RANGE, BA_H_ARMOURY ) )
+	if ( !G_InUseRange( self->client->ps, BA_H_ARMOURY ) )
 	{
 		return;
 	}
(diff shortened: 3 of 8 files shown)


necessarily-equal (changes requested):
I didn't notice that using `G_BuildableInRange` leads to two different ranges in practice in my previous PR :grimacing: Good find!

## The algorithm

I'm not sure buildable radius is a good enough solution. I'd say it is better than what we currently have, but not that much for the extra complexity.

I remember one complaint with the current implementation + range of 64 was that it is wasn't possible anymore to land on armory with jetpack to replenish ammo and fuel. Given how tall is an armory compared to its base radius, this wouldn't help much.

So how about this:

```C++
// G_Distance, but in 2D, and we remove the 2 bbox equivalent radii like you did
radial_distance = sqrtf( pow(player_origin[0] - armory_origin[0], 2) + powf(player_origin[1] - armory_origin[1], 2) )  // pythagora in 2D*
			- RadiusFromBounds( originMins, originMaxs )
			- RadiusFromBounds( targetMins, targetMaxs );

// These might actually be already available in ent->r.abs{min,max}. Not sure
amory_top = armory_origin[2] + armory_max[2];
amory_bottom = armory_origin[2] + armory_min[2];
player_top = player_origin[2] + armory_max[2];
player_bottom = player_origin[2] + armory_min[2];

if (armory_top > player_bottom && armory_bottom < player_top)
    
if (armory_top < player_bottom)
    Δz = player_bottom - armory_top;
    return sqrtf( Square(radial_distance) + Square(Δz) );

if (armory_bottom > player_top)
    Δz = player_top - armory_bottom;
    return sqrtf( Square(radial_distance) + Square(Δz) );

return radial_distance;
```

Let me try to do terrible ascii art drawings to draw the distances. 

```
|-|
|-|   (those two lines) represent an armory

+ represent the player

\, _ and / represent distances 
```

```
+
 \
  |-|    case 1, player is above armory, pythagoras
  |-|

+_|-|    case 1, player is next to armory, straight line
  |-|

+_
  |-|    case 2bis, when player bottom is equal to top of armory, the line is still straight
  |-|

  |-|    case 3, player is below armory, pythagoras
  |-|
 /
+
```

I hope those drawings aren't too terrible. I can also draw it properly, but then the downside is that you can copy-paste them in-code.

## The places where this should also be applied

In addition to where you did the change, you also need to take into account these:
* `ClientThink_real` from src/sgame/sg_active.cpp
* `BotActionBuy` from src/sgame/sg_bot_ai.cpp
* `BotSellWeapons` and `BotSellUpgrades` from src/sgame/sg_bot_util.cpp

Maybe we can ignore the bot ones, but we need `ClientThink_real`.

(*: there are functions here and there like Distance2 in cg_beacon but apparently no clean implementation. Fixing this is probably not a goal anyway, as this is probably best done with glm directly)

bmorel:
> I'd say it is better than what we currently have, but not that much for the extra complexity.

Your proposed changes are way more complex though.

> I remember one complaint with the current implementation + range of 64 was that it is wasn't possible anymore to land on armory with jetpack to replenish ammo and fuel. Given how tall is an armory compared to its base radius, this wouldn't help much.

THis is true, but OTOH I think this makes little sense to keep. This method is just used by camper players AFAIK.
The intent of this PR is twice:

1: to get the range working for any buildings, since currently it hardly works with RC's energy resupply, and if one wanted to get the drills act as energy refueler as repeaters were in trem, which is something I really want to do (to reduce human dependency to armories is a good idea imo) it would be similarly buggy (that is, for drills, the reload range would be way too big).

2: to have the stuff clean enough to be usable by mappers, to that buttons for example could be togglable on specific action instead of "by accident or luck" as is currently the case.


>
* ClientThink_real from src/sgame/sg_active.cpp
* BotActionBuy from src/sgame/sg_bot_ai.cpp
* BotSellWeapons and BotSellUpgrades from src/sgame/sg_bot_util.cpp

Thanks, will do that.

> as this is probably best done with glm directly

Well, I already know this PR will give me troubles next time I rebase my stuff... but it's worth it.

bmorel:
After this PR, usable entities (currently only reactor and armory) are usable from ~2 meters (according to game's HUD).
Aliens can still evolve (yes, some code is linked to aliens).

src/sgame/sg_active.cpp
```diff
@@ -2186,8 +2186,7 @@ void ClientThink_real( gentity_t *self )
 
 		if ( ent && ent->use &&
 		     ( !ent->buildableTeam   || ent->buildableTeam   == client->pers.team ) &&
-		     ( !ent->conditions.team || ent->conditions.team == client->pers.team ) &&
-		     Distance( self->s.origin, ent->s.origin ) < ENTITY_USE_RANGE )
+		     ( !ent->conditions.team || ent->conditions.team == client->pers.team ) )
```
necessarily-equal, line 2189: This makes the use condition different on client than on server. This means that you can have some interesting bugs like this one where you *think* you can build things, but you actually can't. (notice the error message behind the dialog)

image

src/sgame/sg_active.cpp
```diff
@@ -2196,28 +2195,9 @@ void ClientThink_real( gentity_t *self )
...
-					if ( g_debugEntities.Get() > 1 )
-					{
-						Log::Debug("Calling entity->use after an area-search for %s", etos(ent));
-					}
-
-					ent->use( ent, self, self ); // other and activator are the same in this context
-
-					break;
-				}
-			}
```
necessarily-equal, line 2215: I believe deleting this brings us closer to have interactive usable map entities.

necessarily-equal (commented):
Haven't had a close look at the code, but I found some stuff I don't like already :S

There is unfortunately so much somewhat kinda loosely similar code everywhere, and this gets really error-prone. I did check really carefully that it worked somewhat consistently before making my previous PR that changed that. (It was the whole point of doing it actually)

I think the whole idea of having twice the code, once on the server and once in the client gives more trouble in the end. If I had time and would be making my own PR for this, I think I'd merge the server and client logic to have one (1!) instance of the entity code on the server, and send the "is interactive entity" info on a bit inside `STAT_STATE`. I don't care enough about that right now to do a PR though.

necessarily-equal:
Oh yeah, and I don't like that you HAVE TO look at the armory to use it, but that's a more personal thing. Though the fact that everyone but the user using the `q` key is not obligated to look at the armory is especially annoying. It favorises the folks using binds even more.

src/sgame/sg_active.cpp
necessarily-equal: wait, further, I meant that deleting this brings us further from usable map entities

bmorel:
This PR was never intended to be a draft. Imo, only authors should place a PR as draft.

necessarily-equal:
@DolceTriade has asked about the status of this PR. I wanted to answer that some time back, but posted it on the wrong PR instead, as https://github.com/Unvanquished/Unvanquished/pull/1952#issuecomment-1225945792. I've hidden the message there and am reposting it here. Sorry for the mixup.

---

Unless I missed some change, I still don't consider this PR ready.

Here are my points that I made on IRC a month ago:

        About the armory PR: two or three things. First I'm not fond of this "you have to look at the armory to get the thing to open". It forces you to look at the thing with good precision. You can't even look a bit sideways like you would (could) if you care for life

        Second, [snip, not actually a regression]

        Third, I really don't like that the condition for binds is different. I don't want players with binds to be advantaged even more :-/

        I also note that you can refuel if you stand on armory. [About that ppl don't seem to care if it's good or not, I'm fine with this implementation that makes it work.]

Then a bit afterwards in the conversation, I mention another problem: with your implementation: somone can use an armory on the other side of a wall if, and only if they use binds.

Later on, you said "I agree with the binds problem" and "I'll fix that". Which is why I did put this PR on draft.

I don't think issue 1) mentioned here is blocking even if I personally really don't like it. But I find the bind thing blocking.

PR #1840, opened 2022-06-08
Make default font-size much bigger

necessarily-equal:
This is a prerequisite/basis for https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/113.

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -1765,7 +1765,7 @@ public:
 			showTime_ = cg.centerPrintTime;
 			SetInnerRML( Rocket_QuakeToRML( cg.centerPrint, RP_EMOTICONS ) );
 			SetProperty( Rml::Core::PropertyId::FontSize,
-			             Rml::Core::Property( 0.07f * cg.centerPrintSize, Rml::Core::Property::EM ) );
+			             Rml::Core::Property( cg.centerPrintSizeFactor, Rml::Core::Property::EM ) );
 		}
 
 		SetProperty( "opacity", va( "%f", CG_FadeAlpha( cg.centerPrintTime, CENTER_PRINT_DURATION ) ) );
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -1765,7 +1765,7 @@ public:
 			showTime_ = cg.centerPrintTime;
 			SetInnerRML( Rocket_QuakeToRML( cg.centerPrint, RP_EMOTICONS ) );
 			SetProperty( Rml::Core::PropertyId::FontSize,
-			             Rml::Core::Property( 0.2f * cg.centerPrintSize, Rml::Core::Property::EM ) );
+			             Rml::Core::Property( 0.07f * cg.centerPrintSize, Rml::Core::Property::EM ) );
 		}
 
 		SetProperty( "opacity", va( "%f", CG_FadeAlpha( cg.centerPrintTime, CENTER_PRINT_DURATION ) ) );
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -1717,7 +1717,7 @@ public:
 		if ( map != newMap )
 		{
 			map = newMap;
-			SetInnerRML( va( "<img class='levelshot' src='/meta/%s/%s' />", newMap, newMap ) );
+			SetInnerRML( Str::Format( "<img src='/meta/%s/%s' />", newMap, newMap ) );
 		}
 	}
 
--- a/src/cgame/rocket/rocket_documents.cpp
+++ b/src/cgame/rocket/rocket_documents.cpp
@@ -54,16 +54,15 @@ void Rocket_LoadDocument( const char *path )
 			other->Close();
 		}
 	}
-
 }
 
 // Scale the UI proportional to the screen size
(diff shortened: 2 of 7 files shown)


bmorel (approved):
The C++ side is fine to me, except for the comment thing (not really blocking).
I can't review the css one, since reading original source makes my brain segfault, and the patches in those just makes even less sense to me.
If I was to give my opinion on CSS in general, I'd say it simply does not works nicely with tools like git or diff, since the cascading sh^Wstuff just dispatches the context in too many places.

src/cgame/rocket/rocket_documents.cpp
```diff
@@ -61,9 +61,9 @@ void Rocket_LoadDocument( const char *path )
 void Rocket_SetDocumentScale( Rml::Core::ElementDocument& document )
 {
 	// This gets 12pt on 1920�1080 screen, which is libRocket default for 1em
```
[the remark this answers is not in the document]
necessarily-equal: Updated. `/ 60` is just because it looks good enough as a default.

src/cgame/rocket/rocketDataSource.h
```diff
@@ -246,7 +246,7 @@ class RocketDataSource : public Rml::Core::Element, public Rml::Controls::DataSo
 	void AddCancelbutton()
 	{
 		init = true;
-		Rml::Core::Factory::InstanceElementText( this, va( "<button onClick=\"hide %s\">Cancel</button>", GetOwnerDocument()->GetId().c_str() ) );
+		Rml::Core::Factory::InstanceElementText( this, va( "<button onClick=\"hide %s\"><span>Cancel</span></button>", GetOwnerDocument()->GetId().c_str() ) );
```
slipher, line 249: The circlemenu cancel button is generated by a Lua script, so whatever this is, it's not the circlemenu button.

src/cgame/rocket/rocket_documents.cpp
```diff
@@ -60,10 +60,15 @@ void Rocket_LoadDocument( const char *path )
...
-	// This gets 12pt on 1920�1080 screen, which is libRocket default for 1em
-	int size = std::min( cgs.glconfig.vidWidth, cgs.glconfig.vidHeight ) / 90;
+	// This sets the default font size to 18px on a 1920�1080 screen,
+	// and scales appropriately.
+	int size = std::min( cgs.glconfig.vidWidth, cgs.glconfig.vidHeight ) / 60;
+	// Note that this enforces the font-size on the root element. So any
+	// value you would set on that same element (<window> or <body>) would
+	// be forcibly overriden by this.
+	// Use a rule such as `body > * { font-size: 1.2em; }` to adjust size.
 	document.SetProperty( Rml::Core::PropertyId::FontSize,
```
slipher, line 70: We're using `em` units everywhere, so how does this not throw off everything in the entire UI?

necessarily-equal: This PR is to be tested with https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/113. I made sure that thing look better than they currently do.

To answer your question, « We're using em units everywhere, so how does this not throw off everything in the entire UI? », well we had that problem already when switching to RmlUI: it changed the font size, and we can see it broke a fair bit of things. This time the font size changes are made on purpose, at least :smiley:

src/cgame/rocket/rocketDataSource.h
necessarily-equal: Ok then, dropped the commit.

src/cgame/rocket/rocket_documents.cpp
slipher: I think there is a better way in RmlUi that would let us set a font size on a root element without this wacky `body > *` stuff. Instead of setting the font size here we can use the `SetDensityIndependentPixelSize` function on a context. Then the font size in the CSS can be set using the [dp](https://mikke89.github.io/RmlUiDoc/pages/rcss/syntax#dp-unit) unit.

necessarily-equal:
But then we can't just keep having a stylesheet in em, and we need to 
rewrite all the stylesheets with the new sizes, with the downside 
things won't scale with (local) font sizes anymore. This just seems 
like a net loss.

slipher:
> But then we can't just keep having a stylesheet in em

Yes we can. I'm suggesting to use `dp` *just* for root elements, to set up the documenting scaling in a cleaner way that allows root elements to have additional size adjustments in the CSS.

necessarily-equal:
Ok my bad, I misunderstood at first. This sounds like a good thing to. Done.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -1809,7 +1809,7 @@ class CenterPrintElement : public HudElement
 			showTime_ = cg.centerPrintTime;
 			SetInnerRML( Rocket_QuakeToRML( cg.centerPrint, RP_EMOTICONS ) );
 			SetProperty( Rml::Core::PropertyId::FontSize,
-			             Rml::Core::Property( 0.2f * cg.centerPrintSize, Rml::Core::Property::EM ) );
+			             Rml::Core::Property( 0.07f * cg.centerPrintSize, Rml::Core::Property::EM ) );
```
slipher, line 1812: This makes the smallest center prints unreadably small. For an example of the smallest ones (size 10) you can do `/delay 1000 buy lgun` as a human while not having enough funds. To see one of the largest ones (size 30), destroy all your spawns.

So what needs to happen is changing some of the sizes in calls to `CG_CenterPrint` to decrease the size difference between the smallest and largest ones.

PR #1842, opened 2022-06-09
Delete dead libRocket draw functions

necessarily-equal:
Those aren't used anywhere in the RML.


slipher:
LGTM but needs rebase (don't want a merge commit)

PR #1845, opened 2022-06-12
Introduce the RmlUI debugger

necessarily-equal:
It has a companion PR on unvanquished_src.dpkdir to add keybindings for it.


src/cgame/cg_consolecmds.cpp
```diff
@@ -432,6 +432,7 @@ static void CG_BeaconMenu_f()
 	Rocket_DocumentAction( rocketInfo.menu[ ROCKETMENU_BEACONS ].id, "show" );
 }
 
+void CG_RmlUIDebuggerToogle();
```
slipher, line 435: `Toggle` not `Toogle`

src/cgame/cg_main.cpp
```diff
@@ -1184,6 +1185,8 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 	new(&cg) cg_t{};
 	for ( centity_t& ent : cg_entities) { ent = {}; }
 
+	CG_RmlUIDebuggerDisable(); // CHECKME
```
slipher, line 1188: CG_Init is only called after a VM is freshly loaded so this is not needed here. Maybe you meant to put it in a shutdown function?

slipher (commented):
Cool. It doesn't seem to work outside the main menu though. The buttons don't do anything (even if I open a menu so that I can click things).

src/cgame/rocket/rocket_documents.cpp
```diff
@@ -168,3 +169,30 @@ void Rocket_DocumentAction( const char *name, const char *action )
 		}
 	}
 }
+
+static bool debugger_initialised = false;
+void CG_RmlUIDebuggerToogle()
+{
+	if (!debugger_initialised)
+	{
+		Rml::Debugger::Initialise(*cgs.mapname // CHECKME
```
DolceTriade, line 178: You probably want an argument for hud/menu or at least allow for an override because people may want to debug ingame menus and such when a map is loaded. Furthermore, iirc, we never inject mouse/input events into the HUD context, so you can't actually use the debugger in that context (I think...)

DolceTriade, line 178: Also, once you pick  a context, are you locked into that context forever?

necessarily-equal: Ah, wait, are the in-game menu part of the menu or HUD context?

necessarily-equal: That's the best I can do with current RmlUI, yes. More recent versions have introduced `Rml::Debugger::Shutdown();` which would allow switching

DolceTriade: Anything clickable is always the menuContext (ie, if you're in game and press ESC, you are in the menuContext). The hudContext is exclusively the the HUD (ie, things you can't click on because your mouse/keyboard are captured by the game. The only thing you you might actually want to click on is the tab scoreboard, which is part of the hud context.

DolceTriade: Rml::Debugger::SetContext still exists and probably works.

DolceTriade: Looking deeper, it seems we always want to initialize the debugger into the menuContext and then allow switching between the two via `debuggerToggle menu` or `debuggerToggle hud`. Still unsure if debuggerToggle will work with the hudContext though.

necessarily-equal: Then it makes things more complicated for the keybindings in particular. I'm not sure what would be the correct, non-surprising behavior. Use two keys? Press the same key several times (none->menu->hud->none)?

DolceTriade: In games like CSGO, which have a similar scoreboard, if you open a clickable HUD element, you right click to bring up a mouse.

Clickable HUD elements might be interesting if we ever want to be cool like Doom3 and have clickable menus in the map.

DolceTriade: For now, let's just enable the debugger for the menuContext and leave the hudContext as a TODO. This should make the code easier for now.

src/cgame/cg_main.cpp
DolceTriade: Further, I think rmlui will clean up the debugger by itself in the shutdown function. Thus, I think we can probably remove the function all together.

DolceTriade (changes requested):
This is a good start. Let's finish this up and merge it? The debugger is actually quite a cool tool for debugging the menus.

PR #1851, opened 2022-06-24
fix trigger_class and trigger_equipement

bmorel:
Those triggers were broken compared to tremulous, where they could be
used notably to test for a team (if list is empty).
There is no clue at all that the non-deprecated trigger works, and no
clue that it is actually used, this patch only fixes stuff which is
used in practice.
Some actual examples of usage are:
* tremship (used to detect team, but also for some easter eggs I
  believe)
* unvanquished's 0.52 parpax (used to implement a joke "achievement" in
  vents)

### DIFF
--- a/src/sgame/sg_spawn_sensor.cpp
+++ b/src/sgame/sg_spawn_sensor.cpp
@@ -473,7 +473,8 @@ void sensor_player_touch( gentity_t *self, gentity_t *activator, trace_t* )
 	if ( self->conditions.team && ( activator->client->pers.team != self->conditions.team ) )
 		return;
 
-	// handle legacy maps with trigger_class and trigger_equip
+	// handle legacy `trigger_class` and `trigger_equip` entities
+	// by only checking for class on aliens, and equipment on humans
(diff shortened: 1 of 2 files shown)


bmorel:
To test with 0.52's parpax, load it (be careful to *not* load a more recent one! might work with mini-parpax too), go in the ventilation room near human base, and evolve to either tyrant or advanced dragon.
Doing so should give you 2 different texts, which won't trigger for any other class.

Ideally, a map to test various combinations would be built and put to https://github.com/UnvanquishedAssets/UnvanquishedTestAssets for more testing.

src/sgame/sg_spawn_sensor.cpp
```diff
@@ -473,6 +473,16 @@ void sensor_player_touch( gentity_t *self, gentity_t *activator, trace_t* )
 	if ( self->conditions.team && ( activator->client->pers.team != self->conditions.team ) )
 		return;
 
+	// handle legacy maps with trigger_class and trigger_equip
```
necessarily-equal, line 476:
```suggestion
	// handle legacy `trigger_class` and `trigger_equip` entities
	// by only checking for class on aliens, and equipment on humans
```

PR #1853, opened 2022-06-26
Perf improvements

bmorel:
1st commit  avoid doing useless calculations.
2nd commit prevent doing a trace when target is not in front of bot: checking for angles is faster than doing a trace.

There's likely more places in which this could be useful, but... what should actually be done is reuse the information calculated by BotSearchForEnemy() instead, since there's already calculations of distance and visibility check done.
This should still save few cycles, though.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -4987,7 +4987,7 @@ bool G_admin_flag( gentity_t *ent )
 
 bool G_admin_builder( gentity_t *ent )
 {
-	vec3_t     forward, right, up;
+	vec3_t     forward;
 	vec3_t     start, end, dist;
 	trace_t    tr;
 	gentity_t  *traceEnt;
@@ -5005,12 +5005,12 @@ bool G_admin_builder( gentity_t *ent )
 
 	buildlog = G_admin_permission( ent, "buildlog" );
 
-	AngleVectors( ent->client->ps.viewangles, forward, right, up );
+	AngleVectors( ent->client->ps.viewangles, forward, nullptr, nullptr );
 
 	if ( ent->client->pers.team != TEAM_NONE &&
 	     ent->client->sess.spectatorState == SPECTATOR_NOT )
 	{
-		G_CalcMuzzlePoint( ent, forward, right, up, start );
+		G_CalcMuzzlePoint( ent, forward, start );
 	}
 	else
 	{
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1296,6 +1296,11 @@ bool BotTargetIsVisible( const gentity_t *self, botTarget_t target, int mask )
 	G_CalcMuzzlePoint( self, forward, muzzle );
 	target.getPos( targetPos );
 
+	if ( BotAimAngle( self, targetPos ) > g_bot_fov.Get() / 2 )
+	{
+		return false;
+	}
+
 	if ( !trap_InPVS( muzzle, targetPos ) )
 	{
 		return false;
@@ -1489,7 +1494,7 @@ void BotSlowAim( gentity_t *self, vec3_t target, float slowAmount )
 	VectorMA( viewBase, length, skilledVec, target );
 }
 
-float BotAimAngle( gentity_t *self, vec3_t pos )
+float BotAimAngle( const gentity_t *self, vec3_t pos )
 {
 	vec3_t viewPos;
 	vec3_t forward;
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1097,7 +1097,7 @@ bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target )
 	float width = 0, height = 0;
 
 	AngleVectors( self->client->ps.viewangles, forward, right, up );
-	G_CalcMuzzlePoint( self, forward, right, up , muzzle );
+	G_CalcMuzzlePoint( self, forward, muzzle );
 	target.getPos( targetPos );
 	switch ( self->client->ps.weapon )
 	{
@@ -1290,10 +1290,10 @@ bool BotTargetIsVisible( const gentity_t *self, botTarget_t target, int mask )
 
 	trace_t trace;
 	vec3_t  muzzle, targetPos;
-	vec3_t  forward, right, up;
+	vec3_t  forward;
 
-	AngleVectors( self->client->ps.viewangles, forward, right, up );
-	G_CalcMuzzlePoint( self, forward, right, up, muzzle );
+	AngleVectors( self->client->ps.viewangles, forward, nullptr, nullptr );
+	G_CalcMuzzlePoint( self, forward, muzzle );
 	target.getPos( targetPos );
 
 	if ( !trap_InPVS( muzzle, targetPos ) )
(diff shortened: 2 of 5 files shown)


illwieckz:
Can this avoid a bot to do a 180 turn because there is an enemy behind it detected without the model looking behind? If yes I would not be against such patch because having eyes in the back is cheat, but then sometime the bot should randomly look behind to check for enemies.

Out of curiosity, is such code used by buildables like turrets/rocket pods and guided missiles like swarm and rocket? Because while turret, rocket pod and swarm should look everywhere even if not facing, but we may benefit from such optimization for the rocket as by design the rocket isn't expected to do a 180 turn.

bmorel:
This only affect bots.

For the question of rotating bots, no, it should not change this behavior. The trace which is prevented is forward ony anyway.

I also have a patch to improve that, but it's much more complex and requires GLM to be merged.
I expect said patch to be a huge changer in unv's AIs because of the many bugs it fixes, and what it allows to implement. Still, requires a more modern codebase.

slipher:
> The trace which is prevented is forward ony anyway.

I don't see any thing restricted it to a forward direction. The trace simply goes from the muzzle to the enemy, which could be on any side. The trace may go through the player due to the passEntityNum.

illwieckz:
OKOK.

bmorel:
> The trace simply goes from the muzzle to the enemy

Exact. I was probably confused with a different trace. Well then, this means that enemies indeed have eyes in their back. I guess it's also a bugfix then.

This is used by:

* BT's IsVisible condition
* BotSearchForEnemy (not sure, might be my own local changes only)
* BotActionFight (BT)
* BotActionRepair (BT)

It still has to be noted that BT "fight" use bestEnemy as target, and BotActionRepair does a MoveToGoal if not visible anyway, which includes automatically aiming at.
I doubt this would create problems in master's codebase, basically (but yes, I only tested it with my memory rewrite, so can't be sure)

slipher:
The proposed change doesn't seem good in the BotActionRepair case. If the bot is in range but ends up facing away from the building it wants to repair for whatever reason, then it will attempt to continue navigating, instead of going to the code which attempts to aim toward the building.

necessarily-equal:
LGTM, you may want to check the BotActionRepair just to be sure, but I'd think it should mostly be fine

slipher:
I just looked at the BotActionFight case and it seems the behavior would change there too. With this change, if the target enemy goes out of FOV (I think this could easily happen e.g. by a Marauder jumping over the bot's head) and there is some other nearby enemy, then the bot will forgot about the original enemy and the fight action will terminate. (The block commented with `FIXME: this if or what's in it is wrong.`)

bmorel:
> I just looked at the BotActionFight case and it seems the behavior would change there too. With this change, if the target enemy goes out of FOV [...]

Only cases "action fight" can return STATUS_FAILURE are:

* enemy is no more valid
* something about the node being wrong
* there is no navmesh path toward enemy

So I doubt it. But what you mention is an actual bug (or limitation, call it the way you want) of the current enemy memory system, yes.
So, I checked it (several times, as is needed with AI behavior), with that very same branch (I had a doubt). The result you're describing does not seems to happen. Behavior is the same as with current 0.52.1 (except for the many fixes AIs received in between, ofc)

slipher:
> Only cases "action fight" can return STATUS_FAILURE are:

The case I'm talking about returns STATUS_SUCCESS.

bmorel:
Can you produce the bug you're describing? I'm ok to say there could be a regression here, I know by 1st hand experience that AI's code is... say, finicky, but I could not after several minutes of hide and seek with a bot (using the /god command, ofc, otherwise it would be a pain).

slipher:
I wrote some code to print a log statement when the proposed changes would cause BotActionFight to terminate when it wouldn't have before. I put some bots on plat23 and found that `exiting BotActionFight 1` was frequently printed, which suggests that it causes signifcant behavioral changes.

```
index 4b13dd9ac..77d249fc0 100644
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -772,7 +772,8 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 		self->botMind->enemyLastSeen = level.time;
 	}
 
-	if ( !BotTargetIsVisible( self, self->botMind->goal, CONTENTS_SOLID ) )
+	bool answerChanged = false;
+	if ( !BotTargetIsVisible( self, self->botMind->goal, CONTENTS_SOLID, &answerChanged ) )
 	{
 		botTarget_t proposedTarget;
 		proposedTarget = self->botMind->bestEnemy.ent;
@@ -784,10 +785,14 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 		    != self->botMind->bestEnemy.ent )
 		  && BotPathIsWalkable( self, proposedTarget ) )
 		{
+			if (answerChanged)
+				Log::Notice("exiting BotActionFight 1");
 			return STATUS_SUCCESS;
 		}
 		else if ( level.time - self->botMind->enemyLastSeen >= g_bot_chasetime.Get() )
 		{
+			if (answerChanged)
+				Log::Notice("exiting BotActionFight 2");
 			return STATUS_SUCCESS;
 		}
 		else
diff --git a/src/sgame/sg_bot_util.cpp b/src/sgame/sg_bot_util.cpp
index 57d784b5b..c439cd2f6 100644
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1284,7 +1284,7 @@ bool BotEntityIsValidEnemyTarget( const gentity_t *self, const gentity_t *enemy
 	return true;
 }
 
-bool BotTargetIsVisible( const gentity_t *self, botTarget_t target, int mask )
+bool BotTargetIsVisible( const gentity_t *self, botTarget_t target, int mask, bool* answerChanged )
 {
 	ASSERT( target.targetsValidEntity() );
 
@@ -1314,6 +1314,13 @@ bool BotTargetIsVisible( const gentity_t *self, botTarget_t target, int mask )
 				|| trace.fraction == 1.0f )
 			&& !trace.startsolid )
 	{
+		if ( BotAimAngle( self, targetPos ) > g_bot_fov.Get() / 2 )
+		{
+			if (answerChanged)
+				*answerChanged = true;
+			return false;
+		}
+
 		return true;
 	}
 	return false;
@@ -1489,7 +1496,7 @@ void BotSlowAim( gentity_t *self, vec3_t target, float slowAmount )
 	VectorMA( viewBase, length, skilledVec, target );
 }
 
-float BotAimAngle( gentity_t *self, vec3_t pos )
+float BotAimAngle( const gentity_t *self, vec3_t pos )
 {
 	vec3_t viewPos;
 	vec3_t forward;
diff --git a/src/sgame/sg_bot_util.h b/src/sgame/sg_bot_util.h
index 1f43d72c1..80f7e1526 100644
--- a/src/sgame/sg_bot_util.h
+++ b/src/sgame/sg_bot_util.h
@@ -52,12 +52,12 @@ void  BotGetIdealAimLocation( gentity_t *self, botTarget_t target, vec3_t aimLoc
 void  BotAimAtEnemy( gentity_t *self );
 void  BotSlowAim( gentity_t *self, vec3_t target, float slow );
 void  BotAimAtLocation( gentity_t *self, vec3_t target );
-float BotAimAngle( gentity_t *self, vec3_t pos );
+float BotAimAngle( const gentity_t *self, vec3_t pos );
 
 // targets
 bool BotEntityIsValidTarget( const gentity_t *ent );
 bool BotEntityIsValidEnemyTarget( const gentity_t *self, const gentity_t *enemy );
-bool BotTargetIsVisible( const gentity_t *self, botTarget_t target, int mask );
+bool BotTargetIsVisible( const gentity_t *self, botTarget_t target, int mask, bool* answerChanged = nullptr );
 bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target );
 void BotTargetToRouteTarget( const gentity_t *self, botTarget_t target, botRouteTarget_t *routeTarget );
 botTarget_t BotGetRoamTarget( const gentity_t *self );
```

bmorel:
To be honest, I'm not going to invest much time with this PR on this codebase. I noticed this is a problem while hacking my own codebase, which have many, many AI changes.
Imo "this suggests" is not really a good test (as it does not provides any clue the enemy will then be forgotten, which is initial concern), but then, I've also been too lazy to actually test the patch too.

PR #1855, opened 2022-06-27
Reverts all commits related to generating navmeshes in gamelogic

### LINKED ISSUE #1854: Warn: maximum number of navigation meshes exceeded
Commit ffcd67ab87f7c379 "partial revert of 44b29a6 to fix a pathfinding regression" introduced a regression.
Since this commit, it is no longer possible to use the command `/navedit enable`. When doing so, `Warn: maximum number of navigation meshes exceeded` is emitted and nothing happens.
As this commit says, it is itself a regression fix for #1765 which is a production bug, so it can not be reverted itself (it's better to kill debugging features than production ones).

bmorel:
Fix #1854

The reason to no longer try to debug but instead do a bulk revert are
the following ones:

* every time I find a new bug in that code, I'm loosing several
  hours checking if it does not comes from my own code
* the PR litterally only adds bugs, no features nor improvements of any
  sort
* the PR is even used as a pretext to slow down integration of actual
  improvements in pathfinding
* the PR does not follows any coding style already used in gamelogic
* there is no guarantee the feature actually works as no server
  demonstrated yet AFAIK

Reverted commits:
* ffcd67ab87f7c3797ea57c57be9390482658e170.
  "partial revert of 44b29a6 to fix a pathfinding regression"
* 3d106269c95d46da5239c81333c83c80acec1f9d.
  "Load navmeshes on 'navedit enable'"
* d804c034f546e0c6c3fb469699dce3a3eb7c06c8.
  "Deduplicate navmesh file path building"
* e5c417dc0baa2a08174443463b8a9ea5b5376dcb.
  "Tweak navgen log levels"
* 0c1c8e6be38cfb66f4c8144c11cab0c8bb007d9f.
  "'navgen all' and 'navgen missing' commands"
* 668c65a886f60af38dde6397bc3c6197fd4a45f1.
  "Error handling for navmesh generation"
* 231d1618b37a1003bfd0b80ac6195e9ad61bcad7.
  "Preparation for better cache invalidation"
* a005e7e52c02c455fb6a34c2bd1f847518f5ed59.
  "Hard-code list of classes requiring navmesh"
* 44b29a6c45ce23c6afb0142371411fec761ce249.
  "Init navmeshes only when bot is added"
* b790448e539da2e6c1d735d83859c83b2ce63879.
  "Generate navmeshes during cgame loading screen"
* 89dbd5187bf7829bf2e3bad22c45dba24474377e.
  "Don't include sgame files from navgen code"
* d40e4fc7fdd19834094d367df9eb5e3fbeff88a1.
  "Shuffle definitions among some bot headers"
* 6849e625f967aecf482aecc46be32eb28cad676a.
  "Bots: turn MeshProcess into two classes"
* 8295d59e7a341860002fb27c8f1d3784e26b77dd.
  "Split navmesh generation into small chunks of work"
* fd8577d21942511d452f61824530f50fa7346ddf.
  "Implement 'navgen' command proof of concept"
* b2b61e76ad425ef5895cd912453a90959c8ae418.
  "Get navgen code to compile"
* 96fc3291988171f7049b104a06f3eda408c2d626.
  "Import unmodified navgen code from daemonmap"
* bot code requires MIN_WALK_NORMAL


DolceTriade:
I'll defer to @slipher's opinion here...

But, with this change, we'll have to revert back to using daemonmap to generate navmeshes?

necessarily-equal:
Nightly didn't rely on this ever because the maps had it. The tutorial 
server did tough. But I think my pain of setting up daemon map would be 
lower than the pain @bmorel describes so ultimately I'm ok with that if 
it helps.

bmorel:
> Nightly didn't rely on this ever because the maps had it. 

It is even broken on new parpax iteration because daemonmap is not ran on it, and it hosts an old parpax iteration. Result is, old parpax iteration navmeshes are used for new iteration, and geometry changed a lot, hence bots were completely broken there last time I checked (few weeks ago).

Note that I am *not* against the feature, which even comes accompanied, in 0.53 daemon, by a feature I really want for myself: the ability to overwrite navmeshes from a map the same way we currently do with things like layouts.
Now, I think the navmesh generator's codebase is neither understood nor finished enough to includes it in unvanquished's codebase, and in practice it gives me more problems than help.

Problems it gives me in practice:

* it's harder to hack a 133KLoC codebase than a 12KLoC one,
* 1 production-grade regression (bots go buggy), fixed by a partial revert (that I wrote, yes) which introduces itself a new debug-regression (no longer able to debug navmeshes, there may be other consequences I've not detected yet), I also remember needing to have to add a bot to be able to visualise navmeshes at some point (fixed by slipher, though),
* no way to use different settings on a per-map basis, which is something required to find a correct value to support auto-jumping (see https://github.com/bmorel/daemonmap/commit/7388881efaedf135e02a76cc73035a9cab10500b https://github.com/bmorel/daemonmap/commit/bc46f09374f1058c3461ea1ceda01f7b9310337f https://github.com/bmorel/daemonmap/commit/7f878ae640f743ad38e17393654f0c03f16bbf64 https://github.com/bmorel/daemonmap/commit/d6704d2cda3c3d445d56772eca592e2c66106f0b)
* code sprayed in all the folders: cgame, shared, sgame...

And as illwieckz said, it won't even be used for 0.53's build.

illwieckz:
Anyway I'm still having hard time to get what you mean when you say it doesn't work. Despite all the words all I read is “it doesn't work”, but _what_ is not working?

For what I see, the added (removed) code is about generating navmesh. Are you meaning the navmesh produced are bad? Is there another issue? When you say you find a new bug, do you mean the bug is caused by a bad navmesh, or by something else?

> As for the complexity in running daemonmap... well, here's a working shell script: […]

Such script fails at repackaging every map on mod change.

> it's harder to hack a 133KLoC codebase than a 12KLoC one,

You don't say what is the 133KLoC code base and what is the 12KLoC one. Because I've read the chat for months, I assume you're bringing the talk about the need of a small binary dedicated to producing navmesh without having to run the game, right?

necessarily-equal:
> every time I find a new bug in that code, I'm loosing several hours 
checking if it does not comes from my own code

That's unfortunate. Hopefully it'll be better once 0.53.0 is released 
because you could just use the system package or any server to test? It 
would certainly help checking if what is now 0.53.0/sync has the same 
bug or not.

As a side note, are you actually working on a 0.53.0-based branch? (I 
thought you worked on master now?)

 > > Such script fails at repackaging every map on mod change.
 > How many map updates between 0.52.1 and 0.53? One.

This isn't actually true, since there has been improvements in the 
navmesh generator thanks to you, we have to rebuild *every* *single* 
unvanquished navmesh out there. Or we won't profit from the change. 
Which brings me to:

 > the PR litterally only adds bugs, no features nor improvements of 
any sort

Well the promised feature is that it will allow using any map from 
anywhere without caring about which version the navmesh was compiled 
for, or at all, in case of a tremulous map. Yes it's not here yet, but 
it could for 0.54.0. It's a wonderful feature. Imagine if I had to make 
a testing server that took into account the changes for daemonmap. 
Suddenly I also do have to recompile maps! That sounds needlessly 
painful, and I'd really like *not* to have to care about it.

 > the PR is even used as a pretext to slow down integration of actual 
improvements in pathfinding

Well, if daemonmap is the deprecated way of doing things, yes, a PR to 
daemonmap won't be merged. That doesn't sound too surprising to me. But 
I'm sure you can do a PR against Unvanquished and it'll work. It's 
actually much easier then, because you only need to change one repo to 
add a feature! I think you've expressed your dislike of having to 
change different repos for one feature quite a bit already; and an 
important part of the reason the navmesh improvements were hard to push 
upstream is that they are so hard to test. I think the situation would 
be better with a recent Unvanquished where navmesh from maps dpk would 
be ignored (for example with a build/daemon -set g_alwaysBuildNavmesh 
true).

In fact I think the only reason this is a problem for actual 
integration of changes is that you already patched daemonmap to do the 
thing. And yes, that means you would have to "rebase" your change upon 
Unvanquished to overcome this which is a substantial amount of. This is 
unfortunate indeed.

 > the PR does not follows any coding style already used in gamelogic

This boils down to daemonmap not following the coding style already in 
gamelogic, which is true in all cases. I mean, the code style for the 
navmesh generation in unvanquished is actually closer than what's in 
gamelogic, because slipher did some migrations already (like from 
qboolean to C++ bool). So I'm not convinced by this argument. Yeah ok 
maybe it's less bad if the code styles are different in different 
codebase, but the coding style stays just as consistant as it used to 
be *inside of a single file* in Unvanquished so it doesn't sound too 
bad to me.

 > Unvanquished's gamelogic is 133KLoC. Daemonmap is 12KLoC.

Ok fair enough, but then the thing that generates the navmesh in 
unvanquished is 12KLoC, and the thing that generates the navmesh 
outside of daemon is 12KLoC. So I'm not sure I see the point in this 
argument.

--

I'll actually bring two other arguments you brought on IRC, that I 
think are actually good arguments. Hopefully I'll summurize them 
correctly

 > Unvanquished sgame-native-dll.so is slow to link

This can slow down development a bit when you are working on the 
navmesh generator yes. I don't think this is a significant downside, 
but still worthy of being mentioned.

 > Navmesh generation in Unvanquished isn't multithreaded

This is an actual problem for quick development, when you have more 
complex navmesh like better AI has. And I don't think this could be 
fixed easily and don't think it will be fixed soon.

I don't think this is a problem for running a game though. And 
perturbed said all the blocks are in place for progressive navmesh 
generations, so it would be possible for the server to start a map, and 
clients start joining, and in the meantime the server builds the 
navmesh and adds the bots once finished. This would only be needed on 
the first run of course. So this can alleviate the problem.

 > Sure, with the current settings, in a release process, stuff runs 
fast enough. But to me, this means that I should just stop working at 
improving the current settings, with all their already well known 
problems (I'm sure you know about navmeshes on sky?)
I'm not sure I got your point here, as I don't think navmesh on sky 
have anything to do with this. If it's about ease of development I 
think I just talked about it, and if it's about actually have the 
navmesh improvement visible then see "the promised feature" earlier in 
this message.

--

Anyway, this message is already way too long. So I'll summarise my 
views on the matter:

I fail to see why not having this feature in daemon would bring an 
improvement overall, compared to fixing the bug that were introduced in 
the current implementation and finishing the feature. Actually 
finishing the feature sounds very promising to me.

Now if you actually think it really would be better not to have this 
feature and go back / keep using daemonmap, I'm ok with that. I'll 
trust that you can decide best what is easiest for actual AI 
improvements.

To explain why I can deal with daemonmap: I am building something 
totally overkill to test latest version of things that among other 
rebuilds maps. It wouldn't be much work to add support for ONE version 
of daemonmap. Adding support for *more than one* version of daemonmap 
would be significantly more complex and would be so much easier if I 
can assume the navmesh generation code is in Unvanquished. This is 
important to me because I'd really like to have a testing server for PR 
from betterai one day, and it would be much easier not to need any 
daemonmap (what the code currently assumes — so it would be zero work 
to have a betterai testing server if betterai were to support navmesh 
generation by Unvanquished).

necessarily-equal:
Anyway, what I would suggest is to branch off to keep the navmesh generation in unvanquished in a branch, and revert for the 0.53.0 release.

@slipher does that sound good to you?

necessarily-equal:
+1 to what illwieckz proposed, I actually prefer that to keeping the unvanquished navmesh code it in a branch

bmorel:
>  if we expect the new navmesh code to be perfect when we release it we would wait another year for 0.54.

Well then, close this PR. I'll maintain betterai without it and will be able to redirect users to this discussion when they'll ask me why they can't use vanilla to add connections in maps. Or why are vanilla bots suffering weird bugs.

Even without the navmesh built in gamelogic, 0.53 is big enough to be split in several releases, considering the amount of work it includes.
But yes, it would be good to have a *working* navemesh generator in gamelogic. I don't say perfect, but if it was just not introducing any regression (compared to 0.52 gamelogic + 0.52 daemonmap) it would be ok.
But that's not what is happening.

> If we keep this code in tree but disable it at compilation time with a build variable, would this fix the issues for you?

It would. But this PR's changes go across 21 files.

> Or is this code changes non-navmesh things you want to see reverted?

The only things I want to be reverted are the things which made me loosing days wondering about weird AI behaviors or preventing me to even check navmesh's shape.

> if that makes easier for you to port some improvements you would do in daemonmap in the meantime.

Again, the only "real improvements" I did there were allowing to consider jump magnitude and crouch size, in a config file. Are those real, though... it remains to be seen. They're certainly not too stupid hacks, and certainly works, but (again) that's not the proper way to do things.
The problem on that side is that I consider the current status of navmesh generation to be much harder to hack, and much slower to do iterations. I guess I can understand that people not dealing with AI's code would not care though.

As for qboolean or that kind of horror, I'm certainly not asking for them to be back!

bmorel:
> And I still have lot of changes in my pipeline, too, which might be more spectacular than "basic pathfinding" stuff.

Not to mention I'm not the only developer around, and others will also do work. Also, unvanquished is not only about AIs, AIs are just here to bootstrap games in the eyes of many (to me, they're also about training new players, though).

slipher:
I'm confused about which path forward is being suggested here. @necessarily-equal suggests "temporarily" deleting the code, while @bmorel's arguments against the code imply that it should be permanently deleted (except #1854, but that could hardly take more than a half-hour to fix).

@illwieckz the main issue with production-readiness is that you can't generate the navmeshes and then use them in the same playing session. The lazy-loading code which would have allowed this mysteriously didn't work and had to be reverted.

illwieckz:
@bmorel I remind you that daemonmap repository always said it had no future… Any revert, disablement or deletion there should just be temporary. I'm OK to do things that makes life easy for you until this code is ready, but that's _until this code is ready_. Navmeshes are entirely dependent from game parameters, an external tool is the worst solution regarding to the need of having navmeshes match the actual game code.

I clearly understand that it is really faster to rebuild daemonmap and I would be happy if we can produce a stripped down version of the game to quickly produce a navmesh like we can quickly do it with daemonmap… but really… an external tool can't have a future. A navmesh is an _internal_ representation that is specific to the _current_ game. We're not targeting cartridge-based distribution and lifecycle!

It's very OK to me to consider the current code is “not ready” and I don't mind disabling it, but having an in-game navmesh generator really is the _target_.

@slipher: outside of lazy loading and specific things like that, but about the navmesh generation itself, do you know if it's possible to make easy to share some code from daemonmap in a way @bmorel can prototype on daemonmap then later we just copy-paste his code in game code? How much the code had to be ported to work there?

illwieckz:
For example, what if we modify `nav.cpp` in a way it can be copy/pasted straight from one source tree to another? Even if for now parts of it would be a bit ugly to keep that possible, I don't care.

slipher:
> outside of lazy loading and specific things like that, but about the navmesh generation itself, do you know if it's possible to make easy to share some code from daemonmap in a way @bmorel can prototype on daemonmap then later we just copy-paste his code in game code?

I doubt that this would be a fruitful idea. You can look at commit b2b61e76 for a lower bound on the differences.

But any parts of generating navmeshes from gamelogic that are lacking as compared to daemonmap can be improved without much difficulty:
- Single-threaded: we could add multi-threading to the gamelogic navgen (compiling only in dll or exe modes if necessary)
- Configs: it would make sense to have a file read from the VFS and/or `game/` directory to customize the config settings per map.
- Ease of use from command line without starting the game: this could be addressed by a small shell script to do something like `daemonded -set vm.sgame.type 3 -set common.shutdownOnDrop 1 +devmap "$1" +delay 20f navgen "$*2" +delay 21f quit`

bmorel:
> while @bmorel's arguments against the code imply that it should be permanently deleted

No, my point is that it should be removed until it actually work without regressions.

> @bmorel I remind you that daemonmap repository always said it had no future

But it *currently* works.

> but that's until this code is ready

I'm not asking for anything else. Do you really think I wuold have bothered to do the work to remove code and trigger this discussion if the code was not actually causing regressions?
It would even brings features I actually want, it if worked: I would not need to create dpk files for navmeshes because I would be able to host them without that! And this would clearly save me lot of time, too.

I read lot of words about "things can easily be improved". Sure. They can, it only needs someone with all of: time, knowledge, motivation to actually do it.
Most of the AI changes I implemented were rather simple, too.
Currently, this code brings regressions. The commit I did on that which allows to have bots not being broken in games itself brings another regression which prevents one to view navmesh and create offMesh connections (navcon) which is really a big problem, especially since, as @illwieckz  pointed out, releases take more than a year to be done.

This means, a known regression with a known cause which is a feature which currently brings no benefit will be preventing any people interested at improving AI's pathfinding without tech knowledge to contribute for more than a year.
This is what removing the "feature" for the upcoming 0.53 release would prevent.

bmorel:
Better to merge #1858 I guess.

PR #1858, opened 2022-07-02
Fix naveditcommand

### LINKED ISSUE #1854: Warn: maximum number of navigation meshes exceeded

slipher:
Fixes #1854.

### DIFF
--- a/src/sgame/botlib/bot_nav_edit.cpp
+++ b/src/sgame/botlib/bot_nav_edit.cpp
@@ -191,7 +191,6 @@ static bool CheckHost( gentity_t *ent )
 	return false;
 }
 
-bool G_BotNavInit();
 void Cmd_NavEdit( gentity_t *ent )
 {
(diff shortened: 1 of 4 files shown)


bmorel:
This passes the 2 situations I have 100% reproducible bugs on. Still I now think this needs more testing, because debugging bots is painful enough as it is.

bmorel:
tbh, I think I'll be too lazy to test it more. LGTM. (I'll just maintain #1855 in my own work, will solve the problems on my end)

PR #1876, opened 2022-07-07
Add warning when a non-existing progressbar is asked for in the UI files

necessarily-equal:


src/cgame/cg_rocket_progressbar.cpp
```diff
@@ -205,6 +205,11 @@ float CG_Rocket_ProgressBarValue( Str::StringRef name )
 	// Get the progressBar command
 	cmd = ( progressBarCmd_t * ) bsearch( name.c_str(), progressBarCmdList, progressBarCmdListCount, sizeof( progressBarCmd_t ), progressBarCmdCmp );
 
+	if ( !cmd )
+	{
+		Log::Warn("progress bar command %s not found", name.c_str());
```
slipher, line 210: no need for `c_str`

slipher (commented):
As a UI change this should go on 0.53. Otherwise LGTM.

necessarily-equal:
Merging to master as it is compatible with both 0.53.0 and master

PR #1877, opened 2022-07-07
Fix bot traces

necessarily-equal:
There was an exception that allowed them to see through doors, and they often did not see through buildings and other players


slipher:
LGTM

necessarily-equal:
Following a point made by @bmorel, and trying really hard, I could get a bot to shoot at me by putting myself between him and an alien, and forcing him in place. I didn't get much damage this way, and this is not a regression. So merging.

Then a bot took a painsaw. Là j'étais scié (pun)

Maybe we want to do something for painsaw especially because they DO some team damage. But we may want to keep the nice effect that there is when they start firing on sight :)

Issue #29, opened 2012-03-05
cg_shadows makes objects invisible


KayFTW:
OS: Windows Vista 32 bit
Renderer: GL3

On the new renderer (GL3), cg_shadows 2 makes the renderer spazz and things flicker all over the place, making the game unplayable. on cg_shadows 3, all the walls, floors, and ceilings of maps become invisible and see-through. On 4, all the alien player models are invisible, on 6, only the Dragoon, Basilisk, and Marauder models are invisible. 

cg_shadows 3: http://i42.tinypic.com/514hs1.jpg
Replicate: 
`cg_shadows 3`

`````` vid_restart```

I don't have a screenshot of cg_shadows 2 messing up the renderer, how to reproduce:
```cg_shadows 2```
```vid_restart```
I also don't have a screenshot of what cg_shadows 4, and 6 do, however just replicate, obviously, but changing them to 4 or 6. 
``````

PR #1880, opened 2022-07-10
reward grangers for removing fire

bmorel:
Fixes [#29](https://github.com/Unvanquished/gameplay/issues/29) "[Idea] Reward grangers for extinguishing fire"

To be honest, I think the value is too high, but this will make the option configurable. I'd be in favor to have a default setting of `0.f` instead, so that 0.53 server owners can experiment with it, and a value could be agreed on for 0.54.


src/sgame/sg_missile.cpp
```diff
@@ -40,6 +40,10 @@ static Cvar::Cvar<int> g_abuild_blobFireExtinguishRange(
 		"g_abuild_blobFireExtinguishRange",
 		"range around which a granger spit extinguish fires on the ground",
 		Cvar::NONE, 64);
+static Cvar::Cvar<float> g_firefightReward(
+		"g_firefightReward",
+		"how much credits removing fire gives a granger",
```
slipher, line 45: Probably you want this in evolve points rather than human credits? (You'll need to do a unit conversion because G_AddCreditToClient uses human credit units.)

slipher (commented):
As written, the code gives credits for every nearby entity to the impact location, regardless of whether it was on fire. Also it does not give credit if the on-fire entity was directly impacted by the blob.

src/sgame/sg_missile.cpp
bmorel: To be honest, I'm confused by the idea of having 2 representations. Imo this "feature" should be dropped, especially since behind the scene it's the same crap.
Anyway, you're right.

bmorel:
> As written, the code gives credits for every nearby entity to the impact location, regardless of whether it was on fire. Also it does not give credit if the on-fire entity was directly impacted by the blob.

Yeah, I naively thought it would be an easy thing to do. I'll try to fix that soon, or will just drop the effort,since I'm not convinced by the idea when I think more about it.
Advanced granger is already a very powerful thing (can freeze 2 enemies alone) for no cost, and clearly breaks balance. To make a gratis form able to make easy money... it might be a good answer to the fact aliens requires a lot more money to deal serious damages to human bases, but I'm not convinced it's the good answer to the problem.

src/sgame/sg_missile.cpp
necessarily-equal: No, gameplay wise it's good because it feels like aliens are different. Bringing back… capitalistic considerations would be break immersion on alien side.

src/sgame/sg_missile.cpp
```diff
@@ -260,11 +269,18 @@ static int ImpactSlowblob( gentity_t *ent, trace_t *trace, gentity_t *hitEnt )
...
+		if ( neighbor != hitEnt && G_IsOnFire( neighbor ) )
 		{
+			reward = true;
 			neighbor->entity->Extinguish( g_abuild_blobFireImmunityDuration.Get() );
 		}
 	}
+	
+	if ( reward )
+	{
+		float credits = g_firefightReward.Get() * CREDITS_PER_EVO;
```
necessarily-equal, line 281: Reward is yes or no here. Since a single splat can extinguish several fires, it would be good to have some counting

```suggestion
		float credits = g_firefightReward.Get() * extinguished * CREDITS_PER_EVO;
```

src/sgame/sg_missile.cpp
```diff
@@ -260,11 +269,18 @@ static int ImpactSlowblob( gentity_t *ent, trace_t *trace, gentity_t *hitEnt )
 	{
 		// extinguish other entity on fire nearby,
 		// and fires on ground
-		if ( neighbor != hitEnt )
+		if ( neighbor != hitEnt && G_IsOnFire( neighbor ) )
 		{
+			reward = true;
```
necessarily-equal, line 274:
```suggestion
			extinguished++;
```

src/sgame/sg_missile.cpp
```diff
@@ -249,9 +253,14 @@ static int ImpactSlowblob( gentity_t *ent, trace_t *trace, gentity_t *hitEnt )
 	int       impactFlags = MIB_IMPACT;
 	gentity_t *attacker = &g_entities[ ent->r.ownerNum ];
 	vec3_t dir;
+	bool reward = false;
```
necessarily-equal, line 256:
```suggestion
	int extinguished = 0;
```

src/sgame/sg_missile.cpp
bmorel: Yes, reward is yes or no. This is intended. Also, since you're nitpicking, at least use `unsigned` instead of `int`.

> Since a single splat can extinguish several fires, it would be good to have some counting

I don't think we want to make fire removal that OP credit-wise.

slipher: I don't think it's a good gameplay mechanic that if there are 2 nearby fires, it's more profitable to extinguish them with 2 separate shots rather than a single shot. That is, I agree with afontain.

bmorel: Oh, this is what you meant. Ok. It makes sense.

src/sgame/sg_missile.cpp
bmorel: not really. They're the same, it's only representation which changes. There's still that dumb conversion done in various places for basically nothing, but don't care, it's not the topic here.

bmorel:
I believe all issues were fixed

slipher:
LGTM

PR #1881, opened 2022-07-11
bg_misc: Fix off-by-one in BG_WeaponByName.

DolceTriade:
We don't include WP_NONE in bg_weapons, so subtract one. Also see
BG_Weapon() where we do the same.

### DIFF
--- a/src/shared/bg_misc.cpp
+++ b/src/shared/bg_misc.cpp
@@ -610,7 +610,7 @@ const weaponAttributes_t *BG_WeaponByName( const char *name )
 
 	if ( weapon )
 	{
-		return &bg_weapons[ weapon ];
+		return &bg_weapons[ weapon - 1 ];
 	}
 	else
 	{


slipher:
It's not a bug. There's a 0 entry in all these arrays so that you just get a zeroed-out struct if you put in an uninitialized enum instead of crashing.

DolceTriade:
Sorry if the commit message wasn't clear. The bug is that `weapon_t` starts at `WP_NONE = 0`  whereas `weaponAttributes_t bg_weapons[ ARRAY_LEN( bg_weaponsData ) ]` starts at `bg_weapons [ 0 ] = WP_LEVEL0` (https://github.com/Unvanquished/Unvanquished/blob/b125b62596a6774f1441357670d550c18ac0cba0/src/shared/bg_misc.cpp#L562). `WP_LEVEL0 = (weapon_t)1`. That's why we need to subtract one. This is definitely a bug.

bmorel:
There is clearly a bug here, and this patch clearly fixes it at this place, without further refactoring.

We might want to introduce such a zeroed-out first element in arrays, but this means checking consistency across whole codebase, and doing the same for the other arrays which seems to not do that neither. This would save a few branchings and maybe make code easier to maintain, but I'm not sure it's worth the troubles.
The nullWeapon pointer is returned if weapon == 0 btw.

DolceTriade:
I use it in my Lua branch.

bmorel:
nice!

bmorel:
Would be nice if you could add a way to reproduce the bug, I'd be curious to see it. Also, it might be useful to add a "S-Reproducible" tag, for the day someone (maybe @necessarily-equal as I know he talks a lot about it ) do something to get some automated testing.

DolceTriade:
Any call to the function will result in the wrong `weapon_t` being returned.

ie, `BG_WeaponByName("level0")` would return `WP_ALEVEL1`. I didn't test this but I wonder if `BG_WeaponByName("ckit")` causes a crash... 🤔

bmorel:
good question. Do you think you can provide a minimal patch to trigger a crash? That was really my question

DolceTriade:
It doesn't crash thanksfully...

Added this line to sg_main
```
+    Log::Notice("WEAPON %s", BG_WeaponByName("ckit")->name);
```

Then started a map via devmap to see:
```
WEAPON <nullptr>
```

With my fix you see:
```
WEAPON ckit
```

slipher:
Yeah I patched tinyformat to not crash if you print a null string in release mode :laughing:

Issue #1620, opened 2021-12-05
bot action moveTo doesnt work correct


Gireen:
moveTo has a second argument which is supposed to be the radius around the goal to reach.
This has no effect at all. 
```
selector
{
        action moveTo( E_A_OVERMIND, 300 )
}
```
Bots will walk constantly against the goal and ignore the radius.
That should also not happen if no radius is given.

bmorel:
MoveTo and other Move-like actions should be allowed to fail, imo this would be the fix of many problems.

bmorel:
I believe this is fixed. @Gireen could you verify this?

bmorel:
moved to 0.54 anew because it shuold not block 0.53 release, as I'm pretty sure it's now fixed.

necessarily-equal:
It looks fixed in my tests.

PR #1882, opened 2022-07-12
0.52/navigational fixes

bmorel:
This PR prepares the ground for bigger navigational changes which will happen between 0.53 and 0.54, namely:

* fixing the obstacle management (there's several bugs with it, like #1493)
* replacement of complex actions (flee, heal, repair, fight, buy...) by several simpler ones ( most of those can be implemented as `sequence { changeGoal FOOBAR; moveToGoal }` for most of the stuff )
* multiple-goal support, which will start by allowing bots to move in a different direction than the direction they aim at, without moving outside of their navmesh

The reason I do this now is to reduce the risk of code conflicts in the upcoming months. Those changes are also likely to fix currently open issues such as #1620 , #1577 .

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -636,8 +636,11 @@ AINodeStatus_t BotActionAimAtGoal( gentity_t *self, AIGenericNode_t* )
 
 AINodeStatus_t BotActionMoveToGoal( gentity_t *self, AIGenericNode_t* )
 {
-	BotMoveToGoal( self );
-	return STATUS_RUNNING;
+	if ( GoalInRange( self, BotGetGoalRadius( self ) ) )
+	{
+		return STATUS_SUCCESS;
+	}
+	return BotMoveToGoal( self ) ? STATUS_RUNNING : STATUS_FAILURE;
 }
 
 AINodeStatus_t BotActionMoveInDir( gentity_t *self, AIGenericNode_t *node )
@@ -918,13 +921,11 @@ AINodeStatus_t BotActionRoamInRadius( gentity_t *self, AIGenericNode_t *node )
 		self->botMind->currentNode = node;
 	}
 
-	if ( self->botMind->nav.directPathToGoal && GoalInRange( self, 70 ) )
+	if ( GoalInRange( self, BotGetGoalRadius( self ) ) )
 	{
 		return STATUS_SUCCESS;
 	}
-	BotMoveToGoal( self );
-
-	return STATUS_RUNNING;
+	return BotMoveToGoal( self ) ? STATUS_RUNNING : STATUS_FAILURE;
 }
 
 AINodeStatus_t BotActionRoam( gentity_t *self, AIGenericNode_t *node )
@@ -940,15 +941,11 @@ AINodeStatus_t BotActionRoam( gentity_t *self, AIGenericNode_t *node )
 		self->botMind->currentNode = node;
 	}
 
-	if ( self->botMind->nav.directPathToGoal && GoalInRange( self, 70 ) )
+	if ( GoalInRange( self, BotGetGoalRadius( self ) ) )
 	{
 		return STATUS_SUCCESS;
 	}
-	else
-	{
-		BotMoveToGoal( self );
-	}
-	return STATUS_RUNNING;
+	return BotMoveToGoal( self ) ? STATUS_RUNNING : STATUS_FAILURE;
 }
 
 botTarget_t BotGetMoveToTarget( gentity_t *self, AIEntity_t e )
@@ -988,19 +985,16 @@ AINodeStatus_t BotActionMoveTo( gentity_t *self, AIGenericNode_t *node )
 		return STATUS_FAILURE;
 	}
 
-	BotMoveToGoal( self );
-
 	if ( radius == 0 )
 	{
 		radius = BotGetGoalRadius( self );
 	}
 
-	if ( DistanceToGoal2DSquared( self ) <= Square( radius ) && self->botMind->nav.directPathToGoal )
+	if ( GoalInRange( self, radius ) )
 	{
 		return STATUS_SUCCESS;
 	}
-
-	return STATUS_RUNNING;
+	return BotMoveToGoal( self ) ? STATUS_RUNNING : STATUS_FAILURE;
 }
 
 AINodeStatus_t BotActionRush( gentity_t *self, AIGenericNode_t *node )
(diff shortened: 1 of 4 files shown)


src/sgame/sg_bot_nav.cpp
```diff
@@ -613,7 +613,7 @@ Global Bot Navigation
 =========================
 */
 
-void BotMoveToGoal( gentity_t *self )
+bool BotMoveToGoal( gentity_t *self )
```
DolceTriade, line 616: While we are here, can we start documenting some of these functions. Especially now that you are adding a return value? It's not clear to me what return true and return false mean.

bmorel: Yes, all the navigation stuff is heavily *UN*documented. As for everything sadly. I'll document this when I'll have fixed the mess.

bmorel:
This now passes the most easily reproducible problems:

* tyrants going through the Slope of Hell in perseus
* humans going close to another inverted slope in niveus without trying to handle collisions
* humans going through niveus' huge staircase

This still need some more testing though, considering how critical this code is.

DolceTriade (approved):

bmorel:
I'd prefer to have at least another review before merging, considering the fact I consider AI changes quite seriously, especially since 0.53 is very soon and an AI regression would be a really bad thing for the growing number of players. (it's not against you @DolceTriade , just a precaution)

src/sgame/sg_bot_ai.cpp
```diff
@@ -1023,11 +1017,11 @@ AINodeStatus_t BotActionRush( gentity_t *self, AIGenericNode_t *node )
 		return STATUS_FAILURE;
 	}
 
-	if ( !GoalInRange( self, 100 ) )
```
slipher, line 1026: I think BotGetGoalRadius is about the size of the goal entity, while GoalInRange takes any distance away from the goal you want to be. So using BotGetGoalRadius is the desired range basically means you want to be touching the goal entity. This would be bad if you are a human and your goal is an acid tube - better to shoot from a safe distance than to go touch it.

src/sgame/sg_bot_nav.cpp
```diff
@@ -617,13 +615,6 @@ void BotMoveToGoal( gentity_t *self )
 {
 	vec3_t dir;
 	VectorCopy( self->botMind->nav.dir, dir );
-	const playerState_t& ps  = self->client->ps;
-
-	if ( dir[ 2 ] < 0 )
```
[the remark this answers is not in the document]
necessarily-equal: I think this is because climbing would be accounted in the movement while falling isn't, so the distance accounting is slightly different, and thus one might be faster on a downward slope.

I'm not sure why the thing with the negative z axis thing was removed, but to be honest I'm also unsure what it was used for. At least on small slopes like plat23 the speed appears to be the same upwards and downwards.

Since I don't think we have high slopes, and since on stairs the situation is the same as on flat ground, I am ok with this removal.

necessarily-equal (approved):
There is still the acid tube remark, but I think this isn't something that can currently trigger.

LGTM

src/sgame/sg_bot_nav.cpp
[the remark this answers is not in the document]
bmorel: The reason I've removed it is because it was done before in the code IIRC. This "planarisation" is done like 3 times in a row, I just remove one instance here. It won't have huge perf impact, but everything is good to take, as long as it does not introduces bugs.
I'm not sure I've explicitly seen the "<0" in the bunch of changes though. One thing is certain, betterai mod, which have been running a custom version of this code since many months (you can refer to my repo for more details) have some navigation bugs, but they are not tied to this. In practice, it's the 4th time I do massive changes to this, and *this time* I am finally confident about them (1st time was ladder PoC, 2nd time was an attempt to implement some sort of bisect traces, but daemon does not allow it, 3rd time was about doing my best without that idea, 4th time in local repo only, for now, not even deployed)

As for slopes... there *are* some Slopes of Hell that defeat AIs. The most obvious is explained in #1493, but niveus map also have such, in a less dramatically buggy situation (near human base, between the 2 big rooms). Those are the slopes I always use for testing.

src/sgame/sg_bot_ai.cpp
bmorel: Except that "action rush" is not about fighting at all. When a bot sees a target, "action fight" is triggered instead (because alertedToEnemy).
Without this change, it means that if action rush takes a building which is bigger than 100 (currently does not exist, ofc)  with "don't touch building" cvar bots will be stuck trying to move to center.
This change makes things more robust, generally speaking.

slipher: > When a bot sees a target, "action fight" is triggered instead (because alertedToEnemy).

Well, in principle it's possible to write the BT differently.

How about like `if ( !GoalInRange( self, 30 + BotGetGoalRadius(target) ) )` ?

src/sgame/sg_bot_nav.cpp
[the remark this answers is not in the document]
slipher: I see, the 2 code paths that set `dir` both set the z-component to 0. This would be a good fact to include in the comment.

src/sgame/sg_bot_ai.cpp
bmorel: As for the problem of acid tube, or spiker, or turrets even... action rush leads bot to either the unique or a spawn. Never to a defensive structure. When the bot becomes alerted to enemy, then action fight defines the new goal, which is in practice a major issue (prevents fleeing while fighting, notably), that I want to fix later.

What *can* be done, and I have done it in my mod, is to use a `concurrent` node when fighting, so that when fighting, it target is  a building, then move backward as long as target is in range.
Of course, this "solution" create new problems, because the range information of weapons is completely stupid. Also, this only works because of my memory rewrite, as otherwise fighting something hidden by a door would make said thing no longer a target, as the target is removed from the stack.
Still, this is a better solution that the current situation where bots just kiss acid tubes if they happen to have be close enough when they decided to target them.

slipher: OK, in any case it's an improvement over the status quo which just returns STATUS_RUNNING forever.

Since BotMoveToGoal already checks `GoalInRange( self, BotGetGoalRadius( self ) )`, you can delete this if statement - it's duplicated code.

bmorel: Said Ok on IRC.

Issue #1824, opened 2022-06-03
The pain indication broke on 0.53.0/sync


necessarily-equal:
This "green blood" effect on the border for aliens, and "regular blood" for humans are not visible anymore.

It would be nice to fix this before the next release.

necessarily-equal:
This is apparently handled by `CG_PainBlend` in `src/cgame/cg_draw.cpp`.

PR #1886, opened 2022-07-14
Fix pain indication

necessarily-equal:
Integer rounding to zero strike again!

Fixes https://github.com/Unvanquished/Unvanquished/issues/1824


bmorel:
LGTM

Issue #608, opened 2015-02-16
Buy tyrant with 9.9 evos


lamefun:
Sometimes I can buy a tyrant with 9.9 evos, etc. Looks like either counting problem or display problem (ie. I have 10 evos but it still shows 9.9).

DolceTriade:
Yeah, rounding error most likely.

illwieckz:
And sometime we can't earn more than 19.9 evo.

Viech:
Didn't check, *could* be an UI error, too.

DolceTriade:
Yeah, it's just a problem with how it is displayed. Internally, everything is stored correctly.

overvolting:
Also, sometimes I can't get above 19.9 evos stored.

PR #1888, opened 2022-07-14
Fix the second evo rounding bug

necessarily-equal:
Because while 7.99 shouldn't be 8 ([first bug](https://github.com/Unvanquished/Unvanquished/pull/1248/commits/dcf5b0a566403c6102be35a8603ee438396ac9c8)),  20.0 shoudn't be 19.9 either.

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -889,20 +889,26 @@ public:
 	void DoOnUpdate()
 	{
 		playerState_t *ps = &cg.snap->ps;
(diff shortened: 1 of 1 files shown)


necessarily-equal:
This also fixes https://github.com/Unvanquished/Unvanquished/issues/608 for good, since I likely introduced it back in https://github.com/Unvanquished/Unvanquished/pull/1248/commits/dcf5b0a566403c6102be35a8603ee438396ac9c8

slipher:
LGTM

bmorel:
lgtm

PR #1890, opened 2022-07-14
Cherry-pick from #1755

necessarily-equal:
This is some commits from https://github.com/Unvanquished/Unvanquished/pull/1755 that I reviewed and that haven't got a change request.

I've reviewed them and I assume that if bmorel wrotes them, that counts as someone writting it and someone else approving it? In which case I could merge tomorrow if nobody says it's a bad idea?

### DIFF
--- a/src/shared/bg_local.h
+++ b/src/shared/bg_local.h
@@ -52,8 +52,6 @@ extern  pmove_t *pm;
 #define pm_waterfriction     (1.125f)
 #define pm_spectatorfriction (5.0f)
 
-extern  int     c_pmove;
-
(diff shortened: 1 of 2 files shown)


illwieckz:
Just make sure it doesn't break some @bmorel stuff if for example he already rewrote that in another way or something like that. But yes that count as a review.

slipher:
LGTM

bmorel:
well, this will give me conflicts, as anything touching more or less vec3_t stuff, but it should be ok. I'll just have to revert it on my local stuff if I'm in a hurry to rebase, this way things will apply without problem.

PR #1913, opened 2022-07-18
Introduce fallback BT nodes

necessarily-equal:
The name is taken from BT.cpp, https://docs.ros.org/en/kinetic/api/behaviortree_cpp_v3/html/classBT_1_1FallbackNode.html

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -271,6 +271,37 @@ AINodeStatus_t BotSelectorNode( gentity_t *self, AIGenericNode_t *node )
 	return STATUS_FAILURE;
 }
 
+AINodeStatus_t BotFallbackNode( gentity_t *self, AIGenericNode_t *node )
+{
+	AINodeList_t *sequence = ( AINodeList_t * ) node;
+	int start = 0;
+
+	// find a previously running node and start there
+	for ( int i = sequence->numNodes - 1; i > 0; i-- )
+	{
+		if ( NodeIsRunning( self, sequence->list[ i ] ) )
+		{
+			start = i;
+			break;
+		}
+	}
+
+	for ( int i = start; i < sequence->numNodes; i++ )
+	{
+		AINodeStatus_t status = BotEvaluateNode( self, sequence->list[ i ] );
+		if ( status == STATUS_SUCCESS )
+		{
+			return STATUS_SUCCESS;
+		}
+
+		if ( status == STATUS_RUNNING )
+		{
+			return STATUS_RUNNING;
+		}
+	}
+	return STATUS_FAILURE;
+}
+
 AINodeStatus_t BotSequenceNode( gentity_t *self, AIGenericNode_t *node )
 {
 	AINodeList_t *sequence = ( AINodeList_t * ) node;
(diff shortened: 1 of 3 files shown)


bmorel:
I'm not sure I see the point of this, but one thing that is lacking there a lot is: documentation.

necessarily-equal:
Well if behavior trees are supposed to be documented in the wiki, then I can write something in the wiki afterwards. Unless we decide on having the documentation in code.

I'll rebase the branch

DolceTriade:
I think at least providing an example in the commit message is a reasonable thing to do.

DolceTriade:
Also. I'm pretty unfamiliar with the BT code, so sorry if my questions are a little naive:

 - How does this work if we have nested fallback nodes?
 - In your example above, if I didn't have it as a "fallback" node but as a "sequence" node, would the expected behavior then be first try to fight, then try to get to safety, and lastly, fight regardless of what happened before?
 - How are statuses propagated in the behavior tree?

slipher:
LGTM

> How does this work if we have nested fallback nodes?

There is a stack of running nodes so it can remember which ones are running at each level.

bmorel:
> Well if behavior trees are supposed to be documented in the wiki, then I can write something in the wiki afterwards. Unless we decide on having the documentation in code.

Considering the "quality" of BT in wiki, I think this is a rather ridiculous excuse. Even if wiki's doc on BT was of high quality and not out of sync,  it's annoying to have to go reach wiki when someone is trying to make sense of a code snippet. It's actually not always possible.

Btw, I have a great idea. Let's move to SVN. It's less annoying with the conflicts than git, and have pretty nice GUI interfaces. OFC, it requires constant internet connection, but it's not a problem, right?

necessarily-equal:
@DolceTriade:
> In your example above, if I didn't have it as a "fallback" node but as a "sequence" node, would the expected behavior then be first try to fight, then try to get to safety, and lastly, fight regardless of what happened before?

The difference with `sequence` is that this one switches to the next one if the first action *fails*. `sequence` goes to the next action when the current action succeeds and stops when one action fails; `fallback` goes to the next action when the current action fails and stops when one action succeeded.

The difference with `selector` is that the selector falls tries the first one of the list each time to select the first one which doesn't fail, while `fallback` continues where it was on last `ACTION_RUNNING`.

@bmorel complaining about the wiki:
Well you are the one who created the bot documentation on the wiki. If you don't think it's where it should live, fine for me, but it really wasn't my idea in the first place.

So, where should I document it then? Some talked about some other form of wiki, like a markdown wiki checked out in git so that we can `git clone` it, but it's annoying. Some talked about keeping the doc in the code, which is good but also you would be the first to complain it's "coder-first". I feel figuring this out isn't a requirement for this PR.

bmorel:
> @bmorel complaining about the wiki:
Well you are the one who created the bot documentation on the wiki. If you don't think it's where it should live, fine for me, but it really wasn't my idea in the first place.

It is because I tried to use wiki for the documentation that I think it's not suited for this. But, most people are fond of that solution, so, fine.
In any case, I'm going to implement my own parser and behavior tree mechanism, and very likely to redo the whole botMemory_t soonish, so I guess I should not care about missing documentation in upstream code.

PR #1914, opened 2022-07-18
Define ET_PLAYER for BT

necessarily-equal:


necessarily-equal:
Usable like this: `condition goalType == ET_PLAYER`

bmorel:
LGTM

PR #1916, opened 2022-07-18
Add a cvar to display distances

necessarily-equal:
It's a bit hacky / not too practical, but at least it gives a way to estimate distances in-game.

### DIFF
--- a/src/cgame/cg_ents.cpp
+++ b/src/cgame/cg_ents.cpp
@@ -28,6 +28,12 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 #include "cg_local.h"
 
+static Cvar::Range<Cvar::Cvar<float>>
+	cg_drawDebugDistance("cg_drawDebugDistance", "draw a circle around "
+			"yourself to give an idea of distances. Set this "
+			"to 0 to disable", Cvar::CHEAT, 0.0f, 0.0f,
+			5000.0f);
+
 /*
 ======================
 CG_DrawBoxFace
@@ -1372,6 +1378,12 @@ void CG_AddPacketEntities()
 					break;
 			}
 		}
+	}
+
+	if ( cg_drawDebugDistance.Get() > 0.0f )
+	{
+		CG_DrawSphere( cg.snap->ps.origin, cg_drawDebugDistance.Get(),
+		               cgs.media.plainColorShader, Color::Color(0, 0.5, 1, 0.2) );
 	}
 }
 


src/cgame/cg_ents.cpp
```diff
@@ -1373,6 +1373,18 @@ void CG_AddPacketEntities()
 			}
 		}
 	}
+
+	static Cvar::Range<Cvar::Cvar<float>>
```
bmorel, line 1377: Usually, CVars are extern, not static. They're also defined and declared outside of functions.
Any reason to change that convention?

src/cgame/cg_ents.cpp
```diff
@@ -1373,6 +1373,18 @@ void CG_AddPacketEntities()
...
+		cg_drawDebugDistance("cg_drawDebugDistance", "draw a circle around "
+				"yourself to give an idea of distances. Set this "
+				"to 0 to disable", Cvar::CHEAT, 0.0f, 0.0f,
+				5000.0f);
+	//make an attempt at drawing bounding boxes of selected entity types
```
bmorel, line 1382: I don't understand. Is this about drawing a circle to estimate distance, or is it about BBoxes?

src/cgame/cg_ents.cpp
necessarily-equal: Only real reason is build time

src/cgame/cg_ents.cpp
necessarily-equal: Whoops, comment sneaked in. Removing

src/cgame/cg_ents.cpp
bmorel: Yes it is uncommon in gamelogic. Only 3 Cvars from thse are not declared at top of file. Duplicated code is also common, yet it's considered a bad thing.

You need the '-n' parameter to grep to know where the cvars are put in the file, -c is only count occurrences.

necessarily-equal: Just moved it at the beginning of the file.

bmorel:
LGTM

bmorel:
The result is very weird.
This actually does not draws a circle, but a sphere.

2022-07-21-114603_1366x768_scrot

It works, but I'm not sure about this PR still.

necessarily-equal:
Yes it does look a bit weird and isn't too practical. But tbh everything works with spheres in-game (when it's not about bounding boxes), and it's still the only way I know to display distances.

bmorel:
No, not everything works with spheres, and its possible to display circles. Look at /navedit command. Also, you didn't fixed the wording.

slipher:
An alternative is to draw a trace from the player's eye and display the distance to the entity the trace hits, similar to `cg_drawEntityInfo` (or maybe even part of it). Of course this isn't perfect as you can't estimate the distance to something which isn't solid. The sphere thing looks even worse though :laughing:

PR #1917, opened 2022-07-18
Don't protect immortals against damage when suiciding

necessarily-equal:
If something really wants to die, it should be allowed to do it even if it usually doesn't take damage

This is used in the juggernaut mod, which also defines `g_indestructibleBuildables` (see https://github.com/Unvanquished/Unvanquished/pull/1915)


slipher:
LGTM

PR #1931, opened 2022-07-20
bots should not spawn at warmup

### LINKED ISSUE #1236: Bots need a delay before spawning in
Currently, bots spawn in and start playing as soon as the map finishes loading which is significantly faster than player load times. This is detrimental due to bots dying/killing and even ending up inside bases before any player has had a chance to spawn.

bmorel:
Note that this does not prevents them to "select" a team.
Fix #1236

Will impact changelog when merged. We might also change default configuration to have the warmup back by default.

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -503,11 +503,18 @@ void G_BotSpectatorThink( gentity_t *self )
 	self->botMind->futureAimTimeInterval = 0;
 	self->botMind->runningNodes.clear();
 
+	//FIXME: duplicate of sg_cmds.cpp:883 function "void Cmd_Team_f( gentity_t * )"
+	if ( g_doWarmup.Get() && ( ( level.warmupTime - level.time ) / 1000 ) > 0 )
+	{
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


necessarily-equal (approved):

src/sgame/sg_bot.cpp
```diff
@@ -503,11 +503,18 @@ void G_BotSpectatorThink( gentity_t *self )
 	self->botMind->futureAimTimeInterval = 0;
 	self->botMind->runningNodes.clear();
 
+	//FIXME: duplicate of sg_cmds.cpp:883 function "void Cmd_Team_f( gentity_t * )"
+	if ( g_doWarmup.Get() && ( ( level.warmupTime - level.time ) / 1000 ) > 0 )
```
slipher, line 507: This could go a bit earlier, after the PMF_QUEUED if block. Otherwise LGTM

bmorel: By precaution, I'd rather keep the cleanup stage ran each botSpectatorThink. The stuff really feels hackish to me, I won't lie, but I do not understand enough yet the process of spawning bots.

PR #1934, opened 2022-07-20
add Cvar g_bot_ignore_momentum to allow bots to cheat even more

### LINKED ISSUE #1801: Allow bots to ignore momentum when g_bot_infinite_funds is on
Currently, making bots cheat only give them infinite credits. They still can't buy stuff which is still locked.
Doing so would make it possible to let them buy stuff even if locked after some time, in dedicated BTs, which I'm planning to use to create "mob BT" for PvE layouts.

One question is, should it be possible to enable only one kind of cheat mode? That is, either the infinite funds, or the infinite momentum, or both?

Also, it might be interesting to only allow some bots to cheat, for example by making it a BT action.

See #1802.

bmorel:
fix #1801

This is worth a line in a changelog, as it increases control over bots. Also, requires a GUI option.

### DIFF
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -261,6 +261,7 @@ Cvar::Cvar<float> g_bot_fov("g_bot_fov", "bots' \"field of view\"", Cvar::NONE, 
 Cvar::Cvar<int> g_bot_chasetime("g_bot_chasetime", "bots stop chasing after x ms out of sight", Cvar::NONE, 5000);
 Cvar::Cvar<int> g_bot_reactiontime("g_bot_reactiontime", "bots' reaction time to enemies (milliseconds)", Cvar::NONE, 500);
 Cvar::Cvar<bool> g_bot_infinite_funds("g_bot_infinite_funds", "give bots unlimited funds", Cvar::NONE, false);
+Cvar::Cvar<bool> g_bot_infiniteMomentum("g_bot_infiniteMomentum", "allow bots to ignore momentum, but not other restrictions", Cvar::NONE, false);
 
 //</bot stuff>
 
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -68,7 +68,7 @@ struct equipment_t
 	bool allowed( void ) const;
 	bool canBuyNow( void ) const
 	{
-		return allowed() && unlocked();
+		return allowed() && ( g_bot_infiniteMomentum.Get() || unlocked() );
 	}
 	int slots( void ) const;
 	bool operator==( T other ) const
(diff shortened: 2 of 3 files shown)


src/sgame/sg_main.cpp
```diff
@@ -261,6 +261,7 @@ Cvar::Cvar<float> g_bot_fov("g_bot_fov", "bots' \"field of view\"", Cvar::NONE,
 Cvar::Cvar<int> g_bot_chasetime("g_bot_chasetime", "bots stop chasing after x ms out of sight", Cvar::NONE, 5000);
 Cvar::Cvar<int> g_bot_reactiontime("g_bot_reactiontime", "bots' reaction time to enemies (milliseconds)", Cvar::NONE, 500);
 Cvar::Cvar<bool> g_bot_infinite_funds("g_bot_infinite_funds", "give bots unlimited funds", Cvar::NONE, false);
+Cvar::Cvar<bool> g_bot_ignore_momentum("g_bot_ignore_momentum", "allow bots to ignore momentum, but not other restrictions", Cvar::NONE, false);
```
slipher, line 264: I would prefer the name `g_bot_infiniteMomentum` since it does something similar to infinite funds. (But there shouldn't be an underscore after `infinite` since we may want to officially make `_` a namespace separator in tab completion.)

The `but not other restrictions` part of the description seems unnecessary.

slipher: > Ok, but, currently, there seems to be an overall convention to use snake_case in cvar names? 

No, there's only a couple cvars like that. There's a mix of g_bot_foobar, g_bot_fooBar, and g_bot_foo_bar.


> Also, it would play well with the idea of namespaces to have both `g_bot_infinite [ funds | momentum ]` no?

2 cvars is far too few to deserve their own namespace.

bmorel: ok then. I'll rename the other one in the same time then.

bmorel: I'm too lazy to update other stuff. I'll just use the g_bot_infiniteMomentum as suggested.

bmorel:
To make it more obvious if other changes are required: companion is https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/130

Stupid sub-repo organisation!

src/sgame/sg_main.cpp
bmorel: Note that we really need to write down those "conventions" or rather, define conventions, document them, and migrate codebase to those. With that addition of g_bot_infiniteMomentum I just added "yet another convention" in the code (other stuff in g_bot do *not* use uppercase I think).
I did that so that the code will be merged, and won't do another update on this PR for nitpicking, because this 1) is very annoying and 2) makes it more likely to have broken UI (because typos in a subrepo, with no means to check sanity)

If there are actual bugs or real improvements to do, though, I'll be ok with those.

bmorel: Is it good enough now?

slipher: Yes

PR #1938, opened 2022-07-20
add g_bot_minPlayers cvar

### LINKED ISSUE #1786: The need of simple `bot fill` as a cvar aka `minplayers`
Configuring the amount of bots of a server as a cvar has many advantage:

- no need for specific `map/default.cfg` config for simple-enough good-enough configuration (while complex things can still be done);
- makes easy to set-up a configuration entry in UI;
- it's usually the way it is done in other games so people may be familiar.

For example in Smokin' Guns this is configured with `bot_minplayers` cvar, the number is the amount of minimal players a server always have on a map, being bots or real players (so the server fills with bots if needed). Another cvar named `bot_enable` can enable or disable bot filling without setting `bot_minplayers` to `0`.

In ET:Legacy (Wolfenstein: Enemy Territory) the cvar is also named `bot_minplayers` and there is also `bot_enable`.

In Xonotic, this is configured with `minplayers` and `minplayers_per_team` cvars, the first one being the amount of filling bots on non-team games and the second one the amount of filling bots per team on team-based games.

This can be implemented by calling `bot fill` with a cvar value at the start of the game. We may want to do that before actually reading map config files so config files can override this with direct `bot fill` calls.

Additionally Xonotic implements `bot_number` for a minimal amount of bots. I assume it's the equivalent of calling a bunch of `bot add` at every map start.

bmorel:
fix #1786

The current result is buggy, as in, more bots than necessary are created, and excess bots are placed as spectator instead of removed. I do not know why yet. @slipher if you have an idea about that... 

There are roughly 3 times the number of requested bots here.

### DIFF
--- a/dist/configs/config/map/default.cfg
+++ b/dist/configs/config/map/default.cfg
@@ -5,9 +5,3 @@ bot names aliens "^9[bot]^1 Nolwen" "^9[bot]^1 Uriel" "^9[bot]^1 Morgan" "^9[bot
 
 // print a message on screen after 20 seconds
 delay 20000 cp "^2play fair, ^5play nice!"
-
-// fill server with 3 human bots and 3 alien ones,
-// they will leave as soon as real players join,
-// and join as soon as real players leave
-bot fill 3 humans
-bot fill 3 aliens
--- a/dist/configs/config/map/default.cfg
+++ b/dist/configs/config/map/default.cfg
@@ -5,3 +5,9 @@ bot names aliens "^9[bot]^1 Nolwen" "^9[bot]^1 Uriel" "^9[bot]^1 Morgan" "^9[bot
 
 // print a message on screen after 20 seconds
 delay 20000 cp "^2play fair, ^5play nice!"
+
+// fill server with 3 human bots 3 alien ones,
+// they will leave as soon as real players join,
+// and join as soon as real players leave
+// bot fill 3 humans
+// bot fill 3 aliens
--- a/dist/configs/config/server.cfg
+++ b/dist/configs/config/server.cfg
@@ -66,7 +66,7 @@ set g_teamForceBalance 2
 //set g_doWarmup 1
 //set g_warmup 20
 
-set g_bot_minPlayers 3
+set g_bot_defaultFill 3
 
 // load per map config file from this subdirectory
 // it will load map/<mapname>.cfg like map/plat23.cfg
--- a/dist/configs/config/server.cfg
+++ b/dist/configs/config/server.cfg
@@ -66,6 +66,8 @@ set g_teamForceBalance 2
 //set g_doWarmup 1
 //set g_warmup 20
 
+set g_bot_minPlayers 3
+
 // load per map config file from this subdirectory
 // it will load map/<mapname>.cfg like map/plat23.cfg
 // or fallback on map/default.cfg
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -595,26 +595,60 @@ void G_BotFill(bool immediately)
 		return;
 	}
 
-	for (team_t team : {TEAM_ALIENS, TEAM_HUMANS}) {
-		auto& t = level.team[team];
-		int teamSize = t.numClients;
-		if (teamSize > t.botFillTeamSize && t.numBots > 0) {
-			for (int client = 0; client < MAX_CLIENTS; client++) {
-				if (level.clients[client].pers.connected == CON_CONNECTED
-						&& level.clients[client].pers.team == team
-						&& level.clients[client].pers.isFillerBot) {
-					G_BotDel(client);
-					if (--teamSize <= t.botFillTeamSize) {
-						break;
-					}
+	struct fill_t
+	{
+		std::vector<int> current; // list of filler bots
+		int target; // if <0, too many bots, if >0, not enough
+	} fillers[ NUM_TEAMS ] = {};
+	int totalFillers = 0;
+	int missingFillers = 0;
+
+	for (int client = 0; client < MAX_CLIENTS; client++)
+	{
+		auto& pers = level.clients[client].pers;
+		if ( pers.connected == CON_CONNECTED && pers.isFillerBot )
+		{
+			fillers[ pers.team ].current.push_back( client );
+			totalFillers++;
+		}
+	}
+
+	for ( team_t team : {TEAM_ALIENS, TEAM_HUMANS} )
+	{
+		auto& fill = fillers[team];
+		fill.target = level.team[team].botFillTeamSize - level.team[team].numClients;
+		// remove excedent
+		while ( fill.target < 0 && fill.current.size() > 0 )
+		{
+			G_BotDel( fill.current.back() );
+			fill.current.pop_back();
+			fill.target++;
+			totalFillers--;
+		}
+		// remember how much bots are missing
+		if ( fill.target > 0 )
+		{
+			missingFillers += fill.target;
+		}
+	}
+
+	while ( missingFillers )
+	{
+		for ( team_t team : {TEAM_ALIENS, TEAM_HUMANS} )
+		{
+			if ( fillers[ team ].target > 0 )
+			{
+				fillers[ team ].target--;
+				if ( !G_BotAdd( BOT_NAME_FROM_LIST, team, level.team[team].botFillSkillLevel, BOT_DEFAULT_BEHAVIOR, true ) )
+				{
+					//TODO modify the "/bot fill" number so that further warnings will be ignored.
+					//     Note that this means players disconnecting would possibly not be replaced by bot
+					//     so this might be more a problem than an improvement.
+					//TODO if number of clients in both teams is unevent, this is unfair, so kick one of other team.
+					//     to do this, it would be needed to add the new client to fillers[team].current
+					return;
 				}
-			}
-		} else if (teamSize < t.botFillTeamSize) {
-			int additionalBots = t.botFillTeamSize - teamSize;
-			while (additionalBots--) {
-				if (!G_BotAdd(BOT_NAME_FROM_LIST, team, t.botFillSkillLevel, BOT_DEFAULT_BEHAVIOR, true)) {
-					break;
-				}
+				--missingFillers;
 			}
 		}
 	}
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -683,3 +683,13 @@ void botMemory_t::doSprint( int jumpCost, int stamina, usercmd_t& cmd )
 
 	exhausted = exhausted && stamina <= jumpCost * 2;
 }
+
+void G_SetBotFill( int fill )
+{
+	for ( int team = TEAM_NONE + 1; team < NUM_TEAMS; ++team )
+	{
+		level.team[team].botFillTeamSize = fill;
+	}
+
+	G_BotFill(true);
+}
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -586,7 +586,14 @@ void G_BotFill(bool immediately)
 	if (!immediately && level.time < nextCheck) {
 		return;  // don't check every frame to prevent warning spam
 	}
+
 	nextCheck = level.time + 2000;
+	//FIXME this function can actually be called before bots had time to connect
+	//  resulting in filling too many bots.
+	if ( level.matchTime < 2000 )
+	{
+		return;
+	}
 
 	for (team_t team : {TEAM_ALIENS, TEAM_HUMANS}) {
 		auto& t = level.team[team];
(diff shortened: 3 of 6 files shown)


bmorel:
When I join or leave the game, a bot is corrected added or removed from the team I joined. I feel like spectators teams are explicitly filled too, for some reason.

necessarily-equal:
You may want to call `G_BotFill(true);`

slipher:
I don't know why the current code wouldn't work, but there is some missing functionality. It ought to respond to changes in the cvar rather than only using the value at game init. Otherwise the cvar won't work if you put it in `map/default.cfg` for example.

necessarily-equal:
You would also want to set `level.team[team].botFillSkillLevel` to the matching cvar

slipher:
> You may want to call `G_BotFill(true);`

Probably not. Doing stuff in the very first frame is usually a bad idea. And the filling will happen after 2 seconds anyway.

bmorel:
> I don't know why the current code wouldn't work, but there is some missing functionality. It ought to respond to changes in the cvar rather than only using the value at game init. Otherwise the cvar won't work if you put it in map/default.cfg for example.

Right. I should rename it g_bot_default_min_players then.

> You would also want to set level.team[team].botFillSkillLevel to the matching cvar

No, this actually works, because it's handled downstream in the code to add a single bot, which is called by that.

> Doing stuff in the very first frame is usually a bad idea.

I wonder if this is not the problem I'm facing, for some reason...

> And the filling will happen after 2 seconds anyway

Even with "immediately = true" ?

slipher:
> > And the filling will happen after 2 seconds anyway
> 
> Even with "immediately = true" ?

No, I'm saying you don't need to fill it immediately because it will be done later.

> > I don't know why the current code wouldn't work, but there is some missing functionality. It ought to respond to changes in the cvar rather than only using the value at game init. Otherwise the cvar won't work if you put it in map/default.cfg for example.
> 
> Right. I should rename it g_bot_default_min_players then.

I don't think that's an excuse to make a feature that doesn't work reliably...

bmorel:
> I don't think that's an excuse to make a feature that doesn't work reliably...

Of course, I don't intend to push this until it's reliable. Right now it's draft for that very reason.

bmorel:
> Probably not. Doing stuff in the very first frame is usually a bad idea. And the filling will happen after 2 seconds anyway

@slipher You are right. And this is what current code actually does, I think. Or rather, it probably does that even before the 1st frame.
I updated the code to introduce a rather ugly hack, that is, don't `/bot fill` before the game is 2s old, and it fixes it.
The problem was, bots were added once,  and then the code runs `G_BotFill()` anew, but bots didn't had the time to actually connect, so the game runs N*2 more G_BotAdd() calls.

I think the current hack is okayish, it's certainly good enough for me to run on my mod, but I would prefer a cleaner way to merge this into 0.54. If you have an idea, I'm all eyes open (can't use ears on my screen).

bmorel:
FWIW this code should be "correct", or rather good enough. It works with no known bug, but the bug might happen on very slow system where bots take more than 2s to connect.

slipher:
The code as currently written does not work for setting the number of the bots in the current game. This means it would not work in a map config. So it should not be merged in the current state - it's a bad user experience if there is more than one way to configure something but only one of them actually works correctly.

bmorel:
> This means it would not work in a map config. 

Indeed, it would only work if ran *prior* to map config

>  it's a bad user experience if there is more than one way to configure something but only one of them actually works correctly.

On this I agree, but OTOH I don't consider the map config is the correct place for such a feature (setting number of bots), and map rotation is a much better one.

OTOH, the map config allows to affect *individual* bots with specific skill and behavior, and I have no idea how I could get this "upstream", as in, in server's config or map rotation.

If you have any idea as of how to "watch" a cvar's value change and run a `/bot fill` command IG instead, I'd be glad to hear about it...
The main problem is, I need to update some property of the `level.team[];` member.
An "easy" solution would be to have 1 cvar per team, and just replace that thing, but it would mean `/bot fill` would only modify the cvar's content, which feels really bad to me.

slipher:
It's often considered that the appropriate number of players is different for different maps. So it could make sense to set a different fill number depending on the map. Therefore setting the bot fill should work in a map config.

bmorel:
In any case, I'm fine if this is not merged today. The code is here and working for those willing to use it, it's the most important thing.

bmorel:
> The code as currently written does not work for setting the number of the bots in the current game. 

This is now fixed. The only "problem" (or rather missing feature) which remains here is that this applies to all teams, so I consider this still needs improvements, but I'm not sure about the best way for that, as I'm not at all fond of the idea of multiplying by 3 all cvars and related code that applies to team because there are 2 teams (+ the "both teams" situation).

If anyone have an idea to solve this problem, it would be really nice.

src/sgame/sg_main.cpp
```diff
@@ -260,6 +260,7 @@ Cvar::Cvar<bool> g_bot_attackStruct("g_bot_attackStruct", "whether bots target b
 Cvar::Cvar<float> g_bot_fov("g_bot_fov", "bots' \"field of view\"", Cvar::NONE, 125);
 Cvar::Cvar<int> g_bot_chasetime("g_bot_chasetime", "bots stop chasing after x ms out of sight", Cvar::NONE, 5000);
 Cvar::Cvar<int> g_bot_reactiontime("g_bot_reactiontime", "bots' reaction time to enemies (milliseconds)", Cvar::NONE, 500);
+Cvar::Cvar<int> g_bot_defaultFill("g_bot_defaultFill", "fills both teams with that number of bots at start of game", Cvar::NONE, 3);
```
DolceTriade: The commit message here seems wrong? Can we make the code and commit message consistent?

DolceTriade:
With Lua, this becomes a simple Lua script, but I see the value of having this baked into the code itself.

src/sgame/sg_main.cpp
bmorel: Certainly, and I got caught myself at it for the draft I started to only add a way by GUI to do it.
Understand though that the cvar's name was changed after a request, so that it currently just break the average convention (when it comes to bots at least).
I'm not against rewording, but not until the PR is accepted.

bmorel:
> With Lua, this becomes a simple Lua script, but I see the value of having this baked into the code itself.

Imo that kind of "change" should be accessible to non-coders, such as people willing to host a LAN game. There is also currently no official Lua in gamelogic itself, and while I think it might be a useful feature later, it still need to be actually done, and for such a trivial matter I don't think Lua is worth it.
The only important stuff which was asked to me in that PR was really to sync the change, and I never used the Cvar::Callback<>() before.
Now that I used it, I'm even tempted to rewrite the code for the g_bot_default_skill value to change the skill value of bots that were added without explicit skill.

Lua certainly have good (but not only) points, but it's a scripting feature, not something that normal admins are expected to play with.

src/sgame/sg_bot.cpp
```diff
@@ -595,26 +595,60 @@ void G_BotFill(bool immediately)
...
-			int additionalBots = t.botFillTeamSize - teamSize;
-			while (additionalBots--) {
-				if (!G_BotAdd(BOT_NAME_FROM_LIST, team, t.botFillSkillLevel, BOT_DEFAULT_BEHAVIOR, true)) {
-					break;
+	struct fill_t
+	{
+		std::vector<int> current; // list of filler bots
+		int target; // if <0, too many bots, if >0, not enough
+	} fillers[ NUM_TEAMS ] = {};
+	int totalFillers = 0;
```
slipher, line 603: This variable is never used.

src/sgame/sg_bot.cpp
```diff
@@ -595,26 +595,60 @@ void G_BotFill(bool immediately)
...
+		{
+			if ( fillers[ team ].target > 0 )
+			{
+				fillers[ team ].target--;
+				if ( !G_BotAdd( BOT_NAME_FROM_LIST, team, level.team[team].botFillSkillLevel, BOT_DEFAULT_BEHAVIOR, true ) )
+				{
+					//TODO modify the "/bot fill" number so that further warnings will be ignored.
+					//     Note that this means players disconnecting would possibly not be replaced by bot
+					//     so this might be more a problem than an improvement.
+					//TODO if number of clients in both teams is unevent, this is unfair, so kick one of other team.
```
slipher, line 647: s/unevent/uneven/

src/sgame/sg_bot.cpp
```diff
@@ -586,28 +586,69 @@ void G_BotFill(bool immediately)
 	if (!immediately && level.time < nextCheck) {
 		return;  // don't check every frame to prevent warning spam
 	}
+
 	nextCheck = level.time + 2000;
+	//FIXME this function can actually be called before bots had time to connect
+	//  resulting in filling too many bots.
+	if ( level.matchTime < 2000 )
```
slipher, line 593: I wonder if you could get rid of this if you changed the test to `level.time <= nextCheck` to avoid running at time 0 (and set immediately=false on the cvar callback).

dist/configs/config/server.cfg
```diff
@@ -66,6 +66,9 @@ set g_teamForceBalance 2
 //set g_doWarmup 1
 //set g_warmup 20
 
+// allow to fill teams with bots, can also be used in map rotation
```
slipher, line 69: I think the map rotation suggestion is not good because the command will only run if the map is chosen via rotation, and not any other way. So you will have different settings depending whether the next map is picked automatically or by voting.

dist/configs/config/map/default.cfg
```diff
@@ -5,9 +5,3 @@ bot names aliens "^9[bot]^1 Nolwen" "^9[bot]^1 Uriel" "^9[bot]^1 Morgan" "^9[bot
 
 // print a message on screen after 20 seconds
 delay 20000 cp "^2play fair, ^5play nice!"
-
-// fill server with 3 human bots and 3 alien ones,
```
slipher, line 9: The `bot fill` commands should not be deleted without a trace since they serve as documentation that the command exists. How about leaving these here and making the `g_bot_minPlayers` a commented-out alternative?

src/sgame/sg_main.cpp
```diff
@@ -602,6 +603,12 @@ void G_InitGame( int levelTime, int randomSeed, bool inClient )
 
 	// Initialize build point counts for the intial layout.
 	G_UpdateBuildPointBudgets();
+
+	for ( int i = TEAM_NONE + 1; i < NUM_TEAMS; ++i )
```
slipher, line 607: Remove this. The cvar callback should run when it is first initialized.

slipher (commented):
The general idea looks good. Thanks for the improvement to fill the teams more evenly.

dist/configs/config/map/default.cfg
illwieckz: I would comment out this one instead. One of the idea is also that one may require less configuration files to get something working.

DolceTriade (approved):

DolceTriade:
LGTM % slipher's comments

dist/configs/config/server.cfg
```diff
@@ -66,6 +66,9 @@ set g_teamForceBalance 2
...
+set g_bot_minPlayers 3
```
illwieckz, line 70: Did you meant  `g_bot_defaultFill`?

src/sgame/sg_main.cpp
bmorel: Removing this actually breaks the PR

src/sgame/sg_main.cpp
bmorel: I'll rebase to get less commits when things will be ready, and will update the commit messages to be more meaningful in the go.

src/sgame/sg_main.cpp
slipher: The cvar callback does run for the default value. The problem is this happens before G_Init which memsets the global structs.

dist/configs/config/server.cfg
```diff
@@ -66,6 +66,11 @@ set g_teamForceBalance 2
 //set g_doWarmup 1
 //set g_warmup 20
 
+// allow to fill teams with bots, can also be used in map rotation for
+// per-map settings, *as long as the map is choosen by normal rotation*,
```
slipher, line 70: What's the point of suggesting something which doesn't work? Map configs exist and are a working solution for this.

bmorel: The point is to warn people that host many maps of the limitations.

> Map configs exist and are a working solution for this.

I'm not sure creating 114 individual files is a good solution though. This number is *not* ridiculous, mind you:

> find www/unvanquished -name "map-*.dpk"  |  cut -f1 -d_ | sort | uniq | wc -l
114

This is what I have. And I expect the number to keep growing.
But if you insist I can remove the comment, if that's the only thing preventing this feature from being merged.

src/sgame/sg_bot.cpp
bmorel: It seems what *can* be done here is to change the starting value of nextCheck to 2000, and remove this test, actually.
Simply removing the test leads to a segfault, which means it makes it easy to get a segfault if removing this test at all.

I think it would be an error to do that. The "solution" there might be ugly, but at least I think it's rather safe, and it actually works.

dist/configs/config/server.cfg
slipher: Yes please remove the mention of map rotations

bmorel: I removed it

slipher:
LGTM

could use squashing

PR #1947, opened 2022-07-21
Prevent granger from damaging buildables

necessarily-equal:
I think it was originally intended like this

Closes https://github.com/Unvanquished/gameplay/issues/22


bmorel:
LGTM.

Issue #1640, opened 2021-12-20
Add an "oxygen" bar when swimming underwater


bmorel:
When a player is swimming underwater, after some time they drawn, but there is no indication of when it starts.

This can be seen only in the mission maps, AFAIK.

slipher:
If you want a normal Tremulous map with swimming for testing, you can try "Cleanser"

bmorel:
Got a quick look at how this is handled as of 92c3e6111bb627bf3a952052cd91f36f1f76d7d5

All the code is server-side, there is no way the client can get the info about how many time to be out of air, which seems to be 12s for everyone.
So, the easier solution is to similarly hardcode the 12s time in client (or rather, to have the value put in shared as a constant), when client detects the player is underwater print a bar which filling ratio depends on 12s, a lastBreath var, and leveltime.

slipher:
As a dev I kind of dislike having UI elements for rare situations like this. It means when I'm doing something like RmlUi migration, I'll either don't test it (and it probably breaks), or it's a big pain to test.

bmorel:
I hope we won't be going to migrate UI toolkit every 3 versions though. This is not really a priority thing.

slipher:
Well we already have migration to the latest version of RmlUI queued up! And there's simpler stuff like changing font sizes.

bmorel:
> This is not really a priority thing.

I open tickets so that maybe someday they'll be addressed. Not necessarily for the next version. I also sometimes wander in the open tickets, pick one and do it.
I would even rather have rmlUI at latest stable version before adding too much UI elements. But if someone feels bored enough to implement the stuff...

slipher:
Oh I see. I was inspired to comment because @necessarily-equal just opened a PR for it.

necessarily-equal:
I understand the concern. Thankfully this is quite simple and covers an 
area that is already used by many things. I don't think there is any 
chance this would break without something else breaking.

bmorel:
Well, "someone bored enough" was him, it seems.

I think we should use more the "milestone" feature of github btw. This way, one could have milestoned this for i.e. "post-beta".

PR #1949, opened 2022-07-21
Add oxygen bar for when you are underwater

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1640

To be used with https://github.com/Unvanquished/Unvanquished/pull/1949


bmorel:
When running it with both oxygen-bar branches, I have:

> Warn: progress bar command oxygen not found

Spammed *a lot*. So much that I can't even select a line.
And of course, it does not shows.

bmorel:
LGTM.

The progress bar is crude, but that's still an improvement.

PR #1952, opened 2022-07-22
use trace distance instead of origin distance

bmorel:
No real bug is fixed with that, as far as I know, but it should make things better for later. This also avoid calculating a distance between 2 origins, when there is already such distance information in the shape of a trace.

### DIFF
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -2184,10 +2184,11 @@ void ClientThink_real( gentity_t *self )
 
 		ent = &g_entities[ trace.entityNum ];
 
-		if ( ent && ent->use &&
-		     ( !ent->buildableTeam   || ent->buildableTeam   == client->pers.team ) &&
-		     ( !ent->conditions.team || ent->conditions.team == client->pers.team ) &&
-		     Distance( self->s.origin, ent->s.origin ) < ENTITY_USE_RANGE )
+		if ( ent && ent->use
+		     && ( !ent->buildableTeam   || ent->buildableTeam   == client->pers.team )
+		     && ( !ent->conditions.team || ent->conditions.team == client->pers.team )
+		     && trace.fraction < 1.0f
+		   )
 		{
 			if ( g_debugEntities.Get() > 1 )
 			{
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -2188,6 +2188,7 @@ void ClientThink_real( gentity_t *self )
 		     && ( !ent->buildableTeam   || ent->buildableTeam   == client->pers.team )
 		     && ( !ent->conditions.team || ent->conditions.team == client->pers.team )
 		     && trace.fraction < 1.0f
+				 && !( entityType == eType_t::BUILDABLE && Distance( self->s.origin, ent->s.origin ) >= ENTITY_USE_RANGE )
 		   )
 		{
 			if ( g_debugEntities.Get() > 1 )


slipher:
I recall an earlier discussion about this in relation to the armory. I said that origin distance seems better since the armory model is round, not box-shaped.

Other usable entities are very rarely encountered, but changing the algorithm seems like it could only hurt maps that use them.

bmorel:
>  I said that origin distance seems better since the armory model is round, not box-shaped.

Then radiuses of target and user at the very least should be used.

Also note that:

> No real bug is fixed with that, as far as I know,

Is a half-lie. Fact is, without this change, damage entities can not be +activate. I don't consider this a bug, because the real bugs there are:

1. damage entities should not be +activate
2. damage entities should react to damages

But this change *does* make them +activate anew, and at some point I'd like to let mappers do those +activate when using a proper option for stuff in unvanquished. Currently, this seems to "work" only because those

1. do not have a "name"
2. damage thing seems to prevent them from being considered normal
3. so the game fallback on +activate behavior

bmorel:
Result on the damage entities (yes, I know, it's a bug, but it gives an idea of the change) can be tested on [this map](https://github.com/UnvanquishedAssets/UnvanquishedTestAssets/tree/master/src/map-test-ent-damage_src.dpkdir)

bmorel:
> Other usable entities are very rarely encountered, but changing the algorithm seems like it could only hurt maps that use them.

Well, it's easy. Currently, the only usable entities which are usable on-purpose is the armoury. Now, we discovered that damage-triggered entities in unvanquished actually work as `usable`, and this change actually fixes that.

IIRC @cu-kai said on IRC that to have the ability to do actually usable buttons or doors in maps is a desirable feature. This patch allows to abuse the fact damage-entities were broken when CBSE was integrated to get those, and a feeling about how that would look like.
Note that it repairs damages which come from the USE_ENTITY_RANGE changes we did in 0.53, and that this is my 3rd change in order to reduce the problems it caused, the 1st one was a quick'n dirty increase of hard-coded range.

Maybe I should just close #1837 and this PR, until we decide what we want to do about about usable entities?

slipher:
> Now, we discovered that damage-triggered entities in unvanquished actually work as usable, and this change actually fixes that.

What? All this change does is to alter the distances at which something is usable, not whether it is potentially usable at all.

bmorel:
> All this change does is to alter the distances at which something is usable

Yes, and imo it's best to keep in mind that the game does not cares about the models. It's weird and plain wrong to require a distance between origins, as an entity which is bigger than said distance would hence become unusable.
To be more exact, if the BBox pseudo-radius of both entities is bigger than the hard-coded distance, it leads to broken behavior. This is what I suspect happens here.

To be honest, I'm not sure it worth arguing more on that issue and on every single issue related to that. Instead, I'd rather have a higher level discussion so that we can agree on global direction to take, and then do it.

necessarily-equal:
So you recently asked on IRC that you don't understand why this PR isn't moving forward. Unless I missed some change, I still don't consider it ready.

Here are my points that I made on IRC a month ago:

> > About the armory PR: two or three things. First I'm not fond of this "you have to look at the armory to get the thing to open". It forces you to look *at* the thing with good precision. You can't even look a bit sideways like you would (could) if you care for life
> 
> > Second, [snip, not actually a regression]
> 
> > Third, I really don't like that the condition for binds is different. I don't want players with binds to be advantaged even more :-/
>
> > I also note that you can refuel if you stand on armory. [About that ppl don't seem to case if it's good or not, I'm fine with this implementation that makes it work.]

Then a bit afterwards in the conversation, I mention another problem: with your implementation: somone can use an armory on the other side of a wall if, and only if they use binds.

Later on, you said "I agree with the binds problem" and "I'll fix that". Which is why I did put this PR on draft.

I don't think issue 1) mentioned here is blocking even if I personally really don't like it. But I find the bind thing blocking.

bmorel:
> Later on, you said "I agree with the binds problem" and "I'll fix that". Which is why I did put this PR on draft.

You're talking about a *different* PR, that is, #1837 here.

What you likely missed is that I started tickets to actually discuss about this PR (and others), so that I know if I should just stop caring and do the changes only in my own mod: #2054 and #2047

bmorel:
Here is a direct proof that you are wrong about this PR, on the map testing server:

2022-08-24-182447_1366x768_scrot

There's a glass in between.

src/sgame/sg_active.cpp
```diff
@@ -2184,10 +2184,11 @@ void ClientThink_real( gentity_t *self )
...
 		ent = &g_entities[ trace.entityNum ];
 
-		if ( ent && ent->use &&
-		     ( !ent->buildableTeam   || ent->buildableTeam   == client->pers.team ) &&
-		     ( !ent->conditions.team || ent->conditions.team == client->pers.team ) &&
-		     Distance( self->s.origin, ent->s.origin ) < ENTITY_USE_RANGE )
+		if ( ent && ent->use
+		     && ( !ent->buildableTeam   || ent->buildableTeam   == client->pers.team )
+		     && ( !ent->conditions.team || ent->conditions.team == client->pers.team )
+		     && trace.fraction < 1.0f
```
necessarily-equal, line 2190: How about keeping the distance check only for buildables? Something like that (but redacted to be more readable and use the correct symbol names — I am typing this on a phone, can't grep)

```suggestion
		     && !( entityType == eType_t::BUILDABLE && Distance( self->s.origin, ent->s.origin ) >= ENTITY_USE_RANGE )
```

So that this would fix the use of map entities without making the buildable case inconsistent anew

DolceTriade: I think this is a reasonable compromise. I think it's fair to have different modes of "usable" entities. Not all usable entities need to behave the same, especially if the usable entities we are trying to support are legacy entities.

bmorel:
This won't be merged anyway, let's close it.

PR #1953, opened 2022-07-22
Trem compat

bmorel:
This PR aims to automatically replace repeaters from trem maps by a drill, since they have the same role and very close mechanics:

* can't (or rather should avoid, for drill) be too close from other ones
* refuel energy weapons
* help with building more

### DIFF
--- a/src/sgame/sg_spawn.cpp
+++ b/src/sgame/sg_spawn.cpp
@@ -579,11 +579,7 @@ bool G_CallSpawnFunction( gentity_t *spawnedEntity )
 		 *  to allow each spawn function to test and handle for itself,
 		 *  we handle it automatically *after* the spawn (but before it's use/reset)
 		 */
-		if(!G_HandleEntityVersions( spawnedClass, spawnedEntity ))
-			return false;
-
-
-		return true;
+		return G_HandleEntityVersions( spawnedClass, spawnedEntity );
 	}
 
 	//don't even warn about spawning-errors with -2 (maps might still work at least partly if we ignore these willingly)
--- a/src/shared/bg_misc.cpp
+++ b/src/shared/bg_misc.cpp
@@ -60,6 +60,7 @@ static const buildableName_t bg_buildableNameList[] =
 	{ BA_H_MEDISTAT,  "medistat",  "team_human_medistat"  },
  	{ BA_H_DRILL,     "drill",     "team_human_drill"     },
 	{ BA_H_REACTOR,   "reactor",   "team_human_reactor"   },
+	{ BA_H_DRILL,     "repeater",  "team_human_repeater"  },
 };
 
 static const size_t bg_numBuildables = ARRAY_LEN( bg_buildableNameList );
@@ -2272,7 +2273,16 @@ BoundedVector<buildable_t, BA_NUM_BUILDABLES>
 		}
 		else
 		{
-			Log::Warn( "unknown buildable %s", *i );
+			// tremulous' repeater was replaced by drill
+			if ( strcmp( "repeater", *i ) )
+			{
+				results.append( BA_H_DRILL );
+				Log::Warn( "Deprecated %s was replaced by a drill", *i );
+			}
+			else
+			{
+				Log::Warn( "unknown buildable %s", *i );
+			}
 		}
 	}
 


DolceTriade:
What if we don't have this change? We get warnings?

bmorel:
> What if we don't have this change? We get warnings?

Warnings and breakages.
For example, a mapper could have decided to only open a door if a repeater is built somewhere.
Or may have placed repeaters to refuel energy weapon.
But the most likely is probably that mappers used repeaters to build more complex bases. Which means that the base by default requires more BPs. This leads to disabled buildings, which is obviously not the intent of the mapper.
This last one is the most serious, and can be worked around with a per-map configuration.

necessarily-equal:
I'm unconvinced by the reasoning presented in the issue's original post.

First, point by point
* can't (or rather should avoid, for drill) be too close from other ones — yes, sure, but it's not because they have similar pitfals that they are interchangeable. For example, (to make a silly example) even if you should both protect medistation and armory, you can't swap one for another because they have different roles.
* refuel energy weapons — yeah ok that's now true (new), but I think outside of a main base this shouldn't bring anything.
* help with building more — yeah but that does so in totally different ways, see the silly example above where you could make an argument that both medistation and armory help supporting the team

A problem I can see is that replacing repeaters with drills can mean that you are spawning two drills next to another, which was unintended in the original map and *removes* build points. I also note that power distribution and resource mining are quite different.

So to avoid speaking in hypotheticals, do you have example of maps that are "fixed" (automatically ported) by this change? Do some of these maps contain drills already? Does it improve the situation more than just spawning nothing here? (I think you've seen more trem maps than I do)

bmorel:
> For example, (to make a silly example) 

This is indeed silly, since as you pointed out, they have different roles, while both repeater and drill have the same: to allow base expansions.

> yeah ok that's now true (new), 

Some (@illwieckz  to be exact) would say this was forgotten and hence a regression, so this change only addressed a bug, from that PoV. I also think it was a bug, as it would be simply ridiculous to have only the reactor and armory able to reload energy weapons, since they are often quite close anyway.

> see the silly example

You point it anew that it's a silly one. Here, it's more like comparing acid tube and spiker.


> A problem I can see is that replacing repeaters with drills can mean that you are spawning two drills next to another,

No, because tremulous have no drill, and unvanquished have no repeater. So, you can have only one of those in a map.

> So to avoid speaking in hypotheticals, do you have example of maps that are "fixed" (automatically ported) by this change? Do some of these maps contain drills already?

This mitigates the BP disabling buildings in 2 maps I've seen so far:

* tremship, which is completely working
* blackout, which requires more work to have it working, enough so that I'm not sure it's worth completing the work, and even if yes, probably not worth to upstream it

> I think you've seen more trem maps than I do

Maybe. I mostly rely on @Sweeet for that honestly. Repeaters are only likely to be found in scenario-like maps, and I guess those are pretty rare. It's even so pretty rare than unvanquished have none.

illwieckz:
I would say you have more chance to fix things by blindly replacing repeaters with drills than doing nothing or something else.

The case of maps relying on close drills are rarer than maps relying on drills.

cu-kai:
> I'm unconvinced by the reasoning presented in the issue's original post.
> 
> [snipped]
>
> A problem I can see is that replacing repeaters with drills can mean that you are spawning two drills next to another, which was unintended in the original map and _removes_ build points. I also note that power distribution and resource mining are quite different.

In Tremulous, you couldn't build two repeaters next to each other (and indeed in 1.2/GPP if two repeaters were forced next to each other, one of them would explode).

illwieckz (approved):
LGTM.

DolceTriade:
```
Warn: couldn't read file 'configs/buildables/repeater.attr.cfg': No such file 
Warn: Couldn't find image file 'default' 
```

This commit leads to this? Is there a corresponding change to add repeaters?

imo, we shouldn't care about trem compat. If those maps will function just fine without repeaters.

necessarily-equal:
+1 for no trem compat on repeater, I'd rather have a warning that the map layout has to be updated, than partly broken layout and hacks

bmorel:
>> Warn: couldn't read file 'configs/buildables/repeater.attr.cfg': No such file 
>>Warn: Couldn't find image file 'default' 

> This commit leads to this? Is there a corresponding change to add repeaters?
> imo, we shouldn't care about trem compat. If those maps will function just fine without repeaters.

@DolceTriade Please specify the testing conditions, so that I can reproduce and fix it (if that's reproducible). If this warning is introduced by my change, then it means this PR is buggy and that I need to fix it. It's likely I forgot to dynamically replace something, since this PR is meant to put *drills* instead of repeaters.

> than partly broken layout and hacks

Well, then we should revert lot of changes, including the unfinished CBSE which brings various breakages in maps, and no real benefit.
One fact is: there are hundreds of tremulous maps around. There are only 11 unvanquished maps, and of  those, half are not finished, to the point that their arena file even claim they are made for tremulous.
The *only* maps really made for unvanquished are Jota and Fortification. Even those are buggy, or at least engine does not really support them.

To have shitload of warnings that will never be fixed (because it's the truth: people learn to ignore warnings, which is why we have so many of those, instead of thousands of bug reports) just gives the feeling the game is unfinished, unpolished, and will never be any of those.

illwieckz:
> `Warn: couldn't read file 'configs/buildables/repeater.attr.cfg': No such file`

This one looks related to this PR indeed.

> `Warn: Couldn't find image file 'default' `

This one is unrelated to this PR and exists since years, see for details:

- https://github.com/Unvanquished/Unvanquished/issues/1998

> imo, we shouldn't care about trem compat. 

My opinion on that:

- https://github.com/Unvanquished/Unvanquished/issues/1999

DolceTriade:
@freem I just load any map to reproduce. 

I'm all for certain types of trem compat. But the repeater != drill, it has different gameplay mechanics and not having this change doesn't really harm trem compat :-/

PR #1954, opened 2022-07-22
improve door handling for AI (but no func_door_model)

bmorel:
This is improved for several things:

* add support for rotating doors
* doors never really disappear, instead they move
* support all aliases of targetname

Note that the func_door_model is probably not entirely fixed, as I have
no way to test this and no understanding of it.

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -530,6 +530,83 @@ Pos1 is "at rest", pos2 is "activated"
 SetMoverState
 ===============
 */
+
+// most elevators are *not* func_door, so that broken.
+// Some (one on forlorn) is a func_door, with a "targetname",
+// and I suspect this is what makes it different from doors.
+// That distinction means that, doors with a targetname will
+// break for bots.
+// Also, for now, doors without targetname work because bots
+// just walk to kiss (or worse!) them, until they open wide
+// for them to enter.
+// The correct solution here would be to generate navcons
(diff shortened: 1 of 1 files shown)


bmorel:
Only reason this is a draft is because I want to improve even more the situation. There are still things which are not working as they should for bots, for example in cu-kai 's map zittrig-arena, where the button-activated doors in bases have only 1 of the 2 doors working.
From my current investigations, it's because the `targetName` of the 2nd one is removed by unvanquished. I suspect this is be because they are in the same `team`/`group`, so I need to find how to cycle the group or something.

bmorel:
I think this is now ready for review, but to be honest, I'd prefer bunch of tests to be ran on several maps rather than just someone reading the C++ change.

bmorel:
Found a bug already in my latest changes.

bmorel:
This also helps with #1220 (even if the ticket is likely too generic)

DolceTriade (approved):

necessarily-equal:
It seems to work well. My biggest fear of mine which was that they wouldn't pass doors because they wouldn't try to path through them, but they seem to plan ahead. (and if sometime they go towards the door and move away, it looks like it can be due to action roam and a random walk)

It seems that they sometime get stuck on tremor, but this may be due to other things. The bug I've seen most is when 2 bots are stuck looking at each another. But that's on 0.53.0/sync too. And there is one where they stopped before armory, and I guess it was rare enough so that I can't tell if it's new or not. LGTM

PR #1955, opened 2022-07-22
Constexpr check

necessarily-equal:
Whoops, who would have thought keeping two enum lists in sync would be hard.

There are other similar lists in the file. Unfortunately they are harder to static_assert. I wouldn't be surprised [if we find more bugs there](https://github.com/Unvanquished/Unvanquished/pull/1881)


src/shared/bg_public.h
```diff
@@ -1471,6 +1471,18 @@ void     AddFlag(int &flags, int flag);
...
+// true if first is lower than second argument
+// this function is like ` strcmp(first, second)` except constexpr.
+// outside of compile-time evaluations, prefer strcmp.
+constexpr int constexpr_strcmp(const char *first, const char *second)
+{
+       return *first == *second
+               ? (*first == '\0'
+                       ? 0 // strings are equal
+                       : constexpr_strcmp(first+1, second+1) ) // we aren't done y
+               : (*first < *second ? -1 : 1); // these two chars are different, so
```
slipher, line 1483: I don't suppose we'll ever use this on non-ASCII characters, but for portability and consistency with strcmp, the characters should be converted to the 0-255 range before comparison (char may be signed). E.g. `(*first & 0xFF) - (*second & 0xFF)`

src/shared/bg_misc.cpp
```diff
@@ -1350,14 +1348,19 @@ static const char *const eventnames[] =
 BG_EventName
 ===============
 */
-const char *BG_EventName( int num )
+static constexpr const char *constexpr_BG_EventName( int num )
 {
```
bmorel, line 1352: IMO the if block was more readable.

src/shared/bg_public.h
```diff
@@ -1471,6 +1471,18 @@ void     AddFlag(int &flags, int flag);
 void     RemoveFlag(int &flags, int flag);
 void     ToggleFlag(int &flags, int flag);
 
+// true if first is lower than second argument
+// this function is like ` strcmp(first, second)` except constexpr.
+// outside of compile-time evaluations, prefer strcmp.
+constexpr int constexpr_strcmp(const char *first, const char *second)
+{
+       return *first == *second
```
bmorel, line 1479: `foo ? bar ? 0 : 1 : 2` is unreadable to me. And here, you nested 3 ternary operators!

Imo, this would be more readable:

```
if ( *first == *second )
{
  return *first == '\0' ? 0 : constexpr_strcmp(first+1, second+1) );
}
return *first < *second ? -1 : 1;
```

You will notice that it does not even takes more lines...

necessarily-equal: The thing is that this is two statements instead of one, and we need one (1) statement only in the body of a constexpr function

necessarily-equal: (until we upgrade to a more recent C++ compiler, something like C++17)

src/shared/bg_public.h
necessarily-equal: Fixed

slipher:
To get some verification by the compiler, you could make it a switch statement instead of an array. If you do it like 
```
switch (event) {
case EV_NONE: return "EV_NONE";
case EV_FOOTSTEP: return "EV_FOOTSTEP";
...
```
then you can get a compiler warning if one is missing.

Issue #1389, opened 2021-04-22
wall-walking auto-pitch is too fast by default


bmorel:
By default, it is set on "Fast". I believe fast auto-pitch is too fast, especially for new players.
In my opinion, default settings should be adapted to new players, which can then adjust to their skills when they evolve, thus, I would recommend using "Medium" maximum.
For my tastes, "Slow" is too slow, but maybe that would even be a better default than medium.

bmorel:
Most people disagree this is an issue.

PR #1956, opened 2022-07-22
Change default wallwalking default smooth value

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1389

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -97,7 +97,7 @@ Cvar::Cvar<bool> cg_noTaunt("cg_noTaunt", "disable taunt sounds", Cvar::NONE, fa
 Cvar::Cvar<bool> cg_drawSurfNormal("cg_drawSurfNormal", "visualize normal vector of facing surface", Cvar::CHEAT, false);
 Cvar::Range<Cvar::Cvar<int>> cg_drawBBOX("cg_drawBBOX", "show entity bounding boxes (2 = solid)", Cvar::CHEAT, 0, 0, 2);
 Cvar::Cvar<bool> cg_drawEntityInfo("cg_drawEntityInfo", "show number and type of facing entity", Cvar::CHEAT, false);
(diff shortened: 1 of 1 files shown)


DolceTriade:
Can we change this in the UI rather than here?

DolceTriade:
Also I'm not sure medium is the right choice here. https://github.com/Unvanquished/Unvanquished/commit/84d6a234e5756ef8725fda0a8f9c9afad54427b2

I agree with Sixthly here.

bmorel:
> Can we change this in the UI rather than here?

Not really, *this* is actually the UI. This changes the default value, so that the game is more usable for newcommers playing aliens.

> I agree with Sixthly here.

Who? And said what?

cu-kai:
I think medium is too slow. The first thing I always do is set it to fast (at least) if it isn't already. I actually use `cg_wwSmoothTime 50` peresonally.

bmorel:
> I think medium is too slow. The first thing I always do is set it to fast (at least) if it isn't already. I actually use cg_wwSmoothTime 50 peresonally.

Thing is, default configuration should be friendly to *newcommers*. Do you think qualify?

bmorel:
One comment that could be useful considering the UI though: it does not reflects the fact the cvar is an actual number. To set this value to other numbers, people have to know it's a delay-type value, while UI suggest it's only a list-type value. This would be possibly more confusing for new players to have to adjust a number, though.

cu-kai:
> > I think medium is too slow. The first thing I always do is set it to fast (at least) if it isn't already. I actually use cg_wwSmoothTime 50 peresonally.
> 
> Thing is, default configuration should be friendly to _newcommers_. Do you think qualify?

I don't qualify a newcomer, but even when I was all those years ago I found medium very disorientating.

bmorel:
> I don't qualify a newcomer, but even when I was all those years ago I found medium very disorientating.

Wallwalking certainly have this disorienting problem (especially in buggy maps like nano). My opinion, but it's really only mine, is that slower camera moves make it less  disorienting.

PR #1957, opened 2022-07-22
fix target_push when angle is used instead of target

### LINKED ISSUE #1850: pusher entities which only set speed and angle are removed and thus do not work
I'll detail this issue more later.
For reference, a map to reproduce the problematic behavior was [made](https://github.com/UnvanquishedAssets/UnvanquishedTestAssets/tree/master/src/map-test-pushpad_src.dpkdir). The bug can also be seen in zittrig-arena from cu-kai.
On the test map, both jumpers are meant to work the same way. Zittrig-arena works fine in tremulous.

It seems the pusher (push1 in test map) is removed because it can not find what it is associated with.

_Edit by illwieckz:_ I put a better link and fixed typo in Kai's name.

bmorel:
fix #1850

### DIFF
--- a/src/sgame/sg_spawn_shared.cpp
+++ b/src/sgame/sg_spawn_shared.cpp
@@ -69,6 +69,10 @@ void think_aimAtTarget( gentity_t *self )
 	VectorAdd( self->r.absmin, self->r.absmax, origin );
 	VectorScale( origin, 0.5f, origin );
 
+	if ( self->s.origin2[0] != 0.f || self->s.origin2[1] != 0.f || self->s.origin2[2] != 0.f )
+	{
+		return;
+	}
 	pickedTarget = G_PickRandomTargetFor( self );
 
 	if ( !pickedTarget )


src/sgame/sg_spawn_shared.cpp
```diff
@@ -69,6 +69,14 @@ void think_aimAtTarget( gentity_t *self )
 	VectorAdd( self->r.absmin, self->r.absmax, origin );
 	VectorScale( origin, 0.5f, origin );
 
+	if ( false
```
DolceTriade, line 72: why

src/sgame/sg_spawn_shared.cpp
```diff
@@ -69,6 +69,14 @@ void think_aimAtTarget( gentity_t *self )
...
+			|| self->s.origin2[0] != 0.f
```
DolceTriade, line 73: Do you need to compare with epsilon?

DolceTriade (changes requested):

src/sgame/sg_spawn_shared.cpp
bmorel: old trick of mine. Forgot to remove it from unv.
I usually do this when I have a `if` with long lines. This makes it possible to:

1. `yyp` in vim to quickly duplicate stuff,
2. alignment is made with tabs or space: don't care, it just works,
3. you can remove or add components of the if by touching a single line, reduce number of changes in git

src/sgame/sg_spawn_shared.cpp
bmorel: No, because the values are set to 0 before. If it happens that something changed the value and all result in a 0 +- epsilon, then it means something was initialized and thus it's fine (but map is likely buggy).

src/sgame/sg_spawn_shared.cpp
bmorel: Note that this would not have happened with a sane interface, instead of that horribly vec3_t.

DolceTriade (approved):

PR #1973, opened 2022-07-25
add teamstatus sgame command

bmorel:
Dedicated to @cu-kai


bmorel:
Drafting because:

1. log notice not good obviously
2. there's work for same feature, which might trigger less boring discussions I think

bmorel:
This is superseded (as expected from start) by #1974

PR #1974, opened 2022-07-25
port /teamstatus command from https://github.com/enneract/AA-QVM/

cu-kai:
This pull request implements the `/teamstatus` command in sgame, verbatim from Der Bunker (Tremulous).

image

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -3499,6 +3499,116 @@ void Cmd_Reload_f( gentity_t *ent )
 
 /*
 =================
+Cmd_TeamStatus_f
+=================
+*/
+void Cmd_TeamStatus_f(gentity_t * ent)
+{
+	char multiomrc[12];
+	int builders = 0;
+	int arm = 0, mediboost = 0, miner = 0;
+	int omrccount = 0, omrchealth = 0;
+	bool omrcbuild = false;
+	gentity_t *tmp;
+	int i;
+
+	if (!g_teamStatus.Get()) 
+	{
+		trap_SendServerCommand(ent - g_entities, "print \"teamstatus is disabled.\n\"");
+		return;
+	}
+
+	if (ent->client->pers.namelog->muted) 
+	{
+		trap_SendServerCommand(ent - g_entities, "print \"You are muted and cannot use message commands.\n\"");
+		return;
+	}
+
+	if (ent->client->pers.lastTeamStatus &&
+		(level.time - ent->client->pers.lastTeamStatus) <
+	    g_teamStatus.Get() * 1000) 
+	{
+		trap_SendServerCommand(ent - g_entities, va("print \"You may only check your team's status once every %i seconds.\"", g_teamStatus.Get()));
+		return;
+	}
+
+	ent->client->pers.lastTeamStatus = level.time;
+
+	tmp = &g_entities[0];
+	for (i = 0; i < level.num_entities; i++, tmp++) 
+	{
+		if (i < MAX_CLIENTS) {
+			if (tmp->client &&
+			    tmp->client->pers.connected == CON_CONNECTED &&
+			    tmp->client->pers.team ==
+			    ent->client->pers.team && 
+				tmp->entity->Get<HealthComponent>()->Alive()
+			    && (tmp->client->ps.stats[STAT_CLASS] ==
+				PCL_ALIEN_BUILDER0
+				|| tmp->client->ps.stats[STAT_CLASS] ==
+				PCL_ALIEN_BUILDER0_UPG
+				|| BG_InventoryContainsWeapon(WP_HBUILD, tmp->client->ps.stats)))
+				builders++;
+			continue;
+		}
+
+		if (tmp->s.eType == entityType_t::ET_BUILDABLE) 
+		{
+			HealthComponent* health = tmp->entity->Get<HealthComponent>();
+			if (tmp->buildableTeam != ent->client->pers.team || !health->Alive())
+				continue;
+
+			switch (tmp->s.modelindex) {
+			case BA_H_REACTOR:
+			case BA_A_OVERMIND:
+				omrccount++;
+				if (health && health->Health() > omrchealth)
+					omrchealth = health->Health();
+				if (!omrcbuild)
+					omrcbuild = tmp->spawned;
+				break;
+			case BA_H_ARMOURY:
+				arm++;
+				break;
+			case BA_H_MEDISTAT:
+			case BA_A_BOOSTER:
+				mediboost++;
+				break;
+			case BA_H_DRILL:
+			case BA_A_LEECH:
+				miner++;
+				break;
+			default:
+				break;
+			}
+		}
+	}
+
+	if (omrccount > 1)
+		Com_sprintf(multiomrc, sizeof(multiomrc), "^7[x%d]", omrccount);
+	else
+		multiomrc[0] = '\0';
+	
+	if (ent->client->pers.team == TEAM_ALIENS) 
+	{
+		G_Say(ent, SAY_TEAM,
+		      va("^3[overmind]: %s(%d)%s ^3Spawns: ^5%d ^3Builders: ^5%d ^3Boosters: ^5%d ^3Leeches: ^5%d",
+		    	(!omrccount) ? "^1Down" : (omrcbuild) ? "^2Up" : // OM health logic
+		    	"^5Building", omrchealth * 100 / BG_Buildable( BA_A_OVERMIND )->health, multiomrc, // OM health logic part 2
+				level.team[ TEAM_ALIENS ].numSpawns, builders, mediboost, miner)); // spawns, builders, booster, leech
+	} 
+	else 
+	{
+		G_Say(ent, SAY_TEAM,
+		      va("^3[reactor]: %s(%d)%s ^3Spawns: ^5%d ^3Builders: ^5%d ^3Armouries: ^5%d ^3Medistations: ^5%d ^3Drills: ^5%d",
+		    	(!omrccount) ? "^1Down" : (omrcbuild) ? "^2Up" : // RC health logic
+		    	"^5Building", omrchealth * 100 / BG_Buildable( BA_H_REACTOR )->health, multiomrc, // RC health logic part 2
+				level.team[ TEAM_HUMANS ].numSpawns, builders, arm, mediboost, miner)); // spawns, builders, arm, medi, drill
+	}
+}
+
+/*
+=================
 G_StopFromFollowing
 
 stops any other clients from following this one
@@ -4471,6 +4581,7 @@ static const commands_t cmds[] =
 	{ "sell",            CMD_HUMAN | CMD_ALIVE,               Cmd_Sell_f             },
 	{ "setviewpos",      CMD_CHEAT_TEAM,                      Cmd_SetViewpos_f       },
 	{ "team",            0,                                   Cmd_Team_f             },
+	{ "teamstatus",      CMD_TEAM,                            Cmd_TeamStatus_f       },
 	{ "teamvote",        CMD_TEAM | CMD_INTERMISSION,         Cmd_Vote_f             },
 	{ "unignore",        0,                                   Cmd_Ignore_f           },
 	{ "vote",            CMD_INTERMISSION,                    Cmd_Vote_f             },
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -3525,14 +3525,14 @@ void Cmd_TeamStatus_f( gentity_t * ent )
 		return;
 	}
 
-	if ( ent->client->pers.lastTeamStatus
-		&& ( level.time - ent->client->pers.lastTeamStatus ) < g_teamStatus.Get() * 1000 ) 
+	if ( level.team[ ent->client->pers.team ].lastTeamStatus
+		&& ( level.time - level.team[ ent->client->pers.team ].lastTeamStatus ) < g_teamStatus.Get() * 1000 ) 
 	{
-		trap_SendServerCommand( ent - g_entities, va( "print \"You may only check your team's status once every %i seconds.\"", g_teamStatus.Get() ) );
+		trap_SendServerCommand( ent - g_entities, va( "print \"Your team's status may only be checked every %i seconds.\"", g_teamStatus.Get() ) );
 		return;
 	}
 
-	ent->client->pers.lastTeamStatus = level.time;
+	level.team[ ent->client->pers.team ].lastTeamStatus = level.time;
 
 	tmp = &g_entities[ 0 ];
 	for ( int i = 0; i < level.num_entities; i++, tmp++ ) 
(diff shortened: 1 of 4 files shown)


src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,116 @@ void Cmd_Reload_f( gentity_t *ent )
...
 }
 
+/*
+=================
+Cmd_TeamStatus_f
+=================
+*/
+void Cmd_TeamStatus_f(gentity_t * ent)
+{
+	char multiomrc[12];
```
DolceTriade, line 3507: Let's remove multiomrc support.

src/sgame/sg_struct.h
```diff
@@ -520,6 +520,7 @@ struct clientPersistant_t
 	// flood protection
 	int      floodDemerits;
 	int      floodTime;
+	int		 lastTeamStatus;
```
DolceTriade, line 523: Let's try to line this up.

DolceTriade (changes requested):
all in all, looks ok except for those nits.

necessarily-equal:
I'd use "Reactor" with all the letters instead of using the (outdated) emoticon. It'd just be clearer IMO

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,116 @@ void Cmd_Reload_f( gentity_t *ent )
...
+	    g_teamStatus.Get() * 1000) 
+	{
+		trap_SendServerCommand(ent - g_entities, va("print \"You may only check your team's status once every %i seconds.\"", g_teamStatus.Get()));
+		return;
+	}
+
+	ent->client->pers.lastTeamStatus = level.time;
+
+	tmp = &g_entities[0];
+	for (i = 0; i < level.num_entities; i++, tmp++)
```
bmorel, line 3538: use `for ( int i =` instead, it allows to have variables declared closer to where they are used, which heals  reading code.

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,116 @@ void Cmd_Reload_f( gentity_t *ent )
...
+		trap_SendServerCommand(ent - g_entities, va("print \"You may only check your team's status once every %i seconds.\"", g_teamStatus.Get()));
+		return;
+	}
+
+	ent->client->pers.lastTeamStatus = level.time;
+
+	tmp = &g_entities[0];
+	for (i = 0; i < level.num_entities; i++, tmp++) 
+	{
+		if (i < MAX_CLIENTS) {
```
bmorel, line 3540: we put the `{` and `}` on their own line.

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,116 @@ void Cmd_Reload_f( gentity_t *ent )
...
+			if (tmp->client &&
```
bmorel, line 3541: This is just me, maybe, but, to have the `&& on start of line makes later git changes less likely to conflict, and quick'n dirty tweaking like commenting a line are also less changes.
I'll let other talk on this one though.

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,116 @@ void Cmd_Reload_f( gentity_t *ent )
...
+			    tmp->client->pers.connected == CON_CONNECTED &&
+			    tmp->client->pers.team ==
+			    ent->client->pers.team && 
+				tmp->entity->Get<HealthComponent>()->Alive()
+			    && (tmp->client->ps.stats[STAT_CLASS] ==
+				PCL_ALIEN_BUILDER0
+				|| tmp->client->ps.stats[STAT_CLASS] ==
```
bmorel, line 3548: it could be nice to "cache" stuff, like:

```
const playerState* ps = tmp->client->ps;
if ( ps.stats[STAT_CLASS] ...
```

for readability

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,116 @@ void Cmd_Reload_f( gentity_t *ent )
...
+		{
+			HealthComponent* health = tmp->entity->Get<HealthComponent>();
+			if (tmp->buildableTeam != ent->client->pers.team || !health->Alive())
+				continue;
+
+			switch (tmp->s.modelindex) {
+			case BA_H_REACTOR:
+			case BA_A_OVERMIND:
+				omrccount++;
+				if (health && health->Health() > omrchealth)
```
bmorel, line 3565: We put `{}` even for single-line if blocks, it avoid various stupid mistakes like the infamous [apple goto fail](https://dwheeler.com/essays/apple-goto-fail.html)

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,116 @@ void Cmd_Reload_f( gentity_t *ent )
...
+/*
+=================
+Cmd_TeamStatus_f
+=================
+*/
+void Cmd_TeamStatus_f(gentity_t * ent)
+{
+	char multiomrc[12];
+	int builders = 0;
+	int arm = 0, mediboost = 0, miner = 0;
```
bmorel, line 3509: I would rather see an array of ints like I did in #1973, it makes things easier to extend, and requires less individual variables.

bmorel (commented):
This is just basic formatting stuff. Things are ok with me in their current state, but I did only a very quick read.

src/sgame/sg_cmds.cpp
bmorel: To be explicit:

```
			if ( tmp->client &&
￼			    tmp->client->pers.connected == CON_CONNECTED &&
￼			    tmp->client->pers.team ==
￼			    ent->client->pers.team && 
￼				tmp->entity->Get<HealthComponent>()->Alive()
￼```

would become:

```
			const clientPersistant_t &pers = tmp->client->pers;
			if ( tmp->client &&
￼				pers.connected == CON_CONNECTED &&
￼				pers.team == ent->client->pers.team && 
￼				tmp->entity->Get<HealthComponent>()->Alive()
```

Which is more readable.

File: src/sgame/sg_cmds.cpp [hunk 9]
```diff
@@ -3497,6 +3497,111 @@ void Cmd_Reload_f( gentity_t *ent )
...
+	ent->client->pers.lastTeamStatus = level.time;
+
+	tmp = &g_entities[ 0 ];
+	for ( int i = 0; i < level.num_entities; i++, tmp++ ) 
+	{
+		if ( i < MAX_CLIENTS ) 
+		{
+			if ( tmp->client &&
+			    tmp->client->pers.connected == CON_CONNECTED &&
+			    tmp->client->pers.team ==
```
bmorel, line 3542: You can use `G_OnSameTeam( tmp, pers )` here

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,111 @@ void Cmd_Reload_f( gentity_t *ent )
...
+			    ent->client->pers.team && 
+				tmp->entity->Get<HealthComponent>()->Alive()
+			    && (tmp->client->ps.stats[ STAT_CLASS ] ==
+				PCL_ALIEN_BUILDER0
```
bmorel, line 3546: I am not really fond of having the 2 sides of a same condition on different lines without reason. I would prefer `tmp->client->ps.stats[ STAT_CLASS ] == PCL_ALIEN_BUILDER0`

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,111 @@ void Cmd_Reload_f( gentity_t *ent )
...
+				case BA_A_LEECH:
+					miner++;
+					break;
+				default:
+					break;
+			}
+		}
+	}
+	
+	if ( ent->client->pers.team == TEAM_ALIENS )
```
bmorel, line 3587: Note that you can get a buildable's "humanname" by using i.e. `BG_Buildable( type )->humanname`. I could be used to make code here more generic.

slipher:
This ought to be considered as a gameplay proposal because having such detailed information reduces the fog of war quite a bit.

bmorel:
> This ought to be considered as a gameplay proposal because having such detailed information reduces the fog of war quite a bit.

Note that the PR includes a cvar to disable the feature. Maybe it could default to 0, indeed, but since the PR is rather simple and can even be disabled, I think it's ok by itself.

slipher:
> Note that the PR includes a cvar to disable the feature.

That makes little difference. The default value sets the standard that almost everyone will use.

bmorel:
I also said:

> Maybe it could default to 0

Which would fix your concern, I think?

cu-kai:
> > Note that the PR includes a cvar to disable the feature.
> 
> That makes little difference. The default value sets the standard that almost everyone will use.

Sorry yes, the default probably isn't sane and should be 0. That'll be in the next amendment. I also want to make the cooldown global to the team, and not specific to the player (as it is now), so only one person may request /teamstatus every <g_teamStatus> seconds.

DolceTriade:
imo, I've never really seen this feature abuse the fog of war aspect that much in my experience playing with it. Especially with an appropriate delay.

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,98 @@ void Cmd_Reload_f( gentity_t *ent )
...
+	int builders = 0;
+	gentity_t *tmp;
+	struct buildable_status_t
+	{
+		int count;
+		int health;
+		bool spawned;
+	} structures[ BA_NUM_BUILDABLES ] = {};
+
+	if ( !g_teamStatus.Get() )
```
bmorel, line 3516: This is an integer, so you likely want ` < 0` here.

cu-kai: done

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,98 @@ void Cmd_Reload_f( gentity_t *ent )
 	}
 }
 
+/*
```
bmorel, line 3500: Those "headers" are just really useless (they are even annoying when solving git conflicts, as sequences of `=` are used to show where the conflict goes to the 2nd side of the problem), we don't do them anymore, instead we write proper documentation for the functions.

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,98 @@ void Cmd_Reload_f( gentity_t *ent )
...
+		return;
+	}
+
+	if ( ent->client->pers.namelog->muted ) 
+	{
+		trap_SendServerCommand( ent - g_entities, "print \"You are muted and cannot use message commands.\n\"" );
+		return;
+	}
+
+	if ( level.team[ ent->client->pers.team ].lastTeamStatus
```
bmorel, line 3528: `ent->client->pers.team` -> `G_Team( ent )` is much more readable (and future-proof).
Also, you could do `int& lastTeamStatus = level.team[ G_Team( ent ) ].lastTeamStatus;` to make things much more readable.
For example,
`&& ( level.time - level.team[ ent->client->pers.team ].lastTeamStatus ) < g_teamStatus.Get() * 1000 )`
would become `&& ( level.time - lastTeamStatus ) < g_teamStatus.Get() * 1000 )` which could be compacted with previous line to give this:
`if ( lastTeamStatus && ( level.time - lastTeamStatus ) < g_teamStatus.Get() * 1000 )`

As you can see, the whole `if` expression can then stand on a single line and be shorter than the 2nd line of current one. 
￼

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,98 @@ void Cmd_Reload_f( gentity_t *ent )
...
+			}
+			continue;
+		}
+
+		if ( tmp->s.eType == entityType_t::ET_BUILDABLE ) 
+		{
+			int type = tmp->s.modelindex;
+			ASSERT( type < BA_NUM_BUILDABLES );
+			HealthComponent* health = tmp->entity->Get<HealthComponent>();
+			if ( G_OnSameTeam( tmp, ent ) && health->Health() > 0 )
```
bmorel, line 3560: `G_OnSameTeam( tmp, ent ) && health->Health() > 0` and above ` G_OnSameTeam( tmp, ent ) && ( health && health->Alive() )`
Those checks are the same, and could be made before the `if` which check the kind of entity, to simplify code.

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,98 @@ void Cmd_Reload_f( gentity_t *ent )
...
+	}
+
+	level.team[ ent->client->pers.team ].lastTeamStatus = level.time;
+
+	tmp = &g_entities[ 0 ];
+	for ( int i = 0; i < level.num_entities; i++, tmp++ ) 
+	{
+		if ( i < MAX_CLIENTS && tmp->client && tmp->entity ) 
+		{
+			auto health = tmp->entity->Get<HealthComponent>();
```
bmorel, line 3542: better to use `const auto&` since:

1. you do not modify it
2. no need to do a copy

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,98 @@ void Cmd_Reload_f( gentity_t *ent )
...
+				builders++;
+			}
+			continue;
+		}
+
+		if ( tmp->s.eType == entityType_t::ET_BUILDABLE ) 
+		{
+			int type = tmp->s.modelindex;
+			ASSERT( type < BA_NUM_BUILDABLES );
+			HealthComponent* health = tmp->entity->Get<HealthComponent>();
```
bmorel, line 3559: this can be made const too

src/sgame/sg_cmds.cpp
necessarily-equal: Get<> will return a pointer, so this is fine

sweet235:
About the fog of war: your team mates could just report what they know in team chat at any time. Good players were (almost) always the ones chatting a lot.

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,97 @@ void Cmd_Reload_f( gentity_t *ent )
...
+			}
+			continue;
+		}
+
+		if ( tmp->s.eType == entityType_t::ET_BUILDABLE ) 
+		{
+			int type = tmp->s.modelindex;
+			ASSERT( type < BA_NUM_BUILDABLES );
+			const HealthComponent* health = tmp->entity->Get<HealthComponent>();
+			if ( health->Health() > 0 )
```
necessarily-equal, line 3559: I would favor this as I'm not 100% sure every buildabe has a health component all the time
```suggestion
			if ( health && health->Health() > 0 )
```

necessarily-equal (approved):
LGTM

src/sgame/sg_cmds.cpp
```diff
@@ -3497,6 +3497,97 @@ void Cmd_Reload_f( gentity_t *ent )
...
+	for ( int i = 0; i < level.num_entities; i++, tmp++ ) 
+	{
+		if ( !G_OnSameTeam( tmp, ent ) )
+		{
+			continue;
+		}
+		
+		if ( i < MAX_CLIENTS && tmp->client && tmp->entity ) 
+		{
+			const auto& health = tmp->entity->Get<HealthComponent>();
```
necessarily-equal, line 3542:
```suggestion
			const HealthComponent* health = tmp->entity->Get<HealthComponent>();
```

Issue #1939, opened 2022-07-20
sgame crash after game decided


necessarily-equal:
Spotted on experimental. A dretch took some pulse rifle damage, and it seems it did so after the timelimit.

```gdb
Die: 5 1 MOD_LEVEL3_CLAW: [bot] Muxika killed [bot] Malo                                                                                                                
Die: 2 4 MOD_PRIFLE 3 2: [bot] Alistair killed [bot] Nolwen; [bot] Baudry assisted                                                                                      
Die: 3 6 MOD_PAINSAW 2 2: [bot] Baudry killed [bot] Itsaso; [bot] Alistair assisted                                                                                     
Broadcast: Empty teams skip map time exceeded.                                                                                                                          
                                                                                                                                                                        
Exit: Timelimit hit.                                                                
score: 157  ping: 69  client: 1 [bot] Malo                                                                                                                              
score: 150  ping: 64  client: 3 [bot] Baudry                                                                                                                            score: 140  ping: 91  client: 2 [bot] Alistair                                                                                                                          
score: 102  ping: 67  client: 5 [bot] Muxika                                                                                                                            
score: 87  ping: 51  client: 4 [bot] Nolwen                                                                                                                             
score: 78  ping: 76  client: 6 [bot] Itsaso                                                                                                                             
Die: 2 4 MOD_PRIFLE: [bot] Alistair killed [bot] Nolwen                                                                                                                 
]

Thread 38 "daemonded" received signal SIGSEGV, Segmentation fault. [Switching to Thread 0x7ffff32dc640 (LWP 67)]                                                                                                                           0x000000000000007b in ?? ()

Thread 38 (Thread 0x7ffff32dc640 (LWP 67) "daemonded"):
#0  0x000000000000007b in ?? ()
#1  0x00007ffff5877ade in Entity::SendMessage (data=0x7ffff32db150, msg=EntityMessage::Die, this=<optimized out>) at /build/source/src/sgame/backend/CBSEBackend.cpp:27
#2  Entity::Die (meansOfDeath=MOD_PRIFLE, killer=0x7ffff5418790, this=<optimized out>) at /build/source/src/sgame/backend/CBSEBackend.cpp:68
[...]

PR #1976, opened 2022-07-25
sgame: Stop running entity frames after the game ends.

DolceTriade:
Before, even after the game ended, we would still update clients and
such. This can lead to potentially bad states leading to a segfault if
something calls and entity method after the entity is cleared as in #1939.


src/sgame/sg_main.cpp
```diff
@@ -2359,6 +2363,18 @@ void G_RunFrame( int levelTime )
 	// now we are done spawning
 	level.spawning = false;
 
+	// cancel vote if timed out
+	for ( i = 0; i < NUM_TEAMS; i++ )
+	{
+		G_CheckVote( (team_t) i );
+	}
```
necessarily-equal, line 2370: where is this coming from?

DolceTriade: Ooops I forgot to remove it below...Nice catch.

necessarily-equal:
Haven't tested, code looks ok

PR #1979, opened 2022-07-26
Gravity ent fixes

### LINKED ISSUE #1978: trigger_gravity of 0 sets gravity to 800
To reproduce, run the map `mudkip` under gdb, reach the place in the following screenshot, and put a breakpoint in `void env_afx_gravity_touch( gentity_t *ent, gentity_t *other, trace_t* )`.

2022-07-26-123132_1366x768_scrot

The player's gravity is set to 800 when it should be set to 0. This is because this trigger get it's value reset to 800 by `void env_afx_gravity_reset( gentity_t *self )`, this I am not sure why yet.

bmorel:
fix #1978


bmorel:
Can be tested on [this map](https://github.com/UnvanquishedAssets/UnvanquishedTestAssets/tree/master/src/map-test-gravity0_src.dpkdir)

src/sgame/sg_spawn_afx.cpp
```diff
@@ -281,7 +281,9 @@ void SP_env_afx_gravity( gentity_t *self )
 
 	self->touch = env_afx_gravity_touch;
 	self->act = env_afx_toggle;
-	self->reset = env_afx_gravity_reset;
+	//FIXME 0.54 set the reset callback if and only if the property was missing
```
DolceTriade, line 284: What is the wrong required to fix this?

bmorel: I am very unsure about the reason *why* the reset was set and how it is cleared when gravity is not 0, so I'd rather have a comment saying there's a problem around, to give reviewers a hint that there might be a real problem behind the change.

illwieckz: I would write that as a comment in code.

bmorel: Oh, I remember about something else (this patch is several days old).
Here, value would be initialised to 0 if field is missing, but the specs (of quake3 and derivatives) says that in that case, it should be set to 800 (aka, default value, in our case, g_gravity or whatever).
The problem is, value "0" is a legit value, and I have no idea *how* to detect the field was actually missing (which means, the entity should default to 800 or g_gravity).
Hence the fixme

DolceTriade: According to the code, if `G_SpawnInt()` return false, the value is not set.

Therefore we can safely do

```
bool set = G_SpawnInt( "gravity", "0", &self->config.amount );
...
if (!set)
{
  self->reset = env_afx_gravity_reset;
}

bmorel: why not I guess, but without a map to test that, it can hardly be guaranteed to be working. OTOH, setting up a gravity field without value, which would hence default to normal value, I wonder what would be the point...

DolceTriade (approved):

illwieckz (approved):
LGTM

PR #2005, opened 2022-07-30
Merge the rebased 0.53.0/sync into master

DolceTriade:
This squashes a bunch of the initial rmlui commits into a single commit to aid in bisecting.


necessarily-equal:
`$ git diff upstream/0.53.0/sync upstream/ishq/rmlui/rebase` gives me this:

```patch
diff --git a/src/sgame/sg_bot_nav.cpp b/src/sgame/sg_bot_nav.cpp
index 72f930061..40b11ce77 100644
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -339,7 +339,7 @@ void BotWalk( gentity_t *self, bool enable )
 }
 
 // search for obstacle forward, and return pointer on it if any
-static const gentity_t* BotGetPathBlocker( gentity_t *self, const vec3_t dir )
+const gentity_t* BotGetPathBlocker( gentity_t *self, const vec3_t dir )
 {
 	vec3_t playerMins, playerMaxs;
 	vec3_t end;
@@ -369,7 +369,7 @@ static const gentity_t* BotGetPathBlocker( gentity_t *self, const vec3_t dir )
 
 // checks if jumping would get rid of blocker
 // return true if yes
-static bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const vec3_t dir )
+bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const vec3_t dir )
 {
 	vec3_t playerMins;
 	vec3_t playerMaxs;
diff --git a/src/sgame/sg_spawn_mover.cpp b/src/sgame/sg_spawn_mover.cpp
index 629f864e8..3749cba59 100644
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -531,21 +531,21 @@ SetMoverState
 ===============
 */
 
-			// most elevators are *not* func_door, so that broken.
-			// Some (one on forlorn) is a func_door, with a "targetname",
-			// and I suspect this is what makes it different from doors.
-			// That distinction means that, doors with a targetname will
-			// break for bots.
-			// Also, for now, doors without targetname work because bots
-			// just walk to kiss (or worse!) them, until they open wide
-			// for them to enter.
-			// The correct solution here would be to generate navcons
-			// for all movers, with labels indicating the bot how to
-			// handle the situation correctly (including waiting patiently
-			// that elevator reached desired navmesh).
-			// Also, this code is currently seemlingly inverted.
+// most elevators are *not* func_door, so that broken.
+// Some (one on forlorn) is a func_door, with a "targetname",
+// and I suspect this is what makes it different from doors.
+// That distinction means that, doors with a targetname will
+// break for bots.
+// Also, for now, doors without targetname work because bots
+// just walk to kiss (or worse!) them, until they open wide
+// for them to enter.
+// The correct solution here would be to generate navcons
+// for all movers, with labels indicating the bot how to
+// handle the situation correctly (including waiting patiently
+// that elevator reached desired navmesh).
+// Also, this code is currently seemlingly inverted.
 bool IsDoor( const gentity_t *ent )
-			{
+{
 	const char * funcs[] =
 	{
 		"func_door",
@@ -558,7 +558,7 @@ bool IsDoor( const gentity_t *ent )
 		if ( !Q_stricmp( fn, ent->classname ) )
 		{
 			return true;
-			}
+		}
 	}
 	return false;
 }
@@ -597,17 +597,17 @@ bool IsAutomaticMover( const gentity_t *ent )
 void BotHandleDoor( gentity_t *ent )
 {
 	if ( IsDoor( ent ) && !IsAutomaticMover( ent ) )
-			{
-				if ( ent->obstacleHandle )
-				{
-					G_BotRemoveObstacle( ent->obstacleHandle );
-					ent->obstacleHandle = 0;
-				}
+	{
+		if ( ent->obstacleHandle )
+		{
+			G_BotRemoveObstacle( ent->obstacleHandle );
+			ent->obstacleHandle = 0;
+		}
 		G_BotAddObstacle( ent->r.absmin, ent->r.absmax, &ent->obstacleHandle );
-			}
+	}
 }
 
-static void SetMoverState( gentity_t *ent, moverState_t moverState, int time )
+void SetMoverState( gentity_t *ent, moverState_t moverState, int time )
 {
 	vec3_t delta;
 	float  f;
```

It seems harmless enough, but it's still a bit weird

illwieckz:
This is a commit being mistakenly reverted. I got this conflict when merging `master` to `0.53.0/sync` when the file was modified both in `master` and in `0.53.0/sync`.

DolceTriade:
Ok, I messed up in the conflict resolution here. I'll redo this :(

DolceTriade:
The rebase appears to add more problems than it fixes. I'll drop this.

illwieckz:
😥️

PR #2015, opened 2022-08-01
Max miners

sweet235:


src/sgame/sg_buildable.cpp
```diff
@@ -1702,6 +1702,32 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
 		level.numBuildablesForRemoval = 0;
 	}
 
+	gentity_t *tmp = g_entities;
+	int miner = 0;
+	for (int i = 0; i < level.num_entities; i++, tmp++)
```
DolceTriade: Can rewrite this as:

ForEntities<MiningComponent>([](Entity& entity, MiningComponent& mining) {});

src/sgame/sg_buildable.cpp
```diff
@@ -1702,6 +1702,32 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
...
+			case BA_A_LEECH:
+				miner++;
+				break;
+			default:
+				break;
+			}
+		}
+	}
+	int max_miners = g_maxMiners.Get();
+	if (max_miners > 0 && miner >= max_miners)
```
DolceTriade: I think this should be miner > max_miners?

src/sgame/sg_buildable.cpp
DolceTriade: This is an optional change though

DolceTriade: If you choose to not rewrite as a `ForEntities` loop, then you can also start i = MAX_CLIENTS, to save 64 iterations per call since entityNums 0-63 are reserved for players.

DolceTriade (commented):
overall looks good, just minor nits.

slipher:
- It seems wrong that the limit applies to both teams' miners added together. This means that one team could build a room full of miners to prevent the other team from building any (could be a good strategy for aliens).
- G_CanBuild needs to check that the building someone is trying to build is a miner.
- There should be a different error message for this instead of out of BP.

DolceTriade:
It checks the team before counting it (but it'll count either leech or drill for both teams).

src/sgame/sg_buildable.cpp
```diff
@@ -1702,6 +1702,35 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
...
+	if (max_miners > 0)
+	{
+		gentity_t *tmp = g_entities;
+		int miners = 0;
+		for (int i = 0; i < level.num_entities; i++, tmp++) 
+		{
+			if (tmp->s.eType == entityType_t::ET_BUILDABLE) 
+			{
+				HealthComponent* health = tmp->entity->Get<HealthComponent>();
+				if (tmp->buildableTeam != ent->client->pers.team || !health->Alive())
```
DolceTriade, line 1715: Also check for null healthcomponent here to be safe.

slipher: There's `Entities::IsAlive` for this

src/sgame/sg_buildable.cpp
sweet235: I tried miner > max_miners first, but testing reveals miner >= max_miners to do what I intent.

sweet235:
G_CanBuild() is called at 3 locations in the program (if I count correctly). One of these seems to be responsible for deciding whether to draw the "skeleton" in green/yellow/red. I think this call must be made once for every time frame.

G_CanBuild might be in O(1) time without my changes (is it?). The loop over all entities pushes it into O(# of entities). This might introduce performance problems. In commit 002e9cac9, I made sure the loop is only executed when g_maxMiners is non-zero. That should be good enough for an experimental optional feature.

But the clean way to do this would be to do the counting loop only once per time frame. This way, is results can be shared with https://github.com/cu-kai/Unvanquished/tree/cu-kai/teamstatus, and possibly other features. Where would be the correct way to place this loop?

sweet235:
>     * G_CanBuild needs to check that the building someone is trying to build is a miner.
> 
>     * There should be a different error message for this instead of out of BP.

Both points crossed my mind too, since I made the PR.

The error message is less than optimal indeed. However, I'd rather test this with other players before putting effort in adding another error message for something we might dump soon. The error messages could be worse.

I too think that two builders can go above g_maxMiners if they both build a miner in the same time frame. I'm not sure how to resolve this cleanly. Yet.

src/sgame/sg_buildable.cpp
bmorel, line 1715: Codestyle (hence, not blocking):

inconsistent with next if, and usually the old style files use {} even for one-line blocks, {} being on their own lines. I write *usually* because in this simple function there's already 2 different styles, but the one you're going to include would be a 3rd one (no {}) in this function.

`tmp->buildableTeam != ent->client->pers.team` -> `G_OnSameTeam` is what you want here here, I think.
Also, I don't really like double negations (matter of taste, really): I think `if ( !( G_OnSameTeam( tmp, ent ) && IsAlive( tmpt ) ) ){ continue; }` might be more readable, despite the additional parens.

src/sgame/sg_main.cpp
```diff
@@ -213,6 +213,8 @@ Cvar::Cvar<float>  g_devolveMaxBaseDistance("g_devolveMaxBaseDistance", "Max Ove
 
 Cvar::Cvar<bool>   g_autoPause("g_autoPause", "pause empty server", Cvar::NONE, false);
 
+Cvar::Cvar<int> g_maxMiners("g_maxMiners", "set maximum number of miners per team. 0 = disabled.", Cvar::NONE, 0);
```
bmorel, line 216: Since using an int, maybe use `0` for exactly that: 0 miners allowed, and negative value in cvar to disable the feature?

sweet235: If there is any use for a game without miners, which I assume: you are right.

necessarily-equal:
I'd say don't worry too much about performance, this is O(ent) and there are loops like that all over. I'd rather do caching in some O(ent²) loop instead.

If you really want to do it though, there is `level.team[ TEAM_x ].numSpawns` that you could extend with every buildable type

src/sgame/sg_buildable.cpp
```diff
@@ -1702,6 +1702,40 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
 		level.numBuildablesForRemoval = 0;
 	}
 
+	int max_miners = g_maxMiners.Get();
+	if (max_miners >= 0)
```
bmorel, line 1706: Since yuo did the `< 0` => disabled, this needs to be `max_miners > 0` and if `==0` always return `IBE_NOHUMANBP : IBE_NOALIENBP`

sweet235: That would yield the same results, and only optimize the case max_miners=0. I think it's fine without that optimization right now.

bmorel:
As a reminder, this PR should probably be squashed before merge

sweet235:
> I'd say don't worry too much about performance, this is O(ent) and there are loops like that all over. I'd rather do caching in some O(ent²) loop instead.

After some thinking, i believe that my G_CanBuild()'s influence on time is in O(builders * ents). This is somewhere between O(ents) and O(ents²).

src/sgame/sg_buildable.cpp
```diff
@@ -1702,6 +1702,40 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
...
+		// maybe rewrite loop as
+		//   ForEntities([](Entity& entity, MiningComponent& mining) {});
+		gentity_t *tmp = g_entities + MAX_CLIENTS;
+		int miners = 0;
+		for (int i = MAX_CLIENTS; i < level.num_entities; i++, tmp++) 
+		{
+			if (tmp->s.eType == entityType_t::ET_BUILDABLE) 
+			{
+				HealthComponent* health = tmp->entity->Get<HealthComponent>();
+				if (!G_OnSameTeam(tmp, ent) || !health->Alive() || !Entities::IsAlive(tmp))
```
necessarily-equal, line 1717:
```suggestion
				if (!G_OnSameTeam(tmp, ent) || !health->Alive() || !Entities::IsAlive(tmp))
```
```suggestion
				if (!G_OnSameTeam(tmp, ent) || !Entities::IsAlive(tmp))
```
and you can remove the line above. IsAlive uses the healthcomponent and doesn't crash if it's missing

src/sgame/sg_buildable.cpp
necessarily-equal: I agree with Sweet here

slipher:
I still don't see the check that the buildable the player is trying to construct is a miner? The miner counting stuff runs if you are trying to build an egg for example.

src/sgame/sg_buildable.cpp
sweet235: I removed that.

sweet235:
> I still don't see the check that the buildable the player is trying to construct is a miner? The miner counting stuff runs if you are trying to build an egg for example.

This was a serious deficiancy indeed, i fixed that in [18ece92](https://github.com/Unvanquished/Unvanquished/pull/2015/commits/18ece92d22a9946cda5332d06190127b9f4abb0f).

DolceTriade (approved):

src/sgame/sg_buildable.cpp
```diff
@@ -1702,6 +1702,23 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
 		level.numBuildablesForRemoval = 0;
 	}
 
+	int max_miners = g_maxMiners.Get();
+	if ( max_miners >= 0 && ( buildable == BA_H_DRILL || buildable == BA_A_LEECH ) )
+	{
+		int miners = 0;
+		ForEntities<MiningComponent> ( [&](Entity& entity, MiningComponent& mining )
```
slipher, line 1709: Delete the variable name `mining` here to avoid an unused variable warning.

LGTM

sweet235: I'm not sure how to do that...

slipher: Literally delete the word `mining`

src/sgame/sg_buildable.cpp
```diff
@@ -1702,6 +1702,23 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
 		level.numBuildablesForRemoval = 0;
 	}
 
+	int max_miners = g_maxMiners.Get();
+	if ( max_miners >= 0 && ( buildable == BA_H_DRILL || buildable == BA_A_LEECH ) )
```
bmorel, line 1706: I really think `if ( max_miners > 0` would be better here, no need to do anything if no miner is allowed anyway.

src/sgame/sg_buildable.cpp
```diff
@@ -1702,6 +1702,23 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
...
+		ForEntities<MiningComponent> ( [&](Entity& entity, MiningComponent& )
+		{
+			if ( Entities::IsAlive(entity) && G_OnSameTeam( entity.oldEnt, ent ) )
+			{
+				miners++;
+			}
+		});
+		if ( miners >= max_miners )
+		{
+			return ent->client->pers.team == TEAM_HUMANS ? IBE_NOHUMANBP : IBE_NOALIENBP;
```
bmorel, line 1718: just a note for later, probably worth a new issue (that means: I don't mind if this is not fixed, but I'd prefer a `TODO 0.54 bla bla` comment): IBE_NO...BP does not suits well the real error.

bmorel (approved):

bmorel:
LGTM

src/sgame/sg_buildable.cpp
slipher: We use the `TODO(0.54)` notation for things that have to go in the next release due to compatibility breakage. We decided to allow changing the cgame<>sgame protocol at any time, so that does not apply.

src/sgame/sg_buildable.cpp
sweet235: If we do nothing if no miner is allowed, we do in fact allow an infinite number of miners.

PR #2017, opened 2022-08-02
adds cg_drawPosition: position indicator for the HUD

bmorel:
requires https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/144 to take effect

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -917,6 +917,46 @@ private:
 	Color::Color backColor;
 };
 
+class PositionElement : public HudElement
+{
+public:
+	PositionElement( const Rml::String& tag ) :
+			HudElement( tag, ELEMENT_GAME, true )
+	{
+		Rml::XMLAttributes xml;
+		currentPositionElement = dynamic_cast< Rml::ElementText* >( AppendChild( Rml::Factory::InstanceElement(
+			this,
+			"#text",
+			"span",
+			xml) ) );
+		currentPositionElement->SetClass( "position_current", true );
+	}
+
+	void DoOnRender() override
+	{
+		if ( !cg_drawPosition.Get() )
+		{
+			currentPositionElement->SetText( "" );
+		}
+		else
+		{
+			if ( cg.predictedPlayerState.clientNum == cg.clientNum )
+			{
+				// Add text to be configured via CSS
+				vec3_t origin;
+				VectorCopy( cg.predictedPlayerState.origin, origin );
+
+				// HACK: Put extra spaces to separate the children because setting them to block makes them disappear.
+				// TODO: Figure out why setting these two elements to block makes them disappear.
+				currentPositionElement->SetText( va( "%0.0f   %0.0f   %0.0f", origin[0], origin[1], origin[2] ) );
+			}
+		}
+	}
+
+private:
+	Rml::ElementText* currentPositionElement;
+};
+
 class CreditsValueElement : public TextHudElement
 {
 public:
@@ -3637,6 +3677,7 @@ void CG_Rocket_RegisterElements()
 	RegisterElement<CrosshairIndicatorHudElement>( "crosshair_indicator" );
 	RegisterElement<CrosshairHudElement>( "crosshair" );
 	RegisterElement<SpeedGraphElement>( "speedometer" );
+	RegisterElement<PositionElement>( "position_indicator" );
 	RegisterElement<CreditsValueElement>( "credits" );
 	RegisterElement<EvosValueElement>( "evos" );
 	RegisterElement<WeaponIconElement>( "weapon_icon" );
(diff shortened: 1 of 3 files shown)


src/cgame/cg_rocket_draw.cpp
```diff
@@ -917,6 +917,46 @@ class SpeedGraphElement : public HudElement
 	Color::Color backColor;
 };
 
+class PositionElement : public HudElement
```
DolceTriade, line 920: Use TextHudElement

bmorel: What is the difference? Should it also be applied to SpeedGraphElement? I guess not, since it can (only in option 3 IIRC) enable a graph, but could you details the why?

necessarily-equal: That would be because it avoids code such as

```C++
currentPositionElement = dynamic_cast< Rml::Core::ElementText* >( AppendChild( Rml::Core::Factory::InstanceElement(
			this,
			"#text",
			"span",
			xml) ) );
```

DolceTriade: Yes, it avoid the boiler plate of having to instance and append a new child. You can just call SetText() internally.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -3631,6 +3671,7 @@ void CG_Rocket_RegisterElements()
 	RegisterElement<CrosshairIndicatorHudElement>( "crosshair_indicator" );
 	RegisterElement<CrosshairHudElement>( "crosshair" );
 	RegisterElement<SpeedGraphElement>( "speedometer" );
+	RegisterElement<PositionElement>( "position_indicator" );
```
DolceTriade, line 3674: imo, we should call this "player_origin". Position indicator is not correct because it's not an indicator. And since this is more of a debug element, we can be more technical about the names.

bmorel: player_origin talks less to me, even if that's technically true.

bmorel:
Draft because of duplicated code.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -917,6 +917,36 @@ class SpeedGraphElement : public HudElement
...
+		else
+		{
+			if ( cg.predictedPlayerState.clientNum == cg.clientNum )
+			{
+				// Add text to be configured via CSS
+				vec3_t origin;
+				VectorCopy( cg.predictedPlayerState.origin, origin );
+
+				// HACK: Put extra spaces to separate the children because setting them to block makes them disappear.
+				// TODO: Figure out why setting these two elements to block makes them disappear.
```
DolceTriade, line 943: Please fix or remove todos if you're satisfied with the current result.

DolceTriade (approved):

src/cgame/cg_rocket_draw.cpp
bmorel: I actually duplicated code, and merging this would increase tech debt. I don't think it's a good idea.

bmorel: Fun fact, it was actually no longer useful, code works nicely *without* those. I also removed the useless copy...

src/cgame/cg_rocket_draw.cpp
```diff
@@ -917,6 +917,31 @@ class SpeedGraphElement : public HudElement
...
+
+	void DoOnRender() override
+	{
+		if ( !cg_drawPosition.Get() )
+		{
+			SetText( "" );
+			return;
+		}
+
+		if ( cg.predictedPlayerState.clientNum == cg.clientNum )
```
necessarily-equal, line 936: tbh I don't see why we should hide this (and the speedometer) when following someone else, for example a bot. I would just remove the condition and activate it always, but shrug.

necessarily-equal (approved):
LGTM

DolceTriade:
LGTM

DolceTriade (approved):

src/cgame/cg_rocket_draw.cpp
illwieckz: Hmm, I also see no reason to hide something when spectating.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -917,6 +917,31 @@ class SpeedGraphElement : public HudElement
...
 
+class PositionElement : public TextHudElement
+{
+public:
+	PositionElement( const Rml::String& tag ) :
+			TextHudElement( tag, ELEMENT_GAME, true )
+	{
+	}
+
+	void DoOnRender() override
```
slipher, line 928: DoOnRender should only be used for elements that send drawing commands directly to the renderer, bypassing Rmlui rendering. This should be DoUnUpdate

bmorel:
Before someone asks me, yes, this is obviously to be squashed before merge.

bmorel:
@slipher , @illwieckz , @necessarily-equal , I think all your remarks has been addressed.

PR #2032, opened 2022-08-04
Show oxygen bar only when the followed player is underwater

necessarily-equal:
It used to be shown when anyone was underwater

PR #2034, opened 2022-08-04
Water fix

bmorel:
supersedes #2032 by adding the missing part of the patch

### DIFF
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -956,14 +956,20 @@ void CG_EntityEvent( centity_t *cent, vec3_t position )
 			break;
 
 		case EV_WATER_UNDER:
-			cg.underwater = true;
-			cg.underwaterTime = cg.time;
+			if ( clientNum == cg.clientNum )
+			{
+				cg.underwater = true;
+				cg.underwaterTime = cg.time;
+			}
 			trap_S_StartSound( nullptr, es->number, soundChannel_t::CHAN_AUTO, cgs.media.watrUnSound );
 			break;
 
 		case EV_WATER_CLEAR:
-			cg.underwater = false;
-			cg.underwaterTime = cg.time;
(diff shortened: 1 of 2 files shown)


necessarily-equal:
I mean if you're going to nuke the progressive resplenishing you might as well do it fully. Here's a patch with extra things you can remove.

```diff
static float CG_Rocket_GetPlayerOxygenProgress()
{
-	// Hack to show prettier output by flowing the bar back up for a short
-	// time. It doesn't correspond to anything in the game physics.
-	static constexpr int breathAgainDuration = 350; // ms
-	static float lastValue = 1.0f;
-
-	// regular progress bar status
	if ( cg.underwater )
	{
		float underwaterDuration = cg.time - cg.underwaterTime;
-		lastValue = 1.0f - underwaterDuration / static_cast<float>( OXYGEN_MAX_TIME );
-		return lastValue;
+		return 1.0f - underwaterDuration / static_cast<float>( OXYGEN_MAX_TIME );
	}
-
	return 0.0f;
}
```

And you can probably say so in the commit message. e.g: "Remove the progressive oxygen regen display because it somehow makes the previous commit work".

(I'm still quite surprised it would make the previous commit work, but well working is better than pretty)

bmorel:
>  "Remove the progressive oxygen regen display because it somehow makes the previous commit work".

No, because I believe those commits shuold be squashed anyway, they don't make sense individually to me.

slipher:
This is presumably still broken because it doesn't take into account switching followed player while spectating. We need to remove the oxygen bar for 0.53.1.

bmorel:
> This is presumably

hmmm

necessarily-equal:
This isn't shown when spectating for the record. To show it when spectating one would need to use cg.predictedPlayerState.clientNum instead of cg.clientNum

Le 4 août 2022 20:38:09 GMT+02:00, slipher ***@***.***> a écrit :

PR #2042, opened 2022-08-05
Revert "make slowblob for advgrangers lockblob"

cu-kai:
This reverts the absurd decision to change granger spit from slowblobs into lockblobs. This gameplay mechanic is **_broken_** and imo should never have been approved in the first place.

### DIFF
--- a/src/sgame/sg_missile.cpp
+++ b/src/sgame/sg_missile.cpp
@@ -252,7 +252,6 @@ static int ImpactSlowblob( gentity_t *ent, trace_t *trace, gentity_t *hitEnt )
 {
 	int       impactFlags = MIB_IMPACT;
 	gentity_t *attacker = &g_entities[ ent->r.ownerNum ];
-	vec3_t dir;
 	int reward = 0;
 
 	// put out fires on direct hit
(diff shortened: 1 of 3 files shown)


cu-kai:
FYI, changes will be required in https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir such as removing the attribute from configs/weapon/abuild.attr.cfg

Gireen:
The lockblob is really the creation of a madman.  
He must have been completely bonkers adding such a game breaking functions.
The game has become literally unplayable. Servers are only filled with hordes of grangers abusing this cheat like function.
Human player are forced to watch their teammates being destroyed by this savage beasts without any possibility to help.
There is clearly a lack of tools for humans to support each other form afar and kill enemies with some weapon like items.
😄

To prevent such horrible events from happening again we should first discuss gamplay problems, especially controversial topics, in the related issue tracker. 
Freem has already opened an issue for this unvanquished/gameplay#48

DolceTriade:
Was a fun experiment :D

bmorel:
LGTM, but do all other devs agree on the idea of removing this?

DolceTriade:
I'm on board.

bmorel:
Anyone *against* this change?

PR #2045, opened 2022-08-06
Welcome menu

bmorel:
Allows to show a popup when a player connects to a server.
Taken from @cu-kai 's work and added a commit to clean the ugly switch.

This requires changes in unvanquished_src.dpkdir to work.

It also conflicts with @necessarily-equal 's #1696

### DIFF
--- a/src/cgame/cg_servercmds.cpp
+++ b/src/cgame/cg_servercmds.cpp
@@ -415,6 +415,7 @@ static void CG_Menu( int menuType, int arg )
 	switch ( menuType )
 	{
 		case MN_WELCOME:
+			Rocket_DocumentAction( "changelog", "show" );
 			break;
 
 		case MN_TEAM:
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -2257,6 +2257,27 @@ void ClientThink( int clientNum )
 	{
 		ClientThink_real( ent );
 	}
+
+	// ugly hack alert, this cannot run in ClientBegin() as it runs too early, and the menu doesn't show.
+	// display the help menu, if connecting the first time
+	if ( !ent->client->sess.seenWelcome 
+	     && ( ent->client->pers.enterTime + 250 < level.time ) )
+	{
+		ent->client->sess.seenWelcome = 1;
+
+		// 0 - don't show
+		// 1 - always show to all
+		// 2 - show only to unregistered
+		switch ( g_showHelpOnConnection.Get() )
+		{
+		case 0:
+			if (0)
+		default:
+			if ( !ent->client->pers.admin )
+		case 1:
+			G_TriggerMenu( ent->client->ps.clientNum, MN_WELCOME );
+		}
+	}
 }
 
 void G_RunClient( gentity_t *ent )
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -2265,16 +2265,9 @@ void ClientThink( int clientNum )
 	{
 		ent->client->sess.seenWelcome = 1;
 
-		// 0 - don't show
-		// 1 - always show to all
-		// 2 - show only to unregistered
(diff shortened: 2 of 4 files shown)


slipher:
A pop-up every time you join the server sounds pretty annoying...

bmorel:
I agree it would be nice to let clients close those and remember about it until the server includes changes, but I have no idea how this would work.

slipher:
At least turn it off by default so that it could be used only if a mod wanted to write something.

src/cgame/cg_servercmds.cpp
```diff
@@ -415,6 +415,7 @@ static void CG_Menu( int menuType, int arg )
 	switch ( menuType )
 	{
 		case MN_WELCOME:
+			Rocket_DocumentAction( "changelog", "show" );
```
DolceTriade, line 418: Let's not hardcode these things into the game and have them be configurable.

Referring to "changelog"

src/sgame/sg_active.cpp
```diff
@@ -2257,6 +2257,20 @@ void ClientThink( int clientNum )
 	{
 		ClientThink_real( ent );
 	}
+
```
DolceTriade, line 2260: Can we instead fix the root cause of why opening the menu in ClientBegin doesn't work?

src/sgame/sg_main.cpp
```diff
@@ -37,7 +37,7 @@ level_locals_t level;
 gentity_t          *g_entities;
 gclient_t          *g_clients;
 
-Cvar::Range<Cvar::Cvar<int>> g_showHelpOnConnection("g_showHelpOnConnection", "show MN_WELCOME on connect (1 = all, 2 = unregistered)", Cvar::NONE, 1, 0, 2);
+Cvar::Range<Cvar::Cvar<int>> g_showHelpOnConnection("g_showHelpOnConnection", "show MN_WELCOME on connect (1 = all, 2 = unregistered)", Cvar::NONE, 0, 0, 2);
```
DolceTriade, line 40: We should rename this from g_showHelpOnConnection to something more reasonable like g_showWelcome.

We can track state like whether someone has seen it (track a generation or something) in a separate location (admin.dat or something...)

DolceTriade (changes requested):

src/sgame/sg_active.cpp
DolceTriade: The reason that this doesn't work in ClientBegin doesn't work is that when the client connection state becomes `active`, we hide all the windows. We should fix that.

src/cgame/cg_servercmds.cpp
bmorel: I don't really think we want everything to be configurable.
What would be the benefits to change this?
It would only make support harder when someone changes it, and comes saying that their stuff do not work because they typo'ed somewhere.

src/sgame/sg_active.cpp
bmorel: I'd also prefer to fix the root cause, but I don't know much the cgame side of the codebase. As I said, I only stole the original commit and applied a cleanup on the ugly switch() case, because I think this feature is highly desirable.

src/sgame/sg_main.cpp
bmorel: This seems overall out of the scope of this PR, which only tries to repair a feature that existed.

> We should rename this from g_showHelpOnConnection to something more reasonable like g_showWelcome.

I usually try to avoid changing an interface (especially when it's something people can have tweaked, and won't have any way to know they're using a cvar that no longer exists!), even if I do agree with you a lot on the name there.

> We can track state like whether someone has seen it (track a generation or something) in a separate location (admin.dat or something...)

feels out of scope to me.

bmorel:
replaced by https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/151

PR #2051, opened 2022-08-07
welcome: Fix showing welcome window on connection.

DolceTriade:
This used to be g_showHelpOnConnection, but that cvar is too long, so just call it g_showWelcome. Also, with the old code, it would come at the same frame that we go to CA_ACTIVE, which meant that rmlui would hide all the windows. However, since, PRIMED happens just before we get the first data from the server, it makes sense to hide all the menus then so we don't hide any commands we receive from the server.

### DIFF
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -39,7 +39,7 @@ level_locals_t level;
 gentity_t          *g_entities;
 gclient_t          *g_clients;
 
-Cvar::Range<Cvar::Cvar<int>> g_showHelpOnConnection("g_showHelpOnConnection", "show MN_WELCOME on connect (1 = all, 2 = unregistered)", Cvar::NONE, 1, 0, 2);
+Cvar::Range<Cvar::Cvar<int>> g_showWelcome("g_showWelcome", "show MN_WELCOME on connect (0 = none, 1 = all, 2 = unregistered)", Cvar::NONE, 0, 0, 2);
 
 Cvar::Range<Cvar::Cvar<int>> g_gravity(
 		"g_gravity",
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -1275,6 +1275,16 @@ struct cg_t
 	Filter<WeaponOffsets>   weaponOffsetsFilter;
 };
 
+enum rocketElementType_t
+{
+	ELEMENT_ALL,
+	ELEMENT_GAME,
+	ELEMENT_LOADING,
+	ELEMENT_HUMANS,
+	ELEMENT_ALIENS,
+	ELEMENT_BOTH,
+};
+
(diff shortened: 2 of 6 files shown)


bmorel:
Few things we discussed on IRC:

* since this breaks daemon's API, it should go into a 0.54/sync branch
* the hope AFAIK is to move rocket's related stuff out of daemon, since UI stuff have nothing to do in engine  (from current consensus)

Also this comment: since this requires a 0.54 release, I do not think it invalidates #2045. Could not we simply accept #2045 for master, to provide the feature to modders, to people willing to run master, or to people that simply tweak some cvars, and get the proper fix done in next release? This does not seems like there would be too many conflicts to do this.

DolceTriade:
For the record, the only reason that this requires an API change is the inclusion of the MN_WELCOME define. I'd still prefer we fix the bug the way I did here, even if we have to hardcode some stuff.

bmorel:
The problem is not about hardcoding, it's about changing daemon's API. Currently, #2045 is used in one mod, and soon I'll use it in my own mod. Considering the fact there's a 3rd server embedding rather recent changes, it's likely it's owner would also appreciate the feature.
I'm not saying the proper fix should not be integrated, but I say it should be only integrated for 0.54, and that in the meantime, the less proper change is more desirable than current situation.

src/sgame/sg_main.cpp
```diff
@@ -39,7 +39,7 @@ level_locals_t level;
 gentity_t          *g_entities;
 gclient_t          *g_clients;
 
-Cvar::Range<Cvar::Cvar<int>> g_showHelpOnConnection("g_showHelpOnConnection", "show MN_WELCOME on connect (1 = all, 2 = unregistered)", Cvar::NONE, 1, 0, 2);
+Cvar::Range<Cvar::Cvar<int>> g_showWelcome("g_showWelcome", "show MN_WELCOME on connect (0 = none, 1 = all, 2 = unregistered)", Cvar::NONE, 1, 0, 2);
```
slipher, line 42: The default value should be 0

DolceTriade: Oops. Done.

necessarily-equal:
Assuming that's what's on your server, LGTM

I assume pushing the branch again should re-trigger CI and fix the error.

bmorel:
This still does not compiles.

DolceTriade:
I don't understand why the CI continues to fail. It selects an older commit hash for daemon

src/sgame/sg_main.cpp
slipher: It's still 1.

DolceTriade: Done for real. Forgot to push from my main computer...

bmorel:
This finally passes!

PR #2059, opened 2022-08-12
Fix Unicode text entry (RmlUi regression)

### LINKED ISSUE #2025: Non-ASCII letters can't be typed in RmlUi
Since migrating to RmlUi, it is no longer possible to type the full Unicode range in the UI. Instead you get a random ASCII character.

slipher:
Fixes #2025.

### DIFF
--- a/src/cgame/rocket/rocketChatField.h
+++ b/src/cgame/rocket/rocketChatField.h
@@ -116,27 +116,27 @@ public:
 					case Rml::Input::KI_BACK:
 						if ( cursor_character_index > 0 )
 						{
-							text.erase( cursor_character_index - 1, 1 );
+							MoveCursorBackward();
+							text.erase( cursor_character_index, Q_UTF8_Width( text.c_str() + cursor_character_index ) );
 							UpdateText();
-							MoveCursor( -1 );
 						}
 						break;
 
 					case Rml::Input::KI_DELETE:
-						if ( cursor_character_index < (int) text.size() )
+						if ( cursor_character_index < text.size() )
 						{
-							text.erase( cursor_character_index, 1 );
+							text.erase( cursor_character_index, Q_UTF8_Width( text.c_str() + cursor_character_index ) );
 							UpdateText();
 						}
 
 						break;
 
 					case Rml::Input::KI_LEFT:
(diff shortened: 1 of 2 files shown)


src/cgame/rocket/rocketChatField.h
```diff
@@ -273,11 +273,18 @@ class RocketChatField : public Rml::Element, Rml::EventListener
 		Rml::GeometryUtilities::GenerateQuad( &vertices[0], &indices[0], Rml::Vector2f( 0, 0 ), cursor_size, GetProperty< Rml::Colourb >( "color" ) );
 	}
 
-	void MoveCursor( int amt )
+	void MoveCursorForward()
 	{
-		cursor_character_index += amt;
+		cursor_character_index += Q_UTF8_Width( text.c_str() + cursor_character_index );
```
DolceTriade, line 278: Why doesn't text[cursor_character_index] work?

src/cgame/rocket/rocketChatField.h
```diff
@@ -116,27 +116,27 @@ class RocketChatField : public Rml::Element, Rml::EventListener
...
-							MoveCursor( -1 );
 						}
 						break;
 
 					case Rml::Input::KI_DELETE:
-						if ( cursor_character_index < (int) text.size() )
+						if ( cursor_character_index < text.size() )
 						{
-							text.erase( cursor_character_index, 1 );
+							text.erase( cursor_character_index, Q_UTF8_Width( text.c_str() + cursor_character_index ) );
```
DolceTriade, line 128: How come backspace and delete have the same erase code? Shouldn't delete be deleting the character *in front* of the cursor while backspace deletes the character *behind* the cursor?

src/cgame/rocket/rocketChatField.h
slipher: Q_UTF8_Width reads up to 4 `char`s.

src/cgame/rocket/rocketChatField.h
slipher: Backspace is implemented like leftarrow then delete.

DolceTriade (approved):

PR #2065, opened 2022-08-13
rmlui: Clear text if you press ctrl-c

### LINKED ISSUE #2018: Chat box text remains after cancelling the message with esc
When typing text into any messagemode and pressing esc to cancel instead of enter to send, when any messagemode is used again, the text persists. This has only happened since 0.53.0's release.

Screenshot for the avoidance of doubt:
image

DolceTriade:
Fixes #2018

### DIFF
--- a/src/cgame/rocket/rocketChatField.h
+++ b/src/cgame/rocket/rocketChatField.h
@@ -139,6 +139,15 @@ public:
 						MoveCursor( 1 );
 						break;
 
+					case Rml::Input::KI_C:
+						if ( event.GetParameter< int >( "ctrl_key", 0 ) == 1 )
+						{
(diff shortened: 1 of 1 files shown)


slipher:
The discussion on #2018 didn't show a majority supporting this behavior change.

DolceTriade:
It's a trival change either way, let's just decide a direction. My preference was this so I made a PR for it. If we want to go in another direction, that is fine too, but its easier to discuss things with a working PoC rather than nothing.

DolceTriade:
On second thought, this change is pretty uncontoversial and it is a desired feature. we can discuss the esc thing later.

slipher:
LGTM

PR #2069, opened 2022-08-13
admin: Fix quoting around the /restart command

DolceTriade:
Before it would print something like this: `restart: map restarted by console $2t$3$4$5t$6$7$`


slipher:
LGTM

PR #2070, opened 2022-08-13
admin: implement /slap

cu-kai:
This PR implements the /slap command, which throws the victim around a random amount (though considering their mass, max hp) as well as optionally damaging them:

Syntax: `slap [target] (optional damage)`

It will print a message in the chat about somebody being slapped, as long as the slap doesn't kill the victim.

~~The chat print is done in the old Tremulous style (with `AP( va(` style) as I believe there may be a bug in `Quote()` preventing this from working correctly...~~ ok, this is fixed, I missed a `QQ()` somewhere.

I do believe the functionality could be improved upon slightly (such as better calculation of mass/how far the victim should be thrown), but this is a basic implementation and a popular request.


src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+
+	dir[0] = crandom();
+	dir[1] = crandom();
+	dir[2] = random();
+
+	// todo: G_Knockback
+	vec3_t kvel;
+    float mass = BG_Class( vic->client->pers.classSelection )->health;
+
+    VectorScale( dir, ( 250 * ( float ) 200 / mass ), kvel );
```
DolceTriade, line 2411: nit: use c++ floats.

```
VectorScale( dir, ( 250 * static_cast<float>(200) / mass ), kvel );
```

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+		ADMP( va( "%s %s", QQ( N_("^3$1$:^* cannot slap spectators!") ), "slap" ) );
+		return false; 
+	}
+
+	dir[0] = crandom();
+	dir[1] = crandom();
+	dir[2] = random();
+
+	// todo: G_Knockback
+	vec3_t kvel;
```
DolceTriade, line 2408: nit: fix indent

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+	{
+		ADMP( va( "%s %s", QQ( N_("^3$1$:^* cannot slap spectators!") ), "slap" ) );
+		return false; 
+	}
+
+	dir[0] = crandom();
+	dir[1] = crandom();
+	dir[2] = random();
+
+	// todo: G_Knockback
```
DolceTriade, line 2407: Either remove the TODO or implement it. Your velocity change seems to be simalr to G_Knockback?

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+	vec3_t kvel;
+    float mass = BG_Class( vic->client->pers.classSelection )->health;
+
+    VectorScale( dir, ( 250 * ( float ) 200 / mass ), kvel );
+    VectorAdd( vic->client->ps.velocity, kvel, vic->client->ps.velocity );
+
+    // set the timer so that the other client can't cancel
```
DolceTriade, line 2414: fix identation

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+		char dmg_str [ MAX_STRING_CHARS ];
+
+		trap_Argv( 2, dmg_str, sizeof( dmg_str ) );
+		
+		damage = atoi( dmg_str );
+	}
+
+	const HealthComponent* health = vic->entity->Get<HealthComponent>();
+
+	if ( health->Health() - damage > 0 )
```
DolceTriade, line 2431: check if health is null or you will crash

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+
+	if ( level.intermissiontime )
+	{
+		return false;
+	}
+
+	if ( trap_Argc() < 2 )
+	{
+		ADMP( QQ( N_("^3slap:^* usage: slap [^3name|slot#|admin#^7] (^3damage^7)") ) ); 
+		return false; // todo: G_admin_helptext( "slap" ) instead of this copypasta
```
bmorel, line 2377: to introduce a new feature with a TODO usually means the TODO will never be fixed :)

cu-kai: I need to speak to @DolceTriade about this, but basically I want to make a helper function which retrieves the help info automatically from the initial definition... there's several others I'd like to do but that's outside the scope of this PR.

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+	}
+
+	trap_Argv( 1, name, sizeof( name ) );
+
+	if ( ( pid = G_ClientNumberFromString( name, err, sizeof( err ) ) ) == -1 )
```
bmorel, line 2382: please declare the variable as close as possible from where it's used, it's more readable.
This would be:

```
int pid = G_ClientNumberFromString( name, err, sizeof( err ) ) 
if ( pid == -1 )
```

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+	if ( !admin_higher( ent, vic ) )
+	{
+		ADMP( va( "%s %s", QQ( N_("^3$1$:^* sorry, but your intended victim has a higher admin"
+		          " level than you") ), "slap" ) ); // todo: move this print to a single helper function
+		return false; 
+	}
+
+	const HealthComponent* health = vic->entity->Get<HealthComponent>();
+
+	if ( G_Team( vic ) == TEAM_NONE || !health )
```
bmorel, line 2399: instead of heals, maybe use the `IsAlive()` thingy?

bmorel, line 2399: instead of heals, maybe use the `IsAlive()` thingy?

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+	{
+		ADMP( va( "%s %s", QQ( N_("^3$1$:^* cannot slap spectators or dead players!") ), "slap" ) );
+		return false; 
+	}
+
+	dir[0] = crandom();
+	dir[1] = crandom();
+	dir[2] = random();
+
+	// from G_Knockback ...
```
bmorel, line 2409: if this is duplicated code, as the comment suggest, I'd rather see a new function called by the various sites.

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+	// set the timer so that the other client can't cancel
+	// out the movement immediately
+	vic->client->ps.pm_time = 200;
+	vic->client->ps.pm_flags |= PMF_TIME_KNOCKBACK;
+	// G_Knockback finish ...
+
+	if ( trap_Argc() > 2 )
+	{
+		char dmg_str [ MAX_STRING_CHARS ];
+
```
bmorel, line 2425: why the empty lines?

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+		return false; 
+	}
+
+	dir[0] = crandom();
+	dir[1] = crandom();
+	dir[2] = random();
+
+	// from G_Knockback ...
+	vec3_t kvel;
+	float mass = BG_Class( vic->client->pers.classSelection )->health;
```
bmorel, line 2411: ah, no, please... mass != health. If there is a relation here, please document it.

src/sgame/sg_admin.cpp
DolceTriade: Sure, it would be useful, but agree about adding TODOs. Ideally discussions about what you want to do in the future happen in issues or IRC/discord.

If you want a PR, there should be no TODOs

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+	if ( trap_Argc() > 2 )
+	{
+		char dmg_str [ MAX_STRING_CHARS ];
+
+		trap_Argv( 2, dmg_str, sizeof( dmg_str ) );
+		
+		damage = atoi( dmg_str );
+	}
+
+	if ( ( health->Health() - damage ) > 0 )
```
bmorel, line 2431: that is... `if ( not IsAlive() )` right? (you should, ofc, use the C/C++ idiom of `!IsAlive()` I guess, but both are valid C++ code)

src/sgame/sg_admin.cpp
DolceTriade, line 2409: Hm, actually Unv has G_Knockback implemented as part of Damage().

Just call

```
vic->entity->Damage( damage, /*source=*/ent, /*location=*/Util::nullopt, /*direction=*/dir, DAMAGE_KNOCKBACK, MOD_SLAP );
```

Therefore you should remove all the code that is duplicated from G_Knockback().

Can read code here: https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/components/KnockbackComponent.cpp#L10

DolceTriade:
Otherwise change looks fine

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,92 @@ bool G_admin_setlevel( gentity_t *ent )
...
+	{
+		AP( va( "print_tr " QQ( N_( "^3slap:^* $1$^* slapped $2$ ^*$3$" ) ) " %s %s %s", 
+			 G_quoted_admin_name( ent ), Quote( vic->client->pers.netname ), 
+			 ( damage > 0 ? Quote( va( "with %.0f damage", damage ) ) : QQ( "" ) ) ) );
+	}
+
+	CPx( vic - g_entities, va( "cp_tr " QQ( N_("[cross]$1$$2$ is not amused![cross]") ) " %s %s", 
+		G_quoted_admin_name( ent ), 
+		( health->Health() - damage > 0 ) ? "^7" : "^i" ) );
```
bmorel, line 2440: at this point, you should really, really cache this data, it would be easier to maintain. At least put this in a boolean.

src/sgame/sg_admin.cpp
cu-kai: This was discussed on IRC by @DolceTriade and won't be addressed by this PR.

src/sgame/sg_admin.cpp
cu-kai: duplicate comment?

cu-kai: because bots don't have `health` at all until they spawn, and using `health->Alive()` here causes a segfault if you /slap a dead bot.

src/sgame/sg_admin.cpp
cu-kai: as discussed on IRC, this is not a suitable replacement right now.

src/sgame/sg_admin.cpp
cu-kai: no and I have commented the reason why in the latest change.

src/sgame/sg_admin.cpp
cu-kai: now documented

src/sgame/sg_admin.cpp
```diff
@@ -2350,6 +2358,93 @@ bool G_admin_setlevel( gentity_t *ent )
...
+
+	dir[0] = crandom();
+	dir[1] = crandom();
+	dir[2] = random();
+
+	// from G_Knockback ...
+	vec3_t kvel;
+	float mass = BG_Class( vic->client->pers.classSelection )->health;
+	// health != mass however for the purposes of this functionn
+	// using the max health of the class works a lot better
```
DolceTriade, line 2414: Please be clear that we use health as mass to better handle the pure knockback case. That's why we're duplciating this code instead of using the knockbackcomponent.

DolceTriade (approved):

PR #2079, opened 2022-08-15
Allow servers to reward credits to players for destruction

bmorel:
Fix https://github.com/Unvanquished/gameplay/issues/50

The ratio is affected by a cvar, so it can be disabled and adjusted at will.


DolceTriade:
FYI, this will need to be tested on community servers and have good feedback before we will merge it into master.

DolceTriade:
On second thought, since this is opt in and as you mentioned on IRC, it is a pretty trivial change, I think I was wrong to deny it. If you're up for it, I think we can merge this.

DolceTriade (approved):

PR #2090, opened 2022-08-20
Fix spawn messages linger when you leave the spawn queue and become a spectator.

DolceTriade:


necessarily-equal (approved):
LGTM, haven't tested

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2376,6 +2376,50 @@ class SpawnQueueElement : public TextHudElement
...
+			SetText( va( P_( "There is %d spawn remaining", "There are %d spawns remaining", this->spawns ), this->spawns ) );
+		}
+
+		if ( !IsVisible() )
+		{
+			SetProperty( Rml::PropertyId::Visibility, Rml::Property( Rml::Style::Visibility::Visible ) );
+		}
+	}
+
+	int spawns;
```
slipher, line 2419: We decided to go with trailing underscore for class members. https://wiki.unvanquished.net/wiki/Coding_convention#Naming_conventions

slipher (commented):
LGTM other than style nit

src/cgame/cg_rocket_draw.cpp
DolceTriade: Ah I didn't realize that. I was sticking to the Unv convention of no identifier for no class members :D.

Will fix and merge later.

necessarily-equal: Fort he record I don't think this is a problem because it is consistent with the rest of the file.

DolceTriade: Done.

necessarily-equal:
👍

PR #2092, opened 2022-08-21
sg_admin: Don't register commands for the local client. 

### LINKED ISSUE #1884: in devmap: Warn: VM 'CGame' tried to register command 'bot' which is already registered
I just noticed there is a lot of those messages (probably all the commands, actually) emitted when running `./daemon +devmap ...`.
There might be an underlying minor problem behind this, hence the report.

In gdb:
```
^3Warn: VM 'CGame' tried to register command 'adjustban' which is already registered
^3Warn: VM 'CGame' tried to register command 'adminhelp' which is already registered
^3Warn: VM 'CGame' tried to register command 'admintest' which is already registered
^3Warn: VM 'CGame' tried to register command 'allowbuild' which is already registered
^3Warn: VM 'CGame' tried to register command 'allready' which is already registered
^3Warn: VM 'CGame' tried to register command 'ban' which is already registered
^3Warn: VM 'CGame' tried to register command 'bot' which is already registered
^3Warn: VM 'CGame' tried to register command 'builder' which is already registered
^3Warn: VM 'CGame' tried to register command 'buildlog' which is already registered
^3Warn: VM 'CGame' tried to register command 'cancelvote' which is already registered [further messages like this will be suppressed]
```

DolceTriade:
These commands are only added to this array so we can have command
completion for them. There is no need to register them because they will
be registered when the sgame tells the cgame to register them.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -529,6 +529,14 @@ void G_admin_cmdlist( gentity_t *ent )
 	char out[ MAX_STRING_CHARS ] = "";
 	unsigned len, outlen;
 
+	// If this is the local client, no need to send them commands.
+	// The local client will get their commands registered in
+	// G_admin_register_cmds.
+	if ( level.inClient && ent->client->ps.clientNum == 0 )
+	{
+		return;
+	}
+
 	outlen = 0;
 
 	for ( unsigned i = 0; i < adminNumCmds; i++ )


DolceTriade:
On second thought, this doesn't actually fix #1884, but I still this it's correct to do.

bmorel:
Can't say it's good for sure, as I don't know that part of the code, but it makes sense I guess: LGTM.

slipher:
As I understand it the point of commands with no function is to register them for tab completion. When playing on a remote server, the sgame cannot register its own admin commands.

DolceTriade:
Correct. Note that not registering it when this function `CG_InitConsoleCommands` simply means that commands like `buy` will be reported as `No Command Found` *unless* you are connected to a server.

Currently, since we register these commands, but don't include a function, they are registered and are basically a no-op. It's a small trival clean up.

slipher:
> As soon as you connect, the SGame will send these commands and they will be registered then (and unregistered when you disconnect).

Really? How does that work?

necessarily-equal:
Well if sgame sends its command list, then it would be cool to auto complete thel without having to add them explicitely like in https://github.com/Unvanquished/Unvanquished/pull/1462.

Executing the command should work even if the client is not aware they exist though

DolceTriade:
The sgame sends a list of valid commands, not their implementation. The cgamw registers them on behalf of the sgame. Every time a user executes a registered command and the command doesn't have an associated function, it is forwarded to the server for execution. This change prevents registering commands that are not yet available.

Note that the sgame does not send any *implementation*. This means that we are still required to write command completion for server side commands in the cgame.

slipher:
> The sgame sends a list of valid commands

Took a look at the code and I'm pretty sure this is not real

DolceTriade:
Hm, it seems we only do this for admin commands and not for commands in sg_cmds.cpp. Let me modify this change to do that?

DolceTriade:
Adjusted pull request to actually fix the bug in question. Seems I was wrong about us registering sg_cmds.cpp commands. I remember seing the `CG_GameCmds_f` function and assumed it came from sg_cmds.cpp.

src/sgame/sg_admin.cpp
```diff
@@ -529,6 +529,12 @@ void G_admin_cmdlist( gentity_t *ent )
 	char out[ MAX_STRING_CHARS ] = "";
 	unsigned len, outlen;
 
+	// If this is the local client, no need to send them commands.
```
slipher, line 532: You could add a comment about how it's done in G_admin_register_cmds

LGTM

DolceTriade: Done.

PR #2093, opened 2022-08-21
cgame: Fix emoticon in base warning.

DolceTriade:
Previously it showed the RC regardless of which team was attacks.

Also fix up the comments and clarify that both aliens and humans take
the emoticon code path, since the comments suggest that only humans take
that code path.


bmorel:
LGTM

necessarily-equal (approved):
LGTM, haven't tested

PR #2095, opened 2022-08-23
rmlui: Move rocket defs from the engine to rocket.h

DolceTriade:
These defines were in the engine for legacy reasons, but now there is no
reason to keep them there.


bmorel:
One problem this poses is the API break between daemon and gamelogic. I think we got a consensus that the API break between cgame and sgame was abandoned (@slipher can you confirm?) but I think we still need the daemon-gamelogic compat to be respected so that master hosts can be played by regular clients.

slipher:
This doesn't affect any ABIs.

LGTM

bmorel:
Well, if slipher says no API problem, then LGTM

PR #2100, opened 2022-08-26
GLM cherry-picking from bmorel's branch

necessarily-equal:


slipher:
LGTM

slipher:
Oh there was one thing, the submodule commit is pretty old and could be bumped.

illwieckz:
Also the submodule URL should be changed to https://github.com/Unvanquished/glm.git

Having our own URL for the submodule is meant (and only meant) to allow us to tag GLM versions with tags like `unvanquished/0.53.1` when releasing the game like we already do with other submodules. This repository doesn't mean we plan to modify GLM (we don't).

necessarily-equal:
I've updated the submodule url

src/shared/bg_public.h
```diff
@@ -29,6 +29,10 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 // bg_public.h -- definitions shared by both the server game and client game modules
 //==================================================================
 
+#include <glm/vec3.hpp>
+
+#define VEC2GLM( x ) glm::vec3( (x)[0], (x)[1], (x)[2] )
```
DolceTriade, line 34: Can this be glm::make_vec3( x );?

DolceTriade (approved):
% illwieckz and slipher's comments. Thanks for doing this.

src/shared/bg_public.h
necessarily-equal: It seems like it could. The pro of doing it that way is that it is also compatible with Vec3. I'm not too fond of having to rewrite every commit using Vec3.

This VEC2GLM macro is removed in the last commit of the branch anyway

DolceTriade: :shrug: Not something I feel strongly about, so I don't mind either way.

slipher: Rewriting the branch would be annoying but maybe we can at least make it a function so the argument is not evaluated 3 times. `glm::vec3 VEC2GLM(const float* x) { ... }`

necessarily-equal: I guess I could make two inline functions. On the other hand the compiler should already identify those (unless they are function calls, but I don't think so)

DolceTriade: Do you really need to rewrite the branch or can you just rewrite this single commit to update the function? All other commits should then reuse it. Vec3 is going to be deleted anyways.

necessarily-equal: So I tried a bit, and after implementing it for Vec3, vec3_t and glm::vec3 (because it's sometimes called on itself, I get some `error: unknown type name 'Vec3'` somewhere. In the end I think it's just easier to keep the macro.

slipher: Make it a template then. `template<typename T> glm::vec3 VEC2GLM(const T& x) { return {x[0],x[1],x[2]}; }`

necessarily-equal: fair enough

necessarily-equal: rebased

slipher:
LGTM

PR #2101, opened 2022-08-27
g_maxMiners: better error messages

sweet235:
This adds more understandable error messages for the case that g_maxMiners is enabled (it is disabled by default). From my own experiments I conclude that g_maxMiners can solve some of the problems described in https://github.com/Unvanquished/gameplay/issues/23. So having proper error messages seems to be in order.

I'd prefer have to have `longMsg` too, but that would be overlaid by the window informing the builder about build points.

### DIFF
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -1716,7 +1716,7 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
 		if ( miners >= max_miners )
 		{
 			//TODO 0.54: make this a more appropriate error message, or remove g_maxMiners completely
-			return ent->client->pers.team == TEAM_HUMANS ? IBE_NOHUMANBP : IBE_NOALIENBP;
+			return ent->client->pers.team == TEAM_HUMANS ? IBE_NOMOREDRILLS : IBE_NOMORELEECHES;
 		}
 	}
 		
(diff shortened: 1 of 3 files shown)


src/sgame/sg_buildable.cpp
```diff
@@ -1716,7 +1716,7 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
 		if ( miners >= max_miners )
 		{
 			//TODO 0.54: make this a more appropriate error message, or remove g_maxMiners completely
```
slipher, line 1718: Remove the TODO which has been completed.

LGTM otherwise

src/shared/bg_public.h
```diff
@@ -818,7 +821,8 @@ enum dynMenu_t
   MN_H_NOREACTOR,
   MN_H_NOBP,
   MN_H_NOTPOWERED,
-  MN_H_ONEREACTOR
+  MN_H_ONEREACTOR,
+  MN_H_NOMOREDRILLS
 };
```
necessarily-equal, line 826: for the record, it's not a bad idea to add a `,` at the end of the line to avoid diff noise. Or to keep the list dsorted. (nothing blocking)

necessarily-equal (approved):
LGTM

src/shared/bg_public.h
DolceTriade: ack.

DolceTriade: +1 to the comma.

PR #2102, opened 2022-08-27
GLM: Bot Stuff™

necessarily-equal:
Here is a sizeable chunk from the glm branch. It touches mostly the bot stuff. Changes that aren't pure conversion includes inlining a function and making one class member private with a getter and some friend functions.

To be squashed at merge time


src/sgame/sg_bot_util.h
```diff
@@ -155,4 +156,10 @@ bool  BotPathIsWalkable( const gentity_t *self, botTarget_t target );
 //consider bot to be stuck if it does not move farther than this in some period of time
 constexpr float BOT_STUCK_RADIUS = 150;
 
+//Those are not really BG, but hopefully someday they will be.
+glm::vec3 BG_GetClientNormal( const playerState_t *ps );
```
slipher, line 160: Do not write functions prefixed BG_ in sgame.

src/sgame/botlib/bot_types.h
```diff
@@ -60,6 +62,9 @@ struct botNavCmd_t
 	float    dir[ 3 ]; // normalized direction of the target
 	int      directPathToGoal;
 	int      havePath;
+	glm::vec3 glm_dir ( void ) const { return glm::vec3( dir[0] , dir[1] , dir[2] ); };
```
slipher, line 65: Do not write `void` in function prototypes.

src/sgame/sg_bot_local.h
```diff
@@ -115,12 +116,19 @@ struct botMemory_t
...
-	vec3_t      futureAim;
-	botNavCmd_t nav;
+	glm::vec3   futureAim;
+private:
+	botNavCmd_t m_nav;
+public:
+	botNavCmd_t const& nav( void ) const { return m_nav; };
+	void clearNav( void ) { memset( &m_nav, 0, sizeof( m_nav ) ); }
+	void setGoal( botTarget_t target, bool direct ) { goal = target; m_nav.directPathToGoal = direct; }
+	friend void G_BotThink( gentity_t * );
```
slipher, line 126: This friend stuff is very ugly. Maybe if it were going to some sort of sensible end goal but I doubt it...

src/sgame/sg_bot_util.h
necessarily-equal: This is fixed in a later commit. Not sure if you want me to cherry-pick even more commits in this PR or if you are ok with waiting on the next one.

src/sgame/sg_bot_local.h
necessarily-equal: I think the idea is to identify the functions which use it and have a easier time to check they are correct. and progressively migrate to correct setters or smth.

It's not my code so I don't really know though

src/sgame/botlib/bot_types.h
necessarily-equal: Fixed

slipher: There's still some `( void )` left in there

src/sgame/sg_bot_util.h
slipher: I think it's better to assume that a contributor may vanish at any moment. So no bad code should be merged with the promise of being fixed some time in the future.

necessarily-equal: Ok but like, the code is already written.

So would you prefer more commits in this PR or a second one?

src/sgame/botlib/bot_types.h
necessarily-equal: Right, removed the last one. Trusting github to show the whole PR was a bad idea :S

src/sgame/sg_bot_util.h
slipher: Regardless of how you split it I prefer to have the commit that fixes it in the same one that messes it up.

necessarily-equal: I've added a commit for that (to be squashed with everything else)

src/sgame/sg_bot_ai.cpp
```diff
@@ -1282,8 +1280,8 @@ AINodeStatus_t BotActionRepair( gentity_t *self, AIGenericNode_t *node )
 		G_ForceWeaponChange( self, WP_HBUILD );
 	}
 
-	AngleVectors( self->client->ps.viewangles, forward, nullptr, nullptr );
-	VectorMA( self->s.origin, self->r.maxs[1], forward, selfPos );
```
slipher, line 1287: Why is this VectorMA deleted?

src/sgame/sg_bot_util.cpp
```diff
@@ -955,11 +955,18 @@ botTarget_t BotGetRetreatTarget( const gentity_t *self )
 botTarget_t BotGetRoamTarget( const gentity_t *self )
 {
 	botTarget_t target;
-	vec3_t targetPos;
+	glm::vec3 point;
 
-	BotFindRandomPoint( self->s.number, targetPos );
+	int botClientNum = self->s.number;
+	if ( !BotFindRandomPointInRadius( botClientNum, g_entities[ botClientNum ].s.origin, point, 2000 ) )
```
slipher: Use `self` instead of going through g_entities

src/sgame/sg_bot_ai.cpp
necessarily-equal: The result seems to be ignored.

src/sgame/sg_bot_util.cpp
necessarily-equal: done

slipher:
LGTM

DolceTriade:
yay!!

Issue #2107, opened 2022-08-27
updater should provide the `dist/config` folder


bmorel:
The updated is the recommended way to get a working installation, but it does not includes the example files needed to setup a server.

slipher:
Are you suggesting to have an examples folder or something in the universal zip? Or to actually install them into the homepath?

bmorel:
I think at least embedding some documentation, and example files are doc, is important. Installing them in homepath would imply scripts for installation from there, and I'm not sure it's where we want to go

necessarily-equal:
Those days I've been using the `dist/configs/` folder from Unvanquished repo directly for my testing servers. I found way more bug in the default config than you would think.

This is very practical, and in fact I'm using it for every server I host. For example here's the [folder for the juggernaut config](https://github.com/Unvanquished/Unvanquished/tree/juggernaut/deploy/dist/configs). It contains layouts, maprotation, bot add commands and the custom value of the required cvars.

I think on the long term we should seek to use this folder more. I would like to have a run command made by cmake or meson that would use this folder for development purpose. I would like to have this folder used for server installations too.

I'm not so sure the installer is suitable for servers though. But well I also think it could be useful to include this config directory.

Unless someone picks this up though, this won't happen for 0.54, though. I'm putting this is 0.55 for now.

PR #2108, opened 2022-08-27
add documentation

bmorel:
Intended documentations for this PR:

* how to host own dedicated server
* how to package a mod

The files this will add should be provided by the updater and the universal zip, along with the dist/config files (see #2107).

Information provided by this PR have been verified locally, and should work. It is meant to replace following wiki pages:

* https://wiki.unvanquished.net/wiki/Hosting_one's_own_awesome_Mod
* https://wiki.unvanquished.net/wiki/Making_an_awesome_Mod
* https://wiki.unvanquished.net/wiki/Tutorials/Modifying_paks

Please focus on the exactitude of information in there rather than on the wording. Lack of documentation is a real problem of unvanquished, and nitpicking is not going to help addressing this issue.

Currently, the PR only includes documentation for the hosting part, but it it considered complete. I am going to write the 2nd file (PACKAGING) later, maybe in a 2nd PR if this one can be merged quickly enough.
Also, I do NOT intend to document packaging with Urcheon, as the tool is overly complex for the task, and those documents target people willing to quickly setup a server to try their own modifications.
Another point I want to be clear: this is not meant to be exhaustive, but to go to the point.


doc/HOSTING
```diff
@@ -0,0 +1,116 @@
+# introduction
```
DolceTriade, line 1: Capitalize first letter of all headings.

doc/HOSTING
```diff
@@ -0,0 +1,116 @@
...
+
+This document tries to explain what is required to host a
+dedicated server, provided you already have a working
+unvanquished installation and "pak" files (those who have a
+`.dpk` file extension), which are not necessarily the
+official ones.
+
+# non-goals
+It does *not* tries to explain how to compile or modify
```
DolceTriade, line 10: "does *not* try"

doc/HOSTING
```diff
@@ -0,0 +1,116 @@
...
+unvanquised, how to configure a server with modified engine,
```
DolceTriade, line 11: "with a modified engine"

doc/HOSTING
```diff
@@ -0,0 +1,116 @@
...
+how to setup your system or network, and aims at the
```
DolceTriade, line 12: "or how to setup"

DolceTriade, line 12: "network. The aim is the simplest possible configuration."

doc/HOSTING
```diff
@@ -0,0 +1,116 @@
...
+simplest possible configuration.
+
+# pre-requisites
+
+First of all, you need a computer than can be reached by the
```
DolceTriade, line 17: Perhaps this might be more legible as a list rather than a block of text?
```
 - Computer with inbound access from the Internet (either a public IP or port forwarding).
 - HTTP server to host files for players to download
```
Serving on port 80 is not a requirement and may be difficult for people using port forwarding. Unvanquished will respect port numbers in the sv_downloadUrl iirc.

doc/HOSTING
```diff
@@ -0,0 +1,116 @@
...
+* irt_core-x86_64.nexe
+
+# HTTP configuration
+
+You will need to have a directory (I advise you to dedicate
+it to unvanquished, but this is not mandatory) containing
+`.dpk` files and another file named `PAKSERVER`, containing
+the string `ALLOW_UNRESTRICTED_DOWNLOAD`.
+If you get this wrong, your users will experience the
+infamous "UDP download", which is very slow. Slow as in,
```
DolceTriade, line 45: "which will net users 3 KiB/s".

All the follow up text doesn't really add anything to the documentation.

doc/HOSTING
```diff
@@ -0,0 +1,116 @@
...
+less than 30KiB/s, meaning most changes will litterally take
```
DolceTriade, line 46: "litterally" -> "literally"

If you choose to keep the extra words anyways

doc/HOSTING
```diff
@@ -0,0 +1,116 @@
...
+You will also need an HTTP server serving on port 80. While
+this is not strictly necessary, it is highly recommended.
+Not doing this means that if your players lack one of the
+requirements, they might be unable to join at all.
+I'll skip the details on this, as this discussion is out
+of the scope for this current document.
+
+# Daemon installation
+
+Create a directory in which you will place the following
```
DolceTriade, line 29: For the purpose of the document, it would be helpful to give an example of a directory so you can refer to this directory in your subsequent examples.

doc/HOSTING
```diff
@@ -0,0 +1,116 @@
...
+
+## server "homepath" setup
+
+homepath will expect a specific layout. Files there are not
+technically required, but not having them will mean game
+will rely on the built-in alternatives.
+
+* `config/`
+* `config/server.cfg`
+* `config/map/`
```
DolceTriade, line 78: I would say that map specific configs definitely do not fall under "simplest possible set up". Also its confusing that "map" refers to something like "plat23" instead of the word "map"

DolceTriade (commented):
Let's call this file `hosting.md` since it is formatted using markdown?

doc/HOSTING
bmorel: Can't find this cvar when grepping the code.

doc/HOSTING
bmorel: It's basically to put emphasis on the fact it's a bad idea. But I guess you're right.

doc/HOSTING
bmorel: Good point.

doc/HOSTING
bmorel: One thing is, the "map/default.cfg" file is really, really useful currently. Notably for bot names, or to have different number of humans/aliens bots.

doc/HOSTING
DolceTriade: Sorry I meant `sv_wwwBaseURL`

doc/PACKAGING
```diff
@@ -0,0 +1,198 @@
...
+EARLY VERSION
+The "mod" section is not written yet! 
+EARLY VERSION
+EARLY VERSION
+
+This document describes how packages are done in unvanquished.
+
+Games running in Daemon-engine are only able to see files
+inside a VFS (Virtual File System), built mostly from
+packages contents (and some other files server-side).
```
DolceTriade, line 11: "package"

DolceTriade, line 11: "server side files"

doc/PACKAGING
```diff
@@ -0,0 +1,198 @@
...
+
+Packages are compressed archives, which can be generated by
```
DolceTriade, line 13: "compressed zip archives"

doc/PACKAGING
```diff
@@ -0,0 +1,198 @@
...
+standard archivers.
+
+To create such an archive, place all and only the needed
+files (beware hiden files, they can prevent the game to load
+packages!) in a folder, and then run.
```
DolceTriade, line 18: It might be useful to mention conceptually what you are doing is creating a zip file and renaming it to have a .dpk extension.

doc/PACKAGING
```diff
@@ -0,0 +1,198 @@
...
+
+Games running in Daemon-engine are only able to see files
+inside a VFS (Virtual File System), built mostly from
+packages contents (and some other files server-side).
+
+Packages are compressed archives, which can be generated by
+standard archivers.
+
+To create such an archive, place all and only the needed
+files (beware hiden files, they can prevent the game to load
```
DolceTriade, line 17: "beware of"

DolceTriade, line 17: "the game from loading"

doc/HOSTING
bmorel: I simply ended up removing the port information (and writing a PR to change the sv_www cvar_t into Cvar)

slipher:
Why does it need to be in the repository exactly? We could put a link to the wiki pages for hosting etc. in the README.txt which is shipped in the universal zip/updater package.

bmorel:
> Why does it need to be in the repository exactly? We could put a link to the wiki pages for hosting etc. in the README.txt which is shipped in the universal zip/updater package.

I see several reasons:

* it's easier to maintain things in repo than in wiki. For example, if you grep for a cvar because you need to remove it or alter it's behavior, you *will* get references in the doc/ folder, and so might think and take the time to update it, unlike wiki
* with repo, it's possible to find the documentation associated with a specific version of the game
* if the files in repo are served with universal zip or updater, people will *not* have to rely on a browser to read doc, which means it can be present on the servers themselves
* easier to find. In current status, it's very difficult to search for those info in the wiki.

DolceTriade:
I think we need to decide where to put such documentation. We don't want two places for documentation wherein lots of work has already been put into the Unvanquished wiki and now we want a separate place for other documentation. I see the argument for using the github source code for technical documentation, but less so for user facing instructions.

bmorel:
> wherein lots of work has already been put into the Unvanquished wiki

Have you looked at the 3 pages mentioned?

DolceTriade:
From my experience, admins are not technically versed. They require more soft people skills than technical skills.

Of course, having information about modding is very much developer oriented.

illwieckz:
Some cosmetic comments:

- Since that's markdown, it would need `.md` extensions (if it wasn't markdown I would asked for it anyway).
- Better add the file subject as primary heading (`# Hosting`) and other titles as secondary heading (`## Introduction`) (note: If you're mimicking LinuxFr.org even unconsciously, they did it wrong).

For a global comment about this: it looks like a re-implementation of a new wiki, we may want to improve the wiki instead.

bmorel:
Ack, should be done in wiki. I'll close the PR in few days.

illwieckz:
One thing we miss with wiki is the ability to better coordinate the edition. While getting in the full PR review process may be a bit too heavy, the wiki is too light (and I feel mediawiki talk page as unusable for us)

illwieckz:
@bmorel, just to be sure there is no misunderstanding, I'm still thinking we can improve the basic information we provide in repo, for example we may prefer to introduce some key topics and provide links to more detailed wiki pages. But if you feel some pages you wrote can replace existing wiki page, they can replace them there to begin with. 🙂️

bmorel:
One thing is certain about the wiki. It's a total mess, hardly usable, maintained by nobody, and when I write code I certainly don't want to waste my time with a web browser to try to edit that.
You talk about missing features in wiki? Well, an obvious one is: no notifications. Including in IRC.
Another obvious one is: the diff system sucks. Unlike git's.
Also, t does not sync with the code, but I already mentioned that.

Well, have fun with unmaintained documentation. At least this PR made it clear on the upstream stance.

bmorel:
> But if you feel some pages you wrote can replace existing wiki page, they can replace them on the wiki to begin with. slightly_smiling_face

It's not just my pages. It's impossible to have an overview of wiki's content. Many if not all pages are outdated, and not just slightly.

illwieckz:
About notifications, etc. I notice that's the same problem of forums: the same way everything looks like a nail when we only have a hammer, we tend to replace everything with github issues and PR because it happens to do things we need even if not purposed for it (some people would say that means GitHub isn't so bad, at least, not worst 😅️).

illwieckz:
> Many if not all pages are outdated, and not just slightly.

I don't know about the “many”, but I write a lot of stuff on Wiki, which just reveals one of the problem: « yes, but who knows??? »

I started a dedicated thread on that topic there: https://github.com/Unvanquished/unvanquished-infrastructure/issues/27

bmorel:
If I was really using github, then I'd use github's wiki, which happens to *not* suffer from several of the problems official wiki have.
Notably, it can be synced locally (no network access required), through git, which also means it's possible to automatically generate stuff (forgot to mention that about the problems of wiki). This last one would be very precious for things like at least:

* [player documentation](https://wiki.unvanquished.net/wiki/Humans)
* [BT doc](https://wiki.unvanquished.net/wiki/Formats/Behavior_tree)
* [list of maps](https://wiki.unvanquished.net/wiki/Maps) (for which I have a script which gives already more useful info, completely automated)

So, yes, maybe you only have a hammer and to you everything can go in the wiki "coffin nail", but me I know the value of versioned text, and even despite I hate html a lot I can sometimes convince myself it's a good solution.

One thing is certain, I'm the only one thinking the wiki there is bad, some people even said that there was effort put in wiki (maybe it's true, but seriously, how many years ago?)... so this PR will be closed in few days, as I said. That's the consensus, even if it sucks, and I don't like seeing too much open PRs, so at least I should take care of not having too much myself.
Obviously, since I consider the wiki a bad solution, I won't update it myself.

illwieckz:
No one argued against versioned text and the good things that are git-based wikis, neither against wiki families that may be better to generate content. We can even evaluate migrating the wiki to another software, why not. But fore sure if “I consider the wiki a bad solution” is a prerequisite the countless hours I spent on writing the wiki this year can't help.

bmorel:
well. Maybe the wiki can become a good solution. The current one is not, though, and having spent time on something is not a sign of quality. I also have spent time to update wiki. Of all the pages I've linked, there's none I have not touched.
I even tried to do templates or whatever it's called, and this is why I could build my opinion.
The time you've spent on it, while it's still barely used, shows that it's simply not easy to make it a useful asset.
As you know very well, there is no way to have someone really working at improving and keeping updated the wiki.
Especially for technical stuff, only tech persons are able to do that, and will usually prefer to spend their time on actual tech stuff.

Simple text update in same repo requires much less context switching, and that's why I think it's better. But I know I won't convince anyone, so... well, you know my plans.

necessarily-equal:
as for the argument argument of "I'd like to do it in a text editor" and "tracked in Git", GitLab wikis would be pretty good. While still allowing a web interface for those who aren't as old school and just like web interfaces.

The biggest downside I can see is that one needs to be a developer to have edit rights on the wiki. At least until https://gitlab.com/groups/gitlab-org/-/epics/7107 is fixed. It would also help a lot on the "lack of review" problem.

I think GitHub wiki offer a similar feature to allow suggesting a change with a PR, but then I don't really want to depend more on GitHub.

Anyway I think this is fine as a PR to this repo

necessarily-equal:
Actually there is a workaround for MR on GitLab Wikis: https://gitlab.torproject.org/tpo/tpa/team/-/wikis/service/documentation#accepting-merge-requests-on-wikis

illwieckz:
A split documentation is worst than a bad documentation. It adds bad to the bad, and it makes sure more bad or wrong documentation exists and for more longer.

Let's take a simple example: we need to have a readme in repository giving minimal instructions on how to clone and build. This is something we cannot really avoid, it's the minimal exception we have to sustain. And you know what? I edited the wiki page for “getting the source” yesterday (something that is listed on home page), it was missing the instructions for submodules and assumed the latest version was 0.37 in 2015. There are other obsolete stuff in wiki but the other obsolete stuff are things we forget about because specific to some high level topics but this one is different: it is the perfect example of something being both “the very first entry-level information you need to look for” and “the most obsolete one”. The only reason why this page was outdated is that the readme already did the job: we all overlook them. So, while we need to have a readme there for entry level information and we have to deal with it, splitting more the wiki will just lead to more obsolete (and more broken) information to be alive.

illwieckz:
> Note deleting the outdated wiki pages is an option, if we choose to have it here.

Indeed. The same way the readme can link to wiki for extra information.

bmorel:
The wiki is so nice and appropriate for documentation that the recent edits lost the detailed page about actions. Anyway, I said I'd close this, let's do it.

PR #2111, opened 2022-08-28
Introduce g_persistDevmap

slipher:
Make it so that dev mode can continue to be enabled after map changes.
Useful for development scenarios which need to last beyond the end of a
map.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -3392,7 +3392,7 @@ bool G_admin_changemap( gentity_t *ent )
 	admin_log( map );
 	admin_log( layout );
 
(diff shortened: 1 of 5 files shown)


DolceTriade (approved):
I wonder if the right way to do this would be to make a helper function to move to the next map instead? It seems everyone is reimplementing this `<map|devmap> <map> <layout?>` logic.

slipher:
I considered a helper function but it doesn't work for the voting-related cases. For those you have to prepare a command string that is executed if the vote passes.

bmorel:
lgtm

bmorel:
> It seems everyone is reimplementing this <map|devmap> <map> <layout?> logic.

You're referring to  https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/102 ?

DolceTriade:
I was referring to the numerous places in this changes where there was very close code that needed to run map <mapname> <layout>.

PR #2119, opened 2022-08-29
Set g_mapConfigs to "map" by default

slipher:
Get rid of an unnecessary extra step when setting up a server.

### DIFF
--- a/dist/configs/config/map/default.cfg
+++ b/dist/configs/config/map/default.cfg
@@ -1,3 +1,8 @@
+// This file (config/map/default.cfg) is executed each time a new map is loaded.
+// In addition, map- and layout-specific configs will be executed if present:
+// - config/map/plat23.cfg (for map plat23)
+// - config/map/plat23/eggspam.cfg (for map plat23 and layout "eggspam")
+
 // define some bot names
 bot names clear
 bot names humans "^9[bot]^4 Clodomir" "^9[bot]^4 Alaric" "^9[bot]^4 Baudry" "^9[bot]^4 Adelin" "^9[bot]^4 Ferdinand" "^9[bot]^4 Brieux" "^9[bot]^4 Audran" "^9[bot]^4 Alistair" "^9[bot]^4 Renan" "^9[bot]^4 Malo"
--- a/dist/configs/config/server.cfg
+++ b/dist/configs/config/server.cfg
@@ -68,14 +68,6 @@ set g_teamForceBalance 2
 
 set g_bot_defaultFill 3
 
-// load per map config file from this subdirectory
-// it will load all the configs specified:
-//   - map/default.cfg (for every map)
-//   - map/plat23.cfg (for plat23, regardless of the layout)
-//   - map/plat23/layout.cfg (for plat23 and layout "layout")
-// map config files can be used to fill game with bots and set other useful configs.
-set g_mapConfigs "map"
-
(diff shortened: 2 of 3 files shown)


necessarily-equal:
Sounds good. Haven't tested

illwieckz:
One related thing: do we want `map` be the default value?

Everything in the game uses plurals:

`map-name_version.dpkdir/maps`
`map-name_version.dpkdir/textures`
`map-name_version.dpkdir/scripts`
`game/maps/`
`game/layouts/`
`game/stats/`

I don't like that ending `s` but we may want to be consistent.

About why the current example in config sample is `map` without S is that long time ago, years ago, I asked for help in the IRC chat and someone in chat or forums told me that cvar existed and probably gave me `map` as an example of string to use/folder to create. But no one took the time to think about it as a standard. Then lately when we wanted to add a sample configuration in the repository, I probably just copy-pasted my own config folder or someone copy-pasted his own config folder that was just mimicking mine.

slipher:
To my mind `maps` should contain maps. Which our config directory does not. (Yes the navmesh one `game/maps/` also sounds kind of bad)

illwieckz:
> Yes the navmesh one game/maps/ also sounds kind of bad

This does not look bad to me. It looks like `game/` is acting as the folder where you can put files that are accessible via the VFS without being part of any dpk/dpkdir, like the root folder of `base/` (pkg/) behaved in Tremulous.

slipher:
It's not actually part of the VFS, but sometimes we use a filesystem API that searches for a path both in `game/` and the VFS.

I guess navmeshes could arguably be called maps - they're a description of the terrain that may be used to navigate :smile:. You can't say that about the config files.

illwieckz:
> It's not actually part of the VFS, but sometimes we use a filesystem API that searches for a path both in game/ and the VFS.

Then it's useful to have it following the same layout as the VFS. Thanks for the clarification though.

> You can't say that about the config files.

Yes, that's why I never cared that much about `map` because it's not accessible through vfs in any way, but I want to be sure what we do is minded and not just the result of years of mindless cargo cult copy pasting from a random comment on chat or forums years ago.

illwieckz:
An argument in favor of `map` being used in the singular is that `config` is already used in the singular.

bmorel:
An argument to use `map` is that it keeps the current behavior.
LGTM

dist/configs/config/map/default.cfg
```diff
@@ -1,3 +1,6 @@
+// This file is executed each time a new map is loaded.
+// In addition, a map-specific config is executed if present, e.g. config/map/plat23.cfg.
```
DolceTriade, line 2: Keep documentation about layout specific configs?

DolceTriade (approved):
LGTM % nit

dist/configs/config/map/default.cfg
illwieckz: I would report those lines (it can be rewritten if needed):

```
// it will load all the configs specified:
//   - map/default.cfg (for every map)
//   - map/plat23.cfg (for plat23, regardless of the layout)
//   - map/plat23/layout.cfg (for plat23 and layout "layout")
```

because this isn't clear enough to me (and does not mention layout):

```
// In addition, a map-specific config is executed if present, e.g. config/map/plat23.cfg.
```

PR #2126, opened 2022-08-30
improve listrotation display

bmorel:
Before:

2022-08-30-194410_1366x768_scrot

After:

2022-08-30-194502_1366x768_scrot

@cu-kai I know it does not fix much, but it's still a bit better when rotation have maaaaany maps :)

### DIFF
--- a/src/sgame/sg_maprotation.cpp
+++ b/src/sgame/sg_maprotation.cpp
@@ -845,24 +845,24 @@ void G_PrintCurrentRotation( gentity_t *ent )
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):
lgtm

cu-kai:
This is a lot better than before. Thanks @bmorel

bmorel:
yw. It's still bad, but the empty lines were really annoying, and it was easy to fix. On this patch, I had good hopes to not have endless discussions and nitpicks.

PR #2131, opened 2022-08-31
Queue some BP when a miner is built

slipher:
The build points are queued immediately when a miner begins construction.

Do you think having the cvar as a ratio is a good idea? The idea is that you could change g_buildPointBudgetPerMiner without ending up with an unreasonable value for how many BP are queued. But making the cvar value be just the number of BP to queue would be simpler.


src/sgame/sg_buildable.cpp
```diff
@@ -2004,6 +2009,13 @@ static gentity_t *SpawnBuildable( gentity_t *builder, buildable_t buildable, con
 			HealthComponent *healthComponent = built->entity->Get<HealthComponent>();
 			healthComponent->SetHealth(healthComponent->MaxHealth() * BUILDABLE_START_HEALTH_FRAC);
 		}
+
+		if (buildable == BA_A_LEECH || buildable == BA_H_DRILL) {
+			// Building a new miner incurs a queued BP penalty to punish teams who repeatedly
+			// lose and rebuild their miners in the late game.
+			level.team[builder->client->pers.team].queuedBudget +=
```
bmorel, line 2016:
```suggestion
			level.team[G_Team( builder )].queuedBudget +=
```

src/sgame/sg_buildable.cpp
```diff
@@ -35,6 +35,11 @@ static Cvar::Cvar<bool> g_indestructibleBuildables(
 		"g_indestructibleBuildables",
 		"make buildables impossible to destroy", Cvar::NONE, false);
 
+static Cvar::Range<Cvar::Cvar<float>> g_buildPointMinerQueueFraction(
```
bmorel, line 38: I can't help but think this name is very, very long. Would be nice to have shorter names, as I think  long names hurt readability (but ofc, I'd rather have long names than "a", "b", "c"...).

I understand that all BP-related cvars do use that kind of names, but I think it could be nice to do a PR to reduce "buildPoint" to "BP" after this is merged. BP is also the term everyone (players and devs) use, so I do not think it would be a problem, and I believe it would help mitigate the readability issue.

bmorel (commented):
> Do you think having the cvar as a ratio is a good idea? 

Definitely. An absolute number would not scale, and this would get in the way of people tweaking the game to find better balance as they would have to remember to tweak another value.

src/sgame/sg_buildable.cpp
bmorel: Actually, it seems we have a double standard on that point: there are cvars using the BP abbreviation:

* g_BPInitialBudget
* g_BPBudgetPerMiner
* g_BPRecoveryInitialRate
* g_BPRecoveryRateHalfLife

slipher:
Used the shorter cvar prefix and G_Team.

src/sgame/sg_buildable.cpp
```diff
@@ -35,6 +35,11 @@ static Cvar::Cvar<bool> g_indestructibleBuildables(
 		"g_indestructibleBuildables",
 		"make buildables impossible to destroy", Cvar::NONE, false);
 
+static Cvar::Range<Cvar::Cvar<float>> g_BPMinerQueueFraction(
+		"g_buildPointMinerQueueFraction",
```
bmorel, line 39: maybe also update the cvar name? It's not important though.

bmorel:
LGTM

src/sgame/sg_buildable.cpp
```diff
@@ -35,6 +35,11 @@ static Cvar::Cvar<bool> g_indestructibleBuildables(
 		"g_indestructibleBuildables",
 		"make buildables impossible to destroy", Cvar::NONE, false);
 
+static Cvar::Range<Cvar::Cvar<float>> g_BPMinerQueueFraction(
+		"g_BPMinerQueueFraction",
+		"upon miner construction, this fraction of the miner max yield is queued",
+		Cvar::NONE, 0.5, 0, 1);
```
DolceTriade, line 41: I think we should set this to zero for now until it gets some testing.

I think we need a means to change gameplay changes.

DolceTriade (approved):

src/sgame/sg_buildable.cpp
slipher: Well, one way to test is to merge it and then play on nightly or other servers that are rebasing frequently. But if some server owners agree to host it, that would work too.

I don't think we want to have a separate server for each gameplay change, especially for minor tweaks like this. It's like how on a major website they'll have 1000 A/B tests running at a time, and a given user might be in 100 of the experiment arms.

bmorel: > Well, one way to test is to merge it and then play on nightly or other servers that are rebasing frequently. But if some server owners agree to host it, that would work too.

It's interesting to read that from someone not maintaining a server.
I disagree on the idea that changes should be done silently, not on the actual change. Now, I'm one of the least active mod hosters. I think @sweet235  and @cu-kai  should be the ones talking on that topic.

Still, I'd like to remind you that upstream tend to refuse PRs that do change things, and tend to only accept opt-in, and often requires changes that might introduce bugs in code that's working perfectly in prod since months.
To me, and I really talk only about me here, it really feels like a 2 weights, 2 measures thing.

sweet235:
Using this, building a miner and immediately deconstructing it results in BP loss. I tested this using the value 0.5: I started a new game on map-plat23, and joined humans. Then I went to a place where a drill gives the full 50 BP. I built and immediately deconstructed 10 drills in a row. After that, I had around -185 BP.

Maybe this requires a modification to the code that handles miner deconstruction.

src/sgame/sg_buildable.cpp
necessarily-equal: To be honest, I get the impression this "changes that are opt-in" thing is very recent. I have the impression it started one or two months ago. And I'm not sure I like it to be honest. On one hand I would rather have non-broken gameplay for everyone and not just the servers which would have tweaked just the right number of "crutch" cvars even if opt-in changes can mean things are easier to get merged.

sweet235: I'll be glad to host any gameplay experiments. We need more of these if you ask me.

bmorel: >  have the impression it started one or two months ago. And I'm not sure I like it to be honest.

That's the only way people found to contribute without ending in endless political, fruitless discussions, is what I've been told in-game.

PR #2133, opened 2022-08-31
g_bot_defaultFill should default to 0

slipher:
As discussed on IRC, we now use g_bot_defaultFill in the sample configs (`dist/configs/config/server.cfg`), but we don't want bots to be added by default for a server with no config.

```
<perturbed> that's the *example* config though
<perturbed> you shouldn't just have bots automatically if you start a server without any commands
<freem> good point. Guess it needs a PR to address that then
<illwieckz> we can add the autofill cvar in sample server.cfg but this cvar set to 0 by default
<illwieckz> so as a developer you do devmap, no bot
<illwieckz> you're a server owner, you download the sample, done
<freem> yes, so this needs a PR to change that
```


bmorel:
LGTM

necessarily-equal (approved):

Issue #1925, opened 2022-07-19
Add credit sound when extinguishing fire


necessarily-equal:
To make the [credit reward](https://github.com/Unvanquished/Unvanquished/pull/1880) that granger get when they extinguish fire noticeable, we can add a bit of sound when they do so.

PR #2148, opened 2022-09-02
granger: add credit sound and bite mark when "killing" fire

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1925

### DIFF
--- a/src/sgame/sg_missile.cpp
+++ b/src/sgame/sg_missile.cpp
@@ -277,8 +277,10 @@ static int ImpactSlowblob( gentity_t *ent, trace_t *trace, gentity_t *hitEnt )
 
 	if ( reward )
 	{
+		// we extinguished some fire
 		float credits = g_firefightReward.Get() * CREDITS_PER_EVO * reward;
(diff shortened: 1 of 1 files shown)


DolceTriade:
What's a bite mark?

necessarily-equal:
The bite mark is the indicator in the center of the screen that tells you you did hit something

illwieckz:
I believe we call them “weapon indicator”. I'm not saying it's a better name, but you may find some resources with those keywords.

DolceTriade (approved):

illwieckz:
Does teamkilling also make the ding sound? If I'm right teamkilling also display the weapon indicator, meaning someone killing a fire may wrongly believe he teamkilled the building he tried to defend?

necessarily-equal:
I think teamkilling never does the ting sound. And buildings never do either. Of course, someone may not know that, but for the record with https://github.com/Unvanquished/Unvanquished/pull/1947 you shouldn't be able to teamkill a building anymore as granger

PR #2158, opened 2022-09-05
beacons: Don't allow opening beacon menu as a spec.

### LINKED ISSUE #2019: One can open beacon menu as spectator
One can open beacon menu as spectator.

Luckily spectator cannot set beacons.

DolceTriade:
Fixes #2019


necessarily-equal (approved):
Tested it works, thanks!

PR #2160, opened 2022-09-05
sgame: Remove a bunch of excess newlines

DolceTriade:


src/sgame/sg_active.cpp
```diff
@@ -709,9 +709,9 @@ bool ClientInactivityTimer( gentity_t *ent, bool active )
 			if( putSpec )
 			{
 				trap_SendServerCommand( -1,
-				                        va( "print_tr %s %s %s", QQ( N_("$1$^* moved from $2$ to spectators due to inactivity\n") ),
+				                        va( "print_tr %s %s %s", QQ( N_("$1$^* moved from $2$ to spectators due to inactivity.") ),
 				                            Quote( client->pers.netname ), Quote( BG_TeamName( client->pers.team ) ) ) );
-				G_LogPrintf( "Inactivity: %d\n", (int)( client - level.clients ) );
+				G_LogPrintf( "Inactivity: %d.", (int)( client - level.clients ) );
```
illwieckz, line 714: @slipher may disagree with that dot: https://github.com/Unvanquished/Unvanquished/pull/1529#discussion_r728536318

I'm fine with it.

DolceTriade: I don't feel strongly about periods. If @slipher requests them removed, I'll force push an update without them. Otherwise, I'll leave it as it is :)

necessarily-equal (approved):
Sounds good on principle, haven't tested

src/sgame/sg_active.cpp
slipher: Right, I prefer no periods, since it's the predominant convention in the codebase so adding periods puts it into a more inconsistent state. Plus, in this case, `Inactivity: %d` is not even a real sentence.

DolceTriade: Done.

PR #2161, opened 2022-09-05
cg_rocket_draw.cpp: cleanups

necessarily-equal:


src/cgame/cg_rocket_draw.cpp
```diff
@@ -1149,42 +1149,40 @@ class LocationElement : public HudElement
...
+			? CS_LOCATIONS + locent->currentState.generic1
+			: CS_LOCATIONS;
+			location_ = CG_ConfigString( CS_LOCATIONS );
+
+		SetInnerRML( Rocket_QuakeToRML( CG_ConfigString( cs_index ), RP_EMOTICONS ) );
 	}
 
 private:
-	Rml::String location;
+	Rml::String location_;
```
DolceTriade, line 1183: Looks unused?

src/cgame/cg_rocket_draw.cpp
```diff
@@ -1149,42 +1149,40 @@ class LocationElement : public HudElement
...
+	Util::optional<const centity_t *> lastlocation_;
```
DolceTriade, line 1184: optional: Could just store the CS index instead of the entity and avoid the optional, but its no big deal either way.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -921,22 +921,36 @@ class PositionElement : public TextHudElement
...
 		}
 
-		// Add text to be configured via CSS
-		vec3_t const & origin = cg.predictedPlayerState.origin;
-		SetText( va( "%0.0f %0.0f %0.0f", origin[0], origin[1], origin[2] ) );
+		shown_ = true;
+		glm::vec3 origin = VEC2GLM( cg.predictedPlayerState.origin );
+		if ( lastpos_ != origin )
+		{
+			SetText( Str::Format( "%0.0f %0.0f %0.0f",
```
DolceTriade, line 931: Use %g instead?

src/cgame/cg_rocket_draw.cpp
```diff
@@ -797,16 +797,6 @@ class SpeedGraphElement : public HudElement
 		currentSpeedElement->SetClass( "speed_current", true );
 	}
 
-	void OnPropertyChange( const Rml::PropertyIdSet& changed_properties ) override
```
DolceTriade, line 801: I don't think it really does any harm to support this, especially since the goal is to have user modifiable HUDs. The real thing is surfacing documentation on each of these custom elements, their attributes, and their styling properties...

DolceTriade (approved):
LGTM % minor nits

src/cgame/cg_rocket_draw.cpp
necessarily-equal: what the current rcss does is to add the background color on the parent element

src/cgame/cg_rocket_draw.cpp
necessarily-equal: The idea was to use integer rounding

src/cgame/cg_rocket_draw.cpp
necessarily-equal: good catch

src/cgame/cg_rocket_draw.cpp
necessarily-equal: integer rounding also makes it possible to have more stable display as in https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/154

DolceTriade (approved):

src/cgame/cg_rocket_draw.cpp
slipher: `optional` on a pointer is anyway not needed since you can use nullptr; it is already an "optional reference"

src/cgame/cg_rocket_draw.cpp
```diff
@@ -1365,8 +1362,8 @@ class LagometerElement : public HudElement
 public:
 	LagometerElement( const Rml::String& tag ) :
 			HudElement( tag, ELEMENT_GAME, true ),
-			shouldDrawLagometer( true ),
-			adjustedColor( Color::White )
+			shouldDrawLagometer_( true ),
+			adjustedColor_( Color::Alpha )
```
slipher: There's a mistake here in the trailing underscore commit (`Color::Alpha`)

necessarily-equal: I had fixed that but included the fix in the wrong commit. Did the magic to correct the first commit.

Issue #2151, opened 2022-09-04
micro-pounces


bmorel:
When try try to pounce, bots do not aim high enough, resulting to "micro-pounces" that just slows them down.

necessarily-equal:
Before #2162 goes into long debates on whether it fixes the issue or not. Is this issue about micro-pounce while fleeing or attacking?

bmorel:
Both situations are affected.

necessarily-equal:
I've updated the OP to specify that

bmorel:
Comments were marked as outdated because the OP was edited to reflect them. It's important to keep *why* you marked stuff as outdated. Also, I'm going to add the fact that the problem is not only retreating, but globally moving around: skilled players do use pounce and jump a lot to move and explore, and bots should be able to do that, too.

necessarily-equal:
> but pounce is used for other things, that could be kept in mind.

I mean yes but in master it doesn't, so I'm not going to have an issue that has a TODO saying "bot shouldn't do micro-pounce when they try to climb"… because they don't do that. Issues need to be actionable.

If someday someone implements pouncing to reach new parts of the map, then notice it has to be improved further, then let's act on that and open a specialised issue if needed. But I don't think this (bots doing micropounce when navigating) is a problem that we need to track now. I don't think navigating pounces are implemented, even in your fork.

About marking why a comment was thought outdated, unfortunately GitHub doesn't have the feature I think. You can't do better than choosing one simple reason out of 6
![Screenshot_20220912-094404-1](https://user-images.githubusercontent.com/59283660/189599816-ff93919a-5f06-4d9d-b740-7f08c4ab4bed.jpg)

bmorel:
> I mean yes but in master it doesn't, so I'm not going to have an issue that has a TODO saying "bot shouldn't do micro-pounce when they try to climb"… 

I was referring  to both attack and move. Move by pounce is currently pretty bad, because bots often pounce too far.

> About marking why a comment was thought outdated, unfortunately GitHub doesn't have the feature I think. You can't do better than choosing one simple reason out of 6

Just put a comment about the reason then. It's what's done by people in many, many circumstances. Also, you don't have to zoom 400% times your screenshots :D

necessarily-equal:
[...]

PR #2162, opened 2022-09-05
Bots: don't do micro-pounces

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/2151

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1761,7 +1761,9 @@ float CalcAimPitch( gentity_t *self, vec3_t targetPos, vec_t launchSpeed )
 
 	//convert to degrees (ps.viewangles units)
 	angle = RAD2DEG( angle );
-	return angle;
+
+	static constexpr float minimum_angle = 15.0f;
+	return std::max( angle, minimum_angle );
 }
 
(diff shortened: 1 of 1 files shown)


necessarily-equal:
Note 10° is also an acceptable angle. It may be more subtle.

Note also that this may break the usecase where a bot wants to jump on an enemy which has a lower z coord. It would often be broken anyway in that case because the bot would try to jump through the ground, but that's worth a mention.

bmorel:
I'm not fond of magic values. Why 10°? Or 15°?
Was this tested?

slipher:
It looks like when goons are doing movement along the navmesh, they always try to pounce toward `tpos` which is the next path corner. So if it is close, logically they would do a tiny pounce. If I'm right, the correct way to fix it would be not pouncing if `tpos` is too close.

bmorel:
The correct way would be to fix the pounce aiming.

bmorel:
> If I'm right, the correct way to fix it would be not pouncing if tpos is too close.

This would help, yes, probably need to replace pounce in those cases by a jump (since it allows to be faster than walking). Maybe this should be reserved to bots which have `skill >= 6` though (under this, they should just walk?).

necessarily-equal:
superseded by https://github.com/Unvanquished/Unvanquished/pull/2199

Issue #2168, opened 2022-09-06
Suggestion for a simpler implementation of color codes with Rml


illwieckz:
There is an inherent difficulty to implement Quake color codes with Rml as XML codes things with a pair of opening/closing tags (like `<span style="color: red;">a</span>` while the Quake color codes set a state for the rest of the text. Translating `^1` to something like `<span style="color: red;">` (or any other XML variant) will make us produce many opening tags without closing tags.

To prevent that, we [currently track the opened tags to close them](https://github.com/Unvanquished/Unvanquished/blob/a6e9f5a598e5128d5aaec06393f35804cdd3e38a/src/cgame/rocket/rocket.cpp#L586-L607):

```c++
```

One idea that does not require to track any state is to always set the string within an opening and a closing tag of the same kind the one is used to switch the color, and always set a closing tag before adding an opening tag when setting a color.

For example if we have this Rml file:

```xml
<translate quake>default ^1red ^2green ^4blue</translate>
```

We would first translate it to that in game:

```xml
<translate quake><span>default ^1red ^2green ^4blue</span></translate>
```

Then translate every color code to a closing span followed by the opening span, something like `</span><span style="color:#FF0000">`.

Like this:

```
<translate quake><span>default </span><span style="color:#FF0000">red </span><span style="color:#00FF00">green </span><span style="color:#0000FF">blue</span></translate>
```

A reset color code may be implemented by replacing every `^*` with `</span><span>`.

Also we may want to use a custom tag that is not `<span>` to be sure to never conflict with any existing span or other in-game feature producing span.

This way the coloring would be completely stateless with properly enclosed XML tags and the code would be simpler.

Note: this simple implementation only works because we don't have other formatting codes like bold or italic, only one code is applied for a given character (translating IRC codes to XML is likely a nightmare XD).

illwieckz:
[This code](https://github.com/Unvanquished/Unvanquished/blob/a6e9f5a598e5128d5aaec06393f35804cdd3e38a/src/cgame/rocket/rocket.cpp#L555-L560) is likely to break coloring on line break:

```c++
			else if ( c == '\n' )
			{
				out.append( span && spanHasContent ? "</span><br />" : "<br />" );
				span = false;
				spanHasContent = false;
			}
```

The proposed implementation may entirely remove the need to take care of span (a `<br/>` is valid within a span).

DolceTriade:
[...]

PR #2169, opened 2022-09-06
rocket: add a <text> tag that supports “quake” and “emoticons” properties like <translate>

illwieckz:
Add a `<text>` tag that supports `quake` and `emoticons` properties like `<translate>`, but this will not produce translatable strings in `.pot`. This can be useful when displaying user-generated content.

See https://github.com/Unvanquished/Unvanquished/issues/2168#issuecomment-1237589454 by @DolceTriade.

I named this tag `<text>` and it takes `quake` and `emoticons` properties to not have a confusing behaviour differing from `<translate>`.


slipher:
> This can be useful when displaying user-generated content.

It might be useful for displaying some hard-coded content with a lot of emoticons or something. But it is not suitable for user-generated content. How are you planning to use it?

src/cgame/cg_rocket_draw.cpp
```diff
@@ -75,6 +75,46 @@ void Rocket_UpdateLanguage()
...
+			{
+				flags |= RP_QUAKE;
+			}
+
+			if ( HasAttribute( "emoticons" ) )
+			{
+				flags |= RP_EMOTICONS;
+			}
+
+			if ( flags != 0 )
```
DolceTriade, line 104: This behavior is not the same: in the first, inner children will be escaped properly and output as such, in the latter, they will be parsed. I think we should run it through Rocket_QuakeToRML in either case.

<text> is basically akin to <pre>.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -3705,6 +3745,7 @@ void CG_Rocket_RegisterElements()
 	RegisterElement<BeaconOwnerElement>( "beacon_owner" );
 	RegisterElement<PredictedMineEfficiencyElement>( "predictedMineEfficiency" );
 	RegisterElement<BarbsHudElement>( "barbs" );
+	RegisterElement<CustomizableTextElement>( "text" );
```
DolceTriade, line 3748: I still think we should call it `<quake>` or `<qtext>` and have the quake portion implied. Otherwise, it seems kinda pointless? If you want emoticons, you can just do ``.

src/cgame/cg_rocket_draw.cpp
illwieckz: Ah yes, forgot the requirement to escape things.

illwieckz: Or the escaping can be a property.

illwieckz: Hmm, now I notice that our `translate` tag does not escape when there is no `quake` and `emoticon` property, but we may want to not escape it to be able to translate strings like: `^1you have ^2<credits /> [bsuit]`, with such string the code should not be escaped.

So, maybe we better want to add another flag for escaping or not.

slipher: Yes it's intentional that you could use a bold tag inside the translation for example.

illwieckz: I just think we better rename `Rocket_QuakeToRML` to something more generic at some point, and just make it enable some text processing by observing flags like `RP_EMOTICONS`, `RP_QUAKE` and `RP_ESCAPE`.

PR #2170, opened 2022-09-06
RML: Rewrite QuakeToRML

illwieckz:
- use a less convoluted stateless algorithm as described in  
  https://github.com/Unvanquished/Unvanquished/issues/2168
- rewrite html escaping using `std::unordered_map`
- escape more html characters to fully escape things like `<tag p1='c1' p2="c2>`
- fix what looks to be a buffer overflow


src/cgame/rocket/rocket.cpp
```diff
@@ -514,8 +514,21 @@ Rml::String Rocket_QuakeToRML( const char *in, int parseFlags = 0 )
 {
 	Rml::String out;
 	Rml::String spanstr;
-	bool span = false;
-	bool spanHasContent = false;
+
+	// HACK: Guess when a string needs RP_QUAKE.
```
DolceTriade, line 534: Why don't we just remove RP_QUAKE? That's literally the point of this function anyways...

src/cgame/rocket/rocket.cpp
```diff
@@ -525,99 +538,60 @@ Rml::String Rocket_QuakeToRML( const char *in, int parseFlags = 0 )
...
-				if ( span && !spanHasContent )
-				{
-					spanHasContent = true;
-					out.append( spanstr );
-				}
-				out.append( "&amp;" );
-			}
-			else if ( c == '\n' )
+
+			std::unordered_map<char, std::string> htmlEscapeStrings;
```
DolceTriade: Declare this as const static and statically intialize the kv?

src/cgame/rocket/rocket.cpp
```diff
@@ -525,99 +538,60 @@ Rml::String Rocket_QuakeToRML( const char *in, int parseFlags = 0 )
...
+			htmlEscapeStrings['<'] = "&lt;";
+			htmlEscapeStrings['>'] = "&gt;";
+			htmlEscapeStrings['&'] = "&amp;";
+			htmlEscapeStrings['"'] = "&quot;";
+			htmlEscapeStrings['\''] = "&apos;";
+			htmlEscapeStrings['\n'] = "<br />";
+
+			std::string htmlEscapedString = htmlEscapeStrings[ c ];
+
+			if ( htmlEscapedString != "" )
```
DolceTriade, line 563: use .empty()? or do

```
auto it = htmlEscapeStrings.find( c );
if ( htmlEscapeStrings != htmlEscapeStrings.end() )
{
    out.append( it.second );
}
else
{
    // handle emoticon
}

File: src/cgame/rocket/rocket.cpp [hunk 4]
```diff
@@ -525,99 +538,60 @@ Rml::String Rocket_QuakeToRML( const char *in, int parseFlags = 0 )
...
-				if ( span && !spanHasContent )
-				{
-					out.append( spanstr );
-					spanHasContent = true;
-				}
 				out.append( token.Begin(), token.Size() );
 			}
 		}
-		else if ( token.Type() == Color::Token::TokenType::COLOR )
+		else if ( quake && ( token.Type() == Color::Token::TokenType::COLOR ) )
```
DolceTriade, line 580: There is a behavior change here: we will add empty spans, ie

`^1^2Green` in the previous code led to:

```
<span style="color: green;">Green</span>
```

In the new code, this will be:

```
<span style="color: red;"></span><span style="color: green;">Green</span>
```

Should probably avoid doing this else people might abuse `^1^2^3^4^5^6hahah`

DolceTriade (changes requested):

illwieckz:
I noticed there was another code to escape HTML characters (and also missing `'` and `"`) so I deduplicated the code.

src/cgame/rocket/rocket.cpp
illwieckz: 1. Some strings passed to this function should not be translated to quake markups. For example the number field in the server browser will be buggy if we add some markup there.
2. I also tested enabling `RP_QUAKE` anytime `RP_EMOTICONS` is enabled and this makes the game crash.
3. It's possible to use the `<translate>` tag without setting the `quake` property and only translate emoticons, not strings that look like color codes.

illwieckz: The game will crash if we pass `RP_QUAKE | RP_EMOTICONS` instead of `RP_EMOTICONS` there:

https://github.com/Unvanquished/Unvanquished/blob/a6e9f5a598e5128d5aaec06393f35804cdd3e38a/src/cgame/rocket/rocketDataGrid.h#L57

src/cgame/rocket/rocket.cpp
illwieckz: Well, if we want to avoid that abuse, we would need convoluted code.

src/cgame/rocket/rocket.cpp
DolceTriade: Sounds like a bug?

src/cgame/rocket/rocket.cpp
illwieckz: I get:

```
error: ‘struct std::__detail::_Node_const_iterator<std::pair<const char,
  std::__cxx11::basic_string<char> >, false, false>’ has no member named ‘second’
```

illwieckz: Also the compiler doesn't let me make an `unordered_map` with a `const char`.

illwieckz: More precisely, in this situation the compiler does not allow me to make a `const unordered_map` with a `char` (but it also does not allow to make a an `unordered_map` with a `const char`).

I turned it into a static global I never modify anyway.

illwieckz: Also now uses `.empty()`.

src/cgame/rocket/rocket.cpp
illwieckz: The crash is:

```
Thread 56 "daemon" received signal SIGFPE, Arithmetic exception.
0x00007ffefd81af2a in DisplayAspectString (w=0, h=0) at Unvanquished/src/cgame/cg_rocket_dataformatter.cpp:55
55		w /= gcd;

Thread 56 (Thread 0x7ffefffff640 (LWP 1287778) "daemon"):
#0  0x00007ffefd81af2a in DisplayAspectString (w=0, h=0) at Unvanquished/src/cgame/cg_rocket_dataformatter.cpp:55
#1  0x00007ffefd81affd in CG_Rocket_DFResolution (handle=7, data=0x7ffefed9aec0 <CG_Rocket_FormatData(int)::data> "\\1\\<span>3840</span>\\2\\<span>2160</span>") at Unvanquished/src/cgame/cg_rocket_dataformatter.cpp:78
#2  0x00007ffefd81bbb4 in CG_Rocket_FormatData (handle=7) at Unvanquished/src/cgame/cg_rocket_dataformatter.cpp:310
#3  0x00007ffefd8a7dc7 in RocketDataFormatter::FormatData (this=0x7ffef5077fb0, formatted_data="<span>3840</span>", raw_data=std::vector of length 2, capacity 2 = {...}) at Unvanquished/src/cgame/rocket/rocketDataFormatter.h:62
#4  0x00007ffefda3ee6a in Rml::ElementFormControlDataSelect::BuildOptions (this=0x7ffef52c3210) at Unvanquished/libs/RmlUi/Source/Core/Elements/ElementFormControlDataSelect.cpp:207
#5  0x00007ffefda3e677 in Rml::ElementFormControlDataSelect::OnRowAdd (this=0x7ffef52c3210, UNUSED_data_source=0x7ffef5068530, table="default", UNUSED_first_row_added=0, UNUSED_num_rows_added=1) at Unvanquished/libs/RmlUi/Source/Core/Elements/ElementFormControlDataSelect.cpp:117
#6  0x00007ffefda2fb8e in Rml::DataSource::NotifyRowAdd (this=0x7ffef5068530, table="default", first_row_added=0, num_rows_added=1) at Unvanquished/libs/RmlUi/Source/Core/Elements/DataSource.cpp:121
#7  0x00007ffefd8a97cb in RocketDataGrid::AddRow (this=0x7ffef5068530, table=0x7ffefdc72b57 "default", dataIn=0x7ffeffffbd30 "\\width\\3840\\height\\2160") at Unvanquished/src/cgame/rocket/rocketDataGrid.h:69
#8  0x00007ffefd8a9035 in Rocket_DSAddRow (name=0x7ffefdc72b5f "resolutions", table=0x7ffefdc72b57 "default", data=0x7ffeffffbd30 "\\width\\3840\\height\\2160") at Unvanquished/src/cgame/rocket/rocket_datagrid.cpp:74
#9  0x00007ffefd81e3ec in CG_Rocket_BuildResolutionList () at Unvanquished/src/cgame/cg_rocket_datasource.cpp:565
#10 0x00007ffefd822902 in CG_Rocket_BuildDataSource (dataSrc=0x7ffefe170b40 <CG_Argv(int)::buffer> "resolutions", table=0x7ffeffffc180 "") at Unvanquished/src/cgame/cg_rocket_datasource.cpp:1905
#11 0x00007ffefd845c9f in CG_Rocket_BuildDS () at Unvanquished/src/cgame/cg_rocket_events.cpp:93
#12 0x00007ffefd846575 in CG_Rocket_ProcessEvents () at Unvanquished/src/cgame/cg_rocket_events.cpp:333
#13 0x00007ffefd8198b9 in CG_Rocket_Frame (state=...) at Unvanquished/src/cgame/cg_rocket.cpp:554
#14 0x00007ffefd7aa08e in operator() (__closure=0x7ffeffffe877, cs=...) at Unvanquished/src/cgame/cg_api.cpp:82
#15 0x00007ffefd7acc4b in Util::apply_impl<VM::VMHandleSyscall(uint32_t, Util::Reader)::<lambda(cgClientState_t)>, std::tuple<cgClientState_t&&>, 0>(struct {...} &&, std::tuple<cgClientState_t&&> &&, Util::seq<0>) (func=..., tuple=...) at Daemon/src/common/Util.h:125
#16 0x00007ffefd7ac781 in Util::apply<VM::VMHandleSyscall(uint32_t, Util::Reader)::<lambda(cgClientState_t)>, std::tuple<cgClientState_t&&> >(struct {...} &&, std::tuple<cgClientState_t&&> &&) (func=..., tuple=...) at Daemon/src/common/Util.h:130
#17 0x00007ffefd7ab952 in IPC::detail::HandleMsg<VM::VMHandleSyscall(uint32_t, Util::Reader)::<lambda(cgClientState_t)>, IPC::Message<IPC::Id<0, 9>, cgClientState_t>, IPC::Reply<> >(IPC::Channel &, IPC::SyncMessage<IPC::Message<IPC::Id<0, 9>, cgClientState_t>, IPC::Reply<> >, Util::Reader, struct {...} &&) (channel=..., reader=..., func=...) at Daemon/src/common/IPC/Channel.h:217
#18 0x00007ffefd7aaba2 in IPC::HandleMsg<IPC::SyncMessage<IPC::Message<IPC::Id<0, 9>, cgClientState_t> >, VM::VMHandleSyscall(uint32_t, Util::Reader)::<lambda(cgClientState_t)> >(IPC::Channel &, Util::Reader, struct {...} &&) (channel=..., reader=..., func=...) at Daemon/src/common/IPC/Channel.h:241
#19 0x00007ffefd7aa411 in VM::VMHandleSyscall (id=9, reader=...) at Unvanquished/src/cgame/cg_api.cpp:81
#20 0x00007ffefd959cc1 in CommonInit (rootSocket=32) at Daemon/src/shared/VMMain.cpp:66
#21 0x00007ffefd959e49 in vmMain (rootSocket=32) at Daemon/src/shared/VMMain.cpp:104
#22 0x000055555587674d in operator() (__closure=0x555557e8c228) at Daemon/src/engine/framework/VirtualMachine.cpp:379
#23 0x0000555555877e84 in std::__invoke_impl<void, VM::CreateInProcessNativeVM(std::pair<IPC::Socket, IPC::Socket>, Str::StringRef, VM::VMBase::InProcessInfo&)::<lambda()> >(std::__invoke_other, struct {...} &&) (__f=...) at /usr/include/c++/11/bits/invoke.h:61
#24 0x0000555555877e47 in std::__invoke<VM::CreateInProcessNativeVM(std::pair<IPC::Socket, IPC::Socket>, Str::StringRef, VM::VMBase::InProcessInfo&)::<lambda()> >(struct {...} &&) (__fn=...) at /usr/include/c++/11/bits/invoke.h:96
#25 0x0000555555877df4 in std::thread::_Invoker<std::tuple<VM::CreateInProcessNativeVM(std::pair<IPC::Socket, IPC::Socket>, Str::StringRef, VM::VMBase::InProcessInfo&)::<lambda()> > >::_M_invoke<0>(std::_Index_tuple<0>) (this=0x555557e8c228) at /usr/include/c++/11/bits/std_thread.h:253
#26 0x0000555555877dc8 in std::thread::_Invoker<std::tuple<VM::CreateInProcessNativeVM(std::pair<IPC::Socket, IPC::Socket>, Str::StringRef, VM::VMBase::InProcessInfo&)::<lambda()> > >::operator()(void) (this=0x555557e8c228) at /usr/include/c++/11/bits/std_thread.h:260
#27 0x0000555555877dac in std::thread::_State_impl<std::thread::_Invoker<std::tuple<VM::CreateInProcessNativeVM(std::pair<IPC::Socket, IPC::Socket>, Str::StringRef, VM::VMBase::InProcessInfo&)::<lambda()> > > >::_M_run(void) (this=0x555557e8c220) at /usr/include/c++/11/bits/std_thread.h:211
#28 0x00007ffff76ba694 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#29 0x00007ffff734e947 in start_thread (arg=<optimized out>) at pthread_create.c:435
#30 0x00007ffff73dea44 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:100
```

DolceTriade: Divide by 0 is def a bug :D

src/cgame/rocket/rocket.cpp
DolceTriade: Does the iterator support lookahead? (Or can we go backwards?) Can we do something like

```c++
++iter;
if (token.Type() == Color::Token::TokenType::COLOR)
{
  continue;
}
--iter;
char rgb[38];
...
```

src/cgame/rocket/rocket.cpp
illwieckz: But this division by 0 seems to be a consequence of adding rml tags when processing text, so maybe we divide number values processed from rml content or something like that? In such case, even setting RP_EMOTICONS would be wrong.

src/cgame/rocket/rocket.cpp
DolceTriade: I always get mixed up. Maybe `it->second`?

src/cgame/rocket/rocket.cpp
illwieckz: Right, we we may be able to look at the next token.

src/cgame/rocket/rocket.cpp
slipher: RP_QUAKE is completely ignored by the existing code, so if you had crashes from adding it, it must be due to other modifications of the code.

Removing the RP_QUAKE flag and always parsing colors sounds good to me. If you want to have text without color or emoticon parsing, you should use CG_EscapeHtml.

src/cgame/rocket/rocket.cpp
slipher: The hash table approach probably has worse performance than if-else and is harder to read in my view.

src/cgame/rocket/rocket.cpp
illwieckz: The crash probably comes from the fact I always add a `<span>` tag, which probably means somewhere some code does math computation from values taken from Rml text. If that's true, people may try to abuse such rml strings to make the game crash. The server of Gireen is already showcasing some examples of bad data being feed as server properties (name, player count, etc.), what if someones manages to slip in a character in a server property that will make the game crash on listing servers?

illwieckz: The division by zero may be a division by NaN.

src/cgame/rocket/rocket.cpp
```diff
@@ -486,30 +486,32 @@ void Rocket_Update()
...
+static std::unordered_map<char, std::string> htmlEscapeStrings = {
+	{ '<', "&lt;" },
+	{ '>', "&gt;" },
+	{ '&', "&amp;" },
+	{ '"', "&quot;" },
+	{ '\'', "&apos;" },
+};
 
-	std::string htmlEscapedString = htmlEscapeStrings[ c ];
+static std::string CG_EscapeHTMLChar( const char c, bool reuse = true, bool endline = false )
```
slipher, line 497: Constructing a std::string for every single character is going to make the code slower. There's nothing wrong with having a few lines of "duplicated" code in cases where the code is simple and rarely/never needs to be changed.

PR #2171, opened 2022-09-06
make slap velocity independent of mass

sweet235:
Slap is supposed to give a player (or bot) a random velocity boost. This should not depend on the player's mass, as we are looking at velocity, and not newtonian momentum.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -2381,7 +2381,7 @@ bool G_admin_slap( gentity_t *ent )
 
 	if ( trap_Argc() < 2 )
 	{
-		ADMP( QQ( N_( "^3slap:^* usage: slap [^3name|slot#|admin#^7] (^3damage^7)" ) ) ); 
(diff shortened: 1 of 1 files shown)


src/sgame/sg_admin.cpp
```diff
@@ -2420,9 +2420,8 @@ bool G_admin_slap( gentity_t *ent )
 	// using knockback in ent->Damage() won't throw the victim
 	// about if the damage caused to them is 0.
 	vec3_t kvel;
-	float mass = BG_Class( vic->client->pers.classSelection )->health;
```
DolceTriade, line 2423: To be clear, slap velocity is dependent on *health* not mass, which is another thing. Imo, I would rather just have slap take another param for velocity and by default use the existing velocity.

sweet235: Yeah, I think health was just an approximation for mass in that case. Problem with second parameter is that the second (optional) parameter already reduces health, that is important to kill stuck bots. Let me see what I can do...

DolceTriade: Add a 3rd param? `/slap badbot 0 100000`

sweet235: I agree.

PR #2172, opened 2022-09-06
rmlui: Fix inconsistent coloring on names after Follow text.

### LINKED ISSUE #2163: Inconsistentency in name colors
Reported by @sweet235 :

DolceTriade:
Fixes #2163

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -2906,8 +2906,8 @@ static void CG_Rocket_DrawMinimap()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


illwieckz (approved):
LGTM.

necessarily-equal:
I'm afraid that if people do `^*` it will give something weird. On the plus side, ^* is not documented so in a way nobody expects it works correctly

necessarily-equal:
So I had a doubt and checked. `^*` *is* documented, whoops.

Still better than it was before though

DolceTriade:
I changed from ^* to ^7 so its fine

PR #2174, opened 2022-09-06
rocket: prevent overflow

illwieckz:
Commit from #2170

This can be merged independently.


DolceTriade (approved):

slipher:
BG_EmoticonAt only returns true if there is a `]` somewhere. There is no bug.

slipher:
Although there is no bug, this is OK as an extra safety check.

PR #2175, opened 2022-09-07
Small tweaks in BotEvolve and indent in G_NewEntity

necessarily-equal:


DolceTriade (approved):

DolceTriade:
Do we even try to evolve to the same class?

necessarily-equal:
> Do we even try to evolve to the same class?

We did, but it was blocked in bot code by `BotCanEvolveToClass` because it was considered "devolving". If it wasn't blocked we could have a bot evolving (e.g.) from dragoon to adv mara to dragoon or from dragoon to dretch to mantis to … to dragoon every frame. But now it's blocked *and* marked successful if you already have it. This is what (e.g.) `BotBuyWeapon` does.

bmorel:
Note that `BotBuyWeapon` does not prevents buying less evolved weapons, while the evolution code does prevent devolving, while it should not, as since 0.52 we allow players to devolve.
But, it's not a problem, because this requires a specific behavior to be done, that was not implemented yet.

PR #2178, opened 2022-09-08
graphics: avoid ET_LEV2_ZAP_CHAIN undefined behavior

necessarily-equal:
Found by gcc:

```
/build/source/src/cgame/cg_trails.cpp:1514:14: warning: iteration 5 invokes undefined behavior
 1514 |         if ( *ts == nullptr || ( *ts && !( *ts )->valid ) )
      |              ^
/build/source/src/cgame/cg_ents.cpp:1111:40: note: within this loop
 1111 |                         for ( i = 0; i <= LEVEL2_AREAZAP_MAX_TARGETS; i++ )
      |                                        ^
```

(using https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Waggressive-loop-optimizations-Waggressive-loop-optimizations)

level2ZapTS has size LEVEL2_AREAZAP_MAX_TARGETS, and not one more.

### DIFF
--- a/src/cgame/cg_ents.cpp
+++ b/src/cgame/cg_ents.cpp
@@ -1108,7 +1108,7 @@ static void CG_CEntityPVSLeave( centity_t *cent )
 	switch ( es->eType )
 	{
 		case entityType_t::ET_LEV2_ZAP_CHAIN:
-			for ( i = 0; i <= LEVEL2_AREAZAP_MAX_TARGETS; i++ )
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):
I wonder if this is the source of some of our weird zap graphics bugs

necessarily-equal:
> I wonder if this is the source of some of our weird zap graphics bugs

I'm afraid that this shouldn't, the undefined behavior should hit muzzleTS (another trail) instead of level2ZapTS. But maybe, I only had a quick look

PR #2180, opened 2022-09-08
GLM related stuff

necessarily-equal:
This isn't a backport of commits, but some cleanups from me

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1218,7 +1218,7 @@ bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target )
 				range = glm::distance( muzzle, proj );
 
 				// make sure the sign of the range is correct
-				glm::vec3 rdir = glm::normalize( npos - muzzle );
+				glm::vec3 rdir = npos - muzzle;
 				if ( glm::dot( rdir, dir ) < 0 )
 				{
 					range = -range;
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1734,7 +1734,7 @@ float CalcAimPitch( gentity_t *self, glm::vec3 &targetPos, float launchSpeed )
 	glm::vec3 startPos;
 	glm::vec3 forward, right, up;
 	glm::vec3 muzzle;
-	float angle1, angle2, angle;
+
 	float g = self->client->ps.gravity;
 	float v = launchSpeed;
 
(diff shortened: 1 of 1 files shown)


src/sgame/sg_bot_util.cpp
```diff
@@ -1212,7 +1212,7 @@ bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target )
 				range = glm::distance( muzzle, proj );
 
 				// make sure the sign of the range is correct
-				glm::vec3 rdir = glm::normalize( npos - muzzle );
+				glm::vec3 rdir = npos - muzzle;
```
DolceTriade, line 1215: Why is this safe?  Is it because we only care about the sign of the dot product rather than the value? Let's comment about this.

src/sgame/sg_bot_util.cpp
```diff
@@ -1725,45 +1725,36 @@ void BotClassMovement( gentity_t *self, bool inAttackRange )
...
-	y = targetPos[2];
-	v = launchSpeed;
-	g = self->client->ps.gravity;
-
-	//make sure we won't get NaN
-	check = Square( Square( v ) ) - g * ( g * Square( x ) + 2 * y * Square( v ) );
+	// Project everything onto a 2D plane with initial position at (0,0)
+	// dz (Δz) is the difference in height and dr (Δr) the distance forward
+	float dz = targetPos[2] - startPos[2];
+	float dr = glm::length(glm::vec2(targetPos) - glm::vec2(startPos));
```
DolceTriade, line 1741: Does glm::vec2 of a vec3 basically drop the Z value?

src/sgame/sg_bot_util.cpp
```diff
@@ -1752,16 +1751,12 @@ float CalcAimPitch( gentity_t *self, glm::vec3 &targetPos, float launchSpeed )
 		check = Square( Square( v ) ) - g * ( g * Square( dr ) + 2 * dz * Square( v ) );
 	}
 
-	//calculate required angle of launch
-	angle1 = atanf( ( Square( v ) + sqrtf( check ) ) / ( g * dr ) );
-	angle2 = atanf( ( Square( v ) - sqrtf( check ) ) / ( g * dr ) );
+	// the same equation with a + sign would also give a correct result,
```
DolceTriade, line 1754: Is this true? Are there no edge cases which lead to the previous line?

src/sgame/sg_bot_util.cpp
necessarily-equal: Yes exactly. I don't think it makes sense to add a comment in the code though. I guess I can put one in the commit but tbh someone discovering the code shouldn't be helped by that

src/sgame/sg_bot_util.cpp
necessarily-equal: following along with explainations:

```
            sqrt(c)              >             −sqrt(c)                 because sqrt is a positive value
       v² + sqrt(c)              >         v² − sqrt(c)                 we add the same thing on both side
      (v² + sqrt(c))/(g × x)     >        (v² − sqrt(c))/(g × dr)       we divide by a *positive* value on both side
atan( (v² + sqrt(c))/(g × x) )   >  atan( (v² − sqrt(c))/(g × dr) )     atan is strictly monotonous and growing: atan(1)>atan(0)>atan(-0.01) etc
```

necessarily-equal: So the value with the − is always smaller. And if c=0, then this becomes an equality and the result is the same on both sides.

Keeping the information around is practical though, because I might do some tracing to see if the path is actually doable (without obstacles)

necessarily-equal:
rebased

necessarily-equal:
I've checked this is equivalent to the previous version by copying the previous version of the function and adding an assert (not added in this PR). So I'm confident enough that this is fine

src/sgame/sg_bot_util.cpp
DolceTriade: I personally do like these descriptions in the commit message at the very least. I helps people understand *why* certain changes were made. They are discoverable via blame too

DolceTriade (approved):

PR #2184, opened 2022-09-08
GLM: misc small stuff

necessarily-equal:
cherry-picked from bmorel's branch, specifically the parts which don't need more functions

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1508,19 +1508,6 @@ void BotSlowAim( gentity_t *self, glm::vec3 &target, float slowAmount )
 	target = viewBase + length * skilledVec;
 }
 
-float BotAimAngle( gentity_t *self, vec3_t pos )
-{
-	vec3_t viewPos;
-	vec3_t forward;
-	vec3_t ideal;
-
-	AngleVectors( self->client->ps.viewangles, forward, nullptr, nullptr );
-	BG_GetClientViewOrigin( &self->client->ps, viewPos );
-	VectorSubtract( pos, viewPos, ideal );
-
-	return AngleBetweenVectors( forward, ideal );
-}
-
 /*
 ========================
 Bot Team Querys
--- a/src/sgame/sg_spawn_gfx.cpp
+++ b/src/sgame/sg_spawn_gfx.cpp
@@ -35,6 +35,7 @@ Maryland 20850 USA.
 #include "sg_local.h"
 #include "sg_spawn.h"
 
+#include <glm/geometric.hpp>
 /*
 ======================================================================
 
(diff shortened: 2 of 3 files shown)


src/shared/bg_public.h
```diff
@@ -1811,8 +1811,11 @@ void CompleteCommand(int argNum);
 glm::vec3 BG_GetClientNormal( const playerState_t *ps );
 glm::vec3 BG_GetClientViewOrigin( const playerState_t *ps );
 
-void BG_BoundingBox( class_t cl, glm::vec3* mins, glm::vec3* maxs, glm::vec3* cmaxs, glm::vec3* dmins, glm::vec3* dmaxs );
-void BG_BoundingBox( buildable_t buildablel, glm::vec3* mins, glm::vec3* maxs );
+void BG_BoundingBox( class_t cl, glm::vec3 &mins, glm::vec3 &maxs );
+void BG_DeadBoundingBox( class_t cl, glm::vec3 &mins, glm::vec3 &maxs );
+glm::vec3 BG_CrouchBoundingBox( class_t cl );
+
+void BG_BoundingBox( buildable_t buildablel, glm::vec3 &mins, glm::vec3 &maxs );
```
DolceTriade, line 1818: I think we should avoid overloading this function and just call BG_BuildableBoundingBox. WDYT
 
Overloading is kinda weird in this context because CG_Crouching an =d CG_Dead boundingbox dont make sense for buildables

necessarily-equal: I agree that I'm unsure this makes sense to overload the function. Especially since you can't do `BG_BoundingBox( ps->stats[ STAT_CLASS ] )` without a static_cast. I'm unsure if @bmorel has an opinion, but it would probably include "yeah but I'm doing it like that and it would be annoying to switch back". Not sure if he saw it practical or not.

bmorel: > Overloading is kinda weird in this context because CG_Crouching and CG_Dead boundingbox dont make sense for buildables

I don't know which commits were merged or not, but there's a commit that removes the crouchmax and deadmins/maxs from the "normal one".
This allows a more uniform API, and more performances since:

* crouch parameter is used a handful of times
* dead parameters are used only once
* the stuff is mostly used with mins/maxs, which means there's a lot of useless checks happening around

So the solution I had was to have different, specialized functions  for dead and crouch. The only thing which is arguable here is to get the mins and maxs params passed to the function, instead of returned in a pair, but initially I wanted to get that stuff merged quickly so choose the less invasive way.

> Especially since you can't do BG_BoundingBox( ps->stats[ STAT_CLASS ] ) without a static_cast

The problem here is that BG_BoundingBox should be able to take a playerState_t as parameter. So, yes, there's an API problem, but it's not related to required cast.
Fixing the API on that is really out of the scope of the GLM  or betterai changesets though.

> but it would probably include "yeah but I'm doing it like that and it would be annoying to switch back". 

And I think this is a very valid argument. I've been repeatedly told that contributors on unvanquished are not paid. This applies to me, and imo those who do on their free-time should be given some liberties on that kind of changes, especially when it is in such a big changeset.
But, no, this is not the reason behind this change. Sorry to disappoint you.

slipher: I don't think overloading per se is a problem, but it's bad that the new function has a different name from BG_BuildableBoundingBox. If we're going to have 2 versions of a function for different vector APIs, they should at least have the same name so it's clear they are supposed to do the same thing.

bmorel: to me, BG_BuildableBoundingBox and BG_PlayerBoundingBox should be eradicated. They are inherited from C's lack of safety, so they're very verbose while giving absolutely 0 safety. You could pass them the constant 0xFFFF and receive no warning.
Even worst, BG_PlayerBoundingBox takes 3 arguments that are used at most a handful of times, but still need to be checked each call.
Now, I explained why I did that function, which as you can see is not «@bmorel has an opinion, but it would probably include "yeah but I'm doing it like that and it would be annoying to switch back".»

slipher: How about keeping the new function but naming it BG_BuildableBoundingBox?

necessarily-equal:
I've removed the contentious commit from this PR. This should already get three more commits merged.

slipher:
LGTM

PR #2185, opened 2022-09-08
GLM: last functions

necessarily-equal:
This is probably about the last PR on that after https://github.com/Unvanquished/Unvanquished/pull/2184.

It duplicates 2-3 functions that exist in daemon and converts the calling functions.


DolceTriade (approved):

necessarily-equal:
Note to self: a submodule change slipped in the "switch sg_admin.cpp to glm" commit, to fix before merging

PR #2187, opened 2022-09-09
Bot votes

sweet235:
Votes to fill one or both teams with bots to a given number. The votes are called `fillbots`, `fillbots_humans` and `fillbots_aliens`. The reasons for having 3 different votes are:

- There are two seperate cvars for the required percentages: one for `fillbots` and one for `fillbots_*`
- The `reason` vote argument is not used, and can later be extended to denote the bot skill

There is a cvar `g_maxVoteFillBots` to set an upper bound on how many bots can be added by the votes (10 by default).

What do you think about the default values for the cvars? I require 67% for a vote that fills only one team. This way, abuse in a 2v1 game should be ~~impossible~~ rare. All three votes require the quorum.

I would have preferred to use `botfill` instead of `fillbots`, for symmetry with the `bot fill` command. But there is already a vote called `kickbots`.


src/sgame/sg_cmds.cpp
```diff
@@ -1525,22 +1525,25 @@ static const struct {
...
+	{ "map_restart",     true,  V_PUBLIC, T_NONE,    false,  true,     qtrinary::qno,    &g_mapVotesPercent, VOTE_ALWAYS, nullptr, nullptr },
+	{ "map",             true,  V_PUBLIC, T_OTHER,   false,  true,     qtrinary::qmaybe, &g_mapVotesPercent,         VOTE_BEFORE, &g_mapVotesBefore, nullptr },
+	{ "layout",          true,  V_PUBLIC, T_OTHER,   false,  true,     qtrinary::qno,    &g_mapVotesPercent,         VOTE_BEFORE, &g_mapVotesBefore, nullptr },
+	{ "nextmap",         false, V_PUBLIC, T_OTHER,   false,  false,    qtrinary::qmaybe, &g_nextMapVotesPercent, VOTE_ALWAYS, nullptr, nullptr },
+	{ "poll",            false, V_ANY,    T_NONE,    false,  false,    qtrinary::qyes,   &g_pollVotesPercent,        VOTE_NO_AUTO, nullptr, nullptr },
+	{ "kickbots",        true,  V_PUBLIC, T_NONE,    false,  false,    qtrinary::qno,    &g_kickVotesPercent, VOTE_ALWAYS, nullptr, nullptr },
+	{ "fillbots",        true,  V_PUBLIC, T_OTHER,   false,  true,     qtrinary::qno,    &g_fillBotsVotesPercent, VOTE_ALWAYS, nullptr, nullptr },
```
DolceTriade, line 1544: Is it sensible to add an argument to this vote instead of adding 2 additional vote options?

src/sgame/sg_cmds.cpp
```diff
@@ -1849,6 +1859,45 @@ static void Cmd_CallVote_f( gentity_t *ent )
 
 		break;
 
+	case VOTE_BOT_FILL:
+		if ( !botFillVoteArgValid( arg ) )
+		{
+			trap_SendServerCommand( ent - g_entities,
+									va( "print_tr %s %s", QQ( N_("$1$: invalid number") ), cmd ) );
```
DolceTriade, line 1866: Should give a more actionable error message like:

```
 trap_SendServerCommand( ent - g_entities, va( "print_tr %s %s", QQ( N_("$1$: number must be less than $2$") ), cmd, std::to_string(g_maxVoteFillBots.Get()) ) );
```

src/sgame/sg_cmds.cpp
sweet235: Probably, but if we make them a single vote, we cannot simply have different cvars for vote percentages.

DolceTriade: Makes sense.

src/sgame/sg_cmds.cpp
```diff
@@ -1572,6 +1575,13 @@ static bool isDisabledVoteType(const char *vote)
 	return false;
 }
 
+static bool botFillVoteArgValid( char *arg )
```
slipher, line 1578: This should use the same parsing function as BotFillCmd. Which is currently `atoi`, but it would be better to change them both to Str::ParseInt to forbid junk after the number.

sweet235: I changed this to using Str::ParseInt, but did not change BotFillCmd yet.

sweet235: Well, i changed BotFillCmd too. `atoi` is evil.

src/sgame/sg_cmds.cpp
```diff
@@ -1849,6 +1858,42 @@ static void Cmd_CallVote_f( gentity_t *ent )
...
 
+	case VOTE_BOT_FILL:
+		if ( !botFillVoteArgValid( arg ) )
+		{
+			trap_SendServerCommand( ent - g_entities, va( "print_tr %s %s %s", QQ( N_("$1$: number must be non-negative and smaller than $2$") ), cmd, std::to_string( g_maxVoteFillBots.Get() + 1 ).c_str() ) );
+			return;
+		}
+
+		Com_sprintf( level.team[ team ].voteString, sizeof( level.team[ team ].voteString ), "bot fill %s", arg );
+		Com_sprintf( level.team[ team ].voteDisplayString, sizeof( level.team[ team ].voteDisplayString ), N_("Fill with bots to %s"), arg );
```
slipher, line 1869: "Fill each team with bots" maybe?

sweet235: Did that.

bmorel:
Nice. This only needs UI now :)

Issue #1579, opened 2021-11-02
Typo in "/give" command gives no warning


necessarily-equal:
I spent 1h the other day trying to bisect what the hell could have broken `/give momuntum 300`. Well, it was a typo.

There should be a warning of some sort.

Issue #735, opened 2015-05-20
Build point int overflow


hunternet93:
This is a small issue, when build points go over the max for whatever size int it is, it rolls around to a negative value. This never happens unless you play with g_initialMineRate (or similar vars) like I did, so might not need to be fixed.

Kangz:
Thanks for the report; but as you noted this has no chance of being caused by a vanilla setup. I thought build points were in a float though if they wrap then it must be an int: I'd close this, unless @Viech wants to use Cvar::Range to clamp the value of the initial mine rate.

Viech:
It's a network field issue. You do have positive build points but there aren't enough bits to transmit them to the client. I can clamp them before they are transmitted though.

PR #2189, opened 2022-09-09
Add error handling to the /give command

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/1579.

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -588,22 +588,26 @@ Cmd_Give_f
 Give items to a client
 ==================
 */
+static void Cmd_Give_printUsage( gentity_t *ent )
+{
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

illwieckz:
Does it also prevent to give more build points than possible to prevent overflow?

- https://github.com/Unvanquished/Unvanquished/issues/735

necessarily-equal:
> Does it also prevent to give more build points than possible to prevent overflow?

It doesn't. I guess I could add that

necessarily-equal:
Actually I checked https://github.com/Unvanquished/Unvanquished/issues/735 and this is a different problem that should be fixed in another place. I may try to fix the underlying problem one day, but it's not so simple

PR #2191, opened 2022-09-10
negative momentum should not happen

### LINKED ISSUE #1966: momentum can be negative
Currently, momentum can be negative. The faster way to check is: `devmap plat23`, then `/team h; give momentum -50; give momentum 100`. This will only unlock stage 1, and not stage 2.

This is either a bug in gameplay, if negative momentum should not exist, or a bug in UI, as the momentum bar can't show negative momentum.

bmorel:
fix #1966

### DIFF
--- a/src/sgame/sg_momentum.cpp
+++ b/src/sgame/sg_momentum.cpp
@@ -231,10 +231,7 @@ static float AddMomentum( momentum_t type, team_t team, float amount,
 	amount *= MomentumMod( type );
 
 	// limit a team's total
-	if ( level.team[ team ].momentum + amount > MOMENTUM_MAX )
-	{
-		amount = MOMENTUM_MAX - level.team[ team ].momentum;
-	}
+	amount = Math::Clamp( amount, 0.f - level.team[ team ].momentum, MOMENTUM_MAX - level.team[ team ].momentum );
 
 	if ( amount != 0.0f )
 	{


bmorel:
This prevents handicaping a team that needs to move base at start, notably. This is still not perfect because when aliens move base on walls/ceilings, they will still suffer constant stage loss, but I believe this is to be considered a "feature".

DolceTriade (approved):

necessarily-equal:
I have to check, but early reading makes me want to check momentum can still decrease with this PR

bmorel:
I verified in-game that it no longer decreases when a building is deconstructed.

slipher:
> I have to check, but early reading makes me want to check momentum can still decrease with this PR

This. The patch prevents any momentum decreases from AddMomentum. The min argument of the clamp needs to be adjusted similar to the max one.

bmorel:
You were right, this prevented decrease at all. I updated the patch and verified that now, it's impossible to go below 0, while still have decreases.
Maybe a simpler patch would have be to only do the clamp *after* the value is changed, but this works at least.

necessarily-equal (approved):
LGTM.

For the record with yesterday's version I did manage to hack the system pretty easily and reach stage 2 in 1v0 in 2min just by building/replacing turrets

bmorel:
> For the record with yesterday's version I did manage to hack the system pretty easily and reach stage 2 in 1v0 in 2min just by building/replacing turrets

yes, was buggy.

slipher:
LGTM

PR #2195, opened 2022-09-11
bugfix: command showbans could not be found

sweet235:
The slap entry in the array was one place off from being at its lexically sorted position. I suspect that a binary search is done in that array, and that this is the reason the showbans entry could not be found.


DolceTriade (approved):

bmorel:
LGTM

PR #2201, opened 2022-09-12
Remove obsolete shell scripts

slipher:
basepak.sh is so old that it still uses .pk3 instead of .dpk. There is now Urcheon for packaging releases.

tarball.sh contains references to nonexistent paths and a bunch of the options don't do anything or don't make any sense. It's questionable whether the tarball concept even makes sense for Unvanquished since only developers (as opposed to end users who compile their programs) should build Unvanquished from source.


illwieckz:
> basepak.sh is so old that it still uses .pk3 instead of .dpk.

And both `basepak.sh` and `tarball.sh` assume the base folder is `main`! 😜️

> tarball.sh contains references to nonexistent paths and a bunch of the options don't do anything or don't make any sense

If I wanted to redo a tarball and make a script for that, I would start writing a new script from scratch anyway.

illwieckz (approved):
LGTM

necessarily-equal (approved):

PR #2204, opened 2022-09-14
Ignore nobuild

sweet235:
A cvar to ignore nobuild areas in maps, usable only when the map was loaded by /devmap.

Edit:
The cvar is called `g_cheatIgnoreNobuild`. Should it be called `g_cheat_ignoreNobuild` instead?

### DIFF
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -221,6 +221,7 @@ Cvar::Cvar<float>  g_devolveMaxBaseDistance("g_devolveMaxBaseDistance", "Max Ove
 Cvar::Cvar<bool>   g_autoPause("g_autoPause", "pause empty server", Cvar::NONE, false);
 
 Cvar::Cvar<int> g_maxMiners("g_maxMiners", "set maximum number of miners per team. -1 = disabled.", Cvar::NONE, -1);
+Cvar::Cvar<bool> g_cheatIgnoreNobuild("g_cheatIgnoreNobuild", "ignore nobuild area", Cvar::NONE, false);
 
 // <bot stuff>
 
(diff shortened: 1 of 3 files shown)


src/sgame/sg_main.cpp
```diff
@@ -210,6 +210,7 @@ Cvar::Cvar<int> g_combatCooldown("g_combatCooldown", "team change disallowed unt
 Cvar::Range<Cvar::Cvar<int>> g_debugEntities("g_debugEntities", "entity debug level", Cvar::NONE, 0, -2, 3);
 
 Cvar::Cvar<bool> g_instantBuilding("g_instantBuilding", "cheat mode for building", Cvar::NONE, false);
+Cvar::Cvar<bool> g_cheatIgnoreNobuild("g_cheatIgnoreNobuild", "ignore nobuild area", Cvar::NONE, false);
```
necessarily-equal, line 213: `g_instantBuilding` is classified as a regular server option (I think I even saw trem servers with it), but it seems `g_cheatIgnoreBuild` is a cheat :smiley:. I think the easiest is just to mark it as such and then (IIRC, needs testing I guess) you don't need to check for `g_cheat` afterwards.

```suggestion
Cvar::Cvar<bool> g_cheatIgnoreNobuild("g_cheatIgnoreNobuild", "ignore nobuild area", Cvar::CHEAT, false);
```

sweet235: I'm not sure at all if this should be a cheat or not. But it seems i made a mistake here, because i check the value of g_cheats myself when using the cvar's value.

DolceTriade: Don't think it needs to be Cvar::CHEAT. If an admin wants it, why not :shrug:

sweet235: I agree, did that.

DolceTriade (approved):

Issue #1102, opened 2018-12-27
Scoreboard stucked open when mapchange occurs during connecting


t4im:
The Scoreboard sometimes remains stuck open when connecting to a server that shortly after loads a new map. This was confirmed by another user in-game.
Since `/-scores` doesn't free it, it likely is an UI rather than a key issue.

DolceTriade:
Hmm, I thought i had fixed this. I'll investigate

illwieckz:
I confirm this issue, got it multiple time:

```
2018-12-10 01:24 <+illwieckz> interesting, if end of game occurs while map is still downloading, the scoreboard will be displayed once the nextmap is loaded
```

bmorel:
duplicate of https://github.com/Unvanquished/Unvanquished/issues/734

illwieckz:
I'm not sure #734 was a duplicate of this anyway.

DolceTriade:
I cannot reproduce this bug.

Upon changing maps, the scoreboard does not get stuck.

bmorel:
This bug still happens from time to time (I had it few days ago), maybe mapchange is *not* the trigger though.

DolceTriade:
I have found a 100% repro case:

Ensure that sv_wwwFallbackUrl is set badly so the client is never able to download the map.

 - Start `daemonded` on a map the client does not have downloaded.
 - Connect client to server
 - Let client stay stuck on download screen
 - On daemonded, run `aliensWin` and wait for next map to load. The next map should be a map the client has downloaded
 - Let client join the map.
 - Expect to see stuck scoreboard

The reason that this happens is that the new client gets CS_INTERMISSION and CS_WINNER set from the old game, but this never gets reset. We do set `CS_INTERMISSION` to 0 in sg_main.cpp, so I'm not sure why the old commands reach the client but the new ones don't.

I'll debug this tomorrow now that I have a test bench set up to repro this at will.

slipher:
@DolceTriade If you have the repro thing set up still, can you try this?

```
diff --git a/src/cgame/cg_servercmds.cpp b/src/cgame/cg_servercmds.cpp
index c14be9500..bd964766f 100644
--- a/src/cgame/cg_servercmds.cpp
+++ b/src/cgame/cg_servercmds.cpp
@@ -319,10 +319,7 @@ static void CG_ConfigStringModified()
        else if ( num == CS_INTERMISSION )
        {
                cg.intermissionStarted = atoi( str );
[...]

PR #2206, opened 2022-09-16
sgame: Send pmove_params on ClientBegin

DolceTriade:
Before they were being sent on ClientConnect. This is bad because the client doesn't actually process server commands at this time, so they get queued. They get queued with other potentially stale commands, so we don't want to carry them forward to the next game. Moving pmove_params here allows us to drop early commands which can load to bugs like the stuck scoreboard in #1102

### DIFF
--- a/src/sgame/sg_client.cpp
+++ b/src/sgame/sg_client.cpp
@@ -1188,8 +1188,6 @@ const char *ClientConnect( int clientNum, bool firstTime )
 	             client->pers.netname,
 	             client->pers.netname );
 
-	G_SendClientPmoveParams(clientNum);
-
 	// don't do the "xxx connected" messages if they were caried over from previous level
 	if ( firstTime )
 	{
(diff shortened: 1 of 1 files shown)


necessarily-equal (approved):
Sounds good, but I don't know that part of the code well, and haven't tested.

Btw if you push this to a branch named (eg) `pmove-ClientBegin/testing` or `ishq/pmove/testing`, it will be build and deployed automatically, which allows for easier testing of networked bugs

bmorel:
I'd prefer to ask @slipher on that topic, and/or to run some tests with a highly laggy server which quickly (faster than the client can go in game, possibly because UDP downloading) changes maps before accepting this.

DolceTriade:
Well, this change is 100% safe. Keep in mind that in its current state, pmove_params is sent *before* any instructions for downloading missing paks, which seems wrong.

bmorel:
What I keep in mind is that I do not know enough about how this works. Can't say if good or bad.

bmorel:
ping

DolceTriade:
I think we found some bugs with this approach.

slipher:
Yeah there was some PR on another repository which goes together with this change and I described a problem with the approach there

slipher:
We have a proper fix for the config string bugs now so this can be closed.

PR #2212, opened 2022-09-18
bp: Always allow using marked buildables to rebuild things.

DolceTriade:
Even if overall BP is negative, allow using marked buildables to rebuild things.

### DIFF
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -1182,7 +1182,7 @@ void G_DeconstructUnprotected( gentity_t *buildable, gentity_t *ent )
 
 		// Add to build timer.
 		ent->client->ps.stats[ STAT_MISC ] += BG_Buildable( buildable->s.modelindex )->buildTime / 4;
-	}	
+	}
 
 	G_Deconstruct( buildable, ent, MOD_DECONSTRUCT );
 }
@@ -1418,7 +1418,7 @@ static itemBuildError_t PrepareBuildableReplacement( buildable_t buildable, vec3
 	}
 
 	// check if we can already afford the new buildable
-	if ( G_GetFreeBudget( attr->team ) >= cost )
+	if ( Math::Clamp( G_GetFreeBudget( attr->team ), 0, std::numeric_limits<int>::max() ) >= cost )
 	{
 		return IBE_NONE;
 	}
@@ -1471,7 +1471,7 @@ static itemBuildError_t PrepareBuildableReplacement( buildable_t buildable, vec3
 		cost -= G_BuildableDeconValue( ent );
 
 		// check if we have enough resources now
-		if ( G_GetFreeBudget( attr->team ) >= cost )
+		if ( Math::Clamp( G_GetFreeBudget( attr->team ), 0, std::numeric_limits<int>::max() ) >= cost )
 		{
 			return IBE_NONE;
 		}
@@ -1718,7 +1718,7 @@ itemBuildError_t G_CanBuild( gentity_t *ent, buildable_t buildable, int /*distan
 			return ent->client->pers.team == TEAM_HUMANS ? IBE_NOMOREDRILLS : IBE_NOMORELEECHES;
 		}
 	}
-		
+
 	return reason;
 }
 
--- a/src/cgame/cg_rocket_datasource.cpp
+++ b/src/cgame/cg_rocket_datasource.cpp
@@ -1608,7 +1608,8 @@ static Str::StringRef BuildableAvailability( buildable_t buildable )
 	int markedBudget    = cg.snap->ps.persistant[ PERS_MARKEDBUDGET ];
 	int totalBudet      = cg.snap->ps.persistant[ PERS_TOTALBUDGET ];
 	int queuedBudget    = cg.snap->ps.persistant[ PERS_QUEUEDBUDGET ];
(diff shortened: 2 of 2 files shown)


sweet235:
I tested this with changes in `sg_buildpoints.cpp`, see https://github.com/Unvanquished/Unvanquished/commit/d78fcc3b9f24ca2c8996a095b4dda77939b603fe

These seem to be missing here. I don't know if that is a problem.

DolceTriade:
I squashed these changes.

sweet235:
Squashed or not, i have this: https://github.com/sweet235/Unvanquished/blob/map-testing/src/sgame/sg_buildpoints.cpp#L163

DolceTriade:
Yes, I reverted that. That function isn't used anywhere in our code base and I actually don't end up using it in the final fixed patch.

That function was originally used https://github.com/Unvanquished/Unvanquished/commit/397949555deea8959fb0021d278001dba76cfcf8#diff-edacce121b4ccf14dd35c046375ea6233fb6dae58e5bca508059d20fc0b7c5baL1474, but then replaced here: https://github.com/Unvanquished/Unvanquished/commit/494e547d6ed1f5e0e516d704679cc71d978558f9#diff-edacce121b4ccf14dd35c046375ea6233fb6dae58e5bca508059d20fc0b7c5baL1474

sweet235:
Oh lol, didnt notice it is unused.

Gireen:
> This change undermines that by allowing people to freely swap, say, two turrets for an armoury.

I think thats a good thing. Its frustrating with negative bp, no armory and no means to rebuild it despite a otherwise compete base.
And with this its possible to trade some defense in for the chance to counterattack.

A problem i see here may be that one could create a additional BP pool. By building many miners and spending all BP and then lategame with negative BP using the additional buildings to rebuild the base.

DolceTriade:
> I don't like how this could prolong the end game phase.

While this is a side effect, I think it's far more meaningful to improve the feelings of frustration of losing your arm/medi and then basically playing a losing match for the next 10 min than being able to have a fighting chance. I think we have better opportunities to shorten the game.

> I think this is not very meaningful because only the builder would easily notice when the feature is used.

I mean...what would be considered meaningful testing? To me, that is having quality games all around, regardless of who the feature affects directly.

> A problem i see here may be that one could create a additional BP pool. By building many miners and spending all BP and then lategame with negative BP using the additional buildings to rebuild the base.

Perhaps. We could make it so that unpowered marked buildables don't give any BP?

slipher:
> > I don't like how this could prolong the end game phase.
> 
> While this is a side effect, I think it's far more meaningful to improve the feelings of frustration of losing your arm/medi and then basically playing a losing match for the next 10 min than being able to have a fighting chance. I think we have better opportunities to shorten the game.

If the game really lasts another 10 minutes after an armoury was permanently lost, it does seem bad. But surely aliens ought to be able to win faster than that.


> 
> > I think this is not very meaningful because only the builder would easily notice when the feature is used.
> 
> I mean...what would be considered meaningful testing? To me, that is having quality games all around, regardless of who the feature affects directly.

Right. I assumed that players were not watching the overall course of the games and comparing but rather just reporting their experience using the feature directly.

> 
> > A problem i see here may be that one could create a additional BP pool. By building many miners and spending all BP and then lategame with negative BP using the additional buildings to rebuild the base.
> 
> Perhaps. We could make it so that unpowered marked buildables don't give any BP?
Good idea.

Does this code really work? I think it should fail in `PrepareBuildableReplacement` because not enough BP can be freed.

src/sgame/sg_buildable.cpp
```diff
@@ -1418,7 +1418,7 @@ static itemBuildError_t PrepareBuildableReplacement( buildable_t buildable, vec3
 	}
 
 	// check if we can already afford the new buildable
-	if ( G_GetFreeBudget( attr->team ) >= cost )
+	if ( Math::Clamp( G_GetFreeBudget( attr->team ), 0, std::numeric_limits<int>::max() ) >= cost )
```
slipher, line 1421: max(x, 0)

DolceTriade: oops, way better

DolceTriade:
I like that idea too @Gireen I'll implement it here.

Gireen:
> Perhaps. We could make it so that unpowered marked buildables don't give any BP?

Would that not negate the desired effect of this change? if there are negative BP buildings become unpowered and its again not possible to rebuild. Or at least harder, i don't know how unpowering works in detail.

For me both versions are ok to merge. Maybe its in praxis not that big of a problem.

DolceTriade:
I'll test it. Not all things are unpowered with negative BP.

necessarily-equal:
For the record, making recycling BP work only with powered buildables sounds weird/confusing and frustrating from a gameplay point of view. But I'm not definite about this without a test, though. Maybe that would work.

DolceTriade:
I agree. Also to be fair, recycling unpowered buildables doesn't mean they'll become powered again unless they build something worth using. All in all, I think this makes the game less frustrating. I think we should come up with better ways to prevent abuse.

necessarily-equal:
gentle bump

DolceTriade:
I don't think this is for 0.54 yet. While I think it's a good idea, I think we might want to test this on other servers for a bit more...

illwieckz:
@DolceTriade what's the status of this?

Issue #2182, opened 2022-09-08
bots can have "surprise freeze"


bmorel:
When bots meet "unexpected situation", they can freeze, and stop doing anything. I have still no idea what causes that behavior. Opening this thread to have a place to collect infos.
The problem was not around in v0.52.0, and is here in v0.53.2 AFAIK.

PR #2213, opened 2022-09-18
Fix stuck bots

necessarily-equal:
This is a partial revert of the following commits:
* d118e6d37b "handle BT failures in atomic navigation actions"
* 8372cf1245 "BotMoveToGoal can fail"

This makes bot try harder, which makes them less stuck. Fix https://github.com/Unvanquished/Unvanquished/issues/2182.


necessarily-equal:
Obsoleted by https://github.com/Unvanquished/Unvanquished/pull/2214

PR #2214, opened 2022-09-18
Fix stuck bots 2.0

necessarily-equal:
This makes bot try harder to move even when they aren't supposed to be able to avoid the obstacle, which makes them less stuck.

Fix https://github.com/Unvanquished/Unvanquished/issues/2182.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -598,10 +598,10 @@ bool BotMoveToGoal( gentity_t *self )
 	glm::vec3 dir = self->botMind->nav().glm_dir();
 	if ( BotAvoidObstacles( self, dir ) )
 	{
+		BotSeek( self, dir );
 		return false;
 	}
 
-	BotAvoidObstacles( self, dir );
 	BotSeek( self, dir );
 
 	// dumb bots don't know how to be efficient


DolceTriade:
Seems plausible. Don't really understand the impact of this.

DolceTriade:
https://users.unvanquished.net/~modi/bot-2-bot-stuck.dm_86 Seen this in your bot-2 testing server. Not sure if this is sufficient or helpful or even worse

slipher:
Hmm, so the function does something different in a case where it returns failure. So it will make a difference in the cases when the return value is ignored or if the action is retried every frame.

LGTM I guess

necessarily-equal:
@DolceTriade: https://users.unvanquished.net/~modi/bot-2-bot-stuck.dm_86 exhibit two different bugs. The first bug is two bots facing each other and trying to run forward. This PR makes them try to run forward instead of not moving at all, but this indeed doesn't help (because the bounding box are neatly aligned with the desired direction) and they are just as stuck. The second bug is with the turret overhead, and the obstacle collision code that is modified in this PR is just not triggered at all.

The first problem should be fixed by https://github.com/Unvanquished/Unvanquished/pull/2215, the second will need another PR.

The second problem could be fixed by freem's bot crouching code, I think.

DolceTriade:
Note that you do not need to crouch. You can move left and right and go around the turrets just fine. Thanks for checking!

necessarily-equal:
closing in favor of https://github.com/Unvanquished/Unvanquished/pull/2215

PR #2215, opened 2022-09-18
Fix stuck bots 3.0

necessarily-equal:
This makes bot try harder to move even when they aren't supposed to be able to avoid the obstacle, which makes them less stuck.

This version makes them try to run a bit sideway, one time left, one time right until they find their way. It seems more efficient but also a bit more complex. See also #2214 for the simpler version.

Fix https://github.com/Unvanquished/Unvanquished/issues/2182.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -412,9 +412,40 @@ static bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const glm:
 	return blocker->s.modelindex == BA_A_BARRICADE || trace.fraction == 1.0f;
 }
 
-// try to find a path around the obstacle by projecting 5
+// Determine a new direction, which may be unchanged (forward) or sideway. It
+// does try the following every second: one time left, one time right and one
+// time forward until they find their way.
+static void BotFindDefaultSteerTarget( gentity_t *self, glm::vec3 &dir )
+{
+	// turn this into a 2D problem
+	dir[2] = 0;
+	// a vector 90° from dir, pointing right and of same magnitude
+	glm::vec3 right = glm::vec3(dir[1], -dir[0], 0);
+
+	int clientNum = self - g_entities;
+	// We add a delta determined from clientNum to avoid bots
+	// dancing in sync
+	int time = (110 * clientNum + self->client->time1000) % 1000;
+
+	// 280 instead of 700/2=350 because we want some left/right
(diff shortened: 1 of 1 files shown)


src/sgame/sg_bot_nav.cpp
```diff
@@ -598,10 +598,25 @@ bool BotMoveToGoal( gentity_t *self )
...
+
+		int clientNum = self - g_entities;
+		dir[2] = 0;
+		dir = glm::normalize(dir);
+		glm::vec3 up = { 0, 0, 1 };
+		glm::vec3 right = glm::cross(dir, up);
+
+		// invert periodically four times per second. We add a delta
+		// determined from clientNum to avoid bots dancing in sync
+		bool invert = ((110 * clientNum + self->client->time1000) % 125) < 250;
```
slipher, line 611: The condition is always true, barring integer overflows. Did you test this?

necessarily-equal: Good catch!

Bots still moved randomly by changing target, that made it look about right :eyes:. Fixed. I think it looks better now.

bmorel:
This hack should not go in this function, but in BotFindSteerTarget(), which already tries to do what you try to do, or in the unstuck (BT) routine.

src/sgame/sg_bot_nav.cpp
```diff
@@ -412,9 +412,38 @@ static bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const glm:
...
+
+	// 280 instead of 700/2=350 because we want some left/right
+	// asymetry in case it helps
+	bool invert = time < 280;
+	// half of the time we go straight instead
+	bool sideway = time > 700;
+
+	// 2*right + 1*dir means an angle of atan(2/1) = 63°. This
+	// seems about right
+	dir += right * (invert ? -1.0f : 1.0f)
```
slipher, line 438: I think `if (sideways) { ... }` would be more readable. Also it should be "sideways" not "sideway".

LGTM

necessarily-equal: updated

bmorel:
> The two typicals problems are vents that are too tight to navigate

A proper fix to that would be something like that: https://github.com/bmorel/Unvanquished/commit/288a7f5aab286a235b5f71e8f9a34264685bf1b6 which, between other changes, checks for STEPSIZE.

PR #2216, opened 2022-09-20
rmlui: Allow running commands from chat.

DolceTriade:
By prefixing your text with a /, you can run a command. This was the behavior of the trem chat. Since you need to use commands to do a lot of useful things in Unv, it's easy to use the chat to run these commands instead of being forced to open the console.

This behavior can be disabled by using the ``noCmd="1"`` attribute in the <chatField /> element.


slipher:
Eventually I want to forbid the cgame from running arbitrary commands for improved security (so it can't print the user's rcon password for example)

DolceTriade:
Yeah, I'm not sure how thrilled people are by this concept anyways.

PR #2218, opened 2022-09-22
Fix marazap bug where it does hella damage when it shouldn't.

DolceTriade:
Basically we just attack the same entity as many times as there are things to chain to around it. This is a bug that was introduced in some old refactors. Also fix an OOB error in rendering marazap when we zap the max number of targets.


slipher:
LGTM

Issue #2200, opened 2022-09-12
Write scripts in Python?


illwieckz:
[Here](https://github.com/Unvanquished/release-scripts/pull/24#issuecomment-1242820823) @slipher wrote, as I was extending the `build-release` scripts:

> Being opposed to complex shell scripts […]

And in a recent past when I [tried to make the script to build deps](https://github.com/DaemonEngine/Daemon/pull/561) more robust by making it self-aware of what it was doing (and then becoming able to check what it was doing) this effort turned into a disaster because I faced the wall of the belief that a bash program should not become long or complex.

I get the feeling that there is a belief there exist two kind of programs, programs that can do anything if not written in bash, and programs in bash that are not real programs and that should not grow.

To me writing a program in bash language is just the choice of a language among others. I myself often decide to use bash among other choices because 1. it is interpreted, 2. bash is almost ubiquitous (if I restrict to the old bash shipped by default in macos). So I can start long, large, complex projects deciding to go the bash way, never purposing those projects to not be complex but the exact opposite. I can chose bash as an implementation detail _after_ I decide to write something expected to be complex, and I can chose bash in order to write such complex things.

So, at some point, all my attempts to write utilities in a interpreted language to make things that are a bit complex will face opposition if I chose bash among other available languages. And more likely this kind of opposition will be faced years after the utility have been written, when adding this or that over the existing code.

[Here](https://github.com/DaemonEngine/Daemon/pull/561#issuecomment-1138761080) I wrote (in deps rewriting thread):

> Note: I would not be against porting this to another language like python in the future to get rid of the bash wizardry.

Though, python isn't as good as bash and other language with direct access to shell commands in general when it's needed to run a bunch of shell commands. It's even hard to compete with the efficiency of `xargs` when it comes to process in parallel and with efficient job management a bunch of inputs you may even don't know in advance and that may even been feed on the fly.

Some may say that if someone needs more than bash, perl is the best candidate, but unfortunately no one writes perl today, including myself.

One strong argument in favor of Python is that we already require Python for multiple things in the game, so Python is already ubiquitous. Any people who is expected to build the game is expected to already have a working Python installation. So we can unlock full python abilities (and rich and robust data structures like dictionaries that are not available in old bash on macos), and even when restricting to the default python modules we get a lot of features.

So, since Python is considered ubiquitous, why not writing our various utilities in python? I mean those that are expected to be run in development process: building the deps, making a release build, etc.

illwieckz:
[...]

PR #2219, opened 2022-09-22
Rewrite update-version-number in Python

slipher:
I was bored in the airport the other day and did this. See #2200.


DolceTriade (approved):
It would be nice to have a global dict at the top so we don't have to grok the file to add new files whose version we have to update.

Then the code is just magic stuff that we don't need to care about, all you need to edit is the dict.

Otherwise LGTM

illwieckz:
On a side topic, what kind of version we want for such script? BSD? MIT? CC0?

DolceTriade (approved):

slipher:
> It would be nice to have a global dict at the top so we don't have to grok the file to add new files whose version we have to update.

Done.

Also I changed it so it errors only if the regex doesn't match anything, rather than if the replacement doesn't cause a diff.



> On a side topic, what kind of version we want for such script? BSD? MIT? CC0?

I don't know but not MIT I guess since we already use BSD which is similar enough.

PR #2222, opened 2022-09-25
emoji: Add emoji support to rmlui.

DolceTriade:
You can use [:)] or [salute] or other known emoji short codes to use emojis in Unvanquished. These are a critical feature to help player express themselves.

This change generates an emoji map using Google's emoji data to map short names and emoticons to emojis and also provides a script to regenerate this should a new emoji spec come out.


src/cgame/rocket/emojis/generate_emoji_map.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+
+def render_base(base):
+    s = ''
+    for codept in base:
+        s += chr(codept)
+    return s
+
+def escapeq(s):
+    return s.replace('"', '\\"').replace('\\', '\\\\')
```
slipher, line 39: This looks wrong

slipher (commented):
The emoji font is 10 MB, is it worth it?

src/cgame/rocket/emojis/generate_emoji_map.py
DolceTriade: How so? `s.replace()` returns a string which as a method `replace`. There probably is a better way of doing this, but this isn't super performance critical.

DolceTriade:
> The emoji font is 10 MB, is it worth it?

I cannot understate how important this feature is. Also, given that it will be compressed and relatively small in terms of game asset size, I think it's worth while.

src/cgame/rocket/emojis/generate_emoji_map.py
DolceTriade: oic. fixed

slipher:
What's the compressed size?

I suspect there will be limited usage without an emoji picker or something, because it's not discoverable. How are people supposed to know the strings like `[loudly-crying]`?

illwieckz:
Let's welcome `[hand-with-index-finger-and-thumb-crossed]`, `[roasted-sweet-potato]` and `[right-pointing-triangle-with-double-vertical-bar]`. 😅️ And I wonder how people would type `[♪┏(･o･)┛♪]`. 🤣️

Anyway, that's good! And country flags will be appreciated! 🙂️

DolceTriade:
Hm, actually, country flags don't work. Looking closer, I don't think any multi codepoint emojis work... let me see if I can fix that, or whether this is a Freetype limitation.

DolceTriade:
So this is a freetype limitation. We need to use HarfBuzz to layout text properly (ie, parse the emojis comprising of multiple code points and variations that use a [zwj](https://emojipedia.org/zero-width-joiner/). In my script, I can filter out all emojis that comprise of multiple code points and see what the result is. Let me see if I can copy paste these emojis into the game... I doubt it.

DolceTriade:
Also, a size comparison:

```
❯ ls -lh /tmp/unv_*
-rw-r--r-- 1 modi users 25M Sep 25 22:40 /tmp/unv_emoji.dpk
-rw-r--r-- 1 modi users 16M Sep 25 22:43 /tmp/unv_noemoji.dpk
```

Turns out the emoji font **does not** compress well :smile:

necessarily-equal:
I think we we want to ship that without making it insufferable with regard to downloading time, we should move the fonts to another repo, and not unv_src.dpkdir. On the flip side that repo is supposed to be the only one required to load the main menu. Maybe the fonts should be bundled with daemon instead?

DolceTriade:
One thing I think I should do is that since we don't support multicode emojis, if I can find a way to strip them out of the font, that should reduce the size considerably.

PR #2223, opened 2022-09-25
chat: Show a chat bubble when players are typing.

### LINKED ISSUE #2217: implement chat bubble
This means to have a "chat" bubble on top of players, the option should be opt-out. The intent is to have some fair-play: no shooting at chatting people.

DolceTriade:
This uses the usercmd flags which relate to input to note whether the game has input. Currently, AFAICT this field is unused, so we use it to transmit whether the client is typing. WE need to remove the RSF_SPRITE from the chatBubble thing because for some reason the chat bubble doesn't show with it.

Fixes #2217

### DIFF
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -2249,6 +2249,14 @@ void ClientThink( int clientNum )
 
 	ent = g_entities + clientNum;
 	trap_GetUsercmd( clientNum, &ent->client->pers.cmd );
+	if ( ent->client->pers.cmd.flags & UF_TYPING )
+	{
+		ent->client->ps.eFlags |= EF_TYPING;
+	}
+	else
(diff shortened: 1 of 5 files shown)


src/sgame/sg_active.cpp
```diff
@@ -2249,6 +2249,14 @@ void ClientThink( int clientNum )
 
 	ent = g_entities + clientNum;
 	trap_GetUsercmd( clientNum, &ent->client->pers.cmd );
+	if ( ent->client->pers.cmd.flags )
```
slipher, line 2252: Make it really a flag instead of using cmd.flags as a boolean?

DolceTriade: Done

slipher (commented):
LGTM

PR #2224, opened 2022-09-26
Get rid of blueprint miner entity hack

slipher:
A temporary leech entity will no longer be constructed when calculating the potential efficiency of a new miner.

Also implement the TODO to hide under-construction enemy miners from potential efficiency calculations.


src/sgame/components/MiningComponent.h
```diff
@@ -47,15 +46,16 @@ class MiningComponent: public MiningComponentBase {
 		 */
 		static float InterferenceMod(float distance);
 
-		/**
```
DolceTriade, line 50: Why was this remove?

src/sgame/components/MiningComponent.h
```diff
@@ -47,15 +46,16 @@ class MiningComponent: public MiningComponentBase {
...
-		 * @return Whether the miner is currently mining (and interfering with other miners).
-		 */
-		bool Active() { return active; }
+		struct Efficiencies {
```
DolceTriade, line 49: Given that their usage is mutually exclusive, would it be better to expose two functions to fetch these values?

DolceTriade (commented):
nice fix for a hack

src/sgame/components/MiningComponent.h
slipher: Ended up being unused

src/sgame/components/MiningComponent.h
slipher: We have these cases:
- Blueprint miner: calculate only predicted efficiency
- Miner is constructed: need to update only predicted efficiencies
- Miner finishes construction: need to update only current efficiencies
- Miner dies: need to update both predicted and current efficiencies

I could have flags for which ones InformNeighbors should update, but that seemed overly complex so I stuck with always updating both, which is done together for efficiency.

DolceTriade: Got it.

DolceTriade (approved):

Issue #2114, opened 2022-08-28
print map metadata: title, author (version ?) on map loading screen


illwieckz:
Creative people may want to be creative and make complex levelshots with text, symbols, maps or I don't know why, but we should not require mapper to bake map name and map author name on levelshot in order to get the map name and their own name being displayed at map load.

Also we better not require to go the image edition process each time the map change enough to get a new levelshot, taking a new levelshot without more edition should be enough.

The idea is to read map name and map author from `.arena` file and display it on map load screen with some RML.

For the map name, it's a good idea to have a reusable function that reads it from `.arena`, if missing reads it from bsp entity, if missing fallbacks on map bsp base name. This function would be reusable for https://github.com/Unvanquished/Unvanquished/issues/2113 .

We may even add license to .arena file.

illwieckz:
This can be a good thread to also discuss the syntax and key names of that `.arena` file.

Right now, for a map named `castle`, the file is named `meta/castle/castle.arena`.

In tremulous it was named `scripts/castle.arena`

The content would be:

```
{
map "castle"
longname "Castle"
type "tremulous"
}
```

The long name can get spaces, special characters, color codes…

The `name` key is redundant, it exists for historical reasons, probably because some Q3-based games stored metadata for more than one map in one file. For example Smokin' Guns provide a file named `maps/sg_maps.cfg` that contains data for many maps:

```
{
	map	 	"br_backwater"
	author	        "Breli"
	longname	"Backwater"
	desc_line1	"----------------------------------"
	desc_line2	"Out in the nowhere ..."
	desc_line3	"----------------------------------"
	desc_line4	"BR Gametype"
}

{
	map		"dm_backwater"
	author	        "Breli"
	longname	"Backwater"
	desc_line1	"----------------------------------"
	desc_line2	"Out in the nowhere ..."
	desc_line3	"----------------------------------"
	desc_line4	"DM, TDM & RTP Gametypes"
}

{
	map		"dm_dawnfort"
	author	        "Breli"
	longname	"DawnFort"
	desc_line1	"----------------------------------"
	desc_line2	"Dawn has come to the fort ..."
	desc_line3	"----------------------------------"
	desc_line4	"DM, TDM & RTP Gametypes"
}
[…]
```

We don't support that variant, but that explains things…

[...]

Issue #2043, opened 2022-08-05
Other bug `Server command overflow` is still alive


illwieckz:
As seen on #2014, some people also disconnected with some “Server command overflow” log on server, something the RmlUi Lua GC bug did not reproduced:

```
ELCOCO1981^* Server command overflow
```

Seen today on 0.53.1:

```
Extracting VM module sgame-x86_64.nexe from unvanquished_0.53.1/pkg/unvanquished_0.53.1.dpk...

Loading VM module sgame-x86_64.nexe...
Using loader args:  unvanquished_0.53.1/nacl_helper_bootstrap unvanquished_0.53.1/nacl_loader --r_debug=0xXXXXXXXXXXXXXXXX --reserved_at_zero=0xXXXXXXXXXXXXXXXX -v -B unvanquished_0.53.1/irt_core-x86_64.nexe -e -i 100:27 -- unvanquished/sgame-x86_64.nexe 100
Loaded VM module in 54 msec
[…]
------- Game Initialization -------
gamename: base
gamedate: Aug  4 2022
------------------------------------------------------------
InitGame: \g_BPRecoveryInitialRate\16\g_needpass\0\g_unlagged\1\g_BPBudgetPerMiner\50\sv_statsURL\http://gg.illwieckz.net/#stats\sv_hostname\^7[EU] ^9gg.^4ill^7wie^1ckz^9.net ^6{ ^5le frag courtois ^6} ^20.53.1 ^<FIXED ^9#1^*\pakname\map-forlorn\protocol\86\sv_floodProtect\0\g_BPInitialBudget\80\g_maxGameClients\0\timelimit\0\sv_maxclients\31\version\Unvanquished 0.53.0 Linux x86_64 Jul 31 2022\sv_maxRate\0\g_BPRecoveryRateHalfLife\10\sv_privateClients\0\B\-------------------------------\mapname\forlorn\g_devolveMaxBaseDistance\1000.000000\gamename\base\g_unlockableMinTime\30.000000\g_momentumHalfLife\5.000000\P\-------------------------------
RealTime: 2022-08-05 10:24:10 Z
^3readconfig:^* loaded 6 levels, 21 admins, 1 bans, 0 commands
^3Warn: "colorGrade" deprecated. Please use "gradingTexture"
Warmup: 10
[…]
[...]

PR #2226, opened 2022-09-26
Load information from `.arena` file on map load to display it on map loading screen

illwieckz:
Load information from `.arena` file on map load to display it on map loading screen.

Requires this to display the data:

- https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/161

Fixes: 

- https://github.com/Unvanquished/Unvanquished/issues/2114

### DIFF
--- a/src/cgame/cg_gameinfo.cpp
+++ b/src/cgame/cg_gameinfo.cpp
@@ -34,24 +34,19 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 //
 // arena info
 //
-static int  cg_numArenas;
-static char *cg_arenaInfos[ MAX_ARENAS ];
+static std::vector<arenaInfo_t> arenaList;
 
 /*
 ===============
 CG_ParseInfos
 ===============
 */
-static int CG_ParseInfos( const char *buf, int max, char *infos[] )
+static void CG_ParseInfos( const char *buf )
 {
 	char *token;
-	int  count = 0;
 	char key[ MAX_TOKEN_CHARS ];
-	char info[ MAX_INFO_STRING ];
 
-	auto infoPostfixLen = strlen( "\\num\\" ) + strlen( va( "%d", MAX_ARENAS ) );
-
-	while ( 1 )
+	while ( true )
 	{
 		token = COM_Parse( &buf );
 
@@ -62,25 +57,19 @@ static int CG_ParseInfos( const char *buf, int max, char *infos[] )
 
 		if ( strcmp( token, "{" ) )
 		{
-			Log::Notice( "Missing { in info file\n" );
+			Log::Warn( "Missing { in arena file" );
 			break;
 		}
 
-		if ( count == max )
-		{
-			Log::Notice( "Max infos exceeded\n" );
-			break;
-		}
+		arenaInfo_t arenaInfo;
 
-		info[ 0 ] = '\0';
-
-		while ( 1 )
+		while ( true )
 		{
 			token = COM_ParseExt( &buf, true );
 
 			if ( !token[ 0 ] )
 			{
-				Log::Notice( "Unexpected end of info file\n" );
+				Log::Warn( "Unexpected end of arena file" );
 				break;
 			}
 
@@ -98,59 +87,112 @@ static int CG_ParseInfos( const char *buf, int max, char *infos[] )
 				strcpy( token, "<NULL>" );
 			}
 
-			Info_SetValueForKey( info, key, token, false );
+			/* Standard in all games. */
+			if ( strcmp( key, "map" ) == 0 )
+			{
+				arenaInfo.name = token;
+			}
+			/* Standard in all games. */
+			else if ( strcmp( key, "longname" ) == 0 )
+			{
+				arenaInfo.longName = token;
+			}
+			/* Also used in Smokin' Guns and Q3 Rally. */
+			else if ( strcmp( key, "author" ) == 0 )
+			{
+				arenaInfo.authors.push_back( token );
+			}
+
+			/* Some other common keys:
+
+			* type: list of game types as words in a string,
+			  example: "tourney ffa team",
+			  Used by almost games that has various game modes:
+			  Quake 3, Quake Live, World of Padman, Q3 Rally,
+			  Urban Terror, Return to Castle Wolfenstein…
+			* bots: list of bot identity as words in a string,
+			  example: "sarge, uriel",
+			  used by Quake 3, Q3 Rally…
+			* bot_skill: bot skill level,
+			  example: 3,
+			  used by Quake Live.
+			* timelimit: timelimit in minutes,
+			  example: 10,
+			  used by Quake 3, Quake Live, Wild West,
+			  Return to Castle Wolfenstein,
+			  Wolfenstein: Enemy Territory…
+			* fraglimit: amount of frags to end a game,
+			  example: 20.
+			  used by Quake 3, Quake Live…
+			* briefing: a string displayed to the player
+			  before he enters the match,
+			  used by Wolfenstein: Enemy Territory.
+			* desc_line1 (2, 3 ,4): lines of text describing
+			  the map,
+			  used by Smokin' Guns.
+
+			And others.
+
+			They are ignored for now as we don't make use of them.
+
+			As this code does not make use of “Info” storage format,
+			we can also implement list of things by using multiple
+			time the same key name, like the ogg metadata format
+			does.
+
+			Example, instead of:
+
+			  type "tourney ffa team"
+
+			We can implement:
+
+			  type "tourney"
+			  type "ffa"
+			  type "team"
+
+			This is also similar to how it is done in the .mapinfo
+			files from Xonotic (using a different file format). */
 		}
 
-		//NOTE: extra space for arena number
-		infos[ count ] = (char*) BG_Alloc( strlen( info ) + infoPostfixLen + 1 );
-
-		if ( infos[ count ] )
+		if ( !arenaInfo.name.empty() )
 		{
-			strcpy( infos[ count ], info );
-			count++;
+			arenaList.push_back( arenaInfo );
 		}
 	}
-
-	return count;
 }
 
 /*
 ===============
 CG_LoadArenasFromFile
-TODO: consider deleting, unless we want to display 'longname' somewhere
 ===============
 */
-static void CG_LoadArenasFromFile( char *filename )
+static bool CG_LoadArenasFromFile( std::string filename, bool legacy = false )
 {
+	if ( legacy )
+	{
+		Log::Debug( "Loading legacy arena file: %s", filename );
+	}
+	else
+	{
+		Log::Debug( "Loading arena file: %s", filename );
+	}
+
 	std::error_code err;
 	std::string text = FS::PakPath::ReadFile( filename, err );
+
 	if ( err )
 	{
-		Log::Warn( "file not found: %s", filename );
-		return;
+		if ( !legacy )
+		{
+			Log::Warn( "Arena file not found: %s", filename );
+		}
+
+		return false;
 	}
 
-	if ( text.size() >= MAX_ARENAS_TEXT )
-	{
-		Log::Warn( "file too large: %s is %i, max allowed is %i",
-			filename, text.size(), MAX_ARENAS_TEXT );
-		return;
-	}
+	CG_ParseInfos( text.c_str() );
 
-	cg_numArenas += CG_ParseInfos( text.c_str(), MAX_ARENAS - cg_numArenas, &cg_arenaInfos[ cg_numArenas ] );
-}
-
-/*
-===============
-CG_MapLoadNameCompare
-===============
-*/
-static int CG_MapLoadNameCompare( const void *a, const void *b )
-{
-	arenaInfo_t *A = ( arenaInfo_t * ) a;
-	arenaInfo_t *B = ( arenaInfo_t * ) b;
-
-	return Q_stricmp( A->mapLoadName, B->mapLoadName );
+	return true;
 }
 
 /*
@@ -158,45 +200,55 @@ static int CG_MapLoadNameCompare( const void *a, const void *b )
 CG_LoadArenas
 ===============
 */
-void CG_LoadArenas()
+void CG_LoadArenas( std::string mapname )
 {
-	int  numdirs;
-	char filename[ 128 ];
-	char dirlist[ 4096 ];
-	char *dirptr;
-	int  i, n;
-	int  dirlen;
+	arenaList.clear();
 
-	cg_numArenas = 0;
-	rocketInfo.data.arenaCount = 0;
+	std::string filename = Str::Format( "meta/%s/%s.arena", mapname, mapname );
 
-	// get all directories from meta
-	numdirs = trap_FS_GetFileListRecursive( "meta", ".arena", dirlist, sizeof( dirlist ) );
-	dirptr = dirlist;
-
-	for ( i = 0; i < numdirs; i++, dirptr += dirlen + 1 )
+	if ( !CG_LoadArenasFromFile( filename ) )
 	{
-		dirlen = strlen( dirptr );
-		Q_strncpyz( filename, "meta/", sizeof filename );
-		Q_strcat( filename, sizeof filename, dirptr );
-		CG_LoadArenasFromFile( filename );
+		/* Legacy games may store more than one map metadata in a
+		single .arena file, the .arena file not having a basename
+		in common with any map.
+
+		The .arena files (whatever the amount of map described in)
+		may be stored in various folders. Many games use the scripts/
+		folder, some may use the maps/ one (like Smokin' Guns).
+
+		Some games like Quake 3 and Quake Live also reads a file
+		named scripts/arenas.txt and Quake Live also reads files
+		with .sp_arena extension in script/ for single-player
+		training maps. */
+
+		char dirlist[ 4096 ];
+		int numdirs = trap_FS_GetFileListRecursive( "", ".arena", dirlist, sizeof( dirlist ) );
+		char *dirptr = dirlist;
+
+		for ( int i = 0, dirlen = 0; i < numdirs; i++, dirptr += dirlen + 1 )
+		{
+			dirlen = strlen( dirptr );
+
+			CG_LoadArenasFromFile( dirptr, true );
+		}
+
+		CG_LoadArenasFromFile( "scripts/arenas.txt", true );
 	}
 
-	Log::Warn( S_SKIPNOTIFY "%i arenas parsed\n", cg_numArenas );
+	Log::Debug( "%i arenas parsed", arenaList.size() );
+}
 
-	for ( n = 0; n < cg_numArenas; n++ )
+arenaInfo_t CG_GetArenaInfo( std::string mapName )
+{
+	for ( arenaInfo_t &arenaInfo : arenaList )
 	{
-		rocketInfo.data.arenaList[ rocketInfo.data.arenaCount ].mapLoadName = BG_strdup( Info_ValueForKey( cg_arenaInfos[ n ], "map" ) );
-		rocketInfo.data.arenaList[ rocketInfo.data.arenaCount ].mapName = BG_strdup( Info_ValueForKey( cg_arenaInfos[ n ], "longname" ) );
-		rocketInfo.data.arenaCount++;
-
-		if ( rocketInfo.data.arenaCount >= MAX_ARENA_MAPS )
+		if ( arenaInfo.name == mapName )
 		{
-			break;
+			return arenaInfo;
 		}
 	}
 
-	qsort( rocketInfo.data.arenaList, rocketInfo.data.arenaCount, sizeof( mapInfo_t ), CG_MapLoadNameCompare );
+	return {};
 }
 
 /*
--- a/src/cgame/cg_gameinfo.cpp
+++ b/src/cgame/cg_gameinfo.cpp
@@ -42,7 +42,7 @@ static char *cg_arenaInfos[ MAX_ARENAS ];
 CG_ParseInfos
 ===============
 */
-int CG_ParseInfos( const char *buf, int max, char *infos[] )
+static int CG_ParseInfos( const char *buf, int max, char *infos[] )
 {
 	char *token;
 	int  count = 0;
@@ -120,7 +120,7 @@ CG_LoadArenasFromFile
 TODO: consider deleting, unless we want to display 'longname' somewhere
 ===============
 */
-void CG_LoadArenasFromFile( char *filename )
+static void CG_LoadArenasFromFile( char *filename )
 {
 	std::error_code err;
 	std::string text = FS::PakPath::ReadFile( filename, err );
--- a/src/cgame/cg_gameinfo.cpp
+++ b/src/cgame/cg_gameinfo.cpp
@@ -41,9 +41,10 @@ static std::vector<arenaInfo_t> arenaList;
 CG_ParseInfos
 ===============
 */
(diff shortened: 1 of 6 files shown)


illwieckz:
I also created a `cg.mapLongName` variable to be used both by the map loading screen and the score board.

This variable uses the long name from the `.arena` file, or fallback on the worldswpawn's message key from the `.bsp` file (what was previously used in the scoreboard), or fallback on the `.bsp` base name.

This means this PR also fixes map naming on scoreboard on various use case.

illwieckz:
Example with:

```
{
	map "plat23"
	longname "Platform 23"
	author "EmperorJack"
	type "tremulous"
}
```

I get:

[map load screen](https://dl.illwieckz.net/b/unvanquished/wip/map-loading-info/20220926-235341-000.unvanquished-map-load-screen.png)

This means people can avoid baking text in level shot, that also means we can translate the sentence (something we can't do in a baked text).

illwieckz:
In the future I would like to be able to set multiple `author` key for maps having multiple authors, and the game building `name by author, author, and author` string from it, but for now we can start with a simple `author, author, author` string in `author` key.

illwieckz:
On a side note, the original code iterating multiple “arenas” is kept on purpose as some legacy games may have more than one map described in one file, and game would have to read all of the files and all the arena blocks to find a map, which I assume the current code is already doing.

What I want to be make sure to only do is to only look for arena files in currently loaded paks. With non-legacy maps it would only read one file and one block.

src/cgame/cg_gameinfo.cpp
```diff
@@ -31,6 +31,174 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+		}
+	}
+
+	return count;
+}
+
+/*
+===============
+CG_LoadArenasFromFile
+TODO: consider deleting, unless we want to display 'longname' somewhere
```
DolceTriade: Remove todo?

src/cgame/cg_gameinfo.cpp
```diff
@@ -31,6 +31,174 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+===============
+*/
+void CG_LoadArenasFromFile( char *filename )
```
DolceTriade, line 122: Do we want to enforce only one arena per file?

Currently it looks like we can have many arenas in a file. I don't think this is what we want or what we do in practice.

src/cgame/cg_gameinfo.cpp
```diff
@@ -31,6 +31,174 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+
+	return Q_stricmp( A->mapLoadName, B->mapLoadName );
+}
+
+/*
+===============
+CG_LoadArenas
+===============
+*/
+void CG_LoadArenas()
```
DolceTriade, line 160: Let's just do 
```
void CG_LoadArena() {
  CG_LoadArenaFromFile(Str::Format("meta/%s.arena", map));
}
```

DolceTriade (commented):
Before I review further, what's the goal for the change?

Do we want to bring back global arena parsing? Are we **only** loading the arena for the current map?

If we are only loading the arena for the current map, I'd like to refactor the code to not parse everything and only parse the current arena and remove all code that is there to support storing the full map list based on the arenas.

src/cgame/cg_gameinfo.cpp
illwieckz: This is deleted in another commit of the PR.

illwieckz:
@DolceTriade 

Now the code tries `Str::Format("meta/%s/%s.arena", map, map)` and if it fails, fallback on “looking for .arena file” for legacy assets. So this should be fast for ported maps, while keeping it working for legacy assets.

The ability to have more than one map per `.arena` file (which is something that was common in quake 3 scene) is kept, but when we directly load the known meta file, the list is expected to only contain one entry and then be very fast to iterate.
 
I assume it makes things fast while keeping compatibility with legacy assets.

I want to keep compatibility with legacy assets, in fact I was thinking about implementing “multiple files, multiple map per file” old way of doing things as a fallback until I seen this was already implemented. 👍️

It looks like all your comments now had an answer.

DolceTriade:
Why is it important to maintain compatibility with unported maps? As far as I can tell, no one uses unported maps, all the maps are ported first and then put on a server.

illwieckz:
It's easier to test legacy assets (whatever the game) when there is a reference implementation that can load them in Dæmon.

src/cgame/cg_main.cpp
```diff
@@ -1253,9 +1318,15 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 
 	CG_SetMapNameFromServerinfo();
 
+	/* The cgs.mapname variable must already be set at this point. */
+	SetMapInfoFromArena( cgs.mapname );
+
 	// load the new map
 	trap_CM_LoadMap(cgs.mapname);
 
+	/* The map must already be loaded at this point. */
```
slipher, line 1327: By the "map being loaded" what you mean is that configstrings must be received from the server and parsed?

src/cgame/cg_local.h
```diff
@@ -1392,6 +1402,10 @@ struct rocketDataSource_t
 	int demoCount;
 	int demoIndex;
 
+	arenaInfo_t arenaList[ MAX_ARENA_MAPS ];
+	int arenaCount;
+	int arenaIndex;
```
slipher, line 1407: arenaIndex is unused

src/cgame/cg_gameinfo.cpp
```diff
@@ -31,12 +31,195 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+		/* Legacy games may store more than one map metadata in a
+		single .arena file, the .arena file not having a basename
+		in common with any map, and store them in various folders
+		like scripts/ or maps/ folders. */
+
+		char dirlist[ 4096 ];
+		int numdirs = trap_FS_GetFileListRecursive( "", ".arena", dirlist, sizeof( dirlist ) );
+		char *dirptr = dirlist;
+
+		for ( int i = 0, dirlen = 0; i < numdirs; i++, dirptr += dirlen + 1 )
```
slipher, line 191: `i` is useless

src/cgame/cg_gameinfo.cpp
```diff
@@ -31,12 +31,195 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+	{
+		dirlen = strlen( dirptr );
+			Log::Warn( "Loading legacy arena file: %s", dirptr );
+			CG_LoadArenasFromFile( dirptr );
+		}
+	}
+
+	Log::Warn( S_SKIPNOTIFY "%i arenas parsed", cg_numArenas );
```
slipher, line 199: Don't use `Warn` for something normal. Also skipnotify doesn't work

src/cgame/cg_gameinfo.cpp
```diff
@@ -31,12 +31,195 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+		rocketInfo.data.arenaList[ rocketInfo.data.arenaCount ].author = BG_strdup( Info_ValueForKey( cg_arenaInfos[ i ], "author" ) );
+		rocketInfo.data.arenaCount++;
+
+		if ( rocketInfo.data.arenaCount >= MAX_ARENA_MAPS )
+		{
+			break;
+		}
+	}
+
+	qsort( rocketInfo.data.arenaList, rocketInfo.data.arenaCount, sizeof( mapInfo_t ), CG_MapLoadNameCompare );
```
slipher, line 214: The sorting is useless since this list is only accessed by linear search

src/cgame/cg_gameinfo.cpp
```diff
@@ -31,12 +31,195 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+	std::string text = FS::PakPath::ReadFile( filename, err );
+	if ( err )
+	{
+		Log::Warn( "file not found: %s", filename );
+		return 0;
+	}
+
+	if ( text.size() >= MAX_ARENAS_TEXT )
+	{
+		Log::Warn( "file too large: %s is %i, max allowed is %i",
```
slipher, line 134: This is a pointless error unless you are trying to strictly enforce arena files to have backwards compatibility with some old software.

src/cgame/cg_gameinfo.cpp
```diff
@@ -31,12 +31,195 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+	cg_numArenas = 0;
+	rocketInfo.data.arenaCount = 0;
+
+	std::string arena_file = Str::Format( "meta/%s/%s.arena", mapname, mapname );
+
+	Log::Debug( "Loading arena file: %s", arena_file );
+
+	if ( CG_LoadArenasFromFile( arena_file.c_str() ) == 0 )
+	{
+		Log::Warn( "Arena file not found: %s", arena_file );
```
slipher, line 180: You have duplicated file not found warnings here and inside CG_LoadArenasFromFile

src/cgame/cg_main.cpp
illwieckz: The code fallbacks on a string from the bsp itself.

slipher: And the way you're getting this string is by having the *server* read it from the BSP and the client read it from the network.

src/cgame/cg_gameinfo.cpp
illwieckz: Isn't it needed to test for `i < numdirs` to stop iterating? (This is what the legacy code does, I just moved `dirlen` declaration within the `for` block).

src/cgame/cg_gameinfo.cpp
illwieckz: Well, that was what the original code did, I can simplify it but all I did there was a revert.

src/cgame/cg_gameinfo.cpp
slipher: Yes, my mistake

src/cgame/cg_local.h
illwieckz: removed.

src/cgame/cg_main.cpp
illwieckz: That's possible. Do we have a better way to do this?

src/cgame/cg_gameinfo.cpp
illwieckz: Right, that was legacy code I just reverted, I made it a Debug.

src/cgame/cg_gameinfo.cpp
illwieckz: Remove the sorting, we don't use that list anywhere else anyway.

src/cgame/cg_gameinfo.cpp
illwieckz: Removed.

src/cgame/cg_gameinfo.cpp
illwieckz: Right, I only want to read legacy files targeting legacy engine, people are expected to write non legacy files. I removed the test and the error.

This allowed me to turn the function as a boolean one.

illwieckz:
I did some rewrite and simplification, turning global things into local things (local to their files) when there was not other files using them, ported some C arrays to C++ `vector` and some `char*` strings to `std::string`.

One thing I haven't rewritten is the `CG_ParseInfos` function and related variables. This looks dark magic to me so I don't want to delay this PR on this bikeshedding. This code is meant to work as is (it's a straight reintroduction of a code we had for a decade) and this is good enough™. I even have the intuition we may even remove the `cg_arenaInfos` array entirely and feed the `arenaList` vector directly, but I don't want to take the risk to break this and delay the merge of the PR because of this. Rewriting this would deserve its own PR.

src/cgame/cg_main.cpp
illwieckz: I just noticed something… if what you say is right, this string was read from the network on every frame when displaying the scoreboard, as this code was previously run when rendering `<levelname>` tag.

The code was doing this:

```sh
static void CG_Rocket_DrawLevelName()
{
	Rocket_SetInnerRML( CG_ConfigString( CS_MESSAGE ), RP_QUAKE );
}
```

Now it does (reusing a global variable set at map load):

```sh
static void CG_Rocket_DrawLevelName()
{

	Rocket_SetInnerRML( cg.mapLongName.c_str(), RP_QUAKE );

}
```

I wonder if such thing could be a cause of:

- https://github.com/Unvanquished/Unvanquished/issues/2043

It would have not be enough or the only one cause as it is known that issue can happen when connecting (so, when no scoreboard is displayed).

src/cgame/cg_local.h
```diff
@@ -1318,6 +1320,7 @@ struct rocketMenu_t
 #define MAX_OUTPUTS 16
 #define MAX_MODS 64
 #define MAX_DEMOS 256
+#define MAX_ARENA_MAPS 128
```
slipher, line 1323: unused?

src/cgame/cg_gameinfo.cpp
```diff
@@ -31,12 +31,180 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+		}
+	}
+
+	Log::Debug( "%i arenas parsed", cg_numArenas );
+
+	for ( int i = 0; i < cg_numArenas; i++ )
+	{
+		arenaInfo_t arenaInfo;
+
+		arenaInfo.name = BG_strdup( Info_ValueForKey( cg_arenaInfos[ i ], "map" ) );
```
slipher, line 181: Remove strdups

src/cgame/cg_main.cpp
slipher: No it is not doing something on the network every frame. The server sends config strings at the beginning and when one of them changes and the client stores the latest values.

illwieckz:
So, after sleeping a bit, the `CG_ParseInfos` function looked more clear to me. I rewrote it a bit and managed to entirely remove the writing to a “Info” key-value strorage format. Not only that means the code now sets the `arenaList` directly instead of filling an “Info” string and then read it to set the `arenaInfo` and fill the `arenaList`, it means the content of `.arena` files are not suffering from restrictions of the “Info” format: we are expected to now be able to use the `\` character in values, meaning we may implement escaping `"` the `\"` way or things like that to also enable this character in values, and we are expected to now be able to allow the `.arena` files to use multiple keys with the same name (similar to the ogg metadata format, see https://github.com/Unvanquished/Unvanquished/issues/2114#issuecomment-1229478131 ).

Note: the proposed `author` key name is known to be used for map authos by Smokin' Guns and Q3 Rally, so it is good to do the same.

src/cgame/cg_main.cpp
```diff
@@ -1164,6 +1164,87 @@ static void GenerateNavmeshes()
 	cg.loadingNavmesh = false;
 }
 
+static void SetMapInfo( arenaInfo_t arenaInfo )
+{
+	if ( arenaInfo.longName.empty() )
```
slipher, line 1169: It seems wrong to skip making the authors string because there is no long name

src/cgame/cg_main.cpp
```diff
@@ -1253,9 +1334,15 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 
 	CG_SetMapNameFromServerinfo();
 
+	/* The cgs.mapname variable must already be set at this point. */
+	SetMapInfoFromArena( cgs.mapname );
+
 	// load the new map
 	trap_CM_LoadMap(cgs.mapname);
 
+	/* The map must already be loaded at this point. */
```
slipher, line 1343: This comment still needs to be fixed

src/cgame/cg_main.cpp
illwieckz: It was a leftover from when I was producing a single “map by author” string, removed.

src/cgame/cg_main.cpp
illwieckz: This was rewritten and this comment is gone.

DolceTriade (commented):
lgtm % any more of @slipher's comments

src/cgame/cg_gameinfo.cpp
```diff
@@ -31,12 +31,234 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+
+		char dirlist[ 4096 ];
+		int numdirs = trap_FS_GetFileListRecursive( "", ".arena", dirlist, sizeof( dirlist ) );
+		char *dirptr = dirlist;
+
+		for ( int i = 0, dirlen = 0; i < numdirs; i++, dirptr += dirlen + 1 )
+		{
+			dirlen = strlen( dirptr );
+
+			Log::Warn( "Loading legacy arena file: %s", dirptr );
```
slipher, line 232: probably shouldn't be a warning

src/cgame/cg_main.cpp
```diff
@@ -1205,6 +1272,11 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 	// old servers
 	cgs.gameState = gameState;
 
+	/* Set level name and authors string to be used by the UI.
+	For some reason it must be done after cgs.gameState is set
```
slipher, line 1276: It's not a mystery - `cgs.gameState` is the array of config strings.

src/cgame/cg_main.cpp
```diff
@@ -1164,6 +1164,73 @@ static void GenerateNavmeshes()
...
+
+	/* Use longname from .arena file as map name if exists,
+	also read author name if exists. */
+	CG_LoadArenas( mapName );
+
+	arenaInfo_t arenaInfo = CG_GetArenaInfo( mapName);
+
+	if ( arenaInfo.longName.empty() )
+	{
+		/* Use worldswpawn's message key from .bsp file
```
slipher, line 1212: typo

slipher (approved):

src/cgame/cg_main.cpp
illwieckz: Edited the comment.

src/cgame/cg_gameinfo.cpp
illwieckz: This line was even a duplicate of a Debug print elsewhere. Now removed.

illwieckz:
So, I squashed all the wip commits about implementing the feature (and the countless rewrites).

I added two commits to “_use `std::string` more_”, what do you think about them?

src/cgame/cg_gameinfo.cpp
```diff
@@ -41,21 +41,20 @@ static std::vector<arenaInfo_t> arenaList;
...
+static void CG_ParseInfos( std::string text )
 {
-	char *token;
-	char key[ MAX_TOKEN_CHARS ];
+	const char *buf = text.c_str();
 
 	while ( true )
 	{
-		token = COM_Parse( &buf );
+		std::string token = COM_Parse( &buf );
```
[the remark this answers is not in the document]
illwieckz: Well, it makes possible to use easier to read syntax like `str == '"something"` or `str.empty()` instead of a bunch of `strcmp(str, "something") == 0` and others. This is just making code looking better.

src/cgame/cg_gameinfo.cpp
```diff
@@ -41,21 +41,20 @@ static std::vector<arenaInfo_t> arenaList;
 CG_ParseInfos
 ===============
 */
-static void CG_ParseInfos( const char *buf )
+static void CG_ParseInfos( std::string text )
```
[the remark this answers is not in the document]
illwieckz: Right. Also what's the difference between `Str::StringRef text` and `std::string &text` ?

src/cgame/cg_gameinfo.cpp
```diff
@@ -67,38 +66,38 @@ static void CG_ParseInfos( const char *buf )
...
 
-			Q_strncpyz( key, token, sizeof( key ) );
+			std::string key = token;
 
 			token = COM_ParseExt( &buf, false );
 
-			if ( !token[ 0 ] )
+			if ( token.empty() )
 			{
-				strcpy( token, "<NULL>" );
```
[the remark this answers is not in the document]
illwieckz: Well, this looks to be magical code.

The `COM_ParseExt` function is [there](https://github.com/DaemonEngine/Daemon/blob/13e4aca4e6fcce7f521a8c3df50e380500fe422a/src/engine/qcommon/q_shared.cpp#L571), which seems to return a pointer to some memory in a global variable named `com_token` defined [there](https://github.com/DaemonEngine/Daemon/blob/13e4aca4e6fcce7f521a8c3df50e380500fe422a/src/engine/qcommon/q_shared.cpp#L418).

So here maybe I'm duplicating memory.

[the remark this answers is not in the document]
DolceTriade: Yes, Maybe you can use std::string_view there?

src/cgame/cg_gameinfo.cpp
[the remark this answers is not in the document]
DolceTriade: Yes. I think std::string_view is a better way to handle this. std::string makes a copy every time you assign something to it.

slipher:
StringRef in function arguments is good. The other new changes to use std::string are not so great; this means more unneeded copying and allocations. 

The `strcpy(token, "<NULL>")` thing is indeed a bug; COM_Parse ought to return a const pointer. Instead of that you can do `token = "<NULL>"` with a `const char* token`.

src/cgame/cg_gameinfo.cpp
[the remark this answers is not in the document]
DolceTriade: std::string& text passes a reference to the std::string object.

Str::StringRef is wrapper around the pointer to the underlying string data.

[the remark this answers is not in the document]
DolceTriade: Str::StringRef is equivalent to std::string_view, but std::string_view didn't exist when we wrote Str::StringRef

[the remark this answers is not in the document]
slipher: StringRef is NOT equivalent to string_view. StringRef holds a pointer to a null-terminated string ("C string") while string_view can be an arbitrary byte sequence. Also string_view is C++17 so it still doesn't exist now :)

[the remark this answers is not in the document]
DolceTriade: Ah. Right, StringRef points an entire string but string_view can point to arbitrary parts of it.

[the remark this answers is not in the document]
DolceTriade: Ah. Right, StringRef points an entire string but string_view can point to arbitrary parts of it.

src/cgame/cg_gameinfo.cpp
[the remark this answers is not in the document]
illwieckz: It looks like `std::string_view` is C++17, do we have that? On my end it doesn't build.

https://en.cppreference.com/w/cpp/string/basic_string_view

illwieckz:
@slipher, @DolceTriade what do you think about my latest commit?

Note: it looks like `string_view` isn't available anyway.

DolceTriade:
lgtm % @slipher's comments

necessarily-equal:
gentle bump. I can give a review after this is rebased

illwieckz:
@slipher I modified the last commit to only turn `std::string` into `Str::StringRef` and not `const char*` into `std::string`.

slipher:
Last commit LGTM

PR #2227, opened 2022-09-26
tutorial: Can't refill ammo at the RC.

DolceTriade:


necessarily-equal:
For energy weapons like pulse, lasgun and luci, a reactor or a drill is enough. I think we should on the contrary add the drill to the list

DolceTriade:
Yes, I did test with mdriver. I didn't look too closely at the code. Thanks for checking :)

necessarily-equal:
Fixed properly by https://github.com/Unvanquished/Unvanquished/commit/d89b84eb63d627b6c469f63b81052d14decd9905. This PR (which was already closed) is now "finished"

PR #2230, opened 2022-09-28
rmlui: Bounds check the chat cursor.

### LINKED ISSUE #2138: reported crash: Unhandled exception out_of_range in basic_string
```
<Sweet> a player had a client crash while playing on human team, and reports this as the last output line:
<Sweet> Warn: CGame VM: Unhandled exception (St12out_of_range): basic_string
<Sweet> the player cannot remember what he did before that
<Sweet> that is all i know, so i don't think i should write an issue for that
```

DolceTriade:
We let you move the cursor forward every time you press right, even if it is already at the end of the string, this can allow you to go past the end of the string when you try to enter the next character which can lead to a crash.

Thanks to @sweet235 for the report and repro instructions.

Fixes #2138

### DIFF
--- a/src/cgame/rocket/rocketChatField.h
+++ b/src/cgame/rocket/rocketChatField.h
@@ -284,6 +284,10 @@ protected:
 
 	void MoveCursorForward()
 	{
+		if (cursor_character_index >= text.size() )
+		{
+			return;
+		}
 		cursor_character_index += Q_UTF8_Width( text.c_str() + cursor_character_index );
 	}
 


sweet235:
I run that on map testing server, can confirm that it fixes https://github.com/Unvanquished/Unvanquished/issues/2138

necessarily-equal (approved):
LGTM, haven't tested

necessarily-equal:
Yay! Thanks for fixing this

src/cgame/rocket/rocketChatField.h
```diff
@@ -284,6 +284,10 @@ class RocketChatField : public Rml::Element, Rml::EventListener
 
 	void MoveCursorForward()
 	{
+		if (cursor_character_index >= static_cast<size_t>( Q_UTF8_Strlen( text.c_str() ) ) )
```
slipher, line 287: cursor_character_index is actually just a byte index. So you want text.size() not Q_UTF8_Strlen.

DolceTriade: Ahh... I thought that cursor_character_index was the index of the character. I didn't look at the code closely enough. Thanks.

PR #2232, opened 2022-09-30
Small stuff

necessarily-equal:
Various refactoring commit extracted from some bigger branches. To make reviewing those bigger branches easier


DolceTriade (approved):
good stuff lgtm

src/shared/bg_utilities.cpp
```diff
@@ -172,10 +172,15 @@ void ToggleFlag(int &flags, int flag) {
 }
 
 // imported from daemon.
-// Given 3 degree angles, computes yaw, pitch and roll, that is,
-// computes normalized vectors describing a 3D space.
+// Given 3 degree angles (yaw, pitch and roll) computes three vectors that
+// point forwards, rightwards and upwards.
+//
+// The vectors are orthogonal and normalized, but this isn't a proper basis. In
```
slipher, line 178: How is it not a proper basis?

necessarily-equal: I think I explained that in the comment. Basically the idea is that the vectors are not ordered correctly, which means that (with `×` the vector product) you don't have `firstAxis×secondAxis=thirdAxis`, as here `forward×right≠up`. 

On the other hand, `forward×right=down`, or `forward×left=up` are correct though, and AxisVector actually does the latter.

If you prefer, I could write "isn't a proper axis system" maybe.

slipher: Well the usual right X forward = up works. This is just about the order of the function arguments? I don't think it matters that much since they are returned in separate variables not an array, so you can easily use them in any desired order. If you want you could reorder the arguments in the function and its callers to put right first. I don't like this comment creating FUD about the results of the function.

necessarily-equal:
I removed the glm documentation commit before merging

PR #2235, opened 2022-10-04
Refactor Voting code and allow specifying custom votes in the admin.dat

### LINKED ISSUE #2228: Allow defining custom votes
Now that we have active severs, it's clear that it is not scalable to force C++ changes every time someone wants to add a vote. Instead, we should allow creating votes in a config file or something. Ultimately, votes don't need anything fancy:

Something like this in the admin.dat:

```
[vote]
vote = scenario
exec = setScenario.cfg
args = 1

[teamvote]
vote = slap
exec = slap
args = 1
```

The `args` flag would allow that many arguments to be passed to `callvote` and those arguments would be appended to the `exec` command in that order.

In .cfg scripts, we can reference these arguments using cvars `arg_0`, `arg_all`, and `arg_count` using cvar substitution. For example, setScenario.cfg could be something like:

```
set g_layout $arg_1$
exec $arg_1$.cfg
map_restart
pr -1 "^1Loading scenario: ^2$arg_1$"
```

DolceTriade:
Fixes #2228


slipher:
What does the commit `votes: Stringify votes for better portability to custom votes.` do?

I don't see the point of the F macro; you can just add a table header comment at the top of the table to see what the fields are. Well, the final form with the functions inside doesn't resemble a table at all, but I liked the table form better. Also with the F macro the compiler won't warn if you forgot some fields.

The example custom votes suffer from quoting bugs -- you need to quote a variable every time you use it like in Bash.

DolceTriade:
> hat does the commit votes: Stringify votes for better portability to custom votes. do?

Just removes const char* and replaces it with Str::StringRef or std::string for better ergonomics

> I don't see the point of the F macro; you can just add a table header comment at the top of the table to see what the fields are.

I couldn't think of a better macro to to define structs. The table for votes requires scrolling, so you don't have the full context for the votes. Further, you can't define the functions inline in the table, which I think is better because you only need to modify one location to add a new vote instead of two places in sg_cmds.cpp and then add a vote definition in bg_defs.

> Also with the F macro the compiler won't warn if you forgot some fields.

I thought that this might be a feature in that you don't need to set fields unnecessarily.

> The example custom votes suffer from quoting bugs -- you need to quote a variable every time you use it like in Bash.

Hm, I hadn't considered that. Should document it. I'm not sure if we should quote it directly when we do the substitution because that would prevent us from combining arguments in a single command argument...

Ultimately, what would you like to see to make this mergable?

DolceTriade:
On second thought, we should just quote the args prior to insertion to help prevent inject and other attacks. I don't anticipate many usecases for merging these things. And if there are, they can be done using scripts, not using templates.

slipher:
I don't like the functions inside the table; I prefer the old way with one entry per line. The functions could still go in sg_votes, just above the table somewhere. If you really like this we could ask someone for a 3rd opinion.

> Hm, I hadn't considered that. Should document it. I'm not sure if we should quote it directly when we do the substitution because that would prevent us from combining arguments in a single command argument...

My last comment didn't quite make sense because I was assuming it still worked via cvar substitution (like the .cfg-based prototype). So substituting manually and Cmd::Escape()ing anything that goes into the command string (but not the display string) seems better, but there are still potential code injection bugs. Suppose someone tried this:
```
[vote]
name = dayban
target = T_PLAYER
display = Ban $name$ for 1 day
vote = ban $slot$ 1d "Vote-banned by $caller$"
```

Then Cmd::Escape-added quotes could cancel out the one at the beginning of the string and the caller could inject arbitrary code via their username.

The built-in cvar substitution has the problem of being unsafe by default (i.e. with unquoted args), while escaping everything has the problem of not knowing if it is in a quoted string. So in either case a good thing to do would be to ensure that only a single command is executed (instead of a sequence of semicolon- or newline-separated commands like trap_SendConsoleCommand allows). This would greatly limit the damage of escaping bugs if you can only add arguments to a preset command.

DolceTriade:
At the very least I'd like to avoid a mega function of if/else or a huge switch statement. I'm happy to move the functions outside of the definition vote and bring back the table, (although I think I table if it gets too wide or too long starts being a liability).

As for security, I'll see if there is a way to limit what commands are run. fwiw, cvar substitution is not allowed when running commands using `trap_SendConsoleCommand`, so you can't inject commands that way[1].

[1]: We call [Cmd::BufferCommandText](https://github.com/DaemonEngine/Daemon/blob/13e4aca4e6fcce7f521a8c3df50e380500fe422a/src/engine/framework/CommonVMServices.cpp#L247) here, which is defined as having the default argument `parseCvars = false` [here](https://github.com/DaemonEngine/Daemon/blob/736473180ca4b3c3106fa9584750898c94ce0d41/src/engine/framework/CommandSystem.h#L72)

slipher:
The way to do injection is with semicolon-separated commands.

DolceTriade:
@slipher btw, do you know of a good way to automatically maintain a table using automatic formatting. All my attempts to do so got it reformatted...

slipher:
I think you just have to turn formatting off for the table and do it by hand.

DolceTriade:
I don't think we can inject semicolons with my changes:
```
]/callvote max_miners "5;quit" 
savage.genius called a vote: Set max miners to '"\"5;quit\""' 
vote: vote cast 
Vote passed (1 — 0) 
Value '"5;quit"' is not valid for cvar g_maxMiners: value ""5;quit"" is not of type 'int' as expected 
```

slipher:
> I don't think we can inject semicolons with my changes:

I know you can't inject with the most common case. The injection is only possible if the user tries to use a variable inside quotes as demonstrated in [this comment](https://github.com/Unvanquished/Unvanquished/pull/2235#issuecomment-1272141220).

DolceTriade:
Ok,

 - Names cannot have semicolons or quotes
 - I wasn't able to reproduce any injections with quotes...

I also added some command escapes to ensure this.

slipher:
So if we imagine that a reason is allowed for custom votes (currently there is a `$reason$` substitution, but it is impossible to actually set a reason), then my example from the previous comment can be altered slightly:

```
[vote]
name = dayban
target = T_PLAYER
display = Ban $name$ for 1 day
reasonNeeded = yes  // this option is not implemented
vote = ban $slot$ 1d "Vote-banned by $caller$ for $reason$"
```

Then if I set `";quit` as the reason, the command will be split on the semicolon and the `quit` command executed.

DolceTriade:
Reason is not allowed, but using $arg$ is, so let me alter your example:
```
[vote]
name               = echo
stopOnIntermission = 0
adminImmune        = 0
quorum             = 0
type               = V_ANY
target             = T_OTHER
vote               = echo $arg$ 1d Vote-banned by $arg$ for \"$arg$\"
display            = Ban $arg$ for 1 day
```

```
]/callvote echo "\"; quit" 
 -- echo "\"; quit" 1d Vote-banned by "\"; quit" for \""\"; quit"\" 
 -- Ban "\"; quit" for 1 day 
yeahh called a vote: Ban "\"; quit" for 1 day 
vote: vote cast 
Vote passed (1 — 0) 
"; quit 1d Vote-banned by "; quit for ""; quit"  
```

slipher:
What a dreadful parsing function `admin_readconfig_string` is. Why not just read the rest of the line and trim spaces at the end... It won't round-trip with `admin_writeconfig_string` since the latter writes the string verbatim. This could happen for example when a player name has multiple consecutive spaces.

DolceTriade:
Other than that, this looks ok?

slipher:
You mean other than the code injection issues?

DolceTriade:
Does our engine have infra to preload the args in a command?

Could we modify my contrived example in https://github.com/Unvanquished/Unvanquished/pull/2235#issuecomment-1364966119 to try to repro the code injection issue?

I've been trying to see if I can inject it using the easiest human input (the `$arg$` variable) and so far, it hasn't been injecting anything...

DolceTriade:
I simplified it even further:

```
[vote]
name               = echo
stopOnIntermission = 0
adminImmune        = 0
quorum             = 0
type               = V_ANY
target             = T_OTHER
vote               = ban $arg$
display            = Ban $arg$ for 1 day
```

Do you have any ideas of any strings I can try that might cause an injection attack?

slipher:
You should quote it like this: `echo $arg$ 1d Vote-banned by $arg$ for "\"$arg$\""`

slipher:
I thought of an easy way out -- instead of using Cmd::Escape, make your own escaping function which just puts a backslash in front of every character and *doesn't* put quotes at the beginning and end. This will work in both quoted and unquoted contexts.

DolceTriade:
> echo $arg$ 1d Vote-banned by $arg$ for "\"$arg$\""

Note that the non-escaped quotes won't be preserved by the admin read string parsing function.

The custom escaping function is fine I guess, but I'd still like to identify any potential bugs with the current implementation. I haven't been able to find any (not to say that they don't exist...)

In my head, You can only have escaped quotes in the admin read string function and the Cmd::Escape function again will only have escaped quotes, so it doesn't seem like there is a vector for injection

slipher:
> Note that the non-escaped quotes won't be preserved by the admin read string parsing function.

That's the point: you put in `"\"$arg$\""` to get out `"$arg$`

> The custom escaping function is fine I guess, but I'd still like to identify any potential bugs with the current implementation. I haven't been able to find any (not to say that they don't exist...)

The bug is demonstrated by [this](https://github.com/Unvanquished/Unvanquished/pull/2235#issuecomment-1364966119) with [this fix](https://github.com/Unvanquished/Unvanquished/pull/2235#issuecomment-1366270301).

> In my head, You can only have escaped quotes in the admin read string function and the Cmd::Escape function again will only have escaped quotes, so it doesn't seem like there is a vector for injection

If you have an escaped quote in admin.dat then it becomes a non-escaped quote after going through the readstring function. Also if you have a token beginning with a non-quote character, there is no quote or escape processing at all.

DolceTriade:
Ahh, yeah I repro'd it. Pretty sure I tried this before :(

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1340 @@
...
+	}
+	else
+	{
+		return false;
+	}
+}
+
+// Some macro magic to work around the lack of designated initializers.
+#define F( A ) def.A
+#define DEFINE_VOTE( vote, ... )    \
```
slipher, line 52: bad formatting

src/sgame/sg_votes.h
```diff
@@ -0,0 +1,94 @@
...
+	VOTE_AFTER,    // not within the first N minutes
+	VOTE_REMAIN,   // within N/2 minutes before the timelimit
+	VOTE_NO_AUTO,  // don't automatically vote 'yes'
+};
+
+using VoteHandler =
+	std::function<bool( gentity_t* ent, team_t team, std::string& cmd, std::string& arg,
+                        std::string& reason, std::string& name, int clientNum, int id )>;
+
+struct VoteDefinition
```
slipher, line 54: If we're going to do the F macro it would be nice to have default values for optional fields, so we can omit some of them like you said, and make it easier to avoid undefined behavior.

src/sgame/sg_votes.h
```diff
@@ -0,0 +1,94 @@
...
+// $team$ -- plural team name (humans, aliens)
+// $arg$ -- The argument if VoteTarget is not T_NONE. If VoteTarget is T_PLAYER, arg will be
+//          will be whatever the caller chose to enter to match the player (ie, partial name, slot
+//          number, etc)
+// $reason$ -- The reason for the vote. Will be empty is reasonNeeded is set to qno or qmaybe and no
+// reason was passed in. $slot$  -- The target slot $namelogId$ -- The target namelog ID. You
+// probably want the slot, not this. See dist/configs/game/admin.dat for examples of votes.
+bool G_AddCustomVote( std::string vote, VoteDefinition def, std::string voteTemplate,
+                      std::string displayTemplate );
+bool ParseVoteType( Str::StringRef s, VoteType* type );
```
slipher, line 89: `G_` namespace?

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1340 @@
...
+// We only want to escape for commands that are executed. For display only, there is no injection concern,
+// so don't escape those.
+static std::string G_HandleVoteTemplate( Str::StringRef str, gentity_t* ent, team_t team,
+                                         std::string& cmd, std::string& arg, std::string& reason,
+                                         std::string& name, int clientNum, int id, bool escape )
+{
+	std::unordered_map<std::string, std::string> params = {
+		{ "team", BG_TeamNamePlural( team ) },
+		{ "arg", VoteEscape( arg, escape ) },
+		{ "reason", VoteEscape( reason, escape ) },
```
slipher, line 1178: You should either add a way to set reasonNeeded or get rid of `$reason$`

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1340 @@
...
+	};
+	std::string out;
+	out.reserve( str.size() + arg.size() + reason.size() );
+	size_t c = 0;
+	size_t s = 0;
+	size_t e = 0;
+	while ( c < str.size() )
+	{
+		s = str.find( MARKER, c );
+		if ( s != std::string::npos && !IsEscapedMarker( str, s ) )
```
slipher, line 1192: The escaped marker case looks wrong. Needs to skip ahead by 2

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1340 @@
...
+			return "V_PUBLIC";
+		case V_ANY:
+		default:
+			return "V_ANY";
+	}
+}
+
+bool ParseVoteTarget( Str::StringRef s, VoteTarget* type )
+{
+	static const std::unordered_map<std::string, VoteTarget> map = {
```
slipher, line 1280: We usually use case-insensitive parsing. `std::unordered_map<std::string, VoteTarget, Str::IHash, Str::IEqual>`

src/sgame/sg_votes.h
```diff
@@ -0,0 +1,94 @@
...
+	// where a reason requirement is configurable (reasonNeeded must be true)
+	Cvar::Cvar<bool>* reasonFlag;
+	// Vote function that sets the vote text and the vote display
+	VoteHandler handler;
+};
+
+// Handle a vote. Expected to already have args pushed to the arg stack and called from sg_cmds.cpp
+void G_HandleVote( gentity_t* ent );
+// Add a custom vote. Returns false if the vote was not added, which can happen if a vote of the
+// same name already exists or if the vote argument is empty. voteTemplate is the command that is
```
slipher, line 77: Bad formatting

src/sgame/sg_votes.cpp
DolceTriade: This has been deleted.

src/sgame/sg_votes.h
DolceTriade: It's been removed after feedback on IRC.

src/sgame/sg_votes.cpp
DolceTriade: hm, the point is that `$$` -> `$`. In that scenario, `s` will point to the first marker such that `str[s] == '$'` The IsEscapedmarker function checks that `str[s] == '$' && str[s+1] == '$'`. IN that case, yes, i need another case to check `IsEscapedMarker()`.

src/sgame/sg_votes.h
DolceTriade: nvm, my changes were lost after rebase :sob:

bmorel:
ping

DolceTriade:
Rebased. PTAL.

src/sgame/sg_votes.h
DolceTriade: Should be fixed.

src/sgame/sg_votes.cpp
DolceTriade: DOne.

src/sgame/sg_votes.cpp
DolceTriade: Done

src/sgame/sg_votes.h
DolceTriade: Done

src/sgame/sg_votes.h
DolceTriade: Added back

src/sgame/sg_admin.cpp
```diff
@@ -2137,15 +2180,79 @@ bool G_admin_readconfig( gentity_t *ent )
...
 		}
 	}
 
 	BG_Free( cnf2 );
-	ADMP( va( "%s %d %d %d %d", QQ( N_("^3readconfig:^* loaded $1$ levels, $2$ admins, $3$ bans, $4$ commands") ),
-	          lc, ac, bc, cc ) );
+	ADMP( va(
+		"%s %d %d %d %d %zi",
+		QQ( N_(
+			"^3readconfig:^* loaded $1$ levels, $2$ admins, $3$ bans, $4$ commands $5$ votes" ) ),
```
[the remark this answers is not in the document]
DolceTriade: Added missing comma.

src/sgame/sg_votes.cpp
```diff
@@ -1148,18 +1148,38 @@ static bool IsEscapedMarker( Str::StringRef s, size_t pos )
...
 	std::unordered_map<std::string, std::string> params = {
-		{"team",       BG_TeamNamePlural( team )               },
-		{ "arg",       Cmd::Escape( arg )                      },
-		{ "reason",    Cmd::Escape( reason )                   },
-		{ "name",      Cmd::Escape( name )                     },
-		{ "slot",      std::to_string( clientNum )             },
-		{ "namelogId", std::to_string( id )                    },
-		{ "caller",    Cmd::Escape( ent->client->pers.netname )},
+		{ "team", BG_TeamNamePlural( team ) },
+		{ "arg", VoteEscape( arg, escape ) },
```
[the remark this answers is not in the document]
DolceTriade: done.

src/sgame/sg_votes.cpp
```diff
@@ -1148,18 +1148,38 @@ static bool IsEscapedMarker( Str::StringRef s, size_t pos )
 	return pos + 1 < s.size() && s[ pos ] == MARKER && s[ pos + 1 ] == MARKER;
 }
 
+// Use a custom escape to avoid issues with Cmd::Escape unintentionally terminating quotes
+// for strings based on user input.
+static std::string VoteEscape( Str::StringRef s, bool escape )
```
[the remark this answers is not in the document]
DolceTriade: done.

src/sgame/sg_votes.cpp
DolceTriade: done.

DolceTriade:
Rebased again to fix conflicts.

DolceTriade:
@slipher Any additional comments or am I good to merge this?

slipher:
I found a couple of recent changes that were wiped out by the first commit moving stuff to sg_votes.cpp.
- ADMF_NO_GLOBALVOTE and ADMF_NO_TEAMVOTE from 49517ad0
- 093335aa (use cached numBots)

Haven't rechecked the rest yet

DolceTriade:
Hm, I'll audit the code.

DolceTriade:
Audited the code and those were the only two commits I found.

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1198 @@
...
+	{
+		s = str.find( MARKER, c );
+		if ( s != std::string::npos && !IsEscapedMarker( str, s ) )
+		{
+			e = str.find( MARKER, s + 1 );
+			if ( e != std::string::npos )
+			{
+				out += str.substr( c, s - c );
+				const auto& param = params[ str.substr( s + 1, e - s - 1 ) ];
+				out += Quote( Cmd::Escape( escape ? VoteEscape( param ) : param ) );
```
slipher, line 1025: Remove Cmd::Escape and Quote. The VoteEscape is to *replace* the other escaping functions which would allow injection attacks. And you don't want any escaping for display strings.

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1198 @@
...
+	return pos + 1 < s.size() && s[ pos ] == MARKER && s[ pos + 1 ] == MARKER;
+}
+
+// Use a custom escape to avoid issues with Cmd::Escape unintentionally terminating quotes
+// for strings based on user input.
+static std::string VoteEscape( Str::StringRef s )
+{
+	if ( s.empty() )
+	{
+		return "";
```
slipher, line 983: Make this `"\"\""` since we are removing Cmd::Escape

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1198 @@
...
+		return false;
+	}
+	it = voteInfo.emplace( std::move( vote ), std::move( def ) ).first;
+	auto handler = [ vt = move( voteTemplate ), dt = move( displayTemplate ) ](
+					   gentity_t* ent, team_t team, std::string& cmd, std::string& arg,
+					   std::string& reason, std::string& name, int clientNum, int id )
+	{
+		std::string voteCmd =
+			G_HandleVoteTemplate( vt, ent, team, cmd, arg, reason, name, clientNum, id, true );
+		Q_strncpyz( level.team[ team ].voteString, voteCmd.c_str(),
```
slipher, line 1062: It would better to have this be a std::string or have an error when the command is too long

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1198 @@
...
+}
+
+static constexpr char MARKER = '$';
+
+static bool IsEscapedMarker( Str::StringRef s, size_t pos )
+{
+	return pos + 1 < s.size() && s[ pos ] == MARKER && s[ pos + 1 ] == MARKER;
+}
+
+// Use a custom escape to avoid issues with Cmd::Escape unintentionally terminating quotes
```
slipher, line 977: To make this clearer please mention that it needs to work in both quoted and unquoted contexts.

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1198 @@
...
+		            sizeof( level.team[ team ].voteDisplayString ) );
+		return true;
+	};
+	it->second.handler = std::move( handler );
+	return true;
+}
+
+bool G_ParseVoteType( Str::StringRef s, VoteType* type )
+{
+	static const std::unordered_map<std::string, VoteType, Str::IHash> map = {
```
slipher, line 1076: You have to put Str::IEqual also to make it actually case insensitive. This applies to the other 3 cases also

slipher (commented):
I just merged one more commit which needs to be added back: 538a96466e70783b67e21897fe11d4ede9c45c79 (sorry my bad)

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1198 @@
...
+
+// We only want to escape for commands that are executed. For display only, there is no injection
+// concern, so don't escape those.
+static std::string G_HandleVoteTemplate( Str::StringRef str, gentity_t* ent, team_t team,
+                                         std::string& cmd, std::string& arg, std::string& reason,
+                                         std::string& name, int clientNum, int id, bool escape )
+{
+	std::unordered_map<std::string, std::string> params = {
+		{"team",       BG_TeamNamePlural( team )  },
+		{ "arg",       arg						},
```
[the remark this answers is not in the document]
DolceTriade: idk why clang format does this. I'm inclined to not fix it to avoid having clang-format change it back every time :-/

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1198 @@
...
+	{
+		return false;
+	}
+	return level.team[ team ].voteTime && it->second.stopOnIntermission;
+}
+
+/*
+==================
+isDisabledVoteType
+reason[0] = '\0'; // nullify since we've used it here...
```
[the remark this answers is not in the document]
DolceTriade: done.

src/sgame/sg_votes.cpp
DolceTriade: done

src/sgame/sg_votes.cpp
DolceTriade: Added warning + error out

src/sgame/sg_votes.cpp
DolceTriade: noted

src/sgame/sg_votes.cpp
DolceTriade: oops thanks

src/sgame/sg_votes.cpp
```diff
@@ -0,0 +1,1212 @@
...
+	auto handler = [ vt = move( voteTemplate ), dt = move( displayTemplate ) ](
+					   gentity_t* ent, team_t team, std::string& cmd, std::string& arg,
+					   std::string& reason, std::string& name, int clientNum, int id )
+	{
+		std::string voteCmd =
+			G_HandleVoteTemplate( vt, ent, team, cmd, arg, reason, name, clientNum, id, true );
+		std::string display =
+			G_HandleVoteTemplate( dt, ent, team, cmd, arg, reason, name, clientNum, id, false );
+
+		if ( voteCmd.size() > sizeof( level.team[ team ].voteString ) )
```
slipher, line 1064: These two should be `>= sizeof(...)`

slipher:
LGTM modulo comments

src/sgame/sg_votes.cpp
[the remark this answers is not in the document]
DolceTriade: Done.

src/sgame/sg_votes.cpp
DolceTriade: Done.

PR #2236, opened 2022-10-04
Misc fixes to the "give" cmd fixes

necessarily-equal:


necessarily-equal:
I've tested every command and variant of the commands. (For the record I did it the previous time, and except for `/give momentum 100 aliens`, everything worked)

src/sgame/sg_cmds.cpp
```diff
@@ -631,8 +631,8 @@ static void Cmd_Give_f( gentity_t *ent )
 		if ( fabsf(amount) >= 1.0f )
 		{
 			valid = true;
```
DolceTriade, line 633: Is valid still required now?

necessarily-equal: Yes, otherwise it prints an error message

DolceTriade (approved):

PR #1856, opened 2022-07-02
drills can reload energy weapons

bmorel:
fix https://github.com/Unvanquished/gameplay/issues/33

PR #2237, opened 2022-10-04
Fixes the energy resupply hint, for weapon types and reload places

necessarily-equal:
1. We can reload energy weapons at drills too
2. The mass driver isn't an energy weapon. Trying to keep the list up to date is bound to be outdated at times

Fixes what https://github.com/Unvanquished/Unvanquished/pull/2227 wanted to fix, and (hopefully) finishes implementing https://github.com/Unvanquished/gameplay/issues/33 / https://github.com/Unvanquished/Unvanquished/pull/1856 correctly.

### DIFF
--- a/src/cgame/cg_servercmds.cpp
+++ b/src/cgame/cg_servercmds.cpp
@@ -626,9 +626,9 @@ static void CG_Menu( int menuType, int arg )
 			break;
 
 		case MN_H_NOENERGYAMMOHERE:
-			longMsg = _("You must be near a Reactor or a powered Armoury "
+			longMsg = _("You must be near a Reactor, a Drill or a powered Armoury "
 			          "in order to purchase energy ammunition.");
-			shortMsg = _("You must be near a Reactor or a powered Armoury");
+			shortMsg = _("You must be near a Reactor, a Drill or a powered Armoury");
 			break;
 
 		case MN_H_NOROOMARMOURCHANGE:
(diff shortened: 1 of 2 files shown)


necessarily-equal:
Tested with most weapons, notably all the strange ones (blaster, ckit, painsaw, aliens with projectiles), luci and every weapon from smg to lasgun (the others should be fine).

src/cgame/cg_servercmds.cpp
```diff
@@ -626,9 +626,9 @@ static void CG_Menu( int menuType, int arg )
 			break;
 
 		case MN_H_NOENERGYAMMOHERE:
-			longMsg = _("You must be near a Reactor or a powered Armoury "
+			longMsg = _("You must be near a Reactor, a Drill or a powered Armoury "
```
bmorel, line 629: this part can simply be eradicated I'd bet, because `MN_H_NOENERGYAMMOHERE` is **never** used. Well, except, of course, if some magic is done somewhere to refer to `MN_NONE + xx`, but I'd dare say this would be a quite serious bug.

src/cgame/cg_tutorial.cpp
```diff
@@ -351,29 +351,17 @@ static void CG_HumanText( char *text, playerState_t *ps )
 {
 	if ( !ps->ammo && !ps->clips && !BG_Weapon( ps->weapon )->infiniteAmmo )
 	{
-		//no ammo
-		switch ( ps->weapon )
+		// Need to resupply ammo
```
bmorel, line 354: How dare you remove one of those beautifully unmaintainable switches by a sane check? /s

bmorel (approved):

src/cgame/cg_servercmds.cpp
necessarily-equal: I think I've seen you saying that you'd like seing those messages back, so I'm not deleting them for now. Doesn't matter too much I guess

PR #2239, opened 2022-10-04
cp: /cp doesn't conflict with any cgame commands

DolceTriade:
This allows admins to use /cp to center print stuff.


slipher:
LGTM

PR #2240, opened 2022-10-04
rmlui: Properly handle more edge cases in execForm

### LINKED ISSUE #2234: execForm buggy
https://github.com/Unvanquished/Unvanquished/blob/master/src/cgame/cg_rocket_events.cpp#L213

Pretty sure that this code only works when the template is terminated by a $.

DolceTriade:
Before this change, execForm only worked if the Template was terminated by a $ and didn't work otherwise. This properly handles the other cases.

Fixes #2234


slipher:
LGTM

PR #2241, opened 2022-10-04
admin: Fix FIXME where ADMBP_* uses print_tr

DolceTriade:
The strings sent via ADMBP_* are not translatable since they are built up incrementally, so create a new function for use with that function that sends the raw 'print' command instead of 'print_tr'.


necessarily-equal (approved):
Sounds good. (I don't think it fixes a bug tough)

slipher:
LGTM

Issue #2248, opened 2022-10-08
Bots have the chat typing indicator during gameplay


cu-kai:
I've recently rebased my fork on master which has introduced this bug; bots now have a typing indicator above them for no apparent reason. It's inconsistent and does not appear above every bot.

![unvanquished_2022-10-08_205109_000](https://user-images.githubusercontent.com/13281185/194725679-0dfe2bff-4b4a-4aa4-8aa8-e3fd7fb90192.jpg)

DolceTriade:
I probably need to explicitly disable this feature for bots. Thanks for the report.

DolceTriade:
lol wtf, I think it only happens if a bot kills you and the player types... At least that's the only way I repro'd it.

slipher:
Fixed by #2249

PR #2249, opened 2022-10-09
chatbubble: Restrict when chatbubble is down

DolceTriade:
- Ensure bots set the usercmd_t::flag to zero
 - Only set the EF_TYPING flag if the entity is alive and not a bot

This probably helps with #2248


slipher:
LGTM

PR #2251, opened 2022-10-10
Revert "Introduce a new way to iterate entities"

necessarily-equal:
Reverts Unvanquished/Unvanquished#1602


slipher:
LGTM. Squash maybe?

necessarily-equal:
I opened this PR because I thought changing `iterate_entities` to iterate until `level.num_entities` made a huge performance regression, but it's actually fine.

slipher:
```
<Ishq> I think we'll need to revert the entity looping changes
<Ishq> It needs more testing anyways
<afontain[m]> I guess we can do that
```

That's 2 LGTMs I guess

PR #2253, opened 2022-10-10
Component-based entity iteration replacement for ForEntities

slipher:
A drop-in replacement for ForEntities that can be used in a for-each loop (instead of in a lambda) and uses a bit vector to track components instead of `std::set`s. There are companion branches in Daemon (for CountTrailingZeroes) and CBSE (for RegisterComponentCreate/Destroy).

If this is accepted, I'd replace all ForEntities occurrences and remove the component `std::set`s and ForEntities from CBSE.

I made some performance measurements of selected ForEntities occurrences with the existing code and with Entities::Having. I also prepared code to test with the recently introduced `sg_entities_iterator` code, but it crashed so I couldn't produce any results.

```
with ForEntities:
activereac N=24240 min=0us max=25us avg=0us tot=1705us
forbuildable N=7 min=0us max=16us avg=8us tot=57us
forclient N=2114 min=0us max=192us avg=6us tot=12942us

with Entities::Having:
activereac N=24500 min=0us max=24us avg=0us tot=1581us
forbuildable N=7 min=0us max=5us avg=4us tot=33us
forclient N=2116 min=0us max=6us avg=0us tot=158us
```

I did the tests on `operation-dretch`, which has about 450 entities before anyone starts playing, and added 10 bots for 1 minute. `activereac` is a loop over ReactorComponent which would be expected to favor ForEntities since there is only 1 element (but my code was somehow still faster). `forbuildable` is a loop over BuildableComponent which is expected to favor my code because there are over 200 buildable entities. `forclient` is somewhere in the middle (there are 20). On these more dense loops my code is also faster as expected.

I can push the timing code if anyone wants to see it. It only runs on Windows though.

### DIFF
--- a/src/sgame/CBSE.h
+++ b/src/sgame/CBSE.h
@@ -4,14 +4,33 @@
 #ifndef CBSE_H_
 #define CBSE_H_
 
+#include "sg_local.h"
+
+template<typename T> constexpr int ComponentPriority(); // forward declaring from CBSEBackend.h
+
+// Component creating/destruction callbacks. These are used to enable iterating
+// over all entities with a given component.
+void RegisterComponentCreate(int entityNum, int componentNum);
+void RegisterComponentDestroy(int entityNum, int componentNum);
+
+template<typename T>
+void OnComponentCreate(T* component)
+{
+	RegisterComponentCreate(component->entity.oldEnt - g_entities, ComponentPriority<T>());
+}
+
+template<typename T>
+void OnComponentDestroy(T* component)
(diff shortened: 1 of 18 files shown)


DolceTriade:
I do like the ergonomics of this. We don't have a bot component so this code can't be used to iterate over bots?

slipher:
> We don't have a bot component so this code can't be used to iterate over bots?

Yeah, you'd have to use ClientComponent and skip over non-bots.

PR #2255, opened 2022-10-11
merge CG_LoadArenas into CG_Rocket_BuildMapList

bmorel:
Ignore the branch name, this is some start of cleanup for that, but is completely independent. This only removes a cpp file that only includes implementation of a single function, which is used only once.


DolceTriade (approved):

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -1083,7 +1084,19 @@ static void CG_Rocket_SortPlayerList( const char*, const char *sortBy )
 static void CG_Rocket_BuildMapList( const char* )
 {
 	Rocket_DSClearTable( "mapList", "default" );
-	CG_LoadArenas();
+	rocketInfo.data.mapList.clear();
+	auto mapList = FS::GetAvailableMaps(false);
+	rocketInfo.data.mapList.reserve( mapList.size() );
+
+	for (const std::string& mapName : mapList )
```
DolceTriade, line 1091: spacing is wrong here :D

necessarily-equal:
I think we seem to be going towards reading more things from the arena file as in https://github.com/Unvanquished/Unvanquished/pull/2226, and this conflicts so I think this can be closed.

PR #2258, opened 2022-10-12
Fix bot obstacle avoidance

necessarily-equal:
Found by @bmorel


DolceTriade:
Verified it fixes bots jumping over turrets for me. lgtm. Thanks freem for the fix :D

PR #2242, opened 2022-10-05
Implement bot skill and listbots

DolceTriade:

PR #2260, opened 2022-10-13
Fix default fill

necessarily-equal:
based on https://github.com/Unvanquished/Unvanquished/pull/2242 and should be merged alongside https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/164

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -594,6 +594,21 @@ void G_BotCleanup()
 }
 
 // add or remove bots to match team size targets set by 'bot fill' command
+static void G_BotCheckDefaultFill()
+{
+	static Cvar::Modified<Cvar::Cvar<int>> g_bot_defaultFill("g_bot_defaultFill", "fills both teams with that number of bots at start of game", Cvar::NONE, 0);
+
+	Util::optional<int> fillCount = g_bot_defaultFill.GetModifiedValue();
+	if ( fillCount )
+	{
+		for ( int team = TEAM_NONE + 1; team < NUM_TEAMS; ++team )
+		{
+			level.team[team].botFillTeamSize = *fillCount;
+			level.team[team].botFillSkillLevel = 0; // default
+		}
+	}
+}
+
 void G_BotFill(bool immediately)
 {
 	static int nextCheck = 0;
@@ -608,6 +623,8 @@ void G_BotFill(bool immediately)
 	{
 		return;
 	}
+
+	G_BotCheckDefaultFill();
 
 	struct fill_t
 	{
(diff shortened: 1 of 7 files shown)


necessarily-equal:
This is deployed as a testing server.

> Experimental Development Server - fix-defaultFill" / https://play.unvanquished.net/172.86.178.34:27992

DolceTriade:
Maybe you could mention a way to reproduce the issue which motivated you to make the change?

necessarily-equal:
hmm, I'd say that to reproduce the bug the easiest is to start a server with the unmodified default cofig. The one in dist/config.

Since my testing server's code do take the config files straight from dist/config/, they get "broken" bots by default, that have a skill of 0.

necessarily-equal:
I've added a commit that does the same fix to g_bot_infiniteFunds. For now, this is still a draft as wainting on https://github.com/Unvanquished/Unvanquished/pull/2242.

slipher:
Good find with the cvar race. Sadly Cvar::Callback is often unusable because of such initialization problems.

Putting the default skill cvar in a function has an undesirable effect that it won't show up in `listcvars` etc. until a filler bot is added. So maybe it would be better to use Cvar::Modified instead of Cvar::Callback to avoid all this init issues. Like
```
Cvar::Modified<Cvar::Cvar<int>> g_botDefaultFill;

// somewhere in main loop
if (Util::optional<int> fillCount = g_botDefaultFill.GetModifiedValue()) { G_BotFill(fillCount); }
```

As for the cvar renaming, that seems better for a 0.54 branch so as not to break configs.

necessarily-equal:
> Putting the default skill cvar in a function has an undesirable effect that it won't show up in listcvars etc. until a filler bot is added.

Good point. I've decided on calling the function once in the bot initialization code. This should fix it.

> As for the cvar renaming, that seems better for a 0.54 branch so as not to break configs.

I mean, I don't think that's that needed but fine. I dropped those commit and will make a PR later. Hopefully.

DolceTriade:
Honestly, I think as long as engine/gamelogic compat is not affected in master, it should be fair game to merge. Servers using master can fix their configs. Clients don't care about the configs being renamed, and if they do, they can rename them too.

src/sgame/sg_bot_util.cpp
```diff
@@ -2246,8 +2246,20 @@ void BotSellUpgrades( gentity_t *self )
...
+	static Cvar::Range<Cvar::Cvar<int>> g_bot_default_skill( "g_bot_default_skill", "Default skill value bots will have when added", Cvar::NONE, 5, 1, 9 );
+	return g_bot_default_skill.Get();
+}
+
 void BotSetSkillLevel( gentity_t *self, int skill )
 {
+	if ( skill == 0 ) {
+		skill = G_BotDefaultSkill();
+	}
+	ASSERT( skill >= 1 && skill <= 9 );
```
slipher, line 2261: should use MIN_SKILL/MAX_SKILL

necessarily-equal: Fixed

src/sgame/sg_bot.cpp
```diff
@@ -562,6 +562,10 @@ void G_BotIntermissionThink( gclient_t *client )
 
 bool G_BotInit()
 {
+	// we read but discard the result to register the cvar, so that people
```
slipher, line 565: Not going to work with lazy navmesh initialization...

necessarily-equal: Ugh. This is getting quite annoying.

I've made a new version using Cvar::Modified. How about that?

slipher:
LGTM

PR #2262, opened 2022-10-13
Bot heal improvements

necessarily-equal:
This brings the following improvements:
* (commit1,2,3) simplifies code a fair bit
* (commit1) avoid some edge case where a booster that goes unpowered would still be tried for healing. Instead the action heal now fails and can find a better target the next time. And possibly other small bugs that could have been stayed unnoticed in the different code paths
* (commit2) avoid using a egg or leech on the roof or walls for healing in case overmind is gone.
* (commit2) try healing on leechs if overmind and eggs are all gone (not that it would help surviving much)

I intend to squash the first two commits before merging, they are kept separate for now because the first one has been reviewed already.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -221,6 +221,46 @@ Scoring functions for logic
 =======================
 */
 
+botEntityAndDistance_t BotGetClosestBuildingAmongTypes(
+		gentity_t *self, const std::initializer_list<buildable_t> buildables )
+{
+	botEntityAndDistance_t best_choice = { nullptr, 1.0e30f };
+	for ( buildable_t buildable : buildables )
+	{
+		botEntityAndDistance_t candidate =
+			self->botMind->closestBuildings[ buildable ];
+		if ( candidate.ent && candidate.distance < best_choice.distance )
+		{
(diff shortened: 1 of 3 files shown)


src/sgame/sg_bot_util.cpp
```diff
@@ -221,44 +221,31 @@ Scoring functions for logic
...
+	static constexpr glm::vec3 up( 0.0f, 0.0f, 1.0f );
 
-	// Aliens
-	if ( self->botMind->closestBuildings[BA_A_BOOSTER].ent )
-	{
-		// powered booster
-		return self->botMind->closestBuildings[BA_A_BOOSTER].ent;
-	}
-	else
+	for ( buildable_t buildable : { BA_A_BOOSTER, BA_A_OVERMIND, BA_A_SPAWN, BA_A_LEECH })
```
slipher, line 234: Is it intentional to always prefer e.g. overmind over spawn regardless of distance?

necessarily-equal: Here, yes because in the previous version it caused trouble, but since the buildable types are more selected I'd say I could restore it.

I'm really unsure but this is the most conservative option, I guess

slipher:
LGTM

PR #2267, opened 2022-10-20
Cleanups

necessarily-equal:
The first two commits are from the betterai branch


src/cgame/cg_predict.cpp
```diff
@@ -194,14 +194,12 @@ static void CG_ClipMoveToEntities( const vec3_t start, const vec3_t mins,
 			break;
 
 		case traceType_t::TT_BISPHERE:
-			ASSERT(maxs != nullptr);
-			ASSERT(mins != nullptr);
```
slipher, line 203: The pointers have to be non-null because they are dereferenced in the next line.

src/cgame/cg_players.cpp
```diff
@@ -2711,8 +2711,7 @@ static void CG_PlayerSplash( centity_t *cent, class_t class_ )
 	}
 
 	// trace down to find the surface
-	trap_CM_BoxTrace( &trace, start, end, nullptr, nullptr, 0,
-	                  ( CONTENTS_WATER | CONTENTS_SLIME | CONTENTS_LAVA ), 0 );
+	CM_BoxTrace( &trace, start, end, nullptr, nullptr, 0, ( CONTENTS_WATER | CONTENTS_SLIME | CONTENTS_LAVA ), 0 , traceType_t::TT_AABB );
```
slipher, line 2714: bad spacing on every CM_BoxTrace line

necessarily-equal: Fixed twice

necessarily-equal: Ok, fixed two more, hopefully all

src/cgame/cg_predict.cpp
DolceTriade: +1

necessarily-equal: I mean, yes, I agree on that, but what's the difference between crashing and crashing? Like the assert is not revealing a programming error that would have been noticed anyway?

slipher:
LGTM

PR #2269, opened 2022-10-21
Bot Skill Tree

necessarily-equal:
<del>Note: the first few commits are from https://github.com/Unvanquished/Unvanquished/pull/2260, which probably should be reviewed before.</del>

### Principle

This makes some bot tricks hidden behind a skill condition. Some example of skills for humans are using medikit, buying armor correctly, predictive aim or running when fleeing. However, unlike the current code which essentially just checks if skill > 5, this makes each bot have only part of those special tricks and not all at once.

### Expected gameplay change

In the current state, the code makes human bot unlock every skill at level7 and aliens unlock every skill at level8, and bot at level 5 have a majority (more than half) of the skills, although not every skill, which makes them more different from each another. I think it shouldn't nerf bots too hard, while allowing to improve them further.

I think this a prerequisite before improving bot further, otherwise bots just become too strong against beginners. This also helps a lot making weaker bots for the unvanquished tutorial I'm working on at times.

### Note about the skill selection algorithm
```
// The [skill selection] algorithm is stable and deterministic.
// Deterministic:
//     a bot's skill set is determined only by the bot's name and the bot's
//     skill level. A different bot will have a different skillset, but this
//     means that a bot with the same name will have the same skillset from
//     a game to another.
// Stable:
//     when a bot changes skill level, the first skills unlocked are common
//     between both skill levels. This means that a bot whose skill is raised
//     by BotSetSkillLevel will keep the skills it already had, and get new
//     ones.
//     This essentially means that a bot will keep its "personality" after
//     a skill change, although keeping its base skills or unlocking new ones.
```


necessarily-equal:
rebased

src/sgame/sg_bot.cpp
```diff
@@ -263,6 +262,188 @@ bool G_BotSetDefaults( int clientNum, team_t team, int skill, Str::StringRef beh
...
 }
 
+struct botSkillTreeElement_t {
+	Str::StringRef name;
+	enum bot_skill enable;
+	int cost;
+	// function that determines if this skill is available
+	std::function<bool(const gentity_t *self, skillSet_t existing_skills)> predicate;
+	// skills that are unlocked once you unlock this one
+	std::vector<botSkillTreeElement_t> unlocked_skills;
```
DolceTriade, line 272: What does this mean? Total cost of this skill = cost + sum(unlocked_skills.cost)?

DolceTriade: Ah I see now. It's more of a skill tree. You need the parent skill before you can have the "unlocked" skills.

src/sgame/sg_bot.cpp
```diff
@@ -263,6 +262,188 @@ bool G_BotSetDefaults( int clientNum, team_t team, int skill, Str::StringRef beh
...
+//     when a bot changes skill level, the first skills unlocked are common
+//     between both skill levels. This means that a bot whose skill is raised
+//     by BotSetSkillLevel will keep the skills it already had, and get new
+//     ones.
+//     This essentially means that a bot will keep its "personality" after
+//     a skill change, although keeping its base skills or unlocking new ones.
+std::pair<std::string, skillSet_t> BotDetermineSkills(gentity_t *bot, int skill)
+{
+	std::vector<botSkillTreeElement_t> possible_choices = initial_unlockable_skills;
+	// aliens have 50 points to spend max
```
DolceTriade, line 400: What if instead of hardcoding spendable points, we made skill be a percentage of the max possible skills people could buy. For example,

if the total cost of all human skill is 100, then skill level 9 would mean that you would have 90 points to spend?

src/sgame/sg_bot_util.cpp
```diff
@@ -599,6 +599,8 @@ int BotGetDesiredBuy( gentity_t *self, weapon_t &weapon, upgrade_t upgrades[], s
 		upgrades[i] = UP_NONE;
 	}
 
+	auto buyArmors = [&]()
+	{
 	for ( auto const &armor : armors )
```
DolceTriade, line 604: indentation is off wherever you added these lambdas?

src/sgame/sg_bot_util.cpp
```diff
@@ -668,6 +679,27 @@ int BotGetDesiredBuy( gentity_t *self, weapon_t &weapon, upgrade_t upgrades[], s
 			numUpgrades ++;
 		}
 	}
+	};
+
+	if ( self->botMind->botSkillSet[BOT_H_PREFER_ARMOR] )
```
DolceTriade, line 684: What's the difference between BOT_H_PREFER_ARMOR vs BOT_H_BUY_ARMOR? For buy armor, I would expect a randomness factor into deciding armor if BOT_H_BUY_ARMOR is set but not B_H_PREFER_ARMOR?

src/sgame/sg_bot.cpp
```diff
@@ -263,6 +262,188 @@ bool G_BotSetDefaults( int clientNum, team_t team, int skill, Str::StringRef beh
 	return true;
 }
 
+struct botSkillTreeElement_t {
```
DolceTriade, line 265: Honestly, I would like to see the skill tree stuff moved to a separate file. It's conceptually different enough to merit it I think.

DolceTriade (commented):
all in all, code looks good to me. I like the concept a lot.

src/sgame/sg_bot.cpp
necessarily-equal: I initially wanted to do that, but there are two problems to do that:

First, the `botSkillTreeElement_t->predicate` functions makes this problem undecidable in the general sense: you can't count how many skill point you might spend at most because you could have skills that are excluding each other (for example some aggressive skill could be disabled once you unlock a defensive skill using a `bool pred_x_not_unlocked(const gentity_t *, skillSet_t existing_skills) { return existing_skills[BOT_A_XYZ]; }`).

A second problem is a balancing one. I've found that doing "skill 5 has 5/9*max skill points" actually nerfs human bots a bit too much in my opinion. I manually tweaked the value up a bit to compensate for that. I don't think it's a bad idea to keep that freedom.

src/sgame/sg_bot_util.cpp
necessarily-equal: I did it like that to avoid conflicts, I thought at the time there may be other versions of this function around, for example in betterai. Not sure.

src/sgame/sg_bot_util.cpp
necessarily-equal: `PREFER_ARMOR` buys armor first, then maybe weapons if it can; while and `BUY_ARMOR` will buy weapon, then maybe armor if it can.

I've found that buying weapons first is a typical beginner "mistake". That's also something bots did before 0.52 or 0.53. Basically having an armor is IMO more important in combat that having a better gun.

Not knowing armor exists at all is of course an even bigger mistake, that's one more opportunity to make low skill even easier!

src/sgame/sg_bot.cpp
necessarily-equal: Makes sense

src/sgame/sg_bot_util.cpp
necessarily-equal: hmm, in case that wasn't clear in my answer, the intent is that there are three levels in buying armor:

* first is total noob, you don't know it exists.
* second is beginner that thinks big guns are great and spends all its money on a pulse blast without buying armor. But hey if they have enough funds they'll buy an armor
* third is what they currently do, which is buying armor first. That's the best thing to do generally speaking.

DolceTriade: This makes sense now, but this is definitely something that should be commented in the code.

src/sgame/sg_bot.cpp
```diff
@@ -263,6 +262,125 @@ bool G_BotSetDefaults( int clientNum, team_t team, int skill, Str::StringRef beh
...
+//     a skill change, although keeping its base skills or unlocking new ones.
+std::pair<std::string, skillSet_t> BotDetermineSkills(gentity_t *bot, int skill)
+{
+	std::vector<botSkillTreeElement_t> possible_choices = initial_unlockable_skills;
+	int skill_points = skill * 4; // FIXME
+
+	// rng preparation
+	std::string name = bot->client->pers.netname;
+	std::seed_seq seed(name.begin(), name.end());
+	std::mt19937_64 rng(seed);
```
slipher, line 412: Use a lower-quality random number generator. This has an unnecessarily huge state.

src/sgame/sg_bot.cpp
```diff
@@ -263,6 +262,125 @@ bool G_BotSetDefaults( int clientNum, team_t team, int skill, Str::StringRef beh
...
+	if (total_skill_points <= 0)
+		return Util::nullopt;
+
+	std::uniform_int_distribution<int> dist(0, total_skill_points - 1);
+	int random = dist(rng);
+
+	// find which skill this corresponds to (higher point skills are more
+	// likely to be found)
+	int cursor = 0;
+	for (std::vector<botSkillTreeElement_t>::iterator skill = possible_choices.begin(); skill != possible_choices.end(); skill++)
```
slipher, line 356: use `auto`

src/sgame/sg_bot.cpp
```diff
@@ -858,7 +858,11 @@ std::string G_BotToString( gentity_t *bot )
 	{
 		return "";
 	}
-	return Str::Format( "^*%s^*: %s [s=%d b=%s g=%s]",
-			bot->client->pers.netname, BG_TeamName( G_Team( bot ) ), bot->botMind->botSkill.level,
-			bot->botMind->behaviorTree->name, BotGoalToString( bot ) );
+	return Str::Format( "^*%s^*: %s [s=%d ss=\"%s\" b=%s g=%s]",
```
slipher, line 924: This is probably going to have issues with the message being too big for one packet. Not sure what the exact limit is bug if you had 20 bots with 20 skills each and a skill is on average 15 characters long then you have 6000 bytes of skills

src/sgame/sg_bot_local.h
```diff
@@ -104,7 +114,9 @@ struct botMemory_t
 	void doSprint( int jumpCost, int stamina, usercmd_t& cmd );
 	usercmd_t   cmdBuffer;
 
-	botSkill_t botSkill;
+	botSkill_t botSkill; // numberical values
```
slipher: botSkill_t is unused, you can delete it

src/sgame/sg_bot_util.h
```diff
@@ -34,6 +34,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 bool PlayersBehindBotInSpawnQueue( gentity_t *self );
 int      BotGetDefaultSkill();
 void     BotSetSkillLevel( gentity_t *self, int skill );
+std::pair<std::string, skillSet_t> BotDetermineSkills( gentity_t *bot, int skill);
```
slipher, line 37: I don't think it makes sense to return the string and the bitset together. You can have a separate function that converts the bitset to a string (assuming the string will be used at all, given the potential issue with it being too long)

src/sgame/sg_bot.cpp
slipher, line 272: Could skill dependencies be moved into `predicate` and `unlocked_skills` removed? You already call the predicate on every skill in the list every time you choose a new skill anyway

slipher (commented):
We use `camelCase` for variable names and struct members, not `snake_case`.

I have just looked at the infrastructure parts and not specific skills.

src/sgame/sg_bot_local.h
slipher: Never mind, typo in my grep

src/sgame/sg_bot.cpp
necessarily-equal: Heh, I didn't do it in the end, but I wanted to write "performance is a non-goal" on the header of this function.

I initially used a 32bits version of the mersenne twister, but had skill which were too similar from a run to another. I don't know if it was just bad luck and if that would still be a problem given that there are more possible skills now.

src/sgame/sg_bot_util.h
necessarily-equal: The problem with that is that there is no straightforward skill->string mapping, and that one needs to write the code to do that. I don't think it's a big problem to keep some 50chars-long strings for each bots anyway.

src/sgame/sg_bot.cpp
necessarily-equal: I guess since `G_admin_listbots` uses `ADMBP_Begin`, that's a problem. I don't know what to do about it though.

src/sgame/sg_bot.cpp
necessarily-equal: I guess? Initially I made it that way because dependencies are declared as a subelement like this which is IMO more readable:
```C++
	// humans
	{ "buy-armor",          BOT_H_BUY_ARMOR,   10, pred_human, {
		{ "prefer-armor", BOT_H_PREFER_ARMOR, 5, pred_human, {} },
	}},
```

And because otherwise this forces to write a lot of functions, or to do trickeries such as writing a `pred_depends_on()` and a `pred_and()` function to do the logical and so you can start doing weird things such as `std::function predicate = pred_and(pred_human, std::bind(pred_depends_on, BOT_H_BUY_ARMOR))`.

I also fear that if dependencies aren't displayed in a tree like that (allowing for some extra dependency if need be), the tree will become much harder to maintain because it wouldn't be possible to have a global view of what depends on what, unless someone draws it on paper when checking or unless we maintain a separate document.

src/sgame/sg_bot_util.cpp
necessarily-equal: Apparently I did put a comment on it in `src/sgame/sg_bot_local.h`, but I see how it's cut away from the code. I don't know what to do about that, the skill, the skill implementation/effect and the explanation are arguably in 3 places :-/
```diff
	BOT_A_TYRANT_CHARGE_ON_FLEE,
	BOT_H_RUN_ON_FLEE, // when fleeing, RUN
+	BOT_H_BUY_ARMOR, // knows armor exists at all
+	BOT_H_PREFER_ARMOR, // prefer to buy armor rather than guns
	BOT_H_MEDKIT, // knows the medkit even exist
```

src/sgame/sg_bot.cpp
slipher: I guess I would do like `[](gentity_t* ent) { return G_Team(ent) == TEAM_HUMANS && ent->botMind->skills[BOT_H_PREFER_ARMOR]; }` instead of trying to compose predicates.

Heh, I didn't notice until now that the skills are literally stored in a tree. I don't see why it should be expected to be a tree and not a more general digraph, like maybe you have a skill that depends on both fighting and moving.

Anyway I'm not insisting on a particular change right here, just discussion.

src/sgame/sg_bot_util.cpp
necessarily-equal: This is part of the reason why humans get more skill points by default. A similar very important skill is the BOT_A_AIM_HEAD one, because there is quite a radical difference in terms of inflicted damage.

src/sgame/sg_bot.cpp
necessarily-equal: And huh, to do that, is it enough to stop using `ADMBP_Begin` and just send the data using ADMP for each bot in listbots?

src/sgame/sg_bot.cpp
```diff
@@ -263,6 +262,188 @@ bool G_BotSetDefaults( int clientNum, team_t team, int skill, Str::StringRef beh
...
+		}
+	}
+
+	if (total_skill_points <= 0)
+		return Util::nullopt;
+
+	std::uniform_int_distribution<int> dist(0, total_skill_points - 1);
+	int random = dist(rng);
+
+	// find which skill this corresponds to (higher point skills are more
```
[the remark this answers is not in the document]
necessarily-equal: It's a good heuristic to how important a skill is to have. I've been pondering about adding one more field to replace that, but I kind of like it as it's easier

src/sgame/sg_bot.cpp
necessarily-equal: Did that in a new commit to avoid having to do that for every commit.

src/sgame/sg_bot.cpp
necessarily-equal: So it was a lot of pain figuring that out, but I've got a version that doesn't overflow now. And also now I hate `ADMP`.

src/sgame/sg_admin.cpp
```diff
@@ -6077,18 +6076,19 @@ bool G_admin_bot( gentity_t *ent )
...
 	{
-		gentity_t* ent = entity.oldEnt;
-		if ( !( ent->r.svFlags & SVF_BOT ) )
+		gentity_t* bot = entity.oldEnt;
+		if ( !( bot->r.svFlags & SVF_BOT ) )
 		{
 			return;
 		}
-		ADMBP( va( "%zu %s", ent - g_entities, G_BotToString( ent ).c_str() ) );
+		AP( Str::Format( "print_tr %s %d %s",
```
DolceTriade, line 6087: given no translation will be done, no need to use `print_tr`

DolceTriade: If I get this correctly, You will be sending the listbots text and the table header using ADMP, which hardcodes the `ent` variable and only sends it to the `ent` player. However, the actual contents of list command, you will send each line via `AP` (ie, to **all** players). this seems undesirable. Just use trap_SendServerCommand at this rate...

Given that this is a debugging command, I'm not super concerned about the ServerCommand spam.

src/sgame/sg_bot_local.h
```diff
@@ -88,6 +90,37 @@ struct botSkill_t
 	float aimShake;
 };
 
+// boolean flags that tells which skill (compétence) the bot has
+enum bot_skill
```
DolceTriade, line 94: why snake case?

DolceTriade (commented):
Build failure looks legit

src/sgame/sg_admin.cpp
necessarily-equal: I mean I would be using ADMP if I could get it to work. But it just doesn't want to work. I'd be grateful if you could figure that one out for me :-/

I think ADMP is now my #1 hated function.

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -0,0 +1,208 @@
+/*
+===========================================================================
+
+Unvanquished GPL Source Code
+Copyright (C) 2022 Antoine Fontaine
```
slipher, line 5: We don't usually write any specific people's name in the license header

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -0,0 +1,208 @@
...
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+===========================================================================
+*/
+
+#include "sg_bot_util.h"
+
+struct botSkillTreeElement_t {
+	Str::StringRef name;
+	enum bot_skill enable;
```
slipher, line 30: Don't write "enum" there

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -0,0 +1,208 @@
...
+//     This essentially means that a bot will keep its "personality" after
+//     a skill change, although keeping its base skills or unlocking new ones.
+std::pair<std::string, skillSet_t> BotDetermineSkills(gentity_t *bot, int skill)
+{
+	std::vector<botSkillTreeElement_t> possible_choices = initial_unlockable_skills;
+	// aliens have 50 points to spend max
+	// humans have 39 points to spend max
+	// here we give a bit more money to humans because they have more
+	// expensive skills
+	float max = G_Team(bot) == TEAM_ALIENS ? 50.0f : 45.0f;
```
slipher, line 167: Do you have any idea on how to maintain these numbers as skill sets change?

src/sgame/sg_admin.cpp
DolceTriade: ADMP is just a hacky way to batch sending large swaths of strings to the client. Your whole mechanism for avoiding overflow depends on **not** batching anything and sending everything. This is obviously inefficient, but required due to the possibly large amounts of text being sent per line.

Instead you want to replace your AP call with

```
trap_SendServerCommand( ent->num(), Str::Format( blah ).c_str() );
```

`AP(arg)` is basically a define for `trap_SendServerCommand( -1, arg );`, and `AP` stands for "All Players" whereas I think `ADMBP` stands for "ADMin Buffered Print".

src/sgame/sg_bot_skilltree.cpp
necessarily-equal: Removed

src/sgame/sg_bot_local.h
necessarily-equal: Habbits, readability and the consistency with the C++ stdlib I think.

I've tried reverting that but this makes it needlessly painful to deal with conflicts when rebasing.

src/sgame/sg_bot_skilltree.cpp
necessarily-equal: So what should we write?

slipher: "Unvanquished Developers" apparently

src/sgame/sg_bot_local.h
slipher: I find `git filter-branch` useful for renaming variables (or an equivalent external tool because filter-branch is badly designed blah blah)

necessarily-equal: I'll take note of that, but I don't think I'll be doing this for this PR because I don't feel like this is worth it.

src/sgame/sg_bot_skilltree.cpp
necessarily-equal: I'd say there is unfortunately a fair bit of feeling and adjustments. Finding some balance that is about right is also quite important, as we want bots to be about balanced for both teams. Thankfully we don't modify the bot skills that often in theory.

On the flip side, we now have a way of balancing bot strength, which was missing for long.

DolceTriade: Can we add a static assert or something based on the size of the skills struct or something that will fail to remind us to update this value? I feel like this is the type of thing that people will forget about.

DolceTriade: Even if the number doesn't end up changing, the static struct will need to be updated, which is the point when new skills are added/removed.

necessarily-equal: Maybe we can add comments everywhere about that, but I'm afraid a static_assert will need some extra function, and possibly be confusing.

One way to do what you suggest is to have a recursive constexpr function that ignores any condition. That's maybe like 10 lines and that can only be useful here. Do you think that's worth it?

DolceTriade: Hm, I think a comment is sufficient. Probably where you define skills and where the number is defined. In hindsight, a static assert might be pretty annoying.

src/sgame/sg_bot_skilltree.cpp
necessarily-equal: Edited

src/sgame/sg_bot_skilltree.cpp
necessarily-equal: I've added a comment for that in a new commit. I think that should be good.

DolceTriade:
lgtm but would want slipher to lgtm too before merge

src/sgame/sg_bot.cpp
slipher: I still think it would be nice to use a generator that doesn't have a 2.5 kB size.

src/sgame/sg_admin.cpp
```diff
@@ -6076,17 +6076,18 @@ bool G_admin_listbots( gentity_t *ent )
...
-		gentity_t* ent = entity.oldEnt;
-		if ( !( ent->r.svFlags & SVF_BOT ) )
+		gentity_t* bot = entity.oldEnt;
+		if ( !( bot->r.svFlags & SVF_BOT ) )
 		{
 			return;
 		}
-		ADMBP( va( "%i %s", ent->num(), G_BotToString( ent ).c_str() ) );
+		trap_SendServerCommand( ent ? ent->num() : -2, Str::Format( "print %s %d %s",
+			QQ( N_("$1$ $2$") ),
```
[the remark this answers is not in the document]
necessarily-equal: Fixed

src/sgame/sg_bot_util.cpp
necessarily-equal: Also to avoid that, at high level, a alien bot of skill level 8 or a human bot of level 7 will always be maxed out.

src/sgame/sg_bot.cpp
necessarily-equal: Maaaaybe ? But also I tested this one and the generator is destroyed once bot have finished their initialisation step so it really doesn't matter?

src/sgame/sg_bot_util.cpp
slipher: > Also to avoid that, at high level, a alien bot of skill level 8 or a human bot of level 7 will always be maxed out.

But that's dependent on some magic numbers and will change as soon as the skill list is modified.

src/sgame/sg_bot.cpp
slipher: Well that's kind of the point; you're initializing this expensive object and then immediately throwing it away.

Why is this so hard? AFAIU it's just find-and-replace of mt19937 with the name of some other generator. If you use a default configuration and don't try to make up your own parameters there shouldn't be any danger.

src/sgame/sg_bot_util.cpp
necessarily-equal: well, yes, the max number of skill points have to be updated too when you update the skills themselves.

necessarily-equal: Some stats for the default human bots at skill 5, for a sample size of 63.

armor: 15 prefered armor, 35 bought armor and 13 didn't buy armor at all
medkit: 43 used medkit, while 20 pretended it doesn't exist
aim prediction: 29 did aim prediction, for 34 that didn't
running on flee: 38 did run on flee, for 25 that didn't

src/sgame/sg_bot_local.h
```diff
@@ -87,6 +89,40 @@ struct botSkill_t
...
+	BOT_A_LEAP_ON_FLEE, // mantis
+	BOT_A_POUNCE_ON_FLEE, // dragoon and adv dragoon
+	BOT_A_TYRANT_CHARGE_ON_FLEE,
+	BOT_H_RUN_ON_FLEE, // when fleeing, RUN
+	BOT_H_BUY_ARMOR, // knows armor exists at all, if the bot doesn't have BOT_H_PREFER_ARMOR too it will always prefer to buy guns
+	BOT_H_PREFER_ARMOR, // prefer to buy armor rather than guns
+	BOT_H_MEDKIT, // knows the medkit even exist
+
+	// fighting skills
+	BOT_B_BASIC_FIGHT, // doesn't do anything as of now
```
slipher, line 117: Why would you have a skill that doesn't do anything?

necessarily-equal: Because it encourages bot specialisation in some "knowledge domain"

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -0,0 +1,208 @@
...
+std::pair<std::string, skillSet_t> BotDetermineSkills(gentity_t *bot, int skill)
+{
+	std::vector<botSkillTreeElement_t> possible_choices = initial_unlockable_skills;
+	// aliens have 51 points to spend max
+	// humans have 40 points to spend max
+	// here we give a bit more money to humans because they have more
+	// expensive skills
+	float max = G_Team(bot) == TEAM_ALIENS ? 51.0f : 46.0f;
+
+	// unlock every skill at skill 8
```
slipher, line 169: If this is part of the design then it should be calculated by summing up the skill costs rather than counting on people to change these numbers.

src/sgame/sg_bot_local.h
slipher: So there are some skills that depend on it? Which ones are they?

slipher: Never mind I found it

src/sgame/sg_bot.cpp
necessarily-equal: Well the reason I'm not so fond of that change is that I had an impression that the previous RNG did give a bad result. The randomness looked not-so-random. So I switched to a fancier generator. Maybe that was just a feeling due to sample size, I don't know.

About how hard is it to change, well I have to edit things twice because the code was moved to another file and ensure it still builds mid-branch to avoid being annoying on bisect, plus some testing. It's a moderately important cost that I don't feel is needed.

If we want to make bot loading faster anyway, we have to look at the navmesh loading — that's what actually takes time.

src/sgame/sg_bot_skilltree.cpp
necessarily-equal: I don't want to set that into stone because 1. I'd like to allow tuning this, for example raising the human max skill count after I add a second aim skill or a precise goon barb skill and 2. because it's impossible to have a reliable count for arbitrary predicate anyway, for example if you have one skill hidden behind another, but outside of the tree.

src/sgame/sg_bot.cpp
slipher: You can add a new commit at the end rather than rebasing.

src/sgame/sg_bot_skilltree.cpp
slipher: OK you can merge it as-is for now, but we should open an issue about making the skill tree maintainable. It's really hard to understand what I should do to these numbers if I add a new skill.

necessarily-equal: Fair enough, I would even mostly agree on that. This is a problem with most things gameplay related though — it's hard to find a balance.

Issue #2211, opened 2022-09-18
Alien bots don't flee anymore


necessarily-equal:
It seems alien bots don't flee anymore on master, and will fight to death every time. Human bots seem to work fine though.

Building a booster doesn't seem to help.

necessarily-equal:
I think I was observing on plat23 and map size meant sqrt(dist) was too big. Maybe the solution here is to rework the scoring function.

necessarily-equal:
I nearly closed this issue a few days ago after confirming that bots did go to heal (both before and after https://github.com/Unvanquished/Unvanquished/pull/2262) and that the healscore gives reasonable results in geogebra.

Using https://github.com/Unvanquished/Unvanquished/pull/2270, I think I found the real culprit: as soon as alien bots notice an enemy, they check their heal score very rarely, likely as the check is protected by a timer and a dice roll. Maybe tweaking the timer or changing the random chance (30%) would help. Currently bot typically just die too fast to start deciding they want to heal.

PR #2270, opened 2022-10-22
Player colored marker

necessarily-equal:
This adds some new debug tool, that can help visualising what is happening.

### Example Usages

Example debug use (used to troubleshoot https://github.com/Unvanquished/Unvanquished/issues/2211)

```diff
diff --git a/src/sgame/sg_bot_ai.cpp b/src/sgame/sg_bot_ai.cpp
index e904a5040..158c38d8a 100644
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1080,6 +1080,8 @@ AINodeStatus_t BotActionHeal( gentity_t *self, AIGenericNode_t *node )
                             && !BG_InventoryContainsUpgrade( UP_MEDKIT, self->client->ps.stats );
        bool fullyHealed = Entities::HasFullHealth(self) && !needsMedikit;
 
+       self->client->debugColor = Color::Color( 0.0f, 0.0f, 1.0f );
+
        if ( self->botMind->currentNode != node )
        {
                if ( fullyHealed )
diff --git a/src/sgame/sg_bot_parse.cpp b/src/sgame/sg_bot_parse.cpp
index 53f9bbf3d..fdfa0ee3e 100644
--- a/src/sgame/sg_bot_parse.cpp
+++ b/src/sgame/sg_bot_parse.cpp
@@ -166,7 +166,9 @@ static AIValue_t baseRushScore( gentity_t *self, const AIValue_t* )
 
 static AIValue_t healScore( gentity_t *self, const AIValue_t* )
 {
-       return AIBoxFloat( BotGetHealScore( self ) );
+       float score = BotGetHealScore( self );
+       self->client->debugColor = Color::Color( Math::Clamp(score, 0.0f, 1.0f), 0.5f, 0.0f );
+       return AIBoxFloat( score );
 }
 
 static AIValue_t botClass( gentity_t *self, const AIValue_t* )
```

Another example use, testing https://github.com/Unvanquished/Unvanquished/pull/2190.
```diff
diff --git a/bots/default_aliens.bt b/bots/default_aliens.bt
index eb3e6d57..2475db22 100644
--- a/bots/default_aliens.bt
+++ b/bots/default_aliens.bt
@@ -1,8 +1,14 @@
 selector
 {
-       condition ( aliveTime > 1500 && healScore < 0.5 )
+       decorator return( STATUS_FAILURE )
        {
-               action evolve
+               sequence
+               {
+                       action debugColor(1, 0, 0) // red
+                       condition canEvolveTo( PCL_ALIEN_LEVEL2 ) {
+                               action debugColor(0, 1, 0) // green
+                       }
+               }
        }
 
        sequence
```

### Shortcomings

The client-server synchronization of the dot color is done using team status, which means that 1. it only works for the team, 2. it only updates every half-second by default, and 3. it is arguably hacky.

I've you've got a better suggestion as to how sync this, I'll take it.


slipher:
What about using the debug draw API like navedit? This depends on it being a local game run by the person debugging. But I'd value the ability to see it on the opposing team more than being able to use it on the network. Also I don't like how the feature uses bandwidth all the time even when it's not being used.

necessarily-equal:
Ideally we could edit entity fields and use the optimised netcode that only send the changed fields 🥰️

Unfortunately that requires a daemon change so huh, I'm not sure I want to wait a whole release for the feature, and have to recompile ~300 files in the meantime if I change branch locally.

DolceTriade:
The more new features we have for a new release, the more motivated we are to make one!

slipher:
So you decided you don't like the debug draw idea?

necessarily-equal:
> So you decided you don't like the debug draw idea?

I mean, the debug draw idea is cleaner than the current implementation, but also it would be totally unnecessary if we could change entities (entityState_t) without an engine change, and just put a field there. idk

slipher:
Yeah I guess putting it entityState_t would be OK as long as you put it as the very last field in the netcode table: that way it would truly have no impact on performance when not used (the list of field changed/unchanged bits is sent up until the last "1" bit).

Issue #2271, opened 2022-10-22
Enabling navmesh lazy-loading again (Fixing the navmesh regression from #1765)


necessarily-equal:
(parent issue https://github.com/Unvanquished/Unvanquished/issues/2265) Following the integration of the navmesh generator in unvanquished, https://github.com/Unvanquished/Unvanquished/issues/1765 reverts a few things in an attempt to fix it. This PR is about fixing the bug with lazy-loading without the revert. Note some other bugs

The layout @freem sent in https://github.com/Unvanquished/Unvanquished/issues/1765, https://github.com/Unvanquished/Unvanquished/files/8103099/chasm-prison.dat.tar.gz, uncovers a few bugs.

### Alien get stuck in their own base

This one doesn't seem to be affected by the revert. Here's a screenshot from master:

![Screenshot from 2022-10-22 08-54-07](https://user-images.githubusercontent.com/59283660/197325735-0a0d66f7-cf4a-4c62-a354-a7e806127bb5.png)

This issue isn't about fixing this.

### Human bots get to tremble often in the human base

This one is also reproducible in master, and thus is also not addressed in this issue.

### Obstacle avoidance is broken

This one is a regression. The cause is that obstacles are added as the map is loaded, and yet the navmesh isn't around yet.

I've got a fix for this one, I'll make a PR soon.

Issue #2265, opened 2022-10-19
Steering issue: the navmesh generation


necessarily-equal:
So the navmesh generation code was moved into Unvanquished following https://github.com/Unvanquished/Unvanquished/pull/1646. [DaemonMap](https://github.com/DaemonEngine/daemonmap) is destined to be removed but was still used to generate the navmesh for the last release.

On the long-term scope, we should get rid of daemonmap and make the navmesh generator in Unvanquished reliable enough to be used all the time, and enable it by default.

What's the current state of things, and what do we need to make that happen? Can we make a todo list?

ping @slipher, @DolceTriade and @illwieckz I guess, as you are involved in navmesh generation and/or release preparation.

slipher:
The big blocker is that I couldn't figure out why lazy initialization of navmeshes wasn't working (#1765).

necessarily-equal:
Ok! I should have a fix for that in https://github.com/Unvanquished/Unvanquished/pull/2272. What would be the other problem(s)?

slipher:
1. We need a better story for automatic nav generation on servers (it's already working well for local games, cg_navgenOnLoad). A bare-bones setup is possible now by adding a `navgen missing` command in the map config before any `bot fill` commands, but it needs improvement: (a) needs to be more automatic -- this requires issuing a command and won't work with the new g_bot_defaultFill cvar; and (b) causes the server to freeze for up to tens of seconds, so we might want to make it run "in the background".
2. I have a commit to add more info to the navmesh file header to ensure the cached navmesh is invalidated and regenerated if the BSP/navgen algorithm/etc. changes. I haven't merged that since it breaks compatibility with existing navmeshes. I intended this for a next-release branch but if (1) is solved well enough we might merge it to master.

DolceTriade:
> We need a better story for automatic nav generation on servers

Agree. I think we should come up with an end goal too. For instance, is automatic nav generation meant to replace navmeshes shipped with dpks? Is it only meant as a fallback for when navmeshes are missing. Further, some maps might require different parameters to build optimal navmeshes: will we have a mechanism to allow tuning of this?

necessarily-equal:
> Further, some maps might require different parameters to build optimal navmeshes

[...]

PR #2272, opened 2022-10-22
Restore navmesh lazy-loading and fix it

necessarily-equal:
This creates a little freeze when the navmesh is loaded. I'm not 100% sure this is the best UX, but at least according to https://github.com/Unvanquished/Unvanquished/issues/2265#issuecomment-1284616524 that's the next step for navmesh generation in the gamelogic.

This hopefully fixes every bug that isn't already in master. See https://github.com/Unvanquished/Unvanquished/issues/2271 for the investigation.

### DIFF
--- a/src/sgame/botlib/bot_nav.cpp
+++ b/src/sgame/botlib/bot_nav.cpp
@@ -386,11 +386,25 @@ bool G_BotNavTrace( int botClientNum, botTrace_t *trace, const glm::vec3& start_
 	return true;
 }
 
-void G_BotAddObstacle( const glm::vec3 &mins, const glm::vec3 &maxs, qhandle_t *obstacleHandle )
+struct bbox_t {
+	glm::vec3 mins;
+	glm::vec3 maxs;
+};
+static std::map<int, bbox_t> savedObstacles;
+static std::map<int, qhandle_t> obstacleHandles; // handles of detour's obstacles, if any
+
+extern bool navMeshLoaded;
+void G_BotAddObstacle( const glm::vec3 &mins, const glm::vec3 &maxs, int obstacleNum )
 {
 	qVec min = &mins[0];
 	qVec max = &maxs[0];
 	rBounds box( min, max );
+
+	if ( !navMeshLoaded )
+	{
+		savedObstacles[obstacleNum] = { mins, maxs };
+		return;
+	}
 
 	for ( int i = 0; i < numNavData; i++ )
 	{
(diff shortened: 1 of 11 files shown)


src/sgame/botlib/bot_nav.cpp
```diff
@@ -408,25 +422,46 @@ void G_BotAddObstacle( const glm::vec3 &mins, const glm::vec3 &maxs, qhandle_t *
...
+
+void G_BotRemoveObstacle( int obstacleNum )
+{
+	auto obstacle = savedObstacles.find(obstacleNum);
+	if (obstacle != savedObstacles.end()) {
+		savedObstacles.erase(obstacle);
+	}
+
+	auto handle = obstacleHandles.find(obstacleNum);
+	if (handle != obstacleHandles.end()) {
```
slipher, line 455: I guess this should also be erased

slipher (commented):
There's another part to lazy init revert: bbe18d3 so you'll want that as well.

src/sgame/botlib/bot_nav.cpp
necessarily-equal: Fixed

necessarily-equal:
I added the revert to that other commit

slipher:
LGTM

I think the revert ones should be squashed

src/sgame/botlib/bot_nav.cpp
```diff
@@ -386,12 +386,26 @@ bool G_BotNavTrace( int botClientNum, botTrace_t *trace, const glm::vec3& start_
 	return true;
 }
 
-void G_BotAddObstacle( const glm::vec3 &mins, const glm::vec3 &maxs, qhandle_t *obstacleHandle )
+struct bbox_t {
+	glm::vec3 mins;
+	glm::vec3 maxs;
+};
+static std::map<int, bbox_t> savedObstacles;
```
DolceTriade, line 393: Can these be unordered map? I don't see any reason they need to be sorted. If so, please leave a comment on why

src/sgame/botlib/bot_nav.cpp
```diff
@@ -408,25 +422,47 @@ void G_BotAddObstacle( const glm::vec3 &mins, const glm::vec3 &maxs, qhandle_t *
...
 
-void G_BotRemoveObstacle( qhandle_t obstacleHandle )
+// We do lazy load navmesh when bots are added. The downside is that this means
+// map entities are loaded before the navmesh are. This workaround does keep
+// those obstacle (such as doors and buildables) in mind until when navmesh is
+// finally loaded, or generated.
+void BotAddSavedObstacles()
 {
-	for ( int i = 0; i < numNavData; i++ )
+	for ( std::pair<int, bbox_t> obstacle : savedObstacles )
```
DolceTriade, line 440: nit const std::pair<int, bbox_t>& or const auto&

src/sgame/botlib/bot_nav.cpp
```diff
@@ -408,25 +422,47 @@ void G_BotAddObstacle( const glm::vec3 &mins, const glm::vec3 &maxs, qhandle_t *
...
-		NavData_t *nav = &BotNavData[ i ];
-		if ( nav->cache->getObstacleCount() <= 0 )
+		G_BotAddObstacle(obstacle.second.mins, obstacle.second.maxs, obstacle.first);
+	}
+	savedObstacles.clear();
+}
+
+void G_BotRemoveObstacle( int obstacleNum )
+{
+	auto obstacle = savedObstacles.find(obstacleNum);
```
DolceTriade, line 449: is it safe to just run savedObstracle.erase(obstacleNum)?

src/sgame/sg_bot_nav.cpp
```diff
@@ -70,6 +70,8 @@ bool G_BotNavInit()
 		}
 	}
 	navMeshLoaded = true;
+	extern void BotAddSavedObstacles();
```
DolceTriade, line 73: This is hacky. Let's properly add things to header files if they are going to be called elsewhere.

src/sgame/sg_struct.h
```diff
@@ -452,9 +452,6 @@ struct gentity_t
 
 	bool    pointAgainstWorld; // don't use the bbox for map collisions
 
-	// handle of obstacle that was added to prevent bots to
-	// move toward it. Used for movers and big enough buildings.
-	qhandle_t   obstacleHandle;
```
DolceTriade, line 457: Please comment in your commit message your rationale for removing this.

src/sgame/botlib/bot_nav.cpp
necessarily-equal: hum, actually it is, only the iterator one has a « This shall point to a valid and dereferenceable element. » condition. But on the other hand, this makes it consistent with the one below so maybe it's better

src/sgame/sg_bot_nav.cpp
necessarily-equal: It's kind of a function that is only going to be called in one place, ever so that's why it's here. I guess I can move it before this function instead of inside it though.

src/sgame/botlib/bot_nav.cpp
necessarily-equal: IIRC `std::pair<int, bbox_t>&` doesn't compile, somehow. This function is called once per game anyway so it doesn't really matter

src/sgame/botlib/bot_nav.cpp
necessarily-equal: Performance-wise it shouldn't matter because there are like 10 obstacles in a normal game map and absolute worst case there would be 958. I guess this brings a bit more determinism or something?

src/sgame/sg_bot_nav.cpp
necessarily-equal: (done)

src/sgame/sg_struct.h
necessarily-equal: Done

DolceTriade:
lgtm

PR #2273, opened 2022-10-23
Add a bit of doc

necessarily-equal:


src/sgame/sg_buildable.cpp
```diff
@@ -33,7 +33,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 static Cvar::Cvar<bool> g_indestructibleBuildables(
 		"g_indestructibleBuildables",
-		"make buildables impossible to destroy", Cvar::NONE, false);
+		"make buildables impossible to destroy (Note: applies only to buildings that will be built", Cvar::NONE, false);
```
DolceTriade, line 36: What does this mean? Also missing an ending )

necessarily-equal: I've made a longer explaination. It's a big long to my taste, but hopefully clear enough.

DolceTriade (approved):

PR #2274, opened 2022-10-24
Entity and client getter functions

necessarily-equal:
So I've grown a bit annoyed at entityNum being so impractical to get. This provides a way of doing that, and also drops the worst old usages.

### Motivation

`client->ent()` is much more practical than `g_entities[client->ps.clientNum]` and `ent->num()` much more readable than `ent - g_entities`.

It also can avoid stupid bugs. For example, getting the entityNum using `ent->client->ps.clientNum` does work only for clients or trying to use `(gentity_t*)->s.clientNum` which has nothing to do with clients, or nums.

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -2365,15 +2365,15 @@ static bool FindRoomForClassChangeVertically(
 	// before starting the real trace
 	VectorCopy( newOrigin, temp );
 	temp[ 2 ] += nudgeHeight;
-	trap_Trace( &tr, newOrigin, toMins, toMaxs, temp, ent->s.number, MASK_PLAYERSOLID, 0 );
+	trap_Trace( &tr, newOrigin, toMins, toMaxs, temp, ent->num(), MASK_PLAYERSOLID, 0 );
 	temp[ 2 ] = newOrigin[2] + nudgeHeight * tr.fraction;
 
 	// trace down to the ground so that we can evolve on slopes
-	trap_Trace( &tr, temp, toMins, toMaxs, newOrigin, ent->s.number, MASK_PLAYERSOLID, 0 );
+	trap_Trace( &tr, temp, toMins, toMaxs, newOrigin, ent->num(), MASK_PLAYERSOLID, 0 );
 	VectorCopy( tr.endpos, newOrigin );
 
 	// make REALLY sure
-	trap_Trace( &tr, newOrigin, toMins, toMaxs, newOrigin, ent->s.number, MASK_PLAYERSOLID, 0 );
+	trap_Trace( &tr, newOrigin, toMins, toMaxs, newOrigin, ent->num(), MASK_PLAYERSOLID, 0 );
 	return !tr.startsolid && tr.fraction == 1.0f;
 }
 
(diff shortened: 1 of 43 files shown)


src/sgame/sg_cmds.cpp
```diff
@@ -1644,7 +1644,8 @@ static void Cmd_CallVote_f( gentity_t *ent )
 	char   name[ MAX_NAME_LENGTH ] = "";
 	char   caller[ MAX_NAME_LENGTH ] = "";
 	char   reason[ MAX_TOKEN_CHARS ];
-	int    clientNum = -1;
+	int    targetNum = -1;
+	gclient_t *target = nullptr;
```
slipher, line 1648: Better to declare the variables later at the same place they are initialized

src/sgame/sg_utils.cpp
```diff
@@ -781,11 +781,11 @@ void G_TeamToClientmask( team_t team, int *loMask, int *hiMask )
 
 	*loMask = *hiMask = 0;
 
-	for ( clientNum = 0; clientNum < MAX_CLIENTS; clientNum++ )
+	for ( clientNum = 0; clientNum < level.maxclients; clientNum++ )
 	{
-		client = g_entities[ clientNum ].client;
```
slipher, line 786: This looks wrong. You have to check whether the client entity is in use.

src/sgame/sg_main.cpp
```diff
@@ -1556,15 +1555,9 @@ static void GetAverageCredits( int teamCredits[], int teamValue[] )
 		teamValue[ team ] = 0;
 	}
 
-	for ( playerNum = 0; playerNum < MAX_CLIENTS; playerNum++ )
+	for ( playerNum = 0; playerNum < level.maxclients; playerNum++ )
 	{
-		playerEnt = &g_entities[ playerNum ];
-		client = playerEnt->client;
-
-		if ( !client )
```
slipher, line 1564: Needs the check for whether the client exists

slipher (commented):
I'm usually not a fan of member functions on the god structs but it's like something that "should be a member" so I suppose it's OK

src/sgame/sg_cmds.cpp
necessarily-equal: Unfortunately it's used much later in the function inside a if, and used after the if without a check because this is spaghetti code. I'm afraid there is no choice here.

src/sgame/sg_main.cpp
necessarily-equal: The client always exists as as long as clientNum < level.maxclients. The initalisation code does that, and even `ClientDisconnect` sets the value again after a memset: `ent->client = level.clients + clientNum;`

slipher: So it's a pre-existing bug that the test didn't work. It's probably harmless to the calculation's correctness, but better not to read out data from nonexistent clients.

src/sgame/sg_utils.cpp
necessarily-equal: I guess it could be checked, but on the other hand I don't think this is a problem: this is a mark used by daemon to know if it should send entity infos to client, if a client doesn't exist, it doesn't matter if it is allowed or not to access an entity

slipher: Still the function ought to work as advertised and not return wrong results.

src/sgame/sg_main.cpp
necessarily-equal: Seems like it wasn't harmless to the calculation correctness (avegare=sum/number), but the value was also never read. Added a commit for that.

src/sgame/sg_utils.cpp
necessarily-equal: I've added a commit for that too

slipher:
LGTM

necessarily-equal:
I moved the changes to `Cmds_spawn_f` to a last commit.

Issue #2125, opened 2022-08-30
prevent players to use the `[bot]` icon at start of their name


bmorel:
This is a really minor detail, hence the milestone, and I have to say I never seen anyone use that to really troll people.

Bots have, by default, a "chip" icon at the beginning of their name. This is only for automatic names, but is usually followed by admins AFAICT, but nothing prevents a player to do the same, adding to the confusion.
I wonder if it would be ok to prevent players to use the conventions that signal bots.

Honestly, this really is a minor thing, and it could be closed as wontfix, no problem.

sweet235:
> I have to say I never seen anyone use that to really troll people
I have.

DolceTriade:
Seems like a reasonable idea configurable with a cvar.

illwieckz:
We would have to check the name prefix after removing color codes to catch all combinations.

DolceTriade:
Do emojis work with color codes?

necessarily-equal:
So what I'm doing in https://github.com/Unvanquished/Unvanquished/pull/2276 is to forbid the computer chip entirely in players' names. So I guess it resists to color codes, but also is more restrictive.

If somebody else does want to try implementing something, be wary that [bOT] is also a valid way of writting the emoticon.

illwieckz:
> If somebody else does want to try implementing something, be wary that [bOT] is also a valid way of writting the emoticon.

This is probably an undocumented feature you should not rely on. Assume it's because it's implemented like `with x in [x] do /emoticons/x` and it happens that the pkzip implementation of the VFS is case insensitive.

Something working not means something is valid, if that breaks, that would not be a bug.

necessarily-equal:
> This is probably an undocumented feature you should not rely on

Okok, but I don't think a griever is afraid of relying on an undocumented feature. Point is, we want to protect about `[bOT]` and variations too, because they work.

illwieckz:
Yes, I assume a complete way to implement it is to get the non-color-coded variant of the string, lowercase it, and check for `^\[bot\]`.

illwieckz:
well, maybe check for leading spaces too.

necessarily-equal:
The problem is that "checking for leading $x" is not possible because 
we are using unicode. There are many non-printable that are intended, 
or that can used in a way they won't be rendered.

slipher:
It might make sense to implement this by integrating with the `g_emoticonsAllowedInNames` code. IIRC it works by overwriting one of the square brackets.

PR #2276, opened 2022-10-29
No bot impersonation in names

necessarily-equal:
Closes https://github.com/Unvanquished/Unvanquished/issues/2125.

This also includes a commit to use gentity_t->num() more often. (more than half the places changed weren't touched earlier because they were changed in https://github.com/Unvanquished/Unvanquished/pull/2272)


necessarily-equal:
If we're going to disallow `[bot]` in names, maybe it would be good to have some gui feedback on that. I am not bothered enough to do it to be honest.

DolceTriade:
At the very least, we need a cvar to enable/disable this behavior. Also, do you think it is worth it to extend this to a set of regexs as Sweet suggests?

src/sgame/sg_client.cpp
```diff
@@ -750,6 +750,22 @@ static void G_ClientCleanName( const char *in, char *out, size_t outSize, gclien
 		invalid = true;
 	}
 
+	if ( !(client->ent()->r.svFlags & SVF_BOT) )
+	{
+		std::size_t index = 0;
```
slipher, line 755: Q_stristr is what you want

sweet235:
> At the very least, we need a cvar to enable/disable this behavior. Also, do you think it is worth it to extend this to a set of regexs as Sweet suggests?

For the regexp thing, we would have to carefully choose a regexp variant and stick to it in the future. I'm not sure the regexp thing is a good idea, a better way of handling this might be to give each client explicit information about who is a bot.

src/sgame/sg_client.cpp
necessarily-equal: Ok, used that

DolceTriade:
lgtm

src/sgame/sg_client.cpp
```diff
@@ -750,6 +750,18 @@ static void G_ClientCleanName( const char *in, char *out, size_t outSize, gclien
 		invalid = true;
 	}
 
+	if ( !(client->ent()->r.svFlags & SVF_BOT) )
+	{
+		const char *find = "[bot]";
+		const char *found;
+		const char *out_str;
```
slipher, line 757: Doesn't follow variable naming convention

necessarily-equal: I mean it follows the existing convention of  the  function (`out_string`)

slipher:
I just noticed you can chain together different forbidden constructs to get an illegal name... e.g. `[bo//t]`

PR #2280, opened 2022-10-30
Fix rotating door obstacles

necessarily-equal:
I've tested this with the tremship map. I've also added some debug draw to visualise it, but it's not part of this PR.

Here's tremship's main door before it opens:

Screenshot from 2022-10-30 02-02-18

Here's tremship's main door after it opens, without this patch:

Screenshot from 2022-10-30 02-03-32

What you can see on this picture is two huge cubes.

And here's the main door after it opens, with this patch:

Screenshot from 2022-11-06 14-44-51

Bots can now pass through

### DIFF
--- a/src/sgame/sg_cm_world.cpp
+++ b/src/sgame/sg_cm_world.cpp
@@ -444,16 +444,16 @@ void G_CM_LinkEntity( gentity_t *gEnt )
 	// set the abs box
 	if ( gEnt->r.bmodel && ( angles[ 0 ] || angles[ 1 ] || angles[ 2 ] ) )
 	{
-		// expand for rotation
-		float max;
+		glm::mat3 matrix = RotationMatrix( VEC2GLM( angles ) );
 
-		max = RadiusFromBounds( gEnt->r.mins, gEnt->r.maxs );
+		glm::vec3 mins = matrix * VEC2GLM( gEnt->r.mins );
+		glm::vec3 maxs = matrix * VEC2GLM( gEnt->r.maxs );
 
-		for ( int i = 0; i < 3; i++ )
-		{
-			gEnt->r.absmin[ i ] = origin[ i ] - max;
-			gEnt->r.absmax[ i ] = origin[ i ] + max;
-		}
+		glm::vec3 absmin = glm::min(mins, maxs) + VEC2GLM(origin);
+		glm::vec3 absmax = glm::max(mins, maxs) + VEC2GLM(origin);
+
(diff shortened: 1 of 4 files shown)


necessarily-equal:
Here's another test when modifying the spacetracks map to make the doors manual (the door isn't manual and isn't a problem, but what if? — this is to test xy rotations in addition to z rotations):

Door closed:

Screenshot from 2022-10-30 05-33-51

Door open, without this patch:

Screenshot from 2022-10-30 05-33-51

Door open, with this patch:

Screenshot from 2022-10-30 05-22-42

necessarily-equal:
Note that r.absmin and r.absmax are also used to know when precise collision detection is needed. I don't think this should break anything because the calculation is correct, just more precise. If anything, this should bring a slight performance boost.

cu-kai:
More rotating door obstacles on these maps:
`groundsource`
`deathrun`
`kaiji-starsidehotel`
`station12-b5`
`nintendo-land`

sweet235:
I run this on map testing server now, looks good to me. Let's wait a few days and see what we observe.

src/sgame/sg_cm_world.cpp
```diff
@@ -444,16 +460,22 @@ void G_CM_LinkEntity( gentity_t *gEnt )
 	// set the abs box
 	if ( gEnt->r.bmodel && ( angles[ 0 ] || angles[ 1 ] || angles[ 2 ] ) )
 	{
-		// expand for rotation
-		float max;
+		vec3_t matrix[3];
```
slipher, line 463: So the idea is that it used to be OK for G_CM_LinkEntity to use any upper bound of the actual boundaries, but since stuff is being used for bots it needs to be as precise a bound as we can get?

necessarily-equal: Precisely

necessarily-equal:
I'm not 100% sure this doesn't introduce a regression for collision handling while it opens and once it is opened as I've seen bugs, but I think I've seen the same problems on master.

I'd appreciate someone else testing to be sure

necessarily-equal:
I've removed the refactor that touched movers and managed to trigger the rotating mover problem elsewhere. I think I stumbled onto an already existing bug

sweet235:
I did'nt notice any new problems during the last two days.

necessarily-equal:
(The debug draw lives here: https://github.com/Unvanquished/Unvanquished/tree/bot-obstacles-draw)

slipher:
LGTM

PR #2281, opened 2022-10-30
Correct bot aim prediction

necessarily-equal:
This was broken, as apos.trDelta means changes in "angular position". the calculation should use pos.trDelta instead to predict where the target will be.


DolceTriade (approved):
don't fully understand but lgtm

slipher:
I agree that is probably what was intended. But does it really make them aim better?

necessarily-equal:
It makes them aim much better if you go straight. You can kind of trick them to aim wrong if they use a luci, but you could do that already. In my opinion it makes them extremely deadly with pulse blast. (Only a limited set of weapons actually benefit from that)

necessarily-equal:
In my opinion, it can actually make them too deadly, but thankfullty we can make that optional soon :)

slipher:
LGTM

sweet235:
Map testing server uses this now.

PR #2282, opened 2022-11-01
fixes alien players being GIVEN the cost of their requested /class

cu-kai:
to be merged as a matter of urgency as this is gameplay breaking

please & thank you


cu-kai:
specifically: this was broken by commit https://github.com/Unvanquished/Unvanquished/commit/54ea613c796cf8b2f4810cbe8c71e18a1e485b90

necessarily-equal:
Merging now because this is breaking the gameplay and the fix is obviously right (and also since I did the opposite change I can attest that it wasn't intended). Thanks!

DolceTriade:
Thanks for the fix!

cu-kai:
Not a problem!

PR #2284, opened 2022-11-02
Also don't allow opening the beacon menu when following someone alive

necessarily-equal:
This completes https://github.com/Unvanquished/Unvanquished/pull/2158


necessarily-equal:
Steps to reproduce the bug:
* add bots
* follow a bot
* open the beacon menu
* notice that you're not allowed to click on buttons anyway

DolceTriade (approved):

Issue #2247, opened 2022-10-08
Granger spit fire extinguisher gives free and infinite evos


cu-kai:
If you wallclimb as adv granger and spit on whatever you're clinging to, you receive a fractional evo as if you extinguished some fire (you also hear the "ding" confirming payment).

Setting `g_firefightReward` to `0` mitigates this but grangers will no longer be rewarded for extinguishing fires.

bmorel:
Here is the problem: https://github.com/Unvanquished/Unvanquished/blob/master/src/shared/bg_public.h#L462-L468

The proper solution would be to stop stupidly overlapping bits, which is the fix I'll apply on my instances, but this would be rejected as it would break compat between cgame and sgame (which is a ridiculous requirement imo).

slipher:
> The proper solution would be to stop stupidly overlapping bits, which is the fix I'll apply on my instances, but this would be rejected as it would break compat between cgame and sgame (which is a ridiculous requirement imo).

It's OK to break compat between cgame and sgame. We changed the rule.

necessarily-equal:
Heh, that's hillarious. Great find. I'm ok with a PR for both options really.

DolceTriade:
granger$ be rich

PR #2285, opened 2022-11-02
betterai-backports / Fix "Granger spit fire extinguisher gives free and infinite evos #2247"

necessarily-equal:
This a backport from bmorel's work.

Fix https://github.com/Unvanquished/Unvanquished/issues/2247.

### DIFF
--- a/src/shared/bg_public.h
+++ b/src/shared/bg_public.h
@@ -445,7 +445,8 @@ enum persEnum_t
 #define PS_SPRINTTOGGLE       BIT(3)
 
 // entityState_t->eFlags
-// notice that some flags are overlapped, so their meaning depends on context
+// PLEASE DO NOT MAKE THEM OVERLAP FOR NOTHING:
+// this will only be a pain for future changes
 #define EF_DEAD             BIT(0) // don't draw a foe marker over players with EF_DEAD
 #define EF_TELEPORT_BIT     BIT(1) // toggled every time the origin abruptly changes
 #define EF_PLAYER_EVENT     BIT(2) // only used for eType > ET_EVENTS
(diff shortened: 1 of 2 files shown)


necessarily-equal:
LGTM from me, and written by someone else so I guess that I'll merge this in a few days if there's no remarks

src/shared/bg_public.h
```diff
@@ -445,7 +445,8 @@ enum persEnum_t
 #define PS_SPRINTTOGGLE       BIT(3)
 
 // entityState_t->eFlags
-// notice that some flags are overlapped, so their meaning depends on context
+// PLEASE DO NOT MAKE THEM OVERLAP FOR NOTHING:
```
DolceTriade, line 448: What's our policy for overlapping flags? Can you document it here too?

necessarily-equal: > What's our policy for overlapping flags?

I don't think we have one, but the idea of the commit is that we don't *need* to have one so we better don't, as it avoids possible bugs

DolceTriade: Yes, but that's not scalable in the long term. As we add more features, we will end up with overlapping flags.

I was hoping we could come up with one right now and encode it as you are touching this file.

My idea for a policy is something like:

Every time you add an overlapping flag you note what it overlaps with and why this is safe.

necessarily-equal: Crafting a policy sounds good, but also I don't feel like I have a strong opinion on the matter.

DolceTriade: Then let's add this to the code if you don't care and merge it?

necessarily-equal: wfm

PR #2286, opened 2022-11-02
Various misc things from bmorel's tree

necessarily-equal:


src/sgame/sg_bot_local.h
```diff
@@ -131,6 +131,7 @@ struct botMemory_t
 	glm::vec3 stuckPosition;
 
 	int spawnTime;
+private:
```
DolceTriade, line 134: why?

DolceTriade:
lgtm minus the privateness thing, but don't really care that much about it. Maybe @slipher has more thoughts.

src/sgame/sg_bot_local.h
necessarily-equal: it's a value that has a setter function, it's arguably not meant to be public

DolceTriade: fair. lgtm

DolceTriade (approved):

necessarily-equal:
I've dropped the "BotDirectionToUsercmd minor cleanup" commit because it is was improved in https://github.com/Unvanquished/Unvanquished/pull/2291

slipher:
My position is that adding privateness in huge structs cannot help in the long run. If you try to make more fields private, then this just results in more and more member functions on the struct, so there are a huge number of places that can access it. And given the complete lack of a defined responsibility for this struct, I doubt it will prevent someone who wants to touch those fields from just adding to the list of functions which are allowed to access it.

If you want real encapsulation then you should make a StaminaManager object or something with only the sprinting-related fields/functions.

necessarily-equal:
I'm ok with either adding the private: or removing the function if we really think encapsulation is stupid, but the current state doesn't make much sense IMO

Issue #2137, opened 2022-08-31
make bots understand the direction of their enemy


bmorel:
Currently, alien bots strafe when they chase an enemy. This slows them a lot, and makes aliens a lot weaker than they should.
 @DolceTriade  I think you might be interested by this one.

This also applies to turret attacks.

DolceTriade:
Interesting. Bots don't look at their enemy when they move?

bmorel:
No, they don't. It's why dretches are that ineficient: they strafe when their enemy is running away from them. I guess I should create more issues for the stuff I know about AIs :)

PR #2291, opened 2022-11-03
Bot strafe only if enemy is looking at us, plus refactors

necessarily-equal:
I intend to use the *2D functions for other things, like Advanced Dragoon aiming for barbs (already implemented), and possibly flames.

Closes https://github.com/Unvanquished/Unvanquished/issues/2137.

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -795,7 +795,8 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 	}
 
 	// we killed it, yay!
-	if ( !BotEntityIsValidEnemyTarget( self, mind->goal.getTargetedEntity() ) )
+	if ( !mind->goal.targetsValidEntity()
+		|| !BotEntityIsValidEnemyTarget( self, mind->goal.getTargetedEntity() ) )
 	{
 		return STATUS_SUCCESS;
 	}
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -2325,7 +2325,10 @@ botTarget_t& botTarget_t::operator=(const gentity_t *newTarget) {
 
 	if (!targetsValidEntity())
 	{
-		Log::Warn("bot: selecting invalid entity as target");
+		Log::Warn( "bot: selecting invalid entity as target, %s",
+			!newTarget->inuse ? "entity isn't allocated" :
+			!Entities::IsAlive(newTarget) ? "entity is dead" :
+				"for some unspecified reason" );
 	}
 
 	return *this;
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -2377,10 +2377,7 @@ bool botTarget_t::targetsCoordinates() const
 
 bool botTarget_t::targetsValidEntity() const
 {
-	if (ent) {
-		return BotEntityIsValidTarget(ent.get());
-	}
-	return false;
+	return BotEntityIsValidTarget(ent.get());
 }
 
 const gentity_t *botTarget_t::getTargetedEntity() const
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1428,7 +1428,7 @@ void BotAimAtEnemy( gentity_t *self )
 	}
 
 	glm::vec3 viewOrigin = BG_GetClientViewOrigin( &self->client->ps );
-	glm::vec3 desired = VEC2GLM( self->botMind->futureAim ) - viewOrigin;
+	glm::vec3 desired = self->botMind->futureAim - viewOrigin;
 	desired = glm::normalize( desired );
 	glm::vec3 current;
 	AngleVectors( VEC2GLM( self->client->ps.viewangles ), &current, nullptr, nullptr );
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1675,33 +1675,45 @@ void BotFireWeapon( weaponMode_t mode, usercmd_t *botCmdBuffer )
 }
 void BotClassMovement( gentity_t *self, bool inAttackRange )
 {
-	usercmd_t *botCmdBuffer = &self->botMind->cmdBuffer;
+	botMemory_t *mind = self->botMind;
+	usercmd_t *botCmdBuffer = &mind->cmdBuffer;
+	bool botIsSmall = false;
+	bool botIsJumper = false;
+
+	// If we are targetting a buildable or we are targetting a human who
+	// isn't looking at us (likely fleeing), so we try to reach the human
+	// as fast as we can.
+	// Making special tricks for buildables is still useful because turrets
+	// are slow to aim and defenders do more team-damage this way.
+	bool shouldStrafe = true;
+	if ( mind->goal.getTargetType() == entityType_t::ET_PLAYER )
+	{
+		glm::vec3 forward1, forward2;
+		AngleVectors( VEC2GLM( self->client->ps.viewangles ), &forward1, nullptr, nullptr );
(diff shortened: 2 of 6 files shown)


slipher:
We have a policy against adding unused code. New functions should be added in the same PR as the code using them

necessarily-equal:
> We have a policy against adding unused code. New functions should be added in the same PR as the code using them

Fixed

src/sgame/sg_bot_ai.cpp
```diff
@@ -795,8 +795,8 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 	}
 
 	// we killed it, yay!
-	if ( !BotEntityIsValidEnemyTarget( self,
-			self->botMind->goal.getTargetedEntity() ) )
+	if ( !mind->goal.targetsValidEntity()
+		|| !BotEntityIsValidEnemyTarget( self, mind->goal.getTargetedEntity() ) )
```
slipher, line 799: BotEntityIsValidEnemyTarget already does a null check

src/sgame/sg_bot_util.cpp
```diff
@@ -2312,7 +2312,11 @@ botTarget_t& botTarget_t::operator=(const gentity_t *newTarget) {
 
 	if (!targetsValidEntity())
 	{
-		Log::Warn("bot: selecting invalid entity as target");
+		Log::Warn("bot: selecting invalid entity as target, %s",
+			!newTarget->inuse ? "entity isn't allocated" :
+			!Entities::IsAlive(newTarget) ? "entity is dead" :
+				"for some reason" );
+		ASSERT_UNREACHABLE(); // this will be a bug in the calling code, let's warn about it
```
slipher, line 2319: This actually can make the program crash in release mode, probably not what you want

necessarily-equal: Is ASSERT(0) different in crash behaviour?

src/sgame/sg_bot_util.cpp
```diff
@@ -1714,6 +1714,16 @@ void BotClassMovement( gentity_t *self, bool inAttackRange )
 		default:
 			break;
 	}
+
+	if ( shouldDodge && botIsSmall )
+	{
+		BotStrafeDodge( self );
+	}
+
+	if ( shouldDodge && botIsJumper && self->client->time1000 % 300 == 0 && self->botMind->nav().directPathToGoal )
```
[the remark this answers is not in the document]
necessarily-equal: This is existing code though

src/sgame/sg_bot_ai.cpp
necessarily-equal: It doesn't check if the target is stale though, notably if ent.generation changed. I don't think this has a big chance of breaking because if the generation changed, the target is often dead. But we shouldn't rely on this

src/shared/bg_public.h
```diff
@@ -1820,5 +1824,45 @@ void BG_BoundingBox( buildable_t buildablel, glm::vec3* mins, glm::vec3* maxs );
 
 void AngleVectors( const glm::vec3 &angles, glm::vec3 *forward, glm::vec3 *right, glm::vec3 *up );
 
+// a scalar product that handles the z component differently, typically you
+// would use it with z_factor = 0, to ignore the z direction totally
```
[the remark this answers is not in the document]
necessarily-equal: I'd like to use that with non-zero for aliensense range, bot distance evaluation and/or more jumpable flames

src/sgame/sg_bot_util.cpp
slipher: Looks like it is actually different with GCC: ASSERT(0) preprocesses to nothing while ASSERT_UNREACHABLE makes the behavior undefined there (the compiler doesn't ensure that it will jump to the logically next part of the code). 

The main thing is it doesn't make sense to have both a warning and an assert. You should pick one or the other.

src/sgame/sg_bot_ai.cpp
slipher: GentityRef `get()` returns null if the generation changed.

necessarily-equal: Fair enough, I think that's why I didn't check it when I introduced `targetsValidEntity()`. I can remove the commit if you want, but maybe this is better like this.

src/sgame/sg_bot_util.cpp
necessarily-equal: Well the warning and the assert have different purpose: the warning makes it explicit what failed, and the assert makes it possible to have a backtrace.

necessarily-equal: I've changed it to `ASSERT(0);`

src/sgame/sg_bot_util.cpp
```diff
@@ -1674,33 +1674,33 @@ void BotFireWeapon( weaponMode_t mode, usercmd_t *botCmdBuffer )
 }
 void BotClassMovement( gentity_t *self, bool inAttackRange )
 {
-	usercmd_t *botCmdBuffer = &self->botMind->cmdBuffer;
+	botMemory_t *mind = self->botMind;
+	usercmd_t *botCmdBuffer = &mind->cmdBuffer;
+	bool shouldDodge = mind->goal.targetsValidEntity() && Alignment2D( VEC2GLM(self->s.angles), VEC2GLM(mind->goal.getTargetedEntity()->s.angles) ) < -0.5f;
```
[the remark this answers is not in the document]
necessarily-equal: Ok, good point. I've updated that. I've also made it so that the bot doesn't stop jumping anymore either.

src/sgame/sg_bot_util.cpp
slipher: Assertions should only be used for a fatally bad bug like indexing out of bounds. Making it an assert forces the developer to do something about this bot bug even if he is trying to work on something else. If you get a warning and want to debug it, you can put a breakpoint on the log statement for the next time it happens.

necessarily-equal: Ok, fair enough. I've removed the assertion.

src/sgame/sg_bot_util.cpp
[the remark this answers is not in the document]
necessarily-equal: Also note that the bot skilltree branch changes that line too.

src/sgame/sg_bot_util.cpp
```diff
@@ -1675,33 +1675,38 @@ void BotFireWeapon( weaponMode_t mode, usercmd_t *botCmdBuffer )
...
+	usercmd_t *botCmdBuffer = &mind->cmdBuffer;
+	bool botIsSmall = false;
+	bool botIsJumper = false;
+
+	// If we are targetting a buildable or we are targetting a human who
+	// isn't looking at us (likely fleeing), so we try to reach the human
+	// as fast as we can.
+	// Making special tricks for buildables is still useful because turrets
+	// are slow to aim and defenders do more team-damage this way.
+	bool shouldStrafe = mind->goal.getTargetType() != entityType_t::ET_PLAYER
```
slipher, line 1688: The meaning of this variable seems to be should approach directly i.e. *shouldn't* strafe?

Also it seems wrong to use Alignment2D which is made for direction vectors on `angles`.

necessarily-equal: I mean ultimately, what is currently named Alignment2D is just that, an angle. Except it's actually the cosine of an angle. If you want I can rename it Angle2D, putting a warning that you should use acosf(Angle2D(…)) if you actually want the angle

slipher: No, `angles` is like Euler angles, not a vector. Comparing it with a dot product is not even approximately correct since you have 0 degrees vs. 359 degrees etc.

necessarily-equal: Well, here you have an angle from 0 to 180°. But ok for keeping the name.

The Alignment2D lines are equivalent if that's what you were afraid of.

About the strafe thing. Do you mean that the condition is inverted (wrong) or that it would be more readable inverted?

slipher: LGTM

Issue #2293, opened 2022-11-03
Bot could go heal on medistations below ground


necessarily-equal:
![image](https://user-images.githubusercontent.com/59283660/199691193-449a41a9-aa0a-48cb-9078-f023edc7db21.png)

Currently bots will not be able to use it because they think they can't walk to it. Fixing this seems not so hard and fairly hard at the same time.

necessarily-equal:
![image](https://user-images.githubusercontent.com/59283660/199699628-79b9e194-1bdf-4dc8-84b0-65da1ec0b708.png)

Looks like I have a fix

bmorel:
I disagree this would be an improvement.

bmorel:
see my last comment on #2294: try placing a navcon there, it should work. This is hackish, but so is building working through walls anyway.

PR #2295, opened 2022-11-03
Make underground medistat reachable for human bots

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/2293

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1058,8 +1058,7 @@ void BotTargetToRouteTarget( const gentity_t *self, botTarget_t target, botRoute
 
 	// account for building on wall or ceiling
 	if ( entType == entityType_t::ET_PLAYER ||
-			( entType == entityType_t::ET_BUILDABLE
-			  && target.getTargetedEntity()->s.origin2[ 2 ] < MIN_WALK_NORMAL ) )
+		entType == entityType_t::ET_BUILDABLE )
 	{
 		trace_t trace;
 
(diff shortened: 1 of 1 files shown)


slipher:
This might break other cases since you don't always have to stand on top of a medistation to use it. If there is an elevated medistation, you can heal by standing against the side of it.

I don't think we should optimize for impossible glitch builds like the one in the operation-dretch layout...

necessarily-equal:
I don't think I got what you mean. Do you mean a slanted medistation?

slipher:
No. Suppose there's a medistation at a human's chest height. He can stand against its edge and heal.

slipher:
I managed to find a spot where the bots will use the medistation from the side, and it still worked with this change.

However, I found that this change can cause a bot to get stuck:
unvanquished_2022-11-05_132400_000

necessarily-equal:
Ok, that likely means tweaking of the height is needed. I'm actually surprised you could build a medistation here because the roof height is really low. What's the map?

slipher:
It's UTCS. And why not? The height is only 6 units.

```
mins        -35 -35 0
maxs        35 35 6
```

necessarily-equal:
The reason I was surprised this is possible is that it is simply impossible to build a medistation under the stairs in chasm. Here's an example in 3rd person view: (the problem also exists in first person view)

image

necessarily-equal:
So I've tried for a bit, and I still can't reproduce your bug where they go above the platform. Anyhow, if they tried to go below the platform it would also be broken. I think in the worst case this isn't a regression because cases that are already broken will still be broken. It improves other cases though.

image

I've pushed a new version that uses the medistation normal as "up", I think it is slightly better for sloped medistations.

slipher:
The normal doesn't affect the medistation's bounding box or mechanics, so I'm not sure it makes sense to use it.

I took a look at the medistation code and and the max healing height is 32 units above the center of the bbox, so it would make sense to use this instead of origin + 50. 

I propose doing a zero-length trace against the elevated goal point, and if it intersects something solid then falling back to the standard goal point. I think with that I would be happy that it won't introduce too many bugs.

bmorel:
The underground medistation thing is a hack, nothing less, nothing more. I do not think bots should support it: they're fragile enough, no need to add more complexity there.

A valid feature request here would be to have bots go heal in healing areas, which is something they can not do. Then, mappers could replace that underground medistation hack with something cleaner, like a heal area reserved to human players, possibly using a medistat model to make it clear for players it's a heal area.

PR #2297, opened 2022-11-03
Fix obstacle handling

necessarily-equal:
The two errors here are the `[i]` instead of `[obstacleNum]` and the `ent->s.clientNum`.

There are a few problems that I encountered that made this especially annoying to track down.
* that the types lack any kind of type safety and are converted implicitely (`typeof(entityNum) = int = qhandle_t ≃ dtObstacleRef = unsigned int`)
* that the detour removeObstacle function returns `DT_SUCCESS` when the obstacle doesn't exist
* that ent->s.clientNum returns garbage when the entity isn't a client

### DIFF
--- a/src/sgame/botlib/bot_nav.cpp
+++ b/src/sgame/botlib/bot_nav.cpp
@@ -391,7 +391,7 @@ struct bbox_t {
 	glm::vec3 maxs;
 };
 static std::map<int, bbox_t> savedObstacles;
-static std::map<int, qhandle_t> obstacleHandles; // handles of detour's obstacles, if any
+static std::map<int, std::array<dtObstacleRef, MAX_NAV_DATA>> obstacleHandles; // handles of detour's obstacles, if any
 
 extern bool navMeshLoaded;
 void G_BotAddObstacle( const glm::vec3 &mins, const glm::vec3 &maxs, int obstacleNum )
@@ -406,9 +406,10 @@ void G_BotAddObstacle( const glm::vec3 &mins, const glm::vec3 &maxs, int obstacl
 		return;
 	}
 
+	std::array<dtObstacleRef, MAX_NAV_DATA> handles;
+	std::fill(handles.begin(), handles.end(), (unsigned int)-1);
 	for ( int i = 0; i < numNavData; i++ )
 	{
-		dtObstacleRef ref;
 		NavData_t *nav = &BotNavData[ i ];
 
 		const dtTileCacheParams *params = nav->cache->getParams();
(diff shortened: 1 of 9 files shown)


cu-kai:
Seems to fix the issues I was having on bunker with bots not finding the enemy base on the `operation-dretch` map.

src/sgame/botlib/bot_nav.cpp
```diff
@@ -427,7 +427,7 @@ void G_BotAddObstacle( const glm::vec3 &mins, const glm::vec3 &maxs, int obstacl
 		tempBox.mins[ 1 ] -= params->walkableHeight;
 
 		nav->cache->addBoxObstacle( tempBox.mins, tempBox.maxs, &ref );
-		obstacleHandles[i] = ref;
```
slipher, line 430: Seems fishy that we are setting the same entry in the map on every loop iteration.

necessarily-equal: That's true. In principle we could add obstacles per-class and it would still be nice to handle that. Coincidentally this seems to work so idk.

slipher: So it works because if you have added the same number of obstacles to each mesh then you get the same handle? If obstacleHandles is already populated for the given key, this assumption should be checked with a warning or assertion.

necessarily-equal: I just pushed something which uses arrays. So this should no longer be a problem.

src/sgame/botlib/bot_nav.cpp
```diff
@@ -426,9 +427,9 @@ void G_BotAddObstacle( const glm::vec3 &mins, const glm::vec3 &maxs, int obstacl
 		// offset mins down by agent height so obstacles placed on ledges are handled correctly
 		tempBox.mins[ 1 ] -= params->walkableHeight;
 
-		nav->cache->addBoxObstacle( tempBox.mins, tempBox.maxs, &ref );
-		obstacleHandles[i] = ref;
+		nav->cache->addBoxObstacle( tempBox.mins, tempBox.maxs, &handles[i] );
 	}
+	obstacleHandles[obstacleNum] = handles;
```
slipher, line 432: Sorry to keep coming up with new stuff, but I think it would be good to use the map `insert` function here and show a warning if there is already a registered obstacle for the given entity since that would lead to a leak.

Or maybe you could just have BotAddObstacle call BotRemoveObstacle so that leaks can't happen.

necessarily-equal: It's not exactly pretty with .insert, but added that

slipher:
LGTM

PR #2298, opened 2022-11-04
Delete cg_customCrosshair

slipher:
I haven't tested whether the custom crosshair feature works, but at least this allows it to load a file.

The commit message for 6b5df72075e8bf359577d5b15e0ee50a16ff8acd contains the only documentation for this feature :stuck_out_tongue:

### DIFF
--- a/src/cgame/cg_weapons.cpp
+++ b/src/cgame/cg_weapons.cpp
@@ -75,120 +75,6 @@ void CG_RegisterUpgrade( int upgradeNum )
 }
 
 /*
-=======================
-CG_LoadCustomCrosshairs
-
-Load custom crosshairs specified by the user
-=======================
-*/
-
-static void CG_LoadCustomCrosshairs()
-{
-	const char         *text_p, *token;
-	weapon_t     weapon;
-
-	std::string filename = cg_crosshairFile.Get();
-	if ( filename.empty() )
-	{
-		return;
-	}
-	std::error_code err;
-	std::string text = FS::PakPath::ReadFile( filename, err );
-	if ( err )
-	{
-		Log::Warn( "couldn't read custom crosshair file '%s': %s", filename, err.message() );
-		return;
-	}
-
-	// parse the text
-	text_p = text.c_str();
-
-	while ( 1 )
-	{
-		token = COM_Parse2( &text_p );
-
-		if ( !*token )
-		{
-			break;
-		}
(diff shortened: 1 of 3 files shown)


src/cgame/cg_weapons.cpp
```diff
@@ -85,23 +85,31 @@ Load custom crosshairs specified by the user
...
+
+	if ( len < 0 )
 	{
 		return;
 	}
-	std::error_code err;
-	std::string text = FS::PakPath::ReadFile( filename, err );
-	if ( err )
+
+	if ( len == 0 || len + 1 >= (int) sizeof( text ) )
```
DolceTriade, line 100: nit: c++ cast

DolceTriade, line 100: Why not just malloc the length but put a max? I think MAX_TOKEN_CHARS might actually be enough?

slipher:
@DolceTriade I just noticed that the parser doesn't match the purported file format in https://github.com/Unvanquished/Unvanquished/commit/6b5df72075e8bf359577d5b15e0ee50a16ff8acd. It expects some keywords like `crosshairIndicator` while the example has just values with no keywords.

I don't want to spend a lot of time trying to test this or figure out how it works so maybe we should just delete it.

DolceTriade:
I mean, I'm fine deleting it. That purported file format was accurate when that commit was merged. Then https://github.com/Unvanquished/Unvanquished/commit/29c6383d94fb551f5be221972d076900f1a761f9 came along and changed it. 🤷 

Now it's

```
blaster crosshair img
blaster crosshairsize 10
blaster crosshairIndicator img
```

slipher:
Oh and even if the parser worked as promised, the idea 

> I mean, I'm fine deleting it. That purported file format was accurate when that commit was merged. Then [29c6383](https://github.com/Unvanquished/Unvanquished/commit/29c6383d94fb551f5be221972d076900f1a761f9) came along and changed it

Oh, my bad. I was looking for other commits in the git blame but didn't scroll down far enough.

Anyway I will delete it now because there is a more serious issue. You can't have a custom image because the engine only loads images from the VFS (except possibly in the case where the filename includes an extension, but that's presumably a bug).

DolceTriade:
lgtm

necessarily-equal:
I'm also fine with deleting it

Issue #2279, opened 2022-10-30
Segfault in bot BT


necessarily-equal:
I've seen this segfault. I don't know where it comes from. https://paste.gnugen.ch/paste/uu5L.

<del>I've only ever seen it once so it shouldn't be too important.</del> This is 100% reproducible now with default BT.

DolceTriade:
Memory leak? `malloc` shouldn't fail otherwise?

necessarily-equal:
`malloc_printerr` says "corrupted size vs. prev_size in fastbins". I don't think it's an out-of-memory error or it would have returned nullptr instead.

necessarily-equal:
From the calloc documentation:

> This function returns a pointer to the allocated memory, or NULL if the request fails.

This looks more serious than an out-of-memory

necessarily-equal:
I've seen it again. What's interesting is that it seems to happen with the "use_medikit" BT.

```gdb
Starting DNS resolver thread              
][New Thread 0x7ffff272b640 (LWP 34)]
Warn: Hitch: 1643 msec frame time         
]corrupted size vs. prev_size in fastbins                                                                                                                                                                         
Thread 4 "daemonded" received signal SIGABRT, Aborted.

Thread 4 (Thread 0x7ffff32dc640 (LWP 33) "daemonded"):
#0  0x00007ffff7a56c1f in __pthread_kill_implementation () from /nix/store/f6kvkdzp6qfjm6h94d0jgfvm4r06xcaq-glibc-2.34-210/lib/libc.so.6
#1  0x00007ffff7a0c042 in raise () from /nix/store/f6kvkdzp6qfjm6h94d0jgfvm4r06xcaq-glibc-2.34-210/lib/libc.so.6
#2  0x00007ffff79f749c in abort () from /nix/store/f6kvkdzp6qfjm6h94d0jgfvm4r06xcaq-glibc-2.34-210/lib/libc.so.6
#3  0x00007ffff7a4b3f8 in __libc_message () from /nix/store/f6kvkdzp6qfjm6h94d0jgfvm4r06xcaq-glibc-2.34-210/lib/libc.so.6
#4  0x00007ffff7a6029a in malloc_printerr () from /nix/store/f6kvkdzp6qfjm6h94d0jgfvm4r06xcaq-glibc-2.34-210/lib/libc.so.6
[...]

PR #2302, opened 2022-11-06
Define bot includes only once

necessarily-equal:
The patch is ugly, but this really just moves D() and D2() directives up.

This should alleviate https://github.com/Unvanquished/Unvanquished/issues/2279, and maybe hopefully fix it.

The bot include code looks terrible though, and getting rid of it would be worthwhile anyhow.

### DIFF
--- a/src/sgame/sg_bot_parse.cpp
+++ b/src/sgame/sg_bot_parse.cpp
@@ -1247,36 +1247,17 @@ static AIGenericNode_t *ReadNode( pc_token_list **tokenlist )
 	return node;
 }
 
-/*
-======================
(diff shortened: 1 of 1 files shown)


necessarily-equal:
I've tested bots can still use medkits after this, so it looks like it works.

slipher:
The function doesn't actually run only once.

necessarily-equal:
> The function doesn't actually run only once.

Whoops, I left a `static` out. Fixed

slipher:
LGTM

necessarily-equal:
Note to self: test that this new version still works before merging

PR #2304, opened 2022-11-08
Update to new color parser interface

slipher:
Companion to https://github.com/DaemonEngine/Daemon/pull/732.

Fixes the following bugs:
- The chat field displayed the sequence `^^` incorrectly
- The code to prevent a dangling `^` at the end of player names didn't work

### DIFF
--- a/src/cgame/rocket/rocket.cpp
+++ b/src/cgame/rocket/rocket.cpp
@@ -521,9 +521,10 @@ Rml::String Rocket_QuakeToRML( const char *in, int parseFlags = 0 )
 	for ( auto iter = parser.begin(); iter != parser.end(); ++iter )
 	{
 		const Color::Token& token = *iter;
-		if ( token.Type() == Color::Token::TokenType::CHARACTER )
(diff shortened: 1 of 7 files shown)


src/cgame/rocket/rocket.cpp
```diff
@@ -578,10 +579,10 @@ Rml::String Rocket_QuakeToRML( const char *in, int parseFlags = 0 )
 					out.append( spanstr );
 					spanHasContent = true;
 				}
-				out.append( token.Begin(), token.Size() );
+				out.append( text.begin(), text.size() );
 			}
 		}
-		else if ( token.Type() == Color::Token::TokenType::COLOR )
+		else // token.Type() == Color::Token::TokenType::COLOR
```
DolceTriade, line 585: why comment?

DolceTriade (approved):

src/cgame/rocket/rocket.cpp
slipher: I had an awkward if condition of type != COLOR to avoid diffs. I added another commit to swap the if-else branches.

PR #2306, opened 2022-11-09
Add build* to .gitignore and update RmlUi submodule

necessarily-equal:


necessarily-equal:
Looks like the new RmlUi doesn't build in NaCl :-/

illwieckz:
> Add build* to .gitignore

I don't get this one, which purpose it deserves?

Do you know you can do that?

```sh
mkdir -p build/one
cd build/one
cmake ../..
make

cd ../..

mkdir -p build/two
cd build/two
cmake ../..
make

```

necessarily-equal:
Oh well the "Add build* to .gitignore" commit was merged already anyway (in another PR).

About the rmlui upgrade, well, I asked around on the matrix/gitter channel `#RmlUi_community:gitter.im` and I think it's not worth trying for now, and we may want to postpone that until we migrate to wasm.
```
afontain> Well, one tiny trouble we could have at the moment is that RmlUi introduced a small change that is incompatible with our current (nearly obsolete) release toolchain https://dev.azure.com/UnvanquishedDevelopment/Unvanquished/_build/results?buildId=3889&view=logs&j=adc611be-d2d6-58a9-dbb6-278cbdb63baf&t=09b9f419-6765-5e77-4a7d-f7c0c49d5f18&l=522
afontain> to be honest, in the current state of things, it would be possible to just ignore this problem for now and wait until we move on something else
afontain> (this something else could be web assembly with the wasmtime runtime)
afontain> This is a bit of an odd requirement we have, basically we have to use some old clang fork that only supports C++11
with its own bugs
afontain> once we move our sandboxing to something more recent, then we will still be limited to clang, but a standard version and we can have (e.g.) C++20
mikke89 (Michael R. P. Ragazzon)> Oh, I see. Yeah, we can't really keep backward compatibility forever. I think C++14 is a very reasonable requirement these days. 
afontain> I think we use some form of C++14, but it seems to be incomplete, or before the revisions of the standard. We seem to build with set_cxx_flag("-std=gnu++14")
mikke89 (Michael R. P. Ragazzon)> Yeah, there are older compiler versions where you'll find incomplete or buggy support.
```

Issue #2290, opened 2022-11-03
Small devtool regression: g_bot_infiniteMomentum doesn't allow alien bots to evolve anymore


necessarily-equal:
Since https://github.com/Unvanquished/Unvanquished/pull/2190 G_AlienEvolve checks if the conditions are met, and well, it does check if the momentum condition is met :S

PR #2308, opened 2022-11-09
Restore g_bot_infiniteMomentum functionality and cleanups

necessarily-equal:
Closes https://github.com/Unvanquished/Unvanquished/issues/2290

### DIFF
--- a/src/shared/bg_misc.cpp
+++ b/src/shared/bg_misc.cpp
@@ -425,7 +425,6 @@ to allow devolving, or if you can afford the upgrade.
 */
 evolveInfo_t BG_ClassEvolveInfoFromTo( const int from, const int to )
 {
-	bool classIsUnlocked;
 	bool isDevolving;
 	int fromCost, toCost;
 	int evolveCost;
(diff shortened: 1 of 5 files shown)


src/shared/bg_misc.cpp
```diff
@@ -459,33 +459,6 @@ evolveInfo_t BG_ClassEvolveInfoFromTo( const int from, const int to )
 	return { classIsUnlocked, isDevolving, evolveCost };
 }
 
-/*
```
DolceTriade, line 462: Can you add context in your commit messages for why we are making these changes

DolceTriade (commented):
I'm ok with the concept, but please add more context in commit messages for why we are making certain changes, especially when they seem unclear from the PR message

src/shared/bg_misc.cpp
necessarily-equal: I've added some text in the commit message. I thought the comment on the function it removes would be suggestive enough `// [this] function will always return true because it will notice you can devolve to dretch or granger, even when far from the overmind`

DolceTriade: Yes, but its better to have it in the commit message :D

DolceTriade:
LGTM % commit message that says "it has no purpose IMO". Please change it to why it has no purpose.

PR #2312, opened 2022-11-18
Move entityState_t to gamelogic

necessarily-equal:


DolceTriade:
yay! I'm glad you tackled this!

src/shared/bg_public.h
```diff
@@ -144,6 +144,60 @@ struct playerState_t
 	bool IsWeaponReady() const;
 };
 
+struct entityState_t
+{
+	int          number; // entity index
+	entityType_t eType; // entityType_t
+	int          eFlags;
+
+	trajectory_t pos; // for calculating position
```
slipher: I don't see how this could be working as the fields are in a different order from OpaqueEntityState.

slipher: Oh there is another commit. Well, if you really want to add 2 daemon submodule references after merging in daemon, I guess that works

necessarily-equal: Yes, I think the daemon references will unfortunately change quite a bit.

necessarily-equal:
You can find the "gamelogic-entityState_t" testing server on the server list. The bug happens only when the clients needs to download a new dpk. To trigger the bug, you can download https://users.unvanquished.net/~afontain/pkg/nightly/exp/gamelogic-entityState/unv_0.dpk and so some equivalent of `mv Downloads/unv_0.dpk ~/.local/share/unvanquished/pkg/unvanquished_0.54.0-gamelogic-entityState.dpk`. You will thus have the right cgame to try to skip the download message, but still have to download because the name doesn't match.

Now, to explain the bug. I have a bug where I don't manage to skip the right number of bits in the server message when the client needs to download cgame. The offset calculation is likely wrong but I don't know how to fix it. The client has to do this calculation because it has to ignore just the right number of bits to avoid desynchronisation in the protocol.

PR #2314, opened 2022-11-18
Modernize CG_DrawBuildableStatus and add a comment

necessarily-equal:


necessarily-equal:
This is independant from https://github.com/Unvanquished/Unvanquished/pull/2312 and doesn't conflict with it. It finishes the previous migration that made snapshot->entities a vector though.

DolceTriade (approved):

necessarily-equal:
Screenshot from 2022-11-18 19-05-02

I should add that I tested the ordering to be correct, it was initially reversed :)

PR #2318, opened 2022-11-22
Modernize cg_predict to std::vector (+nit)

necessarily-equal:


src/cgame/cg_predict.cpp
```diff
@@ -51,35 +49,23 @@ It also adds content flags to allow for more specific traces in synchronized cod
 */
 void CG_BuildSolidList()
 {
-	centity_t     *cent;
-	snapshot_t    *snap;
-	entityState_t *ent;
+	cg_solidEntities.clear();
```
DolceTriade, line 52: Should you reserve(MAX_ENTITIES_IN_SNAPSHOT]

src/cgame/cg_predict.cpp
```diff
@@ -33,10 +33,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 static  pmove_t   cg_pmove;
 
-static  int       cg_numSolidEntities;
-static  centity_t *cg_solidEntities[ MAX_ENTITIES_IN_SNAPSHOT ];
-static  int       cg_numTriggerEntities;
-static  centity_t *cg_triggerEntities[ MAX_ENTITIES_IN_SNAPSHOT ];
+static std::vector<centity_t *> cg_solidEntities;
```
slipher, line 36: How about BoundedVector<centity_t*, MAX_GENTITIES>

necessarily-equal: This should be fine as entities are known to be duplicated inside a snapshot. On the other hand, I'm not sure it makes sense to have a different handling here than in every other instance. Maybe a BoundedVector migration would have to start with notably `snapshot_t` and `clSnapshot_t`.

src/cgame/cg_predict.cpp
necessarily-equal: I don't know. On one side this means that we need to use `(sizeof(*centity_t) = 2448) * 1024 * 2` entities, that's 5,0MB(!) of `.bss` (heap) memory which is typically not used (only a few entities, while most of those entities are not used because they are not buildables). On the flip side this avoids possible out-of-memory errors.

src/cgame/cg_predict.cpp
necessarily-equal: See also the other comment about the memory usage. I'm really not sure this is a good idea.

src/cgame/cg_predict.cpp
necessarily-equal: Oh wait, this is a `centity_t*`, so it has a size of 8. So that means 1,6kB. That's actually reasonable.

src/cgame/cg_predict.cpp
slipher: This is hardly a "migration" as BoundedVector does the exact same thing as the existing code here.

necessarily-equal: Okok, so I changed it here and in `cg_buildable.cpp` for consistency. I'm not touching snapshot_t because it's in daemon, and because I'd like to avoid conflicts.

slipher:
LGTM

Issue #2085, opened 2022-08-17
"You are at the front of the spawn queue" message stuck on screen after leaving team while in a spawn queue


cu-kai:
Noted just now on Bunker, where I left the team while in the spawn queue. Now I have the message stuck on my screen.

This doesn't go away until I join a team again and spawn. If I leave the team (even while spawned), the message reappears on my screen.

![unvanquished_2022-08-17_195142_000](https://user-images.githubusercontent.com/13281185/185220127-ad6ea5db-7c6d-443a-80dc-9de2286919f7.jpg)

DolceTriade:
Yes I've seen this as well.

PR #2319, opened 2022-11-22
Enable /teamstatus by default

necessarily-equal:
Enabled with a delay of 5s between commands.

### DIFF
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -182,7 +182,7 @@ Cvar::Cvar<float> g_sayAreaRange("g_sayAreaRange", "distance for area chat messa
 
 Cvar::Cvar<int> g_floodMaxDemerits("g_floodMaxDemerits", "client message rate control (lower = stricter)", Cvar::NONE, 5000);
 Cvar::Cvar<int> g_floodMinTime("g_floodMinTime", "mute period after flooding, in milliseconds", Cvar::NONE, 2000);
-Cvar::Cvar<int> g_teamStatus("g_teamStatus", "allow /teamstatus command. 0 = disabled, otherwise you can only /teamstatus every <g_teamStatus> seconds.", Cvar::NONE, 0);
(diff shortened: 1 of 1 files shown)


DolceTriade:
lgtm

necessarily-equal:
(CI failure is unrelated)

slipher:
Personally I dislike it

DolceTriade:
My impression is that the majority of the players like it. Server owners who dislike it can disable it of course, but I think it's almost expected now from Unv players.

slipher:
OK, just casting my vote

necessarily-equal:
Ideally I'd make some improved HUD on tab key with minimap and team status and all; but I don't have the time for that at the moment. I think this is a good start in the meantime :smile:

sweet235:
There is another reason for wanting to have this enabled on all servers: we still did not fix the bug that sometimes tells players there is no spawn left, when in fact there is a single spawn left. This is very confusing, and asking /teamstatus will help, as it counts for itself.

> Ideally I'd make some improved HUD on tab key with minimap and team status

I would prefer this too. At least the output of teamstatus is rather pleasant to look at (compared to other text output).

DolceTriade:
> There is another reason for wanting to have this enabled on all servers: we still did not fix the bug that sometimes tells players there is no spawn left, when in fact there is a single spawn left. This is very confusing, and asking /teamstatus will help, as it counts for itself.

What? Was this not fixed by https://github.com/Unvanquished/Unvanquished/issues/2085?

sweet235:
I was not aware of https://github.com/Unvanquished/Unvanquished/issues/2085, otherwise i would have commented on that earlier. I'm quite sure that I saw this bug on my server a few days ago. My server is currently based on master from around 4 weeks ago.

sweet235:
After reading https://github.com/Unvanquished/Unvanquished/issues/2085 carefully, I'm not sure this is the same bug I was talking about.

necessarily-equal:
I think I'll merge this this weekend if nobody has a strong opinion against this. Overall I think it's a good thing, even if I would gladly take a good minimap or hud over this.

DolceTriade:
Sorry, I meant https://github.com/Unvanquished/Unvanquished/commit/045ca800e8852f89cc08382a61a672b3a15325ce

PR #2322, opened 2022-12-03
Fix Mac CI and zlib link error

slipher:
The pip command is not installing the CBSE requirements in the Python version that is selected by CMake. The error seems to be associated with selecting `/usr/local/Frameworks/Python.framework/Versions/3.11/bin/python3.11` instead of `/Library/Frameworks/Python.framework/Versions/3.11/bin/python3.11`.


slipher:
Besides the Python issues there was an additional problem with a missing zlib dependency.

Also stop installing some Daemon dependencies that are no longer required to appease the build system.

illwieckz (approved):
LGTM.

PR #2324, opened 2022-12-05
Improve bot barbs

necessarily-equal:
* Increase bot barbs safety distance: I have seen alien bots still barbing themselves
* Make barbs handling more clear
* Make advanced goon aim its barbs before firing them. This is devilishly effective (which is why it's hidden behind a skill in the next commit).
* Add bot barb skills
* Give more skill points to alien bots. This adjusts aliens skill point number to have them unlock every skill at skill level 7 like humans already do, and make skill point attribution simpler.

### DIFF
--- a/src/sgame/sg_bot_skilltree.cpp
+++ b/src/sgame/sg_bot_skilltree.cpp
@@ -164,12 +164,10 @@ std::pair<std::string, skillSet_t> BotDetermineSkills(gentity_t *bot, int skill)
 	std::vector<botSkillTreeElement_t> possible_choices = initial_unlockable_skills;
 	// aliens have 61 points to spend max
 	// humans have 40 points to spend max
-	// here we give a bit more money to humans because they have more
-	// expensive skills
-	float max = G_Team(bot) == TEAM_ALIENS ? 61.0f : 46.0f;
+	float max = G_Team(bot) == TEAM_ALIENS ? 61.0f : 40.0f;
 
-	// unlock every skill at skill 8
-	int skill_points = static_cast<float>(skill) / 8.0f * max;
+	// unlock every skill at skill 7
+	int skill_points = static_cast<float>(skill) / 7.0f * max;
 
 	// rng preparation
 	std::string name = bot->client->pers.netname;
--- a/src/sgame/sg_bot_skilltree.cpp
+++ b/src/sgame/sg_bot_skilltree.cpp
@@ -68,6 +68,7 @@ static const std::vector<botSkillTreeElement_t> survival_skills = {
 	{ "mantis-flee-jump",   BOT_A_LEAP_ON_FLEE,          3, pred_alien, {} },
 	{ "goon-flee-jump",     BOT_A_POUNCE_ON_FLEE,        4, pred_alien, {} },
 	{ "tyrant-flee-run",    BOT_A_TYRANT_CHARGE_ON_FLEE, 4, pred_alien, {} },
+	{ "safe-barbs",         BOT_A_SAFE_BARBS,            3, pred_alien, {} },
 
 	// humans
 	{ "buy-armor",          BOT_H_BUY_ARMOR,   10, pred_human, {
@@ -80,6 +81,7 @@ static const std::vector<botSkillTreeElement_t> survival_skills = {
 static const std::vector<botSkillTreeElement_t> fighting_skills = {
 	// aliens
 	{ "aim-head",           BOT_A_AIM_HEAD,      10, pred_alien, {} },
+	{ "aim-barbs",          BOT_A_AIM_BARBS,     7,  pred_alien, {} },
 
 	// humans
 	{ "predict-aim",        BOT_H_PREDICTIVE_AIM, 5, pred_human, {} },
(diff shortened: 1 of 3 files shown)


src/sgame/sg_bot_util.cpp
```diff
@@ -1924,18 +1924,38 @@ void BotFireWeaponAI( gentity_t *self )
 			break;
 		case WP_ALEVEL3_UPG:
 		{
+			glm::vec3 origin = VEC2GLM( self->s.origin );
```
slipher, line 1927: These can be moved closer to the point of use

src/sgame/sg_bot_util.cpp
```diff
@@ -1935,7 +1935,7 @@ void BotFireWeaponAI( gentity_t *self )
 			// hopefully accounts for the movement of the bot and
 			// its target
 			constexpr float barbSafetyFactor = 5.0f/3.0f;
-			bool barbIsSafe = distance > (barbSafetyFactor * BG_Missile(MIS_BOUNCEBALL)->splashRadius);
+			bool barbIsSafe = distance > (barbSafetyFactor * BG_Missile(MIS_BOUNCEBALL)->splashRadius) || !self->botMind->botSkillSet[BOT_A_SAFE_BARBS];
```
slipher, line 1938: It's logically equivalent but it seems more intuitive to put the skill check as the first part of the OR. So that the other code is only executed if the skill is enabled. (This comment applies to both skills)

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -160,14 +162,12 @@ static Util::optional<botSkillTreeElement_t> ChooseOneSkill(const gentity_t *bot
 std::pair<std::string, skillSet_t> BotDetermineSkills(gentity_t *bot, int skill)
 {
 	std::vector<botSkillTreeElement_t> possible_choices = initial_unlockable_skills;
-	// aliens have 51 points to spend max
+	// aliens have 61 points to spend max
 	// humans have 40 points to spend max
-	// here we give a bit more money to humans because they have more
-	// expensive skills
-	float max = G_Team(bot) == TEAM_ALIENS ? 51.0f : 46.0f;
+	float max = G_Team(bot) == TEAM_ALIENS ? 61.0f : 40.0f;
```
slipher, line 167: Why did the amount of human skill points decrease?

necessarily-equal: It's actually about the same, since 40/7 ≃ 46/8.

src/sgame/sg_bot_util.cpp
necessarily-equal: I think this is more clear this way. You are thinking of writting it like this, right?

```
bool barbIsSafe = !self->botMind->botSkillSet[BOT_A_SAFE_BARBS] || distance > (barbSafetyFactor * BG_Missile(MIS_BOUNCEBALL)->splashRadius);
```

src/sgame/sg_bot_util.cpp
necessarily-equal: Updated

src/sgame/sg_bot_skilltree.cpp
necessarily-equal: it's also exactly the same for skill5, since (int)(46.0f÷8×5) = (int)28.75 = 28 and (int)(40÷7×5) = (int)28.57 = 28

src/sgame/sg_bot_util.cpp
slipher: Yes, like that.

src/sgame/sg_bot_skilltree.cpp
slipher: I see. #2325 things :stuck_out_tongue:

src/sgame/sg_bot_util.cpp
necessarily-equal: I find that confusing because you would think the condition is inverted. While it's actually that the bot *think* the barb is safe because it didn't bother to check that it actually isn't safe.

necessarily-equal: Maybe I should change the variable name here.

slipher: My line of thinking has nothing to do with barbs in particular; I think that optional calculations for skills should be gated behind the skill check.

necessarily-equal: I mean most of the calculation is done anyway 🤷‍♂️️

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -160,14 +162,12 @@ static Util::optional<botSkillTreeElement_t> ChooseOneSkill(const gentity_t *bot
...
 
-	// unlock every skill at skill 8
```
DolceTriade, line 169: imo, we shouldn't pick a random max skill level of 7. We should pick either 5 or 10.

DolceTriade (commented):
otherwise lgtm

src/sgame/sg_bot_skilltree.cpp
necessarily-equal: Well, I mean I agree that on paper it sounds good. But now they max out at sk7 and bots are already significantly nerfed at sk5. If we did use sk10 as a max, then bots would be terribly bad by default. (and also bot skill cap at 9)

On the other side, if we did max out skills at sk5, bots would already be surprisingly professional at sk5 already and wouln't have more progression opportunities, beside being more aimbot-y. I think sk7 is fine for now. As we make bots more and more skilled the number will likely have to go further up, but for now this is reasonable enough.

PR #2329, opened 2022-12-07
New navcon text format, with converter, fix #1772

### LINKED ISSUE #1772: migrate navcon files from binary to text format
Currently, files used to store off-mess connections (that is, connections between 2 navmeshes or shorcuts within a navmesh), commonly referred to as `navcon`, use a binary format which does works well with versioning, script edition or even manual edition.

Navcons are linked directly to navmeshes, and some might be possible to generate automatically, for example those of movers (hopefully, if not, please complete) exhaustive list:

* doors;
* elevators;
* teleporters;
* pushers;
* pendulums;
* ladders;
 
But some are very unlikely to be generated automatically one day, namely, the ones which interacts with normal geometry. From this PoV, ladders navcons will also be harder ones to generate, since they just that: geometry, with a special texture applied on it (which explains many of the problems of ladders, even for players):

* jumping over chasms
* jumping upward
* jumping downard

Note that with some tricks, it is possible to get some limited jumping up/down-ward done automatically, but it will be limited in height at agent's height minus one unit (`crouchMaxs[2] - mins[2] - 1` in code), and might lead to new problems itself, like revealing new bugs in generator.

Each "agent" have it own navmesh, and agents are currently defined with those values:

* radius;
* height (crouched height should be used, combined with obstacle detection changes, but currently only max height is);

The current list is hard-coded (list of `class_t`), mostly considering the values above, and without considering their abilities (as an `upgrade_t` can alter the abilities of a `class_t`), which is an error as it effectively prevents to distinguish agents of same size but different abilities, for example: a naked human with a jetpack, a dretch can not pounce but mantises, which are of same size, can.
There are also other different damages modifiers which can be taken depending on the length of the fall, or even only the height magnitude.

Off-mesh connections can at least the following properties set from the editor:

* 3D start coordinates
* 3D end coordinates
* one-way or two-way
* "tunnel" radius
* `class_t` which is able to use it

What is needed to do here is to find a way to describe them so that enough information can be bound to them so that a bot can be able to know if that navcon concerns it or not, which may even depend on their health status (let's dream, but honestly that's one of the things I could not care too much for a 1st implementation).

illwieckz:
Introduce a new navcon text format.
Make the engine loads both old and new format and write new format.
Add a navcon converter to convert existing navcon files to the new format.

Fix #1772.

The output format is a text format, all numbers are written as integers as float precision is believed to be overkill.

Output format is like:

```
navcon 3
x1 z1 y1 x2 z2 y2 rad flags areas dirs
x1 z1 y1 x2 z2 y2 rad flags areas dirs
```

Example:

```
$ hexdump -Cv ~/.local/share/unvanquished/game/maps/plat23-level4.navcon
00000000  02 00 00 00 02 00 00 00  56 c5 7e c3 00 00 00 c0  |........V.~.....|
00000010  b9 bc e4 44 e1 59 0a c3  28 6c c9 3d 48 65 dc 44  |...D.Y..(l.=He.D|
00000020  99 f8 1e 43 00 00 00 c0  91 76 0c 45 9f 45 a3 43  |...C.....v.E.E.C|
00000030  4c 6f e1 c2 24 12 0c 45  00 00 48 42 00 00 48 42  |Lo..$..E..HB..HB|
00000040  01 00 01 00 3f 3f 01 00  00 00 00 00 00 00 00 00  |....??..........|
00000050

$ ./convert-navcon ~/.local/share/unvanquished/game/maps/plat23-level4.navcon > plat23-level4.navcon
Converting /home/illwieckz/.local/share/unvanquished/game/maps/plat23-level4.navcon
Converting from format version 2 to 3
Converting 2 connections

$ ./convert-navcon plat23-level4.navcon
Not converting plat23-level4.navcon
Already in format version 3

$ cat plat23-level4.navcon
navcon 3
-254 -2 1829 -138 0 1763 50 1 63 1
158 -2 2247 326 -112 2241 50 1 63 0

$ du -b ~/.local/share/unvanquished/game/maps/plat23-level4.navcon plat23-level4.navcon
80	/home/illwieckz/.local/share/unvanquished/game/maps/plat23-level4.navcon
80	plat23-level4.navcon
```

### DIFF
--- a/src/sgame/botlib/bot_load.cpp
+++ b/src/sgame/botlib/bot_load.cpp
@@ -39,6 +39,9 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 #include "sgame/sg_local.h"
 #include "bot_local.h"
 
+static const char* NAVCON_HEADER_PREFIX = "navcon";
+static const int NAVCON_VERSION = 3;
+
 static Cvar::Range<Cvar::Cvar<int>> maxNavNodes(
 	"bot_maxNavNodes", "maximum number of nodes in navmesh", Cvar::NONE, 4096, 0, 65535);
 
@@ -70,6 +73,27 @@ void BotAssertionInit()
 #endif
 }
 
+void BotDebugPrintOffMeshConnections( OffMeshConnections &con )
+{
+	#ifdef DEBUG_BUILD
+	for ( int n = 0; n < con.offMeshConCount; n++ )
+	{
+		// userids is unused.
+		Log::Debug( "%0.f %.0f %.0f %.0f %.0f %.0f %.0f %hd %hhd %hhd",
+			con.verts[ ( 6 * n ) + 0 ],
+			con.verts[ ( 6 * n ) + 1 ],
+			con.verts[ ( 6 * n ) + 2 ],
+			con.verts[ ( 6 * n ) + 3 ],
+			con.verts[ ( 6 * n ) + 4 ],
+			con.verts[ ( 6 * n ) + 5 ],
+			con.rad[ n ],
+			con.flags[ n ],
+			con.areas[ n ],
+			con.dirs[ n ] );
+	}
+	#endif
+}
+
 void BotSaveOffMeshConnections( NavData_t *nav )
 {
 	char filePath[ MAX_QPATH ];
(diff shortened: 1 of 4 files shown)


sweet235:
lgtm

tools/navcon-converter/navcon-converter
```diff
@@ -0,0 +1,139 @@
...
+
+Header version number then connection count then arrays of various types.
+
+Example:
+header count {x1 z1 y1 x2 z2 y2}[] rad[] flags[] areas[] dirs[] userids[]
+i      i      f  f  f  f  f  f     f     H       B       B      I
+
+int header;
+int count;
+union coords_t { float x1, z1, y1, x2, z2, y2 };
```
slipher, line 42: You mean struct not union

tools/navcon-converter/navcon-converter
```diff
@@ -0,0 +1,139 @@
...
+"""
+
+"""
+Version 3 text format
+
+Header line with prefix and integer version number
+then one connection per line, all numbers as integers.
+
+Example:
+#navcon 3
```
slipher, line 58: I suspect our tokenizer may not accept `#navcon` as a token, so you might want to pick something else.

tools/navcon-converter/navcon-converter
```diff
@@ -0,0 +1,139 @@
...
+                    value_int = int(struct.unpack(value_format, value_bytes)[0])
+
+                    con_list[i][subkey_name] = value_int
+
+        file_blob.close()
+
+        print("{} {}".format(header_text_prefix, str(current_version)))
+
+        for con in con_list:
+            for key in con.keys():
```
slipher, line 129: This loop could be replaced by `print(*con.values())`

slipher:
`userid` seems to be unused (always 0) so you could omit that from the text format.

tools/navcon-converter/navcon-converter
necessarily-equal: Maybe just "navcon 3" would be enough

tools/navcon-converter/navcon-converter
illwieckz: done.

tools/navcon-converter/navcon-converter
illwieckz: done

tools/navcon-converter/navcon-converter
illwieckz: Wow, what is this dark magic???.
done.

illwieckz:
> `userid` seems to be unused (always 0) so you could omit that from the text format.

I stripped it.

Anyway, that's fun:

```
$ du -b ~/.local/share/unvanquished/game/maps/plat23-level4.navcon plat23-level4.navcon
80	/home/illwieckz/.local/share/unvanquished/game/maps/plat23-level4.navcon
80	plat23-level4.navcon
```

illwieckz:
I added a patch to make the game load both the old and new format and write the new format.

This is a shameless copy paste of the layout parser writer (which uses a similar format), despite being _so C89_, it works.

src/sgame/botlib/bot_load.cpp
```diff
@@ -79,101 +98,175 @@ void BotSaveOffMeshConnections( NavData_t *nav )
...
+		SwapArray( nav->process.con.userids, conCount );
+
+		Log::Debug( "Loaded %d connections from navcon file %s", conCount, filePath );
 		trap_FS_FCloseFile( f );
+
+		BotPrintOffMeshConnections( nav );
 		return;
 	}
+	// New text format, check "navc" from "navcon".
+	else if ( version == 1668702574 )
```
slipher, line 197: For a text format we shouldn't require that the first token is at the exact beginning of the file since whitespace is not supposed to be significant. Just start treating it as text with `sscanf` etc.

tools/navcon-converter/navcon-converter
```diff
@@ -0,0 +1,147 @@
...
+
+    file_blob.close()
+
+    return connection_list
+
+def write_format3(connection_list):
+    print("{} {}".format(header_text_prefix, str(current_format_version)))
+
+    for connection in connection_list:
+        print(*connection.values())
```
slipher, line 127: Actually the `%.0f` thing (see the comment on the C++ side) would be good here too for correct rounding. You can do `print("%.0f %.0f %.0f etc" % tuple(connection.values())`

src/sgame/botlib/bot_load.cpp
```diff
@@ -79,101 +98,175 @@ void BotSaveOffMeshConnections( NavData_t *nav )
...
-	trap_FS_Write( userids, size, f );
-	dtFree( userids );
+	char *s = va( "navcon %d\n", NAVMESHCON_VERSION );
+
+	trap_FS_Write( s, strlen( s ), f );
+
+	for ( int n = 0; n < conCount; n++ )
+	{
+		// userids is unused.
+		s = va( "%d %d %d %d %d %d %d %hd %hhd %hhd\n",
```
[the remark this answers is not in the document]
illwieckz: What is ”correct rounding”? Does it offset the truncation? If yes is there a good reason to offset the truncation?

src/sgame/botlib/bot_load.cpp
illwieckz: how can I restart a `read` from the start of the file without closing/reopening the file?

illwieckz: Also I don't think “white space are not supposed to be significant” is a good idea there, it's not really supposed to be written by hand like a material or a map metadata file, and handling whitespace agnosticism just adds extra complexity that is probably overkill and useless, we would also have to avoid `sscanf`… This being a text format is primarily made to be diffable/mergeable, I don't mind if the format is strict.

src/sgame/botlib/bot_load.cpp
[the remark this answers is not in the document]
illwieckz: I've done it  anyway.

tools/navcon-converter/navcon-converter
```diff
@@ -0,0 +1,147 @@
...
+    header_text_magic = read_unpack(header_text_prefix, "<i")
+
+    if header_magic == 2:
+        connection_list = read_format2(file_blob)
+        write_format3(connection_list)
+    elif header_magic == header_text_magic:
+        print("Not converting {}".format(file_name), file=sys.stderr)
+        print("Already in format version {}".format(current_format_version), file=sys.stderr)
+    else:
+        print("Not converting {}".format(file_name), file=sys.stderr)
```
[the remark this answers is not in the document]
illwieckz: Done.

tools/navcon-converter/navcon-converter
illwieckz: I done it in another way, calling `round()` before converting to `int` (they're already `int` at this point).

illwieckz:
I added the ability to pass `-i` to the converter to rewrite a file in place.

src/sgame/botlib/bot_load.cpp
slipher: We have `trap_FS_Seek`

src/sgame/botlib/bot_load.cpp
```diff
@@ -79,101 +102,176 @@ void BotSaveOffMeshConnections( NavData_t *nav )
...
+		Log::Debug( "Loaded %d connections from navcon file %s", conCount, filePath );
 		trap_FS_FCloseFile( f );
+
+		BotPrintOffMeshConnections( nav );
 		return;
 	}
 
-	int conCount = header.numConnections;
+	// Seek to the beginning of the file.
+	trap_FS_Seek( f, -1 * (int) sizeof( version ), fsOrigin_t::FS_SEEK_CUR );
```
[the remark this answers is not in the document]
slipher: I mean FS_SEEK_SET

src/sgame/botlib/bot_load.cpp
```diff
@@ -79,101 +102,176 @@ void BotSaveOffMeshConnections( NavData_t *nav )
...
+		if ( *navcon == '\n' )
+		{
+			if ( n == nav->process.con.MAX_CON )
+			{
+				Log::Warn( "Too much connections in navcon file %s", filePath );
+				break;
+			}
+
+			i = 0;
+			sscanf( line, "%f %f %f %f %f %f %f %hd %hhd %hhd\n",
```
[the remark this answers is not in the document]
illwieckz: Done.

src/sgame/botlib/bot_load.cpp
```diff
@@ -79,101 +102,176 @@ void BotSaveOffMeshConnections( NavData_t *nav )
...
+	// New text format.
+	char *navcon = (char*) BG_Alloc( len + 1 );
+	char *navconHead = navcon;
+
+	trap_FS_Read( navcon, len, f );
+	trap_FS_FCloseFile( f );
+	navcon[ len ] = '\0';
 
-	if ( conCount > nav->process.con.MAX_CON )
+	// "navc" from "navcon" is already read, read remaining "on".
```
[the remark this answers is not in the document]
illwieckz: Done.

illwieckz:
I added some extra code to make sure the parser doesn't read missing number of a line from the next line, also the `sscanf` calls detect if there is missing fields or too much fields.

src/sgame/botlib/bot_load.cpp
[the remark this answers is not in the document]
illwieckz: Done.

src/sgame/botlib/bot_load.cpp
```diff
@@ -68,6 +71,27 @@ void BotInit()
 #endif
 }
 
+void BotPrintOffMeshConnections( OffMeshConnections &con )
```
slipher, line 74: Remove this. Log verbosity must be controlled by cvars, not build type. But it's too noisy in any case.

src/sgame/botlib/bot_load.cpp
```diff
@@ -79,101 +103,231 @@ void BotSaveOffMeshConnections( NavData_t *nav )
...
+		}
+
+		line[ i++ ] = *navcon;
+		line[ i ] = '\0';
+
+		if ( *navcon == '\n' )
+		{
+			if ( n == con.MAX_CON )
+			{
+				Log::Warn( "Too much connections in navcon file %s", filePath );
```
slipher, line 278: s/much/many/

src/sgame/botlib/bot_load.cpp
```diff
@@ -79,101 +103,231 @@ void BotSaveOffMeshConnections( NavData_t *nav )
...
+	}
 
-	trap_FS_Read( con.areas, sizeof( unsigned char ) * conCount, f );
-	trap_FS_Read( con.dirs, sizeof( unsigned char ) * conCount, f );
+	Log::Debug("Loading format version %d navcon file %s", version, filePath );
+	navcon += strlen( va( "%s %d\n", NAVCON_HEADER_PREFIX, version) );
 
-	trap_FS_Read( con.userids, sizeof( unsigned int ) * conCount, f );
-	SwapArray( con.userids, conCount );
+	char line[ MAX_STRING_CHARS ];
```
slipher, line 261: May I suggest a more economical way to iterate lines:

```
for ( int pos = 0; pos < len; )
{
   int lineLength = strcspn( navcon + pos, "\n" );
   navcon[ pos + lineLength ] ='\0';
   
  sscanf( navcon + pos, ...);
  // etc.

   pos += lineLength + 1;
}

Comment (illwieckz) [reply src/sgame/botlib/bot_load.cpp:261]: Right, if that works we may also backport it to the layout parser (I initially copy pasted the layout parser).

Comment (illwieckz) [reply src/sgame/botlib/bot_load.cpp]: Done,

File: src/sgame/botlib/bot_load.cpp (hunk 11 above)
Comment (illwieckz) [reply src/sgame/botlib/bot_load.cpp]: I really don't want to ever have to debug this code without that debug log.

Comment (slipher) [reply src/sgame/botlib/bot_load.cpp]: What would be the log be like if the entire contents of every file the game opens were printed to the console? It's not reasonable. This has a very low chance of being useful to anyone for how spammy it is.

Comment (illwieckz) [reply src/sgame/botlib/bot_load.cpp]: At first it was just a code I needed in order to do that PR and verify the code works, including when implementing suggested changes like the new way to parse lines. Right now this debug log is only printed if debug log is enabled and this is a debug build, this can be made even more disabled than that, but really, I don't want to edit this code without any way to check for the correct parsing.

Every suggest or request in favor of a rewrite of a part of this parser is just asking for this debug log to exist, even if hidden behind a `#if 0`.

Comment (slipher) [reply src/sgame/botlib/bot_load.cpp]: When you're debugging and want to see it you can hack in a call to `BotSaveOffMeshConnections`, which prints exactly the same thing, and look at the file it generates.

Or just rewrite it which would take all of 5 minutes.

File: src/sgame/botlib/bot_load.cpp [hunk 14]
```diff
@@ -79,101 +103,212 @@ void BotSaveOffMeshConnections( NavData_t *nav )
...
+		Log::Debug("Loading format version %d navcon file %s", version, filePath );
+
+		int conCount;
+		trap_FS_Read( &conCount, sizeof( conCount ), f );
+
+		conCount = LittleLong( conCount );
+
+		if ( conCount > con.MAX_CON )
+		{
+			Log::Warn( "Too much connections in navcon file %s", filePath );
```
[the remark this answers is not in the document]
illwieckz: Done.

illwieckz:
I added a commit to delete the extra debug print, that commit is easy to revert for re-enabling debugging.

src/sgame/botlib/bot_load.cpp
illwieckz: > I added a commit to delete the extra debug print, that commit is easy to revert for re-enabling debugging. -- https://github.com/Unvanquished/Unvanquished/pull/2329#issuecomment-1369301032

illwieckz:
Some navcons ready to import from https://github.com/sweet235/navcon

- https://github.com/UnvanquishedAssets/map-chasm_src.dpkdir/pull/8
- https://github.com/UnvanquishedAssets/map-parpax_src.dpkdir/pull/14
- https://github.com/UnvanquishedAssets/map-perseus_src.dpkdir/pull/7
- https://github.com/UnvanquishedAssets/map-thunder_src.dpkdir/pull/5
- https://github.com/UnvanquishedAssets/map-vega_src.dpkdir/pull/9

src/sgame/botlib/bot_load.cpp
```diff
@@ -81,79 +84,74 @@ void BotSaveOffMeshConnections( NavData_t *nav )
...
 
-	int conCount = header.numConnections;
+		int conCount;
+		trap_FS_Read( &conCount, sizeof( conCount ), f );
+
+		conCount = LittleLong( conCount );
 
 	if ( conCount > con.MAX_CON )
 	{
+			Log::Warn( "Too many connections in navcon file %s", filePath );
```
slipher, line 154: bad spacing

src/sgame/botlib/bot_load.cpp
```diff
@@ -65,7 +68,7 @@ static void FailAssertion(const char* expression, const char* file, int line)
 void BotAssertionInit()
 {
 #ifdef DEBUG_BUILD
-	dtAssertFailSetCustom( FailAssertion );
+	dtAssertFailSetCustom(FailAssertion);
```
slipher, line 71: bad spacing change

slipher (approved):

PR #2330, opened 2022-12-11
Avoid non-initialised value for /devteam angles

necessarily-equal:


slipher:
LGTM, I confirmed it fixes a floating-point exception

PR #2331, opened 2022-12-12
Fix floating point exceptions in sgame

slipher:
Fix "invalid" and "zero divide" floating point exceptions in sgame.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -569,8 +569,16 @@ static bool BotAvoidObstacles( gentity_t *self, glm::vec3 &dir )
 	return true;
 }
 
+// `dir` does not need to be normalized
 static void BotDirectionToUsercmd( gentity_t *self, const glm::vec3 &dir, usercmd_t *cmd )
 {
+	if ( glm::length2( glm::vec2(dir) ) < 1.0e-5f )
+	{
+		cmd->forwardmove = 0;
+		cmd->rightmove = 0;
+		return;
+	}
+
 	glm::vec3 forward;
 	glm::vec3 right;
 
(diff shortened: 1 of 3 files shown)


src/sgame/sg_bot_nav.cpp
```diff
@@ -569,8 +569,16 @@ static bool BotAvoidObstacles( gentity_t *self, glm::vec3 &dir )
 	return true;
 }
 
+// `dir` does not need to be normalized
 static void BotDirectionToUsercmd( gentity_t *self, const glm::vec3 &dir, usercmd_t *cmd )
 {
+	if ( glm::length2( glm::vec2(dir) ) < 1.0e-5f )
```
DolceTriade, line 575: Don't we have some epsilon defined already? Did you pick this number arbitrarily?

DolceTriade (approved):

src/sgame/sg_bot_nav.cpp
slipher: It seems we have dozens of epsilons defined, though none particularly appropriate. In this case I think the epsilon could be many orders of magnitude bigger or smaller in either direction and not matter.

src/sgame/sg_bot_util.cpp
```diff
@@ -1538,6 +1538,10 @@ void BotSlowAim( gentity_t *self, glm::vec3 &target, float slowAmount )
 	//get the Vector from the bot to the enemy (ideal aim Vector)
 	glm::vec3 aimVec = target - viewBase;
 	float length = glm::length( aimVec );
+	if ( length < 1.0f )
+	{
+		return;
+	}
```
necessarily-equal, line 1544: I'm genuinely wondering when this happens, but why not I guess

necessarily-equal (approved):
LGTM

src/sgame/sg_bot_util.cpp
slipher: It does happen. If you want to know, set a breakpoint there :stuck_out_tongue:

PR #2332, opened 2022-12-12
Cvars for different BP budgets for teams

sweet235:
There is some demand for this. One use case is big PvE layouts. Another is to balance maps that favour one team.

This adds two cvars `g_BPInitialBudgetAliens` and `g_BPInitialBudgetHumans`. They are -1 (disabled) by default. Making them non-negative gives them priority over `g_BPInitialBudget`.

Once we add a 3rd team, we can simply add a 3rd cvar for that team. This approach seems easier to me than changing how `g_BPInitialBudget` works.

### DIFF
--- a/src/sgame/sg_buildpoints.cpp
+++ b/src/sgame/sg_buildpoints.cpp
@@ -69,8 +69,21 @@ float G_RGSPredictEfficiencyDelta(vec3_t origin, team_t team) {
  * @brief Calculate the build point budgets for both teams.
  */
 void G_UpdateBuildPointBudgets() {
+	int abp = g_BPInitialBudgetAliens.Get();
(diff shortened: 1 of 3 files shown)


cu-kai:
cc: @illwieckz as NetRadiant will need to be updated to support this (maps may set `g_{human|alien}BuildPoints`

It may even be worth adding support here through some sort of stub cvar for legacy maps which use this.

necessarily-equal (approved):
Not sure which to keep between this PR and https://github.com/Unvanquished/Unvanquished/pull/2332. I of course prefer mine with some reasons in my PR; and cu-kai prefers this one because « it's much closer to the old tremulous version » and because « it will require fewer changes to netradiant ».

PR #2333, opened 2022-12-12
Small tweaks

necessarily-equal:
See individual commits

### DIFF
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -121,7 +121,7 @@ static void CG_Obituary( entityState_t *ent )
 
 	target = ent->otherEntityNum;
 	attacker = ent->otherEntityNum2;
-	assistant = ent->otherEntityNum3;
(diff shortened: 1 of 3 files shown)


src/cgame/cg_event.cpp
```diff
@@ -121,7 +121,7 @@ static void CG_Obituary( entityState_t *ent )
 
 	target = ent->otherEntityNum;
 	attacker = ent->otherEntityNum2;
-	assistant = ent->otherEntityNum3;
+	assistant = ent->groundEntityNum; // we hijack the field for this
```
slipher, line 124: What's the point of this change?

necessarily-equal: It makes it more obvious that the field is hijacked, and it will allow dropping the define later. Note the commit is extracted from https://github.com/Unvanquished/Unvanquished/pull/2312.

slipher:
LGTM

Issue #2275, opened 2022-10-26
Bots do "tremble" from time to time [with reproducer]


necessarily-equal:
Sometimes bot seem to move in-place, like if they had some kind of epilepsy crisis. I think when it happens they are most of the time on top of a buildable and I think they are trying to walk towards a point that is below the buildable, but they can't get closer by walking because the buildable is in the way. Eventually the unstick kicks in and they stop doing that.

A good way to reproduce the problem is using this chasm map layout: [chasm-prison2.tar.gz](https://github.com/Unvanquished/Unvanquished/files/9873385/chasm-prison2.tar.gz). It is based on another layout by bmorel.

Steps to reproduce:
* `/devmap chasm chasm-prison2` load the map with the layout
* `/bot fill 3 h` add human bots (but preferably no alien bots to avoid distracting them)
* wait, looking at the telenodes and turrets in the human base

This layout helps a fair bit reproducing because it forces bot to roam a lot. This also means that they will randomly choose a boint below a buildable more often, triggering the bug more often.

I feel now that I've found a way to reproduce the bug it may help a lot getting a fix.

bmorel:
I still see this on @sweet235 's server. Not sure if the mod have this fix?

necessarily-equal:
If you have a picture or a recording of the place I'm interested, because I've found that my patch did fix the issue consistently for the one reproducer that I've found :-/

I assume Sweet's server should have the commit

bmorel:
No idea, I don't know which commits he had.

PR #2334, opened 2022-12-12
Fix bot "tremble"

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/2275.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -150,7 +150,13 @@ bool GoalInRange( const gentity_t *self, float r )
 
 	if ( self->botMind->goal.targetsCoordinates() )
 	{
-		return ( glm::distance( VEC2GLM( self->s.origin ), self->botMind->nav().glm_tpos() ) < r );
+		// We handle the z direction here differently,
+		// to still count standing over a turret on the
+		// target point or sitting next to a buildable
+		// counts as a "we have reached our destination".
+		glm::vec3 deltaPos = VEC2GLM( self->s.origin ) - self->botMind->nav().glm_tpos();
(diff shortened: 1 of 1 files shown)


src/sgame/sg_bot_nav.cpp
```diff
@@ -150,7 +150,11 @@ bool GoalInRange( const gentity_t *self, float r )
 
 	if ( self->botMind->goal.targetsCoordinates() )
 	{
-		return ( glm::distance( VEC2GLM( self->s.origin ), self->botMind->nav().glm_tpos() ) < r );
+		// We are a bit more lenient in the z direction here,
+		// making it count less so that if we are standing on
+		// a mgturret above our destination point, it still
+		// counts as a "we have reached our destination".
+		return ( Length2D( VEC2GLM( self->s.origin ) - self->botMind->nav().glm_tpos(), 0.2 ) < r );
```
slipher, line 157: I feel like a constant amount of leeway in the z direction would work better than one scaling on the distance, because this will be too relaxed with a larger radius. For example when an alien seeks the booster it uses radius 200. 1000 units would be way too big.

sweet235:
Seems to fix trembling bots on telenodes too. This fix is very welcome, as it makes `camper.bt` usable again for human bots.

src/sgame/sg_bot_nav.cpp
sweet235: I agree in theory. In practice, however, the scaling trick seems to work fine.

necessarily-equal: Fair enough. I've pushed a new version, how about this one ?

slipher: LGTM

PR #2335, opened 2022-12-12
Bot: respect "/g_attackStruct false" a bit more

necessarily-equal:
Before, they would bite or shoot at a buildable if it was in range and aimed at, even if they wouldn't decide to target it.


slipher:
LGTM

Issue #2035, opened 2022-08-04
improve the oxygen bar


bmorel:
As of 0.53, the oxygen bar can be improved in several areas, not counting the few little bugs:

* it should blink in red when drowning instead of disappearing (or at least show it's empty)
* it does not appear for spectator

There are also 2 bugs:

* it shows when other players are underwater, fixed by #2034 
* it does not shows when one spawns underwater

I still thing this is a nice improvement compared to 0.52, if #2034 can be merged for 0.53.1

illwieckz:
So, let's summarize bugs:

- [ ] dying while underwater
  * low priority, dying happens when the bar is empty then not visible, so it is kept invisible
  * not release-blocking
- [ ] not displayed when spectating someone
  * low priority
  * not release-blocking
- [ ] switching followed player while spectating
  * low priority because it is not displayed when spectating to begin with
  * not release-blocking
- [ ] leaving the team
  * it disappears by itself because it goes down to zero and disappear
  * not release-blocking
- [ ] spawning underwater
  * we can live with this bug for now
  * not release-blocking

slipher:
The wrong approach of basing it on submerging/surfacing noises leads to some inevitable bugs. As already mentioned, it is impossible to implement for a spectated player or to start the timer if you spawn underwater. Additionally, events could be dropped due to packet loss or (maybe) too many other events at the same time, making the timer fail to start or stop.

The right way would be to have a netcode field with the time the player entered water.

illwieckz:
1. Right now it does the job despite some minor graphical glitches that doesn't prevent to play.
2. The actual tree is better for gameplay than adding a commit disabling the feature.
3. The feature is already released and the current tree is better than the previous release.
4. We can implement a better implementation later.

With such status quo, I can start building 0.53.1.

illwieckz:
From a player point of view there are only two graphical glitches that doesn't prevent to play:

- oxybar that disappears by itself after switching team or dying underwater by something else than drowning
- missing oxybar when spawining underwater

From the player point of view the features works reliably in all cases.

I don't see anything blocking a release and requiring a new commit and more talk delaying a release.

bmorel:
> by something else than drowning.

This is theoretically fixed by https://github.com/Unvanquished/Unvanquished/issues/2035

PR #2336, opened 2022-12-13
Fix oxygen bar and add disappear animation

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/2035.
Closes https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/148.

### DIFF
--- a/src/cgame/cg_rocket_progressbar.cpp
+++ b/src/cgame/cg_rocket_progressbar.cpp
@@ -89,6 +89,13 @@ static float CG_Rocket_GetStaminaProgress()
 
 static float CG_Rocket_GetPlayerOxygenProgress()
 {
+	// Hack to show prettier output by flowing the bar back up for a short
+	// time. It doesn't correspond to anything in the game physics, but
+	// helps the player understanding that the bar resplenishes on the surface.
+	static constexpr int breathAgainDuration = 350; // ms
+	static float savedLastValue = 1.0f;
+	static int savedLastUnderWaterTime = 0;
+
 	int outTime = cg.snap->ps.lowOxygenTime;
 
 	// regular progress bar status
@@ -104,7 +111,15 @@ static float CG_Rocket_GetPlayerOxygenProgress()
 			timeLeft = (timeLeft + 1000) & ((1 << LOW_OXYGEN_TIME_BITS) - 1);
 		}
 
-		return static_cast<float>( timeLeft ) / static_cast<float>( OXYGEN_MAX_TIME );
+		savedLastValue = static_cast<float>( timeLeft ) / static_cast<float>( OXYGEN_MAX_TIME );
+		savedLastUnderWaterTime = cg.time;
+		return savedLastValue;
+	}
+
+	float overwaterDuration = cg.time - savedLastUnderWaterTime;
+	if ( overwaterDuration < (1.0f - savedLastValue) * breathAgainDuration )
+	{
+		return savedLastValue + overwaterDuration / breathAgainDuration;
 	}
 
 	return 0.0f;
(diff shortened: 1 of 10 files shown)


necessarily-equal:
AFAIK, this fixes all the listed bugs, even the spectating ones.

src/cgame/cg_rocket_progressbar.cpp
```diff
@@ -89,15 +89,27 @@ static float CG_Rocket_GetStaminaProgress()
...
-		float underwaterDuration = cg.time - cg.underwaterTime;
-		return 1.0f - underwaterDuration / static_cast<float>( OXYGEN_MAX_TIME );
+		// the time number is sent cropped, we have to do modulo arithmetics to determine the delta
+		int timeLeft = ( outTime - cg.time ) & ((1 << LOW_OXYGEN_TIME_BITS) - 1);
+
+		if ( timeLeft > OXYGEN_MAX_TIME )
+		{
+			// this happens when we just took damage but we haven't
+			// received the new time yet. Let's do some prediction.
+			timeLeft = (timeLeft + 1000) & ((1 << LOW_OXYGEN_TIME_BITS) - 1);
```
slipher, line 111: If the player is already drowning you should just set the time left to 0. And also make 1000 the zero point of time left since you can't tell whether you're really drowning already or not when it's under 1000. It does not make sense for the progressbar to go back and forth a little while already drowning

slipher (commented):
Thanks for fixing this

src/cgame/cg_rocket_progressbar.cpp
necessarily-equal: Well I like to know when I will take damage next when I'm in game, and this works out nicely for that IMO

necessarily-equal: Also note that what you are suggesting would work only for a few frames, until we receive the new oxygen reserve value from the server — and here we will have to show the bar non-empty

slipher: I mean something like:
```
int timeLeft = ( outTime - cg.time ) & ((1 << LOW_OXYGEN_TIME_BITS) - 1);
if (timeLeft > OXYGEN_MAX_TIME) //underflow
    timeLeft = 0;
else
    timeLeft = std::max(0, timeLeft - 1000);
```

As a user it would seem simply wrong and buggy to me if the bar is going back and forth while drowning.

sweet235:
I did not study the code in detail, but lgtm by testing.

This seems to fix all oxygen bar bugs known to me.

necessarily-equal:
I think I like it like it currently is

src/cgame/cg_rocket_progressbar.cpp
```diff
@@ -89,15 +89,42 @@ static float CG_Rocket_GetStaminaProgress()
...
+	int outTime = cg.snap->ps.lowOxygenTime;
+
 	// regular progress bar status
-	if ( cg.underwater )
+	if ( outTime )
+	{
+		// the time number is sent cropped, we have to do modulo arithmetics to determine the delta
+		int timeLeft = ( outTime - cg.time ) & ((1 << LOW_OXYGEN_TIME_BITS) - 1);
+
+		if ( timeLeft > OXYGEN_MAX_TIME )
```
DolceTriade, line 107: When would this happen?

slipher:
Has anything changed since the last time I reviewed it?

sweet235:
After a few days of testing, I did not notice any problems. One user seemed not to like the animation, but I think it is fine.

necessarily-equal:
> Has anything changed since the last time I reviewed it?

Hasn't changed, but I like it like this :shrug:

necessarily-equal:
Should I go ahead and merge this?

PR #2345, opened 2022-12-25
`listplayers`: don't show any admin information for non-authed admins

cu-kai:
this fixes /listplayers leaking incognito admin details

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -3891,9 +3891,9 @@ bool G_admin_listplayers( gentity_t *ent )
 		           denied,
 		           canseeWarn ? ( p->pers.hasWarnings ? "^3W" : " " ) : "",
 		           p->pers.netname,
-		           ( registeredname && authed ) ? "(a.k.a. " : "",
-		           ( registeredname && authed ) ? registeredname : "",
-		           ( registeredname && authed ) ? "^*)" : "",
+		           ( registeredname ) ? "(a.k.a. " : "",
+		           ( registeredname ) ? registeredname : "",
+		           ( registeredname ) ? "^*)" : "",
 		           ( !authed ) ? "^1NOT AUTHED" : "" ) );
 	}
 
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -3787,7 +3787,7 @@ bool G_admin_listplayers( gentity_t *ent )
 	char            *registeredname;
 	char            lname[ MAX_NAME_LENGTH ];
 	char            bot, muted, denied;
-	int             authed = 1;
+	int             authed;
 	char            namecleaned[ MAX_NAME_LENGTH ];
 	char            name2cleaned[ MAX_NAME_LENGTH ];
 	g_admin_level_t *l;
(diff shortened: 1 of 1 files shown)


sweet235:
I use this on my server now, LGTM.

src/sgame/sg_admin.cpp
```diff
@@ -3787,7 +3787,7 @@ bool G_admin_listplayers( gentity_t *ent )
 	char            *registeredname;
 	char            lname[ MAX_NAME_LENGTH ];
 	char            bot, muted, denied;
-	int             authed = 1;
+	int             authed;
```
necessarily-equal, line 3790: typically one would just remove the declaration here to put it where you initialise the variable. This avoids bugs like the one you fixed

necessarily-equal (approved):

src/sgame/sg_admin.cpp
cu-kai: I think a bigger refactor to the code could resolve this (as well as move the rest of the variables to suitable locations).

As per IRC discussion, will leave this as-is for now.

DolceTriade, line 3790: nit: Make this a bool if you only ever use it as such.

DolceTriade (approved):
lgtm % nit

src/sgame/sg_admin.cpp
cu-kai: Technically this could be defined as `-1` as `pubkey_authenticated` may have 3 values
`-1` for no pubkey
`0` for not authenticated
`1` for authenticated

from: https://github.com/Unvanquished/Unvanquished/blob/6a9ad90a6741b517a757669e579ba0c1b3826b2a/src/sgame/sg_struct.h#L542

For the purposes of this PR I think this is the best way to use it?

slipher:
It appears that the existing code is intended to reveal incognito admin details if and only if the `listplayers` caller is a higher-level admin who has the setlevel privilege. Is this not working as intended? Can non-admins see incognito admin details?

cu-kai:
> It appears that the existing code is intended to reveal incognito admin details if and only if the `listplayers` caller is a higher-level admin who has the setlevel privilege. Is this not working as intended? Can non-admins see incognito admin details?

Incognito admins are leaked while not authed (i.e. while connecting). Any player who runs /listplayers while an incognito admin is joining the game will see that admin's level and registered name.

src/sgame/sg_admin.cpp
```diff
@@ -3888,9 +3891,9 @@ bool G_admin_listplayers( gentity_t *ent )
 		           denied,
 		           canseeWarn ? ( p->pers.hasWarnings ? "^3W" : " " ) : "",
 		           p->pers.netname,
-		           ( registeredname ) ? "(a.k.a. " : "",
-		           ( registeredname ) ? registeredname : "",
-		           ( registeredname ) ? "^*)" : "",
+		           ( registeredname && authed ) ? "(a.k.a. " : "",
```
slipher, line 3894: The authed check is not needed here since registeredname can only be non-null if it was true

LGTM besides that

cu-kai: Good point, I should have checked this before pushing https://github.com/Unvanquished/Unvanquished/pull/2345/commits/a746161432e213dc3ae5c0bcd1e42adbaee1f559. Now fixed.

PR #2346, opened 2022-12-25
layouts: Reset the layout cvar when the builtin layout is used.

### LINKED ISSUE #2176: Sometimes wrong per-layout config files get executed
When having a per-layout config file like `karith/pve.cfg`, this might be executed when loading the map with its default layout. This seems to happen if and only if the last layout changed to was named `pve`.

I think the reason is this line:
https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_main.cpp#L561
It selects the config files to load based on the cvar `layout`. But this cvar seems to keep its value when maps are changed.

DolceTriade:
Previously, we wouldn't update the layout if we return early causing the previous layout's cfg to be loaded.

Fixes #2176

### DIFF
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -2384,6 +2384,7 @@ void G_LayoutSelect()
 		else
 		{
 			//use the map's builtin layout
+			Cvar::SetValue( "layout", "" );
 			return;
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/sgame/sg_buildable.cpp
```diff
@@ -2384,6 +2384,7 @@ void G_LayoutSelect()
 		else
 		{
 			//use the map's builtin layout
+			Cvar::SetValue( "layout", "" );
```
slipher, line 2387: The function has another early return, where it says `"None of the specified layouts could be found, using map default."`

DolceTriade: Forgot to commit it :(

slipher:
LGTM

PR #2348, opened 2022-12-27
Make admin_readconfig_string work better

slipher:
Just grab the rest of the line excluding leading/trailing whitespace, rather than using token functions which collapse internal whitespace and parse some escapes and quotes.

admin_writeconfig_string just copies the string verbatim and makes no attempt to correctly round-trip.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -883,11 +883,16 @@ static bool admin_higher( gentity_t *admin, gentity_t *victim )
 	                           victim->client->pers.admin );
 }
 
-static void admin_writeconfig_string( const char *s, fileHandle_t f )
+static void admin_writeconfig_string( Str::StringRef s, fileHandle_t f )
 {
-	if ( s[ 0 ] )
+	if ( !s.empty() )
 	{
-		trap_FS_Write( s, strlen( s ), f );
+		// (ch & 0xFF) < ' ' is the definition of whitespace used by the COM_Parse* functions
+		if ( ( s.front() & 0xFF ) <= ' ' || ( s.back() & 0xFF ) <= ' ' )
+		{
+			Log::Warn( "String '%s' will be incorrectly serialized in admin config", s );
+		}
+		trap_FS_Write( s.c_str(), s.size(), f );
 	}
 
 	trap_FS_Write( "\n", 1, f );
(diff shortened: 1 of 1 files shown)


src/sgame/sg_admin.cpp
```diff
@@ -863,11 +863,15 @@ static bool admin_higher( gentity_t *admin, gentity_t *victim )
...
-static void admin_writeconfig_string( char *s, fileHandle_t f )
+static void admin_writeconfig_string( Str::StringRef s, fileHandle_t f )
 {
-	if ( s[ 0 ] )
+	if ( !s.empty() )
 	{
-		trap_FS_Write( s, strlen( s ), f );
+		if ( ( s.front() & 0xFF ) <= ' ' || ( s.back() & 0xFF ) <= ' ' )
+		{
+			Log::Warn( "String '%s' cannot be serialized in admin config", s );
```
DolceTriade, line 872: return, or update warning to note that the output maybe mangled.

src/sgame/sg_admin.cpp
```diff
@@ -863,11 +863,15 @@ static bool admin_higher( gentity_t *admin, gentity_t *victim )
...
 }
 
-static void admin_writeconfig_string( char *s, fileHandle_t f )
+static void admin_writeconfig_string( Str::StringRef s, fileHandle_t f )
 {
-	if ( s[ 0 ] )
+	if ( !s.empty() )
 	{
-		trap_FS_Write( s, strlen( s ), f );
+		if ( ( s.front() & 0xFF ) <= ' ' || ( s.back() & 0xFF ) <= ' ' )
```
DolceTriade, line 870: Should you just use something like `!iscntrl()` or `isprint()`?

slipher: The standard library functions are all unusable since they are supposedly locale-dependent (hence `Str::cislower` etc.).

I did this since it's the same thing that `SkipWhitespace` does

src/sgame/sg_admin.cpp
slipher: Done

src/sgame/sg_admin.cpp
DolceTriade: Got it. Maybe leave a comment about  what is going on here.

DolceTriade:
I assume this doesn't break backwards compat? Does this change the escaping behavior of this function for things like quotes?

slipher:
If you had funny stuff like quotes or line comments in the string value line, then it would just wiped out by the naive serialization of admin_writeconfig_string. Backwards compatibility is only broken in the implausible scenario that you have a custom hand-written admin.dat and never ever use ban or setlevel, so that the gamelogic doesn't rewrite the file.

DolceTriade:
IN that case, lgtm

src/sgame/sg_admin.cpp
slipher: Done.

DolceTriade (approved):

PR #2349, opened 2022-12-27
fix level name alignment in `listplayers` & `listadmins`, and add a legend to `listplayers`

cu-kai:
This PR fixes the broken alignment and additional padding in `listplayers`, as well as fixes the alignment in `listadmins`.

It also adds a variable legend to the `listplayers` command, as some of the hints are not immediately obvious to new admins.
The `setlevel` and `warn` description will only show if the admin has permission to see these hints in `listplayers` in the first place.

To fix the alignment, I change the `admin_level_maxname` variable to be the name length _including_ the colours, as we don't use this variable for any other reason other than to format these lists (and the method previously used was broken). This fixes the calculation of characters required to align the level names correctly.

Here is a comparison:
**BEFORE**
listplayers-alignment

listadmins-alignment

**AFTER**
listplayers-fixed

listadmins-fixed

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -1302,14 +1302,16 @@ static int admin_out( void *admin, char *str )
 	l = G_admin_level( a->level );
 
 	int lncol = Color::StrlenNocolor( l->name );
+	int namelen  = strlen( l->name );
 
 	if ( a->lastSeen.tm_mday )
 	{
 		trap_GetTimeString( lastSeen, sizeof( lastSeen ), "%Y-%m-%d", &a->lastSeen );
 	}
 
-	Com_sprintf( str, MAX_STRING_CHARS, "%-6d %*s %s %s",
-	             a->level, admin_level_maxname + lncol, l ? l->name : "(null)",
+	Com_sprintf( str, MAX_STRING_CHARS, "%-6d %*s ^*%s %s",
+	             a->level, namelen + ( admin_level_maxname - lncol ), 
+	             l ? l->name : "(null)",
 	             lastSeen, a->name );
 
 	return 0;
@@ -3877,6 +3879,7 @@ bool G_admin_listplayers( gentity_t *ent )
 		}
 
 		int colorlen = Color::StrlenNocolor( lname );
+		int namelen  = strlen( lname );
 
 		ADMBP( va( "%2i %s%c^7 %-2i^2%c^7 %*s^* ^5%c^1%c%c%s^7 %s^* %s%s%s %s",
 		           i,
@@ -3884,7 +3887,7 @@ bool G_admin_listplayers( gentity_t *ent )
 		           t,
 		           l ? l->level : 0,
 		           hint ? '*' : ' ',
-		           admin_level_maxname + colorlen,
+		           namelen + ( admin_level_maxname - colorlen ),
 		           lname,
 		           bot,
 		           muted,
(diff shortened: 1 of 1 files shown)


src/sgame/sg_admin.cpp
```diff
@@ -2014,7 +2012,7 @@ bool G_admin_readconfig( gentity_t *ent )
 			{
 				admin_readconfig_string( &cnf, l->name, sizeof( l->name ) );
 				// max printable name length for formatting
-				len = Color::StrlenNocolor( l->name );
+				len = strlen( l->name );
```
DolceTriade, line 2015: Why change this?

src/sgame/sg_admin.cpp
```diff
@@ -3897,6 +3893,15 @@ bool G_admin_listplayers( gentity_t *ent )
 		           ( !authed ) ? "^1NOT AUTHED" : "" ) );
 	}
 
+	ADMBP( va( "\n^3listplayers:^* legend:" ) );
+
+	if ( canset ){
```
DolceTriade, line 3898: nit style:
```
if ( canset )
{
    ...
}
```

src/sgame/sg_admin.cpp
cu-kai: Already explained in the initial comment, the previous code made no sense as it was adding the length of the decoloured level name in the listplayers entry to `admin_level_maxname` resulting in a double-length alignment in some cases; that code never worked properly.

I changed this because in every instance `admin_level_maxname` was used, it was being compared to another decoloured admin level name which is essentially a double negative. For formatting, this makes the most sense anyway as you'd never not want to display coloured admin levels in any relevant formatting.

The only two places this is used is `listplayers` and `listadmins`, where I have fixed the same issue in both commands with this PR.

src/sgame/sg_admin.cpp
cu-kai: fixed in https://github.com/Unvanquished/Unvanquished/pull/2349/commits/56ffb617c5896ffec35ba73491c22465259e4d14

src/sgame/sg_admin.cpp
slipher: The original code is right here. The change only appears to work because all the names in the examples are ASCII-only and have the same number of characters in color tokens.

src/sgame/sg_admin.cpp
```diff
@@ -1301,15 +1301,13 @@ static int admin_out( void *admin, char *str )
 
 	l = G_admin_level( a->level );
 
-	int lncol = Color::StrlenNocolor( l->name );
```
slipher, line 1304: What we want is to have `admin_level_maxname` non-color code points in the field. `Color::StrlenNocolor( l->name )` is the number of non-color code points in the name. So we want to add `admin_level_maxname - Color::StrlenNocolor( l->name )` spaces. So the width specifier should be bigger than the name's byte length by that amount. Therefore the fix is `lncol = strlen( l->name ) + ( admin_level_maxname - Color::StrlenNocolor( l->name ) )` (and revert the change to admin_level_maxname).

cu-kai (commented):
the reason for this change is to fix the following warning during compilation:

`src/sgame/sg_admin.cpp:1305:30: warning: ‘size_t strlen(const char*)’ reading 1 or more bytes from a region of size 0 [-Wstringop-overread]`

slipher:
It seems these "stringop" warnings are really broken and only give false positives - we already had to disable `stringop-overflow`. Just turn off the warning [here](https://github.com/Unvanquished/Unvanquished/blob/85dee939a83780b7fd790c6d1b707b89d24577c2/CMakeLists.txt#L153).

slipher:
LGTM

sweet235:
lgtm

PR #2352, opened 2023-01-03
More bot navigation-related cleanup

slipher:


src/sgame/botlib/bot_nav_edit.cpp
```diff
@@ -439,15 +439,12 @@ void Cmd_NavTest( gentity_t *ent )
 	}
 }
 
-void NavEditInit()
-{
-	memset( &cmd, 0, sizeof( cmd ) );
```
necessarily-equal, line 444: Is there an assurance somewhere that static-global variables get zero-initialised ? I know this is true for Linux elf and MacOS, but I'm not sure this is portable

necessarily-equal (approved):
LGTM with one question. Haven't tested.

DolceTriade (approved):

src/sgame/botlib/bot_nav_edit.cpp
slipher: C++ standard

Issue #2358, opened 2023-01-08
Bots try to walk straight through an open door pane


necessarily-equal:
Since https://github.com/Unvanquished/Unvanquished/pull/2259, bot don't try to "dodge" the door when walking trough it. This means that bots can get stuck around an already opened door, just like if they expected the door to open further.

![Screenshot from 2023-01-08 11-34-41](https://user-images.githubusercontent.com/59283660/211191553-aeb486f1-6f56-430a-b080-a44f909f613d.png)

necessarily-equal:
https://github.com/Unvanquished/Unvanquished/pull/2360

PR #2360, opened 2023-01-08
BotHandleDoor: fix some bugs and refactor

necessarily-equal:
This brings two functional changes:
1. IsAutomaticMover is only checked for *_POS1: you don't want bots to think they can open more an already opened door. This fixes https://github.com/Unvanquished/Unvanquished/issues/2358.
2. HandleDoor is now called after trap_LinkEntity, so ent->r.absmin (and max) should be more correct now

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -594,14 +594,41 @@ bool IsAutomaticMover( const gentity_t *ent )
 	return true;
 }
 
-void BotHandleDoor( gentity_t *ent )
+// This function adds obstacles for doors that the bot can't open,
+// that is doors that are not automatic, or doors that are already opened.
+void BotHandleDoor( gentity_t *ent, moverState_t moverState )
 {
-	if ( IsDoor( ent ) && !IsAutomaticMover( ent ) )
+	switch ( moverState )
 	{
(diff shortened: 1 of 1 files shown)


src/sgame/sg_spawn_mover.cpp
```diff
@@ -594,14 +594,41 @@ bool IsAutomaticMover( const gentity_t *ent )
 	return true;
 }
 
-void BotHandleDoor( gentity_t *ent )
+// This function adds obstacles for doors that the bot can't open,
+// that is doors that are not automatic, or doors that are already opened.
```
DolceTriade, line 598: Why are open doors obstacles?

necessarily-equal: Open door are obstacles because they are not going to open more if you get closer, so you need to avoid walking on the door pane if it's sticking out of the wall. See also https://github.com/Unvanquished/Unvanquished/issues/2358.

DolceTriade: Ah interesting.

DolceTriade:
lgtm

necessarily-equal:
This brings a regression for rund "iris" doors like in spacetracks, cutting off the navmesh when it the door is open.

image

That means that if you stand behind the door you are effectively blocking the only bot path between the two bases.

PR #2363, opened 2023-01-09
Allow setting g_BPInitialBudget to a value per team, in one cvar

necessarily-equal:
Setting "g_BPInitialBudget 10,50" will give 10 bp to the first team (aliens) and 50bp to the second (humans).

If the list isn't long enough, the last value will be used. This means that this is backward compatible as if there is only one value, it will be used for every team.

This approach has the advantage of extending to more teams should we add support for that; and doesn't need a change in netradiant for each team we add either.

### DIFF
--- a/src/sgame/sg_buildpoints.cpp
+++ b/src/sgame/sg_buildpoints.cpp
@@ -23,6 +23,7 @@ along with Unvanquished. If not, see <http://www.gnu.org/licenses/>.
 */
 
 #include "sg_local.h"
+#include "shared/parse.h"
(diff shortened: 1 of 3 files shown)


DolceTriade:
imo, we should just create different cvars as that's more clear. This is just more confusing. I don't think we'll need to add more teams often enough for this.

necessarily-equal:
Let's keep https://github.com/Unvanquished/Unvanquished/pull/2332 instead. I think people like it more, and I think there is another argument in favor of it: it's easier to make UI for it.

PR #2365, opened 2023-01-10
Remove dead code for favorite servers

slipher:
Companion to https://github.com/DaemonEngine/Daemon/pull/775

### DIFF
--- a/src/cgame/cg_rocket_datasource.cpp
+++ b/src/cgame/cg_rocket_datasource.cpp
@@ -244,83 +244,74 @@ void CG_Rocket_BuildServerList( const char *args )
 	rocketInfo.serversLastRefresh = rocketInfo.realtime;
 	rocketInfo.currentNetSrc = netSrc;
 
-	if ( netSrc != AS_FAVORITES )
+	int numServers;
+
+	rocketInfo.data.retrievingServers = true;
+
+	Rocket_DSClearTable( "server_browser", args );
+	CG_Rocket_CleanUpServerList( args );
+
+	trap_LAN_MarkServerVisible( netSrc, -1, true );
+
+	numServers = trap_LAN_GetServerCount( netSrc );
+
+	// Still waiting for a response...
+	if ( numServers == -1 )
 	{
-		int numServers;
-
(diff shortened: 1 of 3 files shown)


DolceTriade:
We should remove the favorite code from the master server too if we want to do this.

slipher:
> We should remove the favorite code from the master server too if we want to do this.

Maybe you are thinking of "featured" servers. Favorite servers were the user's personal favorites stored in a file in the homepath.

DolceTriade:
Ah, both should be removed then :D this lgtm

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -427,7 +418,7 @@ void CG_Rocket_CleanUpServerList( const char *table )
 	int j;
 	int netSrc = CG_StringToNetSource( table );
 
-	for ( i = AS_LOCAL; i <= AS_FAVORITES; ++i )
+	for ( i = AS_LOCAL; i <= AS_NUM_TYPES; ++i )
```
necessarily-equal, line 421: Is this on purpose or did you mean this instead?
```suggestion
	for ( i = AS_LOCAL; i < AS_NUM_TYPES; ++i )
```

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -244,83 +244,74 @@ void CG_Rocket_BuildServerList( const char *args )
...
+		if ( rocketInfo.data.retrievingServers )
+		{
+			rocketInfo.data.retrievingServers = false;
 		}
-
-	}
-
-	else if ( !Q_stricmp( args, "serverInfo" ) )
-	{
-		CG_Rocket_BuildServerInfo();
```
necessarily-equal, line 323: Was this deletion intended?

necessarily-equal (commented):
Not sure why the build fails. I've made comments for two oddities

src/cgame/cg_rocket_datasource.cpp
slipher: Thanks, fixed

PR #2367, opened 2023-01-10
add worldspawn strings for per-team BP budgets

cu-kai:
Addendum to #2332 
This commit allows the new per team BP budget cvars to be set by worldspawn keys in the BSP.

The documentation for NetRadiant needs to be updated, but for now, this works.


necessarily-equal (approved):
LGTM, although I haven't tested.

cu-kai:
I tested this with a testing map and it worked fine.

Issue #2337, opened 2022-12-13
Navmesh debug draw is unreliable and can draw lines between wrong elements


necessarily-equal:
See the screenshots in https://github.com/Unvanquished/Unvanquished/issues/1448 for some examples.

slipher:
Should be fixed by https://github.com/DaemonEngine/Daemon/pull/925.

necessarily-equal:
Thanks!

PR #2368, opened 2023-01-11
Add obstacle debug draw

necessarily-equal:
This is a cleanup version of the code I used to debug bots' obstacles. I found it to be very useful.

Screenshots:
Screenshot from 2023-01-08 12-33-23
Screenshot from 2023-01-10 10-06-34
Screenshot from 2022-11-06 14-44-51

### DIFF
--- a/src/sgame/botlib/bot_nav_edit.cpp
+++ b/src/sgame/botlib/bot_nav_edit.cpp
@@ -130,6 +130,65 @@ static void DrawPath( Bot_t *bot, DebugDrawQuake &dd )
 	dd.depthMask(true);
 }
 
+// This draws a red box around every obstacle. Each face are crossed.
+// The high number of line doesn't look too ugly, (IMO) convey well the idea
+// that it is impossible to go there, are actually helpful because more lines
+// means it will be more likely that one line will be displayed right
+// (see https://github.com/Unvanquished/Unvanquished/issues/2337 for the bug)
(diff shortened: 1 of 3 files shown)


src/sgame/botlib/bot_nav_edit.cpp
```diff
@@ -160,6 +219,14 @@ void BotDebugDrawMesh()
 		duDebugDrawNavMeshPortals( &dd, *cmd.nav->mesh );
 	}
 
+	//if ( cmd.showobstacles )
```
necessarily-equal, line 222: I'm not sure if there is any point making this optional. I think it's fine as is.

necessarily-equal:
To test, you need to use e.g. `/navedit enable level0` after a /devmap.

slipher:
Why does it draw a different shape in the first screenshot than the other two?

necessarily-equal:
That'd be https://github.com/Unvanquished/Unvanquished/issues/2337, there is a bug in the renderer that makes some lines disappear or that makes some lines between wrong vertex. Sometime you need a bit of patience to be sure. This is also why this is drawing so many lines, to have more chances some will be drawn correctly 😅

src/sgame/botlib/bot_nav_edit.cpp
DolceTriade: Unless there is a severe performance impact, I think it's fine to make it on by default. If we find a performance impact, we can add the check back.

necessarily-equal:
Should I go ahead and merge this?

DolceTriade:
lgtm

PR #2369, opened 2023-01-12
Move gameplay value definitions from bg_gameplay.h to bg_gameplay.cpp

necessarily-equal:
This allows tweaking the gameplay without recompiling 153 files. In that sense, this is a great improvement.

### DIFF
--- a/src/sgame/components/RocketpodComponent.cpp
+++ b/src/sgame/components/RocketpodComponent.cpp
@@ -3,8 +3,6 @@
 
 #include <glm/geometric.hpp>
 
-constexpr float ATTACK_RANGE         = (float)ROCKETPOD_RANGE; // cgame needs to know this.
-constexpr int   ATTACK_PERIOD        = ROCKETPOD_ATTACK_PERIOD; // cgame needs to know this.
 constexpr int   SHUTTER_OPEN_TIME    = 1000;
 constexpr int   TARGET_SEARCH_PERIOD = 500;
 constexpr int   LOCKON_TIME          = 500;
@@ -17,9 +15,9 @@ RocketpodComponent::RocketpodComponent(Entity& entity, HumanBuildableComponent& 
 	, lastTargetSearch(-TARGET_SEARCH_PERIOD)
 	, openingShuttersSince(-SHUTTER_OPEN_TIME)
 	, lockingOnSince(-LOCKON_TIME)
-	, lastShot(-ATTACK_PERIOD)
+	, lastShot(-ROCKETPOD_ATTACK_PERIOD)
 {
-	GetTurretComponent().SetRange(ATTACK_RANGE);
+	GetTurretComponent().SetRange(ROCKETPOD_RANGE);
 
 	REGISTER_THINKER(Think, ThinkingComponent::SCHEDULER_BEFORE, 0);
 }
@@ -131,7 +129,7 @@ void RocketpodComponent::Think(int timeDelta) {
 				// The lockon timer has expired and it's safe to shoot, so do so.
 				firing = true;
 
-				if (lastShot + ATTACK_PERIOD <= level.time) {
+				if (lastShot + ROCKETPOD_ATTACK_PERIOD <= level.time) {
 					Shoot();
 				}
 			} else if (!lockingOn) {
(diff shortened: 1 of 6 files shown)


necessarily-equal:
One thing I'm unsure about is that comments are now essentially duplicated between the .h and .cpp file. Maybe I should delete half of them so that it's easier to keep them in sync?

slipher:
Is the compiling thing actually important? How often do you need to change gameplay values really?

necessarily-equal:
> Is the compiling thing actually important? How often do you need to change gameplay values really?

When you are tweaking values, very often I would say. We don't currently tweak values a often, I'll admit, but this is rather a subject of complains (If not the most frequent subject of complains).

I would like to make tweaking the gameplay easier.

DolceTriade:
am ok with the change. Agree that it is helpful when iterating gameplay values to not recompile the entire cgame/sgame.

src/sgame/components/RocketpodComponent.cpp
```diff
@@ -3,11 +3,11 @@
 
 #include <glm/geometric.hpp>
 
-constexpr float ATTACK_RANGE         = (float)ROCKETPOD_RANGE; // cgame needs to know this.
-constexpr int   ATTACK_PERIOD        = ROCKETPOD_ATTACK_PERIOD; // cgame needs to know this.
-constexpr int   SHUTTER_OPEN_TIME    = 1000;
-constexpr int   TARGET_SEARCH_PERIOD = 500;
-constexpr int   LOCKON_TIME          = 500;
+static const float &ATTACK_RANGE        = ROCKETPOD_RANGE; // cgame needs to know this.
```
slipher, line 6: This reference stuff is kinda ugly; how about just replacing ATTACK_RANGE with ROCKETPOD_RANGE everywhere instead of proxying through another constant?

Or even better, don't put it in a global header at all; define the constant in this file only.

necessarily-equal: > Or even better, don't put it in a global header at all; define the constant in this file only.

That can't be done, cgame needs to be aware of the value.

> How about just replacing ATTACK_RANGE with ROCKETPOD_RANGE everywhere […]?

I could, although then the constants are in different places.

slipher: Let's go with replacing ATTACK_RANGE with ROCKETPOD_RANGE everywhere, it sucks to have multiple names for the same constant anyway.

necessarily-equal:
rebased

slipher (commented):
Reviewed this a few days ago but forgot to publish the comments apparently...

The non-comment parts look OK

src/shared/bg_gameplay.cpp
```diff
@@ -0,0 +1,294 @@
...
+ * process. The second downside is that this requires you to deal with .cfg
+ * file packaging.
+ *
+ * Cvars have the upside of allowing runtime edits. The process of adding one
+ * is more complicated: after creating the cvar with the SERVERINFO, you must
+ * parse it in cgame (see g_devolveMaxBaseDistance for an example).
+ * Use this solution if you expect server admins to want to change this value.
+ * This solution has downsides too of this method is that large scale
+ * modifications are pretty unpractical and one looses the ability to have
+ * proper tools such as a text editor and git, or the ability to personal
```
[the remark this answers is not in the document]
necessarily-equal: Modified, hopefully it should be more clear

src/shared/bg_gameplay.cpp
```diff
@@ -0,0 +1,294 @@
...
+ * comments on the values.
+ */
+
+
+
+/*
+ * Common constants for the build weapons for both teams.
+ */
+
+// Pressing +deconstruct for less than this many milliseconds toggles the deconstruction mark
```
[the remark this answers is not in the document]
necessarily-equal: I've removed the line comments in the .cpp, although I kept the structural comments

slipher:
LGTM

bmorel:
> We don't currently tweak values a often

Upstream does not, and it's why gameplay very rarely moves. Modders do though. Thanks for the change.

PR #2372, opened 2023-01-16
tactic command: show allowed behaviors, group functions in header

sweet235:
This shows the allowed behaviors from `g_tacticBehaviors` in the usage message. Without this, there is no way for a user to find out what the allowed behaviors are.

PR also groups the tactic functions in bg_public.h.

### DIFF
--- a/src/shared/bg_public.h
+++ b/src/shared/bg_public.h
@@ -1717,9 +1717,9 @@ bool BG_BuildableDisabled( int buildable );
 
 weapon_t BG_PrimaryWeapon( int const stats[] );
 
-void BG_SetTacticBehaviors(std::string tacticCsv);
-bool BG_TacticBehaviorAllowed(std::string behavior);
-std::string BG_TacticBehaviorsToString(std::string sep);
+void BG_SetTacticBehaviors( Str::StringRef tacticCsv );
+bool BG_TacticBehaviorAllowed( Str::StringRef behavior );
+std::string BG_TacticBehaviorsToString( Str::StringRef sep );
 
 // bg_voice.c
 #define MAX_VOICES             8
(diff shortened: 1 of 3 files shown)


cu-kai (approved):
Seems like a sensible implementation to me, previously it was not obvious which tactics were available.

I think once this is implemented we should document this on the wiki somewhere, whether in an existing page or by creating a new one, that way server owners can configure this without needing to look at the code.

In the long term perhaps the list of available tactics could be made available to the client so we can have a real menu for it.

sweet235:
Let me explain what this PR does. If i join a team and enter `/tactic`, i see something like:

```]/tactic 
tactic: usage: tactic <behavior> [how-many-bots] 
   <behavior> is one of: camper, default, defend
```

Without this PR, I only see:
```]/tactic 
tactic: usage: tactic <behavior> [how-many-bots] 
```

src/shared/bg_public.h
```diff
@@ -1709,18 +1709,18 @@ void BG_SetForbiddenEquipment(std::string forbidden_csv);
...
 bool BG_ClassDisabled( int class_ );
 bool BG_BuildableDisabled( int buildable );
 
-bool BG_TacticBehaviorAllowed(std::string behavior);
-
 weapon_t BG_PrimaryWeapon( int const stats[] );
 
+void BG_SetTacticBehaviors(std::string tacticCsv);
+bool BG_TacticBehaviorAllowed(std::string behavior);
+std::string BG_TacticBehaviorsToString(std::string sep);
```
DolceTriade, line 1722: nit: Use `Str::StringRef` to take input strings

DolceTriade (approved):

DolceTriade (approved):

PR #2374, opened 2023-01-16
cgame: better viewpos command

illwieckz:
- also print the pitch  
something I needed for years
- do a float round and not an in truncation  
make sure numbers printed by `viewpos` is the same as the one set with `setviewpos`

Before printing the pitch:

```
]/viewpos
(0 2688 192) : -90
```

Before doing a float round:

```
]/viewpos
-1013 -378 130 -47 -30
]/setviewpos -1013 -378 130 -47 -30
]/viewpos
-1013 -378 130 -46 -29
```

### DIFF
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -62,9 +62,9 @@ Debugging command to print the current position
 */
 static void CG_Viewpos_f()
 {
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

slipher:
LGTM

PR #2380, opened 2023-01-21
cgame: read map name from server config string instead of cvar, fix #2376

### LINKED ISSUE #2376: Map loading screen displays incomplete information when connecting to a server
I probably put the code to fetch the data too early in map load code to get it while connecting to a server that is not itself.

Loading a map with a client :

20230117-211633-000 unvanquished

Joining a server with a client:

20230117-211757-000 unvanquished

illwieckz:
cgame: read map name from server config string instead of cvar, fix #2376

- cvar is only reliable when hosting map with the client
- server config string also work when joining a server with a client

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1198,7 +1198,13 @@ static void SetMapInfoFromArenaInfo( arenaInfo_t arenaInfo )
 
 static void SetMapInfo()
 {
-	std::string mapName = Cvar::GetValue( "mapname" );
+	/* When SetMapInfo() is called,
+
+	- cgs.mapname may not work because it is not set yet.
+	- Cvar::GetValue( "mapname" ) may not work because
+	  only client used as a server has it set. */
+
+	std::string mapName = Info_ValueForKey( CG_ConfigString( CS_SERVERINFO ), "mapname" );
 
 	/* Use longname from .arena file as map name if exists,
 	also read author name if exists. */


src/cgame/cg_main.cpp
```diff
@@ -1198,7 +1198,7 @@ static void SetMapInfoFromArenaInfo( arenaInfo_t arenaInfo )
 
 static void SetMapInfo()
 {
-	std::string mapName = Cvar::GetValue( "mapname" );
```
slipher, line 1201: There's `cgs.mapname` you can use

illwieckz: The map loading screen is painted before `cgs.mapname` is set and we better want to have the name displayed as soon as possible to not see the map load screen for some time without text, then see the text appearing.

illwieckz:
Note: The fact the cvar value is not set when the client is not also a server may be a bug, what do you think about it?

src/cgame/cg_main.cpp
illwieckz: For reference the code to display the levelshot also uses `Info_ValueForKey( CG_ConfigString( CS_SERVERINFO ), "mapname" )`, I assume it's also to display it as soon as possible:

https://github.com/Unvanquished/Unvanquished/blob/756925e9ec1a4d31e680511423f2ba3609057ba0/src/cgame/cg_rocket_draw.cpp#L1797

DolceTriade:
Agree, if we need to read the cvar earlier, we should set it earlier.

illwieckz:
> Agree, if we need to read the cvar earlier, we should set it earlier.

@DolceTriade The cvar is never set at all when joining a server. I don't know if it's a bug or not, but if it's a bug I would consider this another topic than what this PR tries to fix anyway. And there is high chance this is not a bug. I created a specific thread for that topic: #2381

illwieckz:
I see no other way to implement this since `cgs.mapname` isn't usable yet when the map load screen starts to paint (or we would have to delay the painting of the map load screen after the game reached some initialization steps), and cvar is simply not set and then not meant to be used for that.

PR #2383, opened 2023-01-25
In-game navmesh generation final touches

slipher:


illwieckz:
See https://github.com/DaemonEngine/daemonmap/pull/13/files for better cell height adjustment code that was known to work with anthill.

illwieckz (approved):
LGTM

PR #2384, opened 2023-01-29
Last-minute fixes

slipher:
Most importantly:
- Fix `listbots`
- Disable chat bubble (see commit message for explanation)

### DIFF
--- a/src/cgame/cg_servercmds.cpp
+++ b/src/cgame/cg_servercmds.cpp
@@ -1193,7 +1193,7 @@ CG_Print_f
 */
 static void CG_Print_f()
 {
-	Log::Notice( CG_Argv( 1 ) );
+	Log::CommandInteractionMessage( CG_Argv( 1 ) );
 }
 
 /*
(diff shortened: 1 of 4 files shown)


DolceTriade:
I disagree with the chat bubble change. Maybe the "chat bubble" is inaccurate, but the point of the chat bubble is to explain to people that the person is unable to attack and they can do with that information what they want.

slipher:
The purpose of the chat bubble is to suggest that it might be polite not to attack someone. But there is absolutely nothing wrong with attacking someone using the armory.

DolceTriade:
I mean...maybe, but it is true that they are defenseless in that situation. Imo, this is hardly a reason to remove it all together though.

slipher:
We should fix it by activating on the chat menu and the big options menu but not other menus. It seems a little complicated though so I'm doing a quick fix for now.

illwieckz:
I may prefer to still having it even if wrong sometime, but I'm fine with merging it. I'll remove the mention of this feature in the blog post. I'm also concerned by disclosing a gameplay information anyway. Until we fix this people can revert this in their mod if they prefer having it on their servers.

illwieckz (approved):
LGTM

slipher:
> I'm also concerned by disclosing a gameplay information anyway.

Hahaha I didn't even think of that. You can use the chat bubble to know when is the perfect time to attack the human in their base. Could be good for ckit users as well as armoury users.

DolceTriade:
Sure. We can add it back as a cvar after release and some deeper looks at how to distinguish between different types of mens.

PR #2385, opened 2023-01-29
Invalidate navmesh cache when map version has changed

slipher:
I have tested it with dpk and dpkdir packages.


necessarily-equal:
On principle LGTM, haven't tested

illwieckz (approved):
I've tested it with dpk, it works.

LGTM

PR #2386, opened 2023-01-29
Add entity debug field

necessarily-equal:
This will allow for a better implementation of https://github.com/Unvanquished/Unvanquished/pull/2270

This has [a daemon counterpart too](https://github.com/DaemonEngine/Daemon/pull/790), because the intent was to modify entityState_t, and apparently entityState_t has to be kept in sync with playerState_t, and entityState_t lies in daemon for now. Yep.


necessarily-equal:
Marking as draft to avoid merging by accident before 0.55.

This will also need to be merged against 0.55.0/sync, when the time comes.

DolceTriade:
I will close this for now, since I don't think this will be merged any time soon. We can reopen when its rebased

PR #2388, opened 2023-01-30
Force "BOT_H_BUY_ARMOR" skill enabled for sk5+ bots

necessarily-equal:
This is just too impairing otherwise


slipher:
LGTM

IIRC I raised this point on the original skill PR :stuck_out_tongue:

cu-kai:
Yes, this predictably caused me problems on bunker, when I had a human team consisting of flamer/luci + no armour at all.

PR #2396, opened 2023-01-31
Improve bot trace navigation

necessarily-equal:
This PR should help bot better navigate despite imperfect navmesh or crowding. It is not a valid replacement for good quality navmesh (for example this doesn't fix UTCS' boxes and guardrail), but should help nonetheless.

See individual commits and code comments for explanations.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -590,13 +590,13 @@ static void BotFindDefaultSteerTarget( gentity_t *self, glm::vec3 &dir )
 
 	// We add a delta determined from clientNum to avoid bots
 	// dancing in sync
-	int time = (110 * self->num() + self->client->time1000) % 1000;
+	int time = (110 * self->num() + self->client->time10000) % 1600;
 
-	// 280 instead of 700/2=350 because we want some left/right
+	bool invert = time < 800;
+	// part of the time we go straight instead
+	// 600 instead of 800 because we want some left/right
 	// asymetry in case it helps
-	bool invert = time < 280;
-	// half of the time we go straight instead
-	bool sideways = time > 700;
+	bool sideways = std::abs(time - 600) < 500;
 
 	if ( sideways )
 	{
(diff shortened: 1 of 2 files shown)


src/sgame/sg_bot_nav.cpp
```diff
@@ -797,10 +804,32 @@ Global Bot Navigation
 bool BotMoveToGoal( gentity_t *self )
 {
 	glm::vec3 dir;
-	if ( BotAvoidObstacles( self, dir ) )
+	constexpr int ignoreGeometryThreshold = 1700; // 1.7s, we seem stuck
```
slipher, line 807: Note that stuck time is only recalculated once per second, so it seems a little fishy to use such a low threshold for something.

src/sgame/sg_bot_nav.cpp
```diff
@@ -590,13 +590,13 @@ static void BotFindDefaultSteerTarget( gentity_t *self, glm::vec3 &dir )
 
 	// We add a delta determined from clientNum to avoid bots
 	// dancing in sync
-	int time = (110 * self->num() + self->client->time1000) % 1000;
+	int time = (110 * self->num() + self->client->time10000) % 1600;
 
-	// 280 instead of 700/2=350 because we want some left/right
+	// 600 instead of 800 because we want some left/right
```
slipher, line 595: Comment doesn't match the code which does 800 and symmetrical

necessarily-equal: it's 600 on the line below, but I've just moved the comment around for clarity

src/sgame/sg_bot_nav.cpp
necessarily-equal: Well, it seems to work well enough like that. I guess that would mean 1.7 to 2.7s then, but I don't think that's a problem.

slipher: Maybe you could change stuck time to be recalculated every 500ms or something so that it would be more reliable.

LGTM

Issue #2399, opened 2023-01-31
possibly wrong lastCombatTime update


ghost:
https://github.com/Unvanquished/Unvanquished/blob/1449f3c9bca3787462a8dc997700916020c43c1d/src/sgame/components/HealthComponent.cpp#L148

This line really only updates the entity receiving damages. I'm pretty sure the intent is to update both attacker and defender's values.

bmorel:
A quick test shows that there is no delay when taking or receiving damages from a building. Need to test with AIs to see what is the current behavior and then decide if this is really what is expected

slipher:
This was resolved with #2875.

PR #2401, opened 2023-01-31
More correctness in lastCombatTime update

necessarily-equal:
Found by bmorel
Fixes https://github.com/Unvanquished/Unvanquished/issues/2399.


slipher:
Not all cases of giving damage really correspond to being "in combat". For example poison and ground flames. In particular it would not make sense that these can make you be "in combat" after you have already died.

necessarily-equal:
Good point, I didn't think it was a problem but that would be a regression indeed

necessarily-equal:
Maybe we should just promote the current behavior to "feature" and forget about it?

slipher:
> Maybe we should just promote the current behavior to "feature" and forget about it?

Works for me

bmorel:
In that case, one of the affectations can be removed, or someone might wonder again why the heck the same variable is affected a value twice, with different names...

slipher:
Cleaning up the double assignment of lastCombatTime in #2875.

Issue #2389, opened 2023-01-30
Extra navmesh parts in Chasm base


necessarily-equal:
```
/devmap chasm
/navedit level3upg
```

![Screenshot from 2023-01-30 09-50-51](https://user-images.githubusercontent.com/59283660/215430971-43eac8e6-7d94-4598-ab72-7d3e953848ec.png)
![Screenshot from 2023-01-30 09-50-44](https://user-images.githubusercontent.com/59283660/215430976-d7e3bb71-96fb-485f-9a50-a058c5f846ad.png)

These connections shouldn't exist, because otherwise bots will get stuck trying to go through them.

I suspect this is a consequence of disabling the patch blocks here: https://github.com/Unvanquished/Unvanquished/blob/master/src/shared/navgen/nav.cpp#L313. @slipher do you want to have a shot at porting this?

slipher:
OK, I'll take a look at porting that bit

bmorel:
As I said on IRC this morning, I remember when I still had the embedded navgen in my codebase that aliens did not had problems moving out of their base, and chasm is the map I had a layout to prove a pair of bugs in this mechanism.

So, I do not think the problem is that old. But maybe my memory does not serves me well.

necessarily-equal:
Should be fixed by https://github.com/Unvanquished/Unvanquished/pull/2402

PR #2402, opened 2023-02-01
Navgen: use BSP patches

slipher:
The navmesh in the location under the Chasm pipes as shown in #2389 indeed looks different when using the patches. I haven't attempt to test whether bot navigation is really better.

On most maps the number of triangles increases only modestly, but on Perseus it increases 72%. Maybe that would be a good candidate to set up a configuration to ignore patches (assuming it doesn't actually fix the navigation bugs there).

### DIFF
--- a/src/shared/navgen/nav.cpp
+++ b/src/shared/navgen/nav.cpp
@@ -187,14 +187,21 @@ void NavmeshGenerator::LoadBSP()
 	{
 		((int*)header)[i] = LittleLong(((int*)header)[i]);
 	}
-	for (int lump: {LUMP_MODELS, LUMP_BRUSHES, LUMP_SHADERS, LUMP_BRUSHSIDES, LUMP_PLANES})
+	for (int lump: {LUMP_MODELS, LUMP_BRUSHES, LUMP_SHADERS, LUMP_BRUSHSIDES, LUMP_PLANES, LUMP_SURFACES, LUMP_DRAWVERTS})
 	{
+		if ((lump == LUMP_SURFACES || lump == LUMP_DRAWVERTS) && !config_.generatePatchTris)
+		{
+			continue;
+		}
 		unsigned nBytes = header->lumps[lump].fileofs;
 		char *base = &mapData_[0] + header->lumps[lump].filelen;
 		for (unsigned i = 0; i < nBytes; i += 4)
 		{
 			if (lump == LUMP_SHADERS && i % sizeof(dshader_t) < offsetof(dshader_t, surfaceFlags))
 				continue;
+			// There is one more exception to 4-byte fields, byte color[4] in drawVert_t, but we don't use it
+#define color DO_NOT_USE_BYTE_SWAPPED
+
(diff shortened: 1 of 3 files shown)


necessarily-equal:
I've tested this improves the situation on UTCS, making boxes and guardrails obstacles again

src/shared/navgen/nav.cpp
```diff
@@ -308,27 +319,22 @@ void NavmeshGenerator::LoadBrushTris( std::vector<float> &verts, std::vector<int
...
-	char surfaceparm[16];
-
-	strcpy( surfaceparm, "default" );
-	ApplySurfaceParm( surfaceparm, &solidFlags, NULL, NULL );
+	if ( !config_.generatePatchTris )
+	{
+		return;
+	}
+	LOG.Debug( "Generated %d brush tris", tris.size() / 3 );
+	LOG.Debug( "Generated %d brush verts", verts.size() / 3 );
```
necessarily-equal, line 328: We might want to put this above (nit)

src/shared/navgen/nav.cpp
```diff
@@ -339,11 +345,12 @@ static void LoadPatchTris( std::vector<float> &verts, std::vector<int> &tris ) {
...
 
 	// calculate bounds of all verts
+	vec3_t mins, maxs;
 	rcCalcBounds( &verts[ 0 ], verts.size() / 3, mins, maxs );
 
 	// convert from recast to quake3 coordinates
-	recast2quake( mins );
-	recast2quake( maxs );
+	std::swap( mins[ 1 ], mins[ 2 ] );
+	std::swap( maxs[ 1 ], maxs[ 2 ] );
```
necessarily-equal, line 353: I like the `recast2quake` call better, but this is mostly cosmetic

necessarily-equal (approved):

src/shared/navgen/nav.cpp
slipher: The total number of tris is already logged elsewhere so it only provides non-duplicate information if patches are enabled

src/shared/navgen/nav.cpp
slipher: recast2quake currently lives in sgame and it didn't seem worth the restructuring effort

necessarily-equal: fair enough

bmorel:
what is the measured performance problem behind the 72% triangles increase?

slipher:
> what is the measured performance problem behind the 72% triangles increase?

Nothing, the map didn't take particularly long to generate that I noticed. In all likelihood it's not as serious problem.

illwieckz:
Before:

unvanquished_2023-02-02_033756_000

After:

unvanquished_2023-02-02_034136_000

Note: the fact there is no navmesh drawn around the egg is a rendering bug, on other frames there is one.

But I confirm there is no navmehs under the pipes.

bmorel:
Vega's alien base was also affected. Does this fixes it? (same for station 15 but describing where it happens is harder, and I can't do a screenshot now)

PR #2410, opened 2023-02-02
Replace bot "buy armor" skill by a "buy modern armor" skill

necessarily-equal:
This skill will buy armor up to the second best instead of not at all. This should help with the balancing a bit.

Superseeds https://github.com/Unvanquished/Unvanquished/pull/2388.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -557,6 +557,26 @@ AINodeStatus_t BotActionEvolve( gentity_t *self, AIGenericNode_t* )
 	return STATUS_FAILURE;
 }
 
+static bool BotSkillArmorIsDesirable( const gentity_t *self, upgrade_t armor )
+{
+	if ( self->botMind->botSkillSet[BOT_H_BUY_MODERN_ARMOR] ) {
+		return true;
+	}
+
+	// let's pretend that we have one less armor rank unlocked:
+	switch ( armor )
+	{
+	case UP_LIGHTARMOUR:
+		return BG_UpgradeUnlocked( UP_MEDIUMARMOUR );
+	case UP_MEDIUMARMOUR:
+		return BG_UpgradeUnlocked( UP_BATTLESUIT );
+	case UP_BATTLESUIT:
+		return false;
+	default:
+		ASSERT_UNREACHABLE();
+	}
+}
+
 // Allow human bots to decide what to buy
 // pre-condition:
 // * weapon is a valid pointer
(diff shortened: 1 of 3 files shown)


src/sgame/sg_bot_util.cpp
```diff
@@ -557,6 +557,28 @@ AINodeStatus_t BotActionEvolve( gentity_t *self, AIGenericNode_t* )
 	return STATUS_FAILURE;
 }
 
+// Ideally this function would be part of equipment_t<upgrade_t>::unlocked(),
```
slipher, line 560: Disagree with this comment since `unlocked` should just tell if the bot is legally allowed to have it. If a behavior tree requested buying it directly then `unlocked` should allow it even if there is no armor skill.

src/sgame/sg_bot_util.cpp
```diff
@@ -557,6 +557,28 @@ AINodeStatus_t BotActionEvolve( gentity_t *self, AIGenericNode_t* )
...
+	if ( self->botMind->botSkillSet[BOT_H_BUY_MODERN_ARMOR] ) {
+		return true;
+	}
+
+	// let's pretend that we have one less armor rank unlocked:
+	switch ( armor )
+	{
+	case UP_BATTLESUIT:
+		return false;
+	case UP_LIGHTARMOUR:
```
slipher, line 573: This could be split into two separate cases instead of doing some hacky math on the enum

src/sgame/sg_bot_util.cpp
```diff
@@ -557,6 +557,28 @@ AINodeStatus_t BotActionEvolve( gentity_t *self, AIGenericNode_t* )
...
+	case UP_MEDIUMARMOUR:
+		// check for the upgraded version instead
+		return BG_UpgradeUnlocked( Util::ordinal(armor) + 1 );
+	default:
+		return true;
```
slipher, line 578: ASSERT_UNREACHABLE?

src/sgame/sg_bot_util.cpp
necessarily-equal: hmm, ok, the line count change isn't significant enough I guess

src/sgame/sg_bot_util.cpp
necessarily-equal: I've removed the comment

src/sgame/sg_bot_util.cpp
necessarily-equal: done also

slipher:
LGTM

PR #2419, opened 2023-02-06
Use a config file to customize navgen options

slipher:
A configuration file can be placed in the VFS or in <homepath>/game/ that will
customize the values in NavgenConfig (the former daemonmap flags).
For example:
```
/* C or C++-style comments can be used in here since it's based on COM_Parse */
filtergaps 1
excludeCaulk 0
cellheight 2.5
```

The path `maps/$map.navcfg` is a placeholder; hopefully someone will suggest something better. Actually it would be nice to reorganize the paths somehow so that all of the cache stuff (`*.navMesh`) is not in the same directory as the config or non-cached data (navcons).

This is stacked on #2409

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1102,6 +1102,7 @@ static void GenerateNavmeshes()
 {
 	std::string mapName = Cvar::GetValue( "mapname" );
 	std::vector<class_t> missing;
+	NavgenConfig config = ReadNavgenConfig( mapName );
 	for ( class_t species : RequiredNavmeshes() )
 	{
 		fileHandle_t f;
@@ -1112,7 +1113,7 @@ static void GenerateNavmeshes()
 			continue;
 		}
 		NavMeshSetHeader header;
-		std::string error = GetNavmeshHeader( f, header, mapName );
+		std::string error = GetNavmeshHeader( f, config, header, mapName );
 		if ( !error.empty() )
 		{
 			Log::Notice( "Existing navmesh file %s can't be used: %s", filename, error );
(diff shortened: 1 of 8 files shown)


DolceTriade:
idk if this is overkill or trying to solve a problem that doesn't exist:

Should the mod or some other factor have a bearing on the config name? If for example, two different mods are sharing the same home path, it might cause problems. 

Secondly, do you think it is worth it to add these options per class?

Something like this:
```
global {
 option1 a
 option2 b
}

level1 {
 option1  c  // override option1 from the global config
  // option2 is inherited
}
```

slipher:
> Should the mod or some other factor have a bearing on the config name? If for example, two different mods are sharing the same home path, it might cause problems.

Well we currently have zero support of any kind for mod-specific homepath storage (keybinds etc.) and navgen doesn't seem like the most compelling place to start. In this specific case you could work around it by putting the nav configs in a mod-specific pak.



> Secondly, do you think it is worth it to add these options per class?

Let's ask the people who were using these options. @sweet235 @bmorel Have you ever need different daemonmap options depending on which alien class you are generating?

bmorel:
>  @bmorel Have you ever need different daemonmap options depending on which alien class you are generating?

Not on per class fashion, but with my lesser-auto-jump feature I noticed that maps might react correctly with different safety values, and the bigger the safety value, the more useful the feature is.
So this feature could be useful to upstream, if people implements the same feature, the same way I did.

sweet235:
> Have you ever need different daemonmap options depending on which alien class you are generating?

I never tried that. So I'm not sure if it would have been useful.

slipher:
Sounds like nobody needs per-class options for now. We can add it later if needed.

illwieckz:
Yes, I doubt per-class options can be useful, at least for the current available options, none are class-specific.

sweet235:
lgtm

Issue #2362, opened 2023-01-09
Flooding protection behaves differently when game is paused


sweet235:
It is more restrictive when the game is paused. I'm not sure if this is a feature. But I know that players tend to talk more when the game gets paused. The server makes it hard for them.

To reproduce, start a server and do this on a client with default level 0 permission:
```
]/say a; delay 1s say a; delay 2s say a; delay 3s say a; delay 4s say a 
[S] level0player: a 
[S] level0player: a 
[S] level0player: a 
[S] level0player: a 
[S] level0player: a
```
Then restart the server (to be sure), and do this:
```
pause: console paused the game. 
]/say a; delay 1s say a; delay 2s say a; delay 3s say a; delay 4s say a 
[S] level0player: a 
[S] level0player: a 
[S] level0player: a 
You are flooding: please wait 1s before trying again 
You are flooding: please wait 3s before trying again
```

Edit:
In practice, the player talking an admin into pausing the game will be censored even more, as they will have some recent talking on their account already. This creates the unfortunate situation that they cannot explain what they expect to be fixed.

bmorel:
I can reproduce.
This is clearly a bug, as [this](https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_cmds.cpp#L4739-L4785) does not shows any direct relation with pause. Even if it was a feature, I would fail a lot to see the rationale behind it.
I guess this is the same kind of problems which led to make the auto pause of servers when starting a game opt-in instead of opt-out, and generates a bunch of other problems when changing the pace of the game.

Maybe you can work around the flood protection by changing the value of g_floodMinTime, which defaults to 2s.

necessarily-equal:
I agree this is a bug. Fixing this may be as simple as replacing level.time with another value, such as Sys::Milliseconds.

sweet235:
I noticed another problem: when you have permission for `pause`, but not `NOCENSORFLOOD`, the flooding protection will prevent you from unpausing.

necessarily-equal:
I'm not sure I understand the implications of this

Le 8 février 2023 10:57:27 GMT+01:00, sweet235 ***@***.***> a écrit :

DolceTriade:
> I noticed another problem: when you have permission for pause, but not NOCENSORFLOOD, the flooding protection will prevent you from unpausing.

Should be fixed by this change right?

> I'm not sure I understand the implications of this

[...]

PR #2430, opened 2023-02-07
Use Sys::Milliseconds instead of level.time for rate-limiting of chat messages

necessarily-equal:
Fixes https://github.com/Unvanquished/Unvanquished/issues/2362.


necessarily-equal:
CI failure is unrelated, compiles fine locally.

slipher:
Looks like you could get the "real level.time" by doing `level.time + level.pausedTime` (regardless of whether the game is paused). Would rather not start introducing a lot of real time calls since now everything consistently works based on level.time and friends.

sweet235:
LGTM

DolceTriade:
lgtm

PR #2434, opened 2023-02-09
Add cvars to control the bot skillset

sweet235:
Currently, all bot skillset skills are allowed. Add three cvars:

- `g_allowedSkillset`
- `g_skillsetBudgetAliens`
- `g_skillsetBudgetHumans`

`g_allowedSkillset` is a comma-separated list of allowed skills. `g_skillsetBudget{Aliens,Humans}` replace the magic numbers 61 and 40 in `sg_bot_skilltree.cpp`.

This will allow server owners to enable/disable certain skills. We will be able to add skills that fail on some maps, and simply disable them for these maps. The skilltree code can become a unified interface to program and control experimental bot behaviors.

For example: to give every skill to each bot, set `g_skillsetBudget{Aliens,Humans}` to 1000. This is easy to do when testing a new bot skill.

### DIFF
--- a/src/sgame/sg_bot_skilltree.cpp
+++ b/src/sgame/sg_bot_skilltree.cpp
@@ -29,7 +29,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 // aliens have 61 points to spend max
 // humans have 40 points to spend max
 Cvar::Cvar<int> g_skillsetBudgetAliens( "g_skillsetBudgetAliens", "the skillset budget for aliens.", Cvar::NONE, 61 );
-Cvar::Cvar<int> g_skillsetBudgetHumans( "g_skillsetBudgetHumans", "the skillset budget for humans.", Cvar::NONE, 42 );
+Cvar::Cvar<int> g_skillsetBudgetHumans( "g_skillsetBudgetHumans", "the skillset budget for humans.", Cvar::NONE, 40 );
 
 static std::set<std::string> bg_allowedSkillset;
 
--- a/src/sgame/sg_bot_skilltree.cpp
+++ b/src/sgame/sg_bot_skilltree.cpp
@@ -38,7 +38,7 @@ static void G_UpdateSkillsets()
 	}
 }
 
-static std::set<std::string> G_ParseSkillsetList( const std::string &skillsCsv )
+static std::set<std::string> G_ParseSkillsetList( Str::StringRef skillsCsv )
 {
 	std::set<std::string> skills;
 
--- a/src/sgame/sg_bot_skilltree.cpp
+++ b/src/sgame/sg_bot_skilltree.cpp
@@ -26,6 +26,11 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "sg_bot_util.h"
 #include "../shared/parse.h"
 
+// aliens have 61 points to spend max
+// humans have 40 points to spend max
+Cvar::Cvar<int> g_skillsetBudgetAliens( "g_skillsetBudgetAliens", "the skillset budget for aliens.", Cvar::NONE, 61 );
+Cvar::Cvar<int> g_skillsetBudgetHumans( "g_skillsetBudgetHumans", "the skillset budget for humans.", Cvar::NONE, 42 );
+
 static std::set<std::string> bg_allowedSkillset;
 
 static std::set<std::string> BG_ParseAllowedSkillsetList( const std::string &behaviorsCsv )
@@ -195,9 +200,8 @@ static Util::optional<botSkillTreeElement_t> ChooseOneSkill(const gentity_t *bot
 std::pair<std::string, skillSet_t> BotDetermineSkills(gentity_t *bot, int skill)
 {
 	std::vector<botSkillTreeElement_t> possible_choices = initial_unlockable_skills;
-	// aliens have 61 points to spend max
-	// humans have 40 points to spend max
-	float max = G_Team(bot) == TEAM_ALIENS ? 61.0f : 40.0f;
+
+	float max = G_Team(bot) == TEAM_ALIENS ? static_cast<float>( g_skillsetBudgetAliens.Get() ) : static_cast<float>( g_skillsetBudgetHumans.Get() );
 
 	// unlock every skill at skill 7
 	int skill_points = static_cast<float>(skill) / 7.0f * max;
--- a/src/sgame/sg_bot_skilltree.cpp
+++ b/src/sgame/sg_bot_skilltree.cpp
@@ -33,11 +33,11 @@ Cvar::Cvar<int> g_skillsetBudgetHumans( "g_skillsetBudgetHumans", "the skillset 
 
 static std::set<std::string> bg_allowedSkillset;
 
-static std::set<std::string> BG_ParseAllowedSkillsetList( const std::string &behaviorsCsv )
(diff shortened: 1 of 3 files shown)


src/sgame/sg_bot_skilltree.cpp
```diff
@@ -24,6 +24,44 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+{
+	bg_allowedSkillset = BG_ParseAllowedSkillsetList( behaviorsCsv );
+}
+
+static bool BG_SkillAllowed( Str::StringRef behavior )
+{
+	return bg_allowedSkillset.find( behavior ) != bg_allowedSkillset.end();
+}
+
+static Cvar::Callback<Cvar::Cvar<std::string>> g_allowedSkillset(
```
DolceTriade, line 58: Might it be better to do disabled skill set instead of allowed? This also means that we don't have to keep this in sync if we add new ones. Also if we decide that allowed skillset is better, let's at least generate this list dynamically.

sweet235: It is either a blacklist or a whitelist. After some discussion on IRC, I went for the whitelist. Creating the list dynamically would be nice, but it intimately tied to the two magic numbers right now.

DolceTriade: I can guarantee that this will be the reason why someone spends hours trying to figure out why their skill isn't working 😄 

I'm a vote in favor of a denylist but if others feel otherwise, I'll defer.

sweet235: I have no strong opinion on this.

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -24,6 +24,39 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
 #include "sg_bot_util.h"
+#include "../shared/parse.h"
+
+static std::set<std::string> bg_allowedSkillset;
```
necessarily-equal, line 34: I feel like there could be a constructor race condition here, in doubt I would recommend this:

```suggestion
static std::set<std::string>& G_GetAllowedSkillset()
{
	static std::set<std::string> allowedSkillset;
	return allowedSkillset;
}
```

And then replacing every `bg_allowedSkillset` with `G_GetAllowedSkillset`

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -24,6 +24,39 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+
+static std::set<std::string> BG_ParseAllowedSkillsetList( const std::string &behaviorsCsv )
```
necessarily-equal: This is sgame, so the convention is to use a `G_` prefix (nitpick)

```suggestion
static std::set<std::string> G_ParseAllowedSkillsetList( const std::string &behaviorsCsv )
```

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -24,6 +24,39 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+
+	for (Parse_WordListSplitter i(behaviorsCsv); *i; ++i)
+	{
+		skills.insert( *i );
+	}
+
+	return skills;
+}
+
+static void BG_SetAllowedSkillset( Str::StringRef behaviorsCsv )
```
necessarily-equal:
```suggestion
static void G_SetAllowedSkillset( Str::StringRef behaviorsCsv )
```

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -24,6 +24,39 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+{
+	bg_allowedSkillset = BG_ParseAllowedSkillsetList( behaviorsCsv );
+}
+
+static bool BG_SkillAllowed( Str::StringRef behavior )
```
necessarily-equal, line 53:
```suggestion
static bool G_SkillAllowed( Str::StringRef behavior )
```

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -24,6 +24,39 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+		skills.insert( *i );
+	}
+
+	return skills;
+}
+
+static void BG_SetAllowedSkillset( Str::StringRef behaviorsCsv )
+{
+	bg_allowedSkillset = BG_ParseAllowedSkillsetList( behaviorsCsv );
+}
```
necessarily-equal, line 51: I would recommend writing a `G_UpdateSkillsets` function that would loop over every bot and set back their skill level to the same level (using `BotSetSkillLevel( bot, bot->botMind->botSkill.level )`).

And then you can call it in this (`BG_SetAllowedSkillset`) function.

This would make the change reflected in the already added bots.

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -24,6 +24,44 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
 #include "sg_bot_util.h"
+#include "../shared/parse.h"
+
+// aliens have 61 points to spend max
+// humans have 40 points to spend max
+Cvar::Cvar<int> g_skillsetBudgetAliens( "g_skillsetBudgetAliens", "the skillset budget for aliens.", Cvar::NONE, 61 );
+Cvar::Cvar<int> g_skillsetBudgetHumans( "g_skillsetBudgetHumans", "the skillset budget for humans.", Cvar::NONE, 40 );
```
necessarily-equal, line 32: I would make them callback cvars too which would call `G_UpdateSkillsets` (see the other comment) on update. This would make the changes live.

necessarily-equal (changes requested):
You would need to loop over every bot here to update their skillset after the list is updated. And I think there are race conditions here in the current state.

I've added comments for all of these, here's a review

src/sgame/sg_bot_skilltree.cpp
sweet235: I changed to a a denylist. There are already several `g_disabled*` cvars, so this should follow this convention.

sweet235:
Summary of changes:

- Converted to a denylist, as recommended by Ishq. Reasons: there are already several `g_disabled...` cvars, and the default value is easier to set. Also, there will usually be more enabled skills than disabled skills. So this is more convinient to set.

- There is now a 4th cvar to specify a list of skills each bot shall receive unconditionally. I found this to be very useful during testing.

bmorel:
All requested changes have been made. I've read the patches too and I think this is a (very needed and ) good change. LGTM

bmorel (approved):

src/sgame/sg_bot_skilltree.cpp
```diff
@@ -24,6 +24,114 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
...
+	{
+		gentity_t *bot = &g_entities[ i ];
+		if ( bot->r.svFlags & SVF_BOT )
+		{
+			BotSetSkillLevel( bot, bot->botMind->botSkill.level );
+		}
+	}
+}
+
+static std::set<std::string> G_ParseSkillsetList( const std::string &skillsCsv )
```
DolceTriade, line 41: nit: Str::StringRef

Issue #2403, opened 2023-02-01
Navmesh generation problems in 0.54


sweet235:
I'll try to collect all problems with the ingame navmesh generator here. Maybe I'll split this into several issues later.

- [x] Use BSP Patches
  * https://github.com/Unvanquished/Unvanquished/issues/2389
  * https://github.com/Unvanquished/Unvanquished/pull/2402
- [ ] RC_MAX_LAYERS is hit
- [x] Tweaking of cell height would help
  * https://github.com/Unvanquished/Unvanquished/pull/2407
- [ ] Only partial navmeshes
- [x] Navmesh generation can take a long time
- [x] A local game seems to create navmeshes twice

### Use BSP Patches

Navgen should be blocked by curvy objects.

### RC_MAX_LAYERS is hit
In the past, some maps required increasing `RC_MAX_LAYERS` in `recastnavigation/Recast/Source/RecastLayers.cpp` to successfully create navmeshes. This "fix" suggested an investigation, which I never performed. But it did give us navmeshes that seemed to work.

One example for this is [map-transparency](https://users.unvanquished.net/~sweet/pkg/map-transparency_0%2Balpha2.dpk). I observe this when loading it:

```Generating navmesh for level0 
Generating navmesh for level1 
Generating navmesh for level2 
rcBuildHeightfieldLayers: layer overflow (too many overlapping walkable platforms). Try increasing RC_MAX_LAYERS. 
Generating navmesh for level2upg 
rcBuildHeightfieldLayers: layer overflow (too many overlapping walkable platforms). Try increasing RC_MAX_LAYERS. 
Generating navmesh for level3 
rcBuildHeightfieldLayers: layer overflow (too many overlapping walkable platforms). Try increasing RC_MAX_LAYERS. 
Generating navmesh for level3upg 
Generating navmesh for level4 
Generating navmesh for human_naked 
Generating navmesh for human_bsuit
```

The old navmeshes for this map were created using this in `RecastLayers.cpp`:
```
-static const int RC_MAX_LAYERS = 63;
+static const int RC_MAX_LAYERS = 255;
```

Other maps with this problem are: [map-hq-beta28](https://users.unvanquished.net/~sweet/pkg/map-hq-beta28_0%2B0.dpk), [map-arkadia](https://users.unvanquished.net/~sweet/pkg/map-arkadia_0%2B0.dpk).

### Tweaking of cell height would help
[...]

PR #2435, opened 2023-02-09
Auto navgen quality of life improvements

slipher:
Fix some of the issues touched on in #2403:
- Don't generate navmeshes twice if `g_bot_defaultFill` and `cg_navgenOnLoad` are used together
- Warn if `bot fill` or `bot add` is used in a local game map config which might make navmeshes be generated twice
- Don't require navmeshes on `bot fill 0`
- Improve the responsiveness of the server during navgen. It was not quite as "background" as intended and could hang the server for several seconds at a time.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -5871,12 +5871,13 @@ static bool BotFillCmd( gentity_t *ent, const Cmd::Args& args )
 		BotUsage( ent );
 		return false;
 	}
-	int count = 0;
-	if ( !Str::ParseInt( count, args[2].data() ) || count < 0 )
+	int count;
+	if ( !Str::ParseInt( count, args[2].data() ) )
 	{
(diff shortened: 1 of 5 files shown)


sweet235:
LGTM, however:

- I can see the occasional `Warn: Hitch: 1190 msec frame time?`, but this is definitely an improvement.
- There is still the doubled generation when a config file has `bot fill <non-zero>`. How can we prevent this?

slipher:
>     * I can see the occasional `Warn: Hitch: 1190 msec frame time`, but this is definitely an improvement.

BSP loading is not split across frames so this can still happen once at the beginning of navgen. It's a bug if it happens anywhere else.


>     * There is still the doubled generation when a config file has `bot fill <non-zero>`. How can we prevent this?

In this case there should be a warning printed about the possibility of double navgen. I decided fixing this is a non-goal because map configs in a local game are a rare use case.

PR #2436, opened 2023-02-09
Put useful info in bot pings

slipher:
- 998 if navmesh generation in progress
- skill level otherwise

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -400,6 +400,18 @@ Bot Thinks
 =======================
 */
 
+static void SetPing( gentity_t *self )
+{
+	if ( navMeshLoaded == navMeshStatus_t::GENERATING )
+	{
+		self->client->ps.ping = 998;
+	}
+	else
+	{
+		self->client->ps.ping = self->botMind->botSkill.level;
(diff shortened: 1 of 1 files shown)


illwieckz:
I believe this is more ugly than the ugliest hack I got refused.

As said in the other PR, we actually have a cvar for bot status as `slotBots("B")`. We can store there a letter (for example `G` for “Generating”) or a number (for example, to store a bot skill level) or any other symbol that is not `-` (this one already means the slot is not occupied by a bot).

slipher:
How is it more ugly than the existing hack of `rand() % 50 + 50`?

necessarily-equal:
If we are afraid the value might be interpreted elsewhere, we could also do 50+skillLevel. But this feels reasonably safe

illwieckz:
See this other PR that implements the slotBots thing:

- https://github.com/Unvanquished/Unvanquished/pull/2437

For display it also reuses the “Ping” column in scoreboard but this is only client specific: it's up to the client to decide where and how to display the information, but the server doesn't abuse ping to display things on client in the back of the client, and properly use an existing dedicated field for bot status that is reusable by any client (including third party ones).

PR #2437, opened 2023-02-09
bot: store bot skill and generation status in slotBots B cvar and display it on scoreboard (reuse ping cell)

illwieckz:
Better implementation than:

- https://github.com/Unvanquished/Unvanquished/pull/2436

Right now the data is stored in scoreboard in “Ping” column, but we can do other things like appending `(generating…)` to bot name or things like that.

### DIFF
--- a/src/cgame/cg_rocket_datasource.cpp
+++ b/src/cgame/cg_rocket_datasource.cpp
@@ -936,7 +936,28 @@ void CG_Rocket_BuildPlayerList( const char* )
 
 		Info_SetValueForKey( buf, "num", va( "%d", score->client ), false );
 		Info_SetValueForKey( buf, "score", va( "%d", score->score ), false );
-		Info_SetValueForKey( buf, "ping", va( "%d", score->ping ), false );
+
+		const char* B = Info_ValueForKey( CG_ConfigString( CS_SERVERINFO ),"B" );
+		const char bot_status = B[ score->client ];
+
+		// HACK: Display bot status in “Ping” column.
+		if ( bot_status == '-' )
+		{
+			// This player is not a bot.
+			Info_SetValueForKey( buf, "ping", va( "%d", score->ping ), false );
+		}
+		// Bot skill can be 0 while spawning, just before skill is set.
+		else if ( bot_status >= '0' && bot_status <= '9' )
+		{
+			// Bot skill.
+			Info_SetValueForKey( buf, "ping", va( "sk%c", bot_status ), false );
+		}
+		else
+		{
+			// Example: “G” is for “Generating navmeshes”.
+			Info_SetValueForKey( buf, "ping", va( "%c", bot_status ), false );
+		}
+
 		Info_SetValueForKey( buf, "weapon", va( "%d", score->weapon ), false );
 		Info_SetValueForKey( buf, "upgrade", va( "%d", score->upgrade ), false );
 		Info_SetValueForKey( buf, "time", va( "%d", score->time ), false );
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -1193,7 +1193,20 @@ void CalculateRanks()
 
 			if ( bot )
 			{
-				B[ clientNum ] = 'b';
+				if ( navMeshLoaded == navMeshStatus_t::GENERATING )
+				{
+					B[ clientNum ] = 'G';
+				}
+				else
+				{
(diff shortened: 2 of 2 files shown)


illwieckz:
If reusing the “Ping” column may still be considered a hack, the hack is only client-specific.

illwieckz:
See also on Mantis side:

- _unvquery: all non `-` slot in `B` key is bot_ [gitlab.com:Viech/Mantis!5](https://gitlab.com/Viech/Mantis/-/merge_requests/5)

illwieckz:
Example:

unvanquished_2023-02-09_125619_000

Bot skill is prefixed by `S`, we can do more complex things (add an emoticon, use colors…).

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -936,7 +936,27 @@ void CG_Rocket_BuildPlayerList( const char* )
...
+		// HACK: Display bot status in “Ping” column.
+		if ( bot_status == '-' )
+		{
+			// This player is not a bot.
+			Info_SetValueForKey( buf, "ping", va( "%d", score->ping ), false );
+		}
+		else if ( bot_status >= '0' && bot_status <= '9' )
+		{
+			// Bot skill.
+			Info_SetValueForKey( buf, "ping", va( "S%c", bot_status ), false );
```
necessarily-equal, line 952: Maybe we could set this to "sk%c", I think it would be more obvious

```suggestion
			Info_SetValueForKey( buf, "ping", va( "sk%c", bot_status ), false );
```

necessarily-equal (approved):
LGTM, haven't tried

slipher:
Disadvantages of this:
- You have to update the Mantis and Xonpress code
- Bot skill range is now locked in as never being higher than 9

What are the advantages? I don't see any. (You could also put different text in the ping column with my approach but I think `sk5` is ugly anyway and prefer just `5`).

illwieckz:
>     * You have to update the Mantis and Xonpress code

It's OK. Even the Mantis code was just too strict anyway.

>     * Bot skill range is now locked in as never being higher than 9

You can use `[a-z]` and others.

> What are the advantages? I don't see any. (You could also put different text in the ping column with my approach but I think `sk5` is ugly anyway and prefer just `5`).

We already a have a dedicated field to store per-bot status, so this field has an unbeatable advantage over any hack reusing something that is not bot status to store bot status…

illwieckz:
About Xonpress:

```
$ grep -r 'bot ='
engine.php:                $player->bot = $player->ping == 0;
```

We better not want to abuse ping to store something else than ping.

slipher:
> About Xonpress:
> 
> ```
> $ grep -r 'bot ='
> engine.php:                $player->bot = $player->ping == 0;
> ```
> 
> We better not want to abuse ping to store something else than ping.

Actually Xonpress would not see the fake ping... the engine (server) is the ultimate source of the ping info. All of the fake ping meddling is only visible on the in-game scoreboard; the server always uses the real ping when responding to `getstatus` requests.

But I would consider that this represents a real advantage: the `B` info string is visible to external status requests, while fake pings are only visible to connected clients. So that's a good reason to go with your approach.

illwieckz:
> But I would consider that this represents a real advantage: the B info string is visible to external status requests.

Yes, this is how it looks in XQF, for example, without any patch to actually process the botness of clients:

[unvanquished bot status](https://dl.illwieckz.net/b/xqf/shots/unvanquished-bot-status/20230211-070050-000.xqf-unvanquished-bot-status.png)

> Actually Xonpress would not see the fake ping... the engine (server) is the ultimate source of the ping info.

Oh we have two pings? Well the fake ping is less bad then… But I really prefer that bot status string.

src/cgame/cg_rocket_datasource.cpp

illwieckz:
This is how it now looks:

[bot skill](https://dl.illwieckz.net/b/unvanquished/shots/bot-skill/unvanquished_2023-02-11_072504_000.jpg)

The Mantis bot has been patched, and restarted, so it is ready for game servers running master.

illwieckz:
For displaying the generation in scoreboard, we can also display an emoticon or a simple “…” instead of a letter.

slipher:
You could use [this icon](https://fontawesome-v4.github.io/icon/spinner.html) like `"<icon>\xEF\x84\x90</icon>"`. But you might have some trouble if HTML escaping is done on these strings...

As I said earlier, personally I would prefer just the number like `5` rather than `sk5`. LGTM

PR #2442, opened 2023-02-10
Add more alien skills and balance the skill tree a bit (2/3)

necessarily-equal:
<del>This includes Sweet's https://github.com/Unvanquished/Unvanquished/pull/2434 PR for now too for ease of testing, I'll remove it before merging.</del>

### DIFF
--- a/src/sgame/sg_bot_local.h
+++ b/src/sgame/sg_bot_local.h
@@ -106,7 +106,8 @@ enum bot_skill
 {
 	// movement skills
 	BOT_B_BASIC_MOVEMENT, // doesn't do anything as of now
-	BOT_A_MARA_JUMP_ON_ATTACK,
+	BOT_A_SMALL_JUMP_ON_ATTACK, // small aliens (dretch and mantis) do jump from time to time when attacking, this helps dodging
+	BOT_A_MARA_JUMP_ON_ATTACK, // marauder jumps when in a fight
 	BOT_A_LEAP_ON_ATTACK, // mantis
 	BOT_A_POUNCE_ON_ATTACK, // dragoon and adv dragoon
 	BOT_A_TYRANT_CHARGE_ON_ATTACK,
--- a/src/sgame/sg_bot_local.h
+++ b/src/sgame/sg_bot_local.h
@@ -91,11 +91,6 @@ struct enemyQueue_t
 	int back;
 };
 
-struct botSkill_t
-{
-       int level;
-};
-
 // boolean flags that tells which skill (compétence) the bot has.
 //
 // When you add a skill, add it to the skill tree in sg_bot_skilltree.cpp and
@@ -151,9 +146,9 @@ struct botMemory_t
 	void doSprint( int jumpCost, int stamina, usercmd_t& cmd );
 	usercmd_t   cmdBuffer;
 
-	botSkill_t  botSkill;
-	skillSet_t  botSkillSet; // boolean flags
-	std::string botSkillSetExplaination;
+	int         skillLevel;
+	skillSet_t  skillSet; // boolean flags
+	std::string skillSetExplaination;
 
 	botEntityAndDistance_t bestEnemy;
 	botEntityAndDistance_t closestDamagedBuilding;
--- a/src/sgame/sg_bot_local.h
+++ b/src/sgame/sg_bot_local.h
@@ -108,6 +108,7 @@ enum bot_skill
 	BOT_B_BASIC_MOVEMENT, // doesn't do anything as of now
 	BOT_A_SMALL_JUMP_ON_ATTACK, // small aliens (dretch and mantis) do jump from time to time when attacking, this helps dodging
 	BOT_A_MARA_JUMP_ON_ATTACK, // marauder jumps when in a fight
+	BOT_A_STRAFE_ON_ATTACK, // strafe when attacking to dodge bullets better
 	BOT_A_LEAP_ON_ATTACK, // mantis
 	BOT_A_POUNCE_ON_ATTACK, // dragoon and adv dragoon
 	BOT_A_TYRANT_CHARGE_ON_ATTACK,
--- a/src/sgame/sg_bot_local.h
+++ b/src/sgame/sg_bot_local.h
@@ -127,6 +127,7 @@ enum bot_skill
 
 	// fighting skills
 	BOT_B_BASIC_FIGHT, // doesn't do anything as of now
+	BOT_B_FAST_AIM,
 	BOT_A_AIM_HEAD,
 	BOT_A_AIM_BARBS, // precisely target barbs
 	BOT_H_PREDICTIVE_AIM, // predict where to aim depending on weapon and enemy speed
--- a/src/sgame/sg_bot_local.h
+++ b/src/sgame/sg_bot_local.h
@@ -93,9 +93,7 @@ struct enemyQueue_t
 
 struct botSkill_t
 {
-	int level;
-	float aimSlowness;
-	float aimShake;
+       int level;
 };
 
 // boolean flags that tells which skill (compétence) the bot has.
@@ -153,7 +151,7 @@ struct botMemory_t
 	void doSprint( int jumpCost, int stamina, usercmd_t& cmd );
 	usercmd_t   cmdBuffer;
 
-	botSkill_t  botSkill; // numerical values
+	botSkill_t  botSkill;
 	skillSet_t  botSkillSet; // boolean flags
 	std::string botSkillSetExplaination;
 
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -177,7 +177,7 @@ int G_BotGetSkill( int clientNum )
 		return 0;
 	}
 
-	return bot->botMind->botSkill.level;
+	return bot->botMind->skillLevel;
 }
 
 void G_BotSetSkill( int clientNum, int skill )
@@ -730,7 +730,7 @@ void botMemory_t::willSprint( bool enable )
 // walking, and stamina not recharging at all.
 void botMemory_t::doSprint( int jumpCost, int stamina, usercmd_t& cmd )
 {
-	exhausted = exhausted || ( botSkill.level >= 5 && stamina <= jumpCost + jumpCost / 10 );
+	exhausted = exhausted || ( skillLevel >= 5 && stamina <= jumpCost + jumpCost / 10 );
 	if ( !exhausted && wantSprinting )
 	{
 		usercmdPressButton( cmd.buttons, BTN_SPRINT );
@@ -775,6 +775,6 @@ std::string G_BotToString( gentity_t *bot )
 			BG_TeamName( G_Team( bot ) ),
 			bot->botMind->behaviorTree->name,
 			BotGoalToString( bot ),
-			bot->botMind->botSkill.level,
-			bot->botMind->botSkillSetExplaination );
+			bot->botMind->skillLevel,
+			bot->botMind->skillSetExplaination );
 }
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -628,7 +628,7 @@ AINodeStatus_t BotActionActivateUpgrade( gentity_t *self, AIGenericNode_t *node 
 	AIActionNode_t *action = ( AIActionNode_t * ) node;
 	upgrade_t u = ( upgrade_t ) AIUnBoxInt( action->params[ 0 ] );
 
-	if ( u == UP_MEDKIT && !self->botMind->botSkillSet[BOT_H_MEDKIT] )
+	if ( u == UP_MEDKIT && !self->botMind->skillSet[BOT_H_MEDKIT] )
 	{
 		// we don't know how to use it
 		return STATUS_FAILURE;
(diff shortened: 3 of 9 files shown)


DolceTriade:
LGTM. Saw no obvious regressions on the testing server

bmorel:
What was tested? That alien bots do not have weird behaviors, or that the balance between AI teams is improved? Last time I tried an "official mod" (last week), said balance was terribly broken by the skill-tree, which is why I'm asking about what was tested.
Keep in mind that players are quite attached to bots, so it would be good to not worsen the situation.

DolceTriade:
I was able to play a game and have fun. Bots didn't seem to be doing obviously wrong stuff. Bots seemed to be using good tactics. Was able to see bots do cool things.

imo, I really like the skill concept for adjusting bot skills. Still some ironing out to do wrt total skill points, but we can fix that over time.

slipher:
What does "lag prediction" mean here?

necessarily-equal:
> What does "lag prediction" mean here?

"lag prediction" means that bots do have self-inflicted "lag": `self->botMind->futureAimTimeInterval = BotGetAimTime( self );` this fuses two instance of the same "prediction", where one is about adding where the player will be by the time of the next aim time, and the other is about adding the projectile fly time (e.g. for the luci blast).

(Note the "Bot: refactor BotGetIdealAimLocation to include lag prediction" is a non-functional change, and while there are weird things about this that are now more obvious, e.g. regarding `frac` which only talks about the first but affects both, fixing this and testing this is outside the scope of this PR.)

necessarily-equal:
So about balance, I'd say this makes things better.

some win-lose statistics for bot-only games with and without this PR:

```
afontain@daemon:~$ sh stats.sh 0.54-stable
     63  Aliens win.  antares
     66  Aliens win.  chasm
     45  Aliens win.  forlorn
     65  Aliens win.  parpax
     65  Aliens win.  perseus
     64  Aliens win.  plat23
     65  Aliens win.  station15
     23  Aliens win.  yocto
      1  Humans win.  forlorn
      1  Humans win.  yocto
afontain@daemon:~$ sh stats.sh more-skills
     30  Aliens win.  antares
     43  Aliens win.  chasm
     13  Aliens win.  forlorn
     45  Aliens win.  parpax
     45  Aliens win.  perseus
     29  Aliens win.  plat23
     37  Aliens win.  station15
      4  Aliens win.  yocto
      2  Humans win.  chasm
      8  Humans win.  yocto
```

So well, it's an better, even if the gap is still strong. Note that bot-only statistics may be very unrepresentative of a real-game experience because human bots are extremely bad at attacking bases.

I've also had some feedback from a player on the testing server, which hints that this is working fine:

> "Player#8994062": haha, yes. the few games I've played have been pretty evenly matched with the bots.

I think this PR is an improvement.

necessarily-equal:
I'll wait a few days (e.g. slipher may want a further review), and if nobody complains, I intend to merge this in the following days because I believe this helps balance.

bmorel:
well, current balance is badly broken, so I'm not sure this can break it more. Still, I would have preferred to see master back to a sane balance before introducing new changes. From what I understand, adding new skills in that stuff will silently break balance of mods configurations, which  I see as a pretty bad thing.
Fixing the current balance problems before merging this would make it easier for modders to suffer less from upstreams decisions, which would be a pretty important thing, considering the fact that currently both the official server and the testing servers are quite far in config from the actually played servers. Those played servers try to stay on master, and thus are the most impacted servers. Most impacted, and most played... I'm think you see the same problem as I do?

bmorel:
well, current balance is badly broken, so I'm not sure this can break it more. Still, I would have preferred to see master back to a sane balance before introducing new changes. From what I understand, adding new skills in that stuff will silently break balance of mods configurations, which  I see as a pretty bad thing.
Fixing the current balance problems before merging this would make it easier for modders to suffer less from upstreams decisions, which would be a pretty important thing, considering the fact that currently both the official server and the testing servers are quite far in config from the actually played servers. Those played servers try to stay on master, and thus are the most impacted servers. Most impacted, and most played... I'm think you see the same problem as I do?

necessarily-equal:
That's a fair point. Let's see what the interested people think about that (they can comment on this issue already, but I didn't ping them explicitely on this issue — I think I did on IRC).

@Sweet and @cu-kai: I think merging this would help balance, but do you think merging this would make balancing harder for you?

necessarily-equal:
(rebased)

cu-kai:
I'd question whether such granularity is required and I think it would make it more difficult to balance, but bot behaviours are more @sweet235's domain.

bmorel:
> I think merging this would help balance,

It will not. What helps balance is fine tuning, not adding more chaos or making things harder to tweak.

sweet235:
I am still not sure on how to proceed with the granularity. Make a fine-grained skillset, and you can selectively disable more things on maps where they go wrong. But it is not quite easy to maintain indeed.

SeanScherer:
As a semi-random passerby (please tell me if the comment is not welcomed!):

If I understand correctly, there is a certain degree of conflicting goals here -
a) One is to provide bots that are as human-like as possible
b)  The other is to have bots that duke it out with each other (in addition to a lowish number of human players), and will result in relatively even matches (ie both sides have a good chance of winning - "balance").

So my (naive) take here would be: If there is code ready to provide a) , *plus* there are still quite a few things missing for bot-behaviour that will effect b) anyway (eg Navcons for maps, better human base-attacking behaviour, etc.), then I would suggest implementing (ie merging this PR). *Especially* if the changes provided are essentially optional (ie setup config to employ skills or no).
<br>
P.S. In this context, (I also think this for other areas of the game) what *would* be a good / important thing to do (ie, ideally by the person making the PR) is to provide documentation for the implemented skillsets (even if the status-quo skillsets are missing these also). I mean on the wiki or so. But I wouldn't make that a condition of merging.

--> that way you can have more people more easily mess around with them (eg for bugtesting, finding good working combinations, etc.).

src/sgame/sg_bot_local.h
```diff
@@ -125,6 +120,7 @@ enum bot_skill
 
 	// fighting skills
 	BOT_B_BASIC_FIGHT, // doesn't do anything as of now
+	BOT_B_FAST_AIM, // doesn't do anything as of now
```
bmorel, line 123: "does not do anything as for now" this should block the merge of this PR.

bmorel (dismissed):
So, I just noticed this, on the topic of heavy maintenance and unfinished things:

```
src/sgame% grep -r FIGHT   
sg_bot_skilltree.cpp:	{ "fighting", BOT_B_BASIC_FIGHT, 3, pred_always, fighting_skills },
sg_bot_local.h:	BOT_B_BASIC_FIGHT, // doesn't do anything as of now
```

and

```
src/sgame% grep -r BOT_B_BASIC_MOVEMENT     
sg_bot_skilltree.cpp:	{ "movement", BOT_B_BASIC_MOVEMENT, 2, pred_always, movement_skills },
sg_bot_local.h:	BOT_B_BASIC_MOVEMENT, // doesn't do anything as of now
```

So I really think the current set should be **finished** before adding more problems and dead code.
I would also point the fact that preventing bots to move or to fight can be done without touching C++, by the use of specific BT.

I want to point out the fact that I also see granularity of dubious usefulness, i.e. pouncing to flee OR to attack OR to do both?

And then I looked more carefully the code, and see the exact same problems:

* duplicated things for the sake of granularity
* code that is never used

@SeanScherer see? That's called "doing the first 90% but not the remaining 90%". I'm not even sure one can tell the first 90% have been done, here.

src/sgame/sg_bot_local.h
necessarily-equal: This was a copy-paste error. Good catch.

SeanScherer:
@bmorel - since you're tagging me, I'll respond:
I think I've made it clear here (and in private with you) that even if I may agree with some of the issues you have, I object to the way you are approaching the matter, that is Lautre, here.

If you are asking me, I find GitHub issues more or less the worst forum to try to clarify things like that (I find voice chat, or even better, Video chat - ie Jitsi or [Chatology](https://stage.chatmosphere.cc/) - the right way and place to sort issues like this). _[of course **in person** is actually the ideal, but even if you're both from France, I doubt this solution is easily at hand :P ]_ 

In case I've not made it obvious, please don't tag me again with "I told you so" messages... (I do get notifications for responses on issues I've commented on).

Regards, Sean

sweet235:
> I was thinking the best answer to that would be to actually push down the question of granularity out of the code, and into config / game variables

Certainly. In fact, we already have that. Look at `g_disabledSkillset` and `g_preferredSkillset`. And also `g_skillsetBudgetAliens`, `g_skillsetBudgetHumans`.

> ideally which are not "fixed" for a game but can be changed while running the server - but I guess this may not be possible

Sure this is possible 😀. We already have that too. Each time you change one of these 4 cvars, the bots' skills will be adapted instantly.

But in the end, this might still not be easy to maintain.

necessarily-equal:
So I'm not going to address questions about the principle of the skilltree, on making some skills optional (which is a problem Sweet solved already), or making the skilltree opt-in (which you technically canby enabling every skill). I'm also not going to address if we should create yet another type of config file, because this is outside the scope of this PR. I also think that has been discussed and overdiscussed a bit since.

I also want to say @SeanScherer, constructive criticism is welcome, sure :) It may help to put general criticism that is not specific to a PR in an issue, apologies if I didn't answer, this post is lengthy enough as is.

### What is this PR about, and answer to the criticism about this PR

So I think it was a bit unclear what this PR was about, and it was a bit big, but I'll try to explain for now.

> [This is a big PR]

Fair enough. For the next time I'll try to do batch of PR instead of big PR like this one. Hopefully that would make it easier to follow. To avoid making the testing burden too high though, I may only do the balancing tests only on some of those PR in a series. This is more work (it also makes it harder to maintain a set of changes as you need to rebase several branches, keep the various PR updated, etc). But let's try, we'll see if that helps.

> It [This PR] will not [help balance]. What helps balance is fine tuning, not adding more chaos or making things harder to tweak.

I'm not sure if I should answer this one, because this looks like a troll. But I'll try anyway.

I think not all the balancing problems introduced in 0.54 are due to the skilltree, which is why this doesn't just tweak skills.

So, if by "adding more chaos" you mean tweaking other features, then I disagree, because I think the bot aiming changes are actually part of the issue (I explain why below). This is a multifaceted problem IMO, which is why I ended up changing several things.

About "making things harder to tweak", you mean that adding a few skills will make the tweaking of the skills more complicated, I disagree. But this was discussed on IRC before, gotta agree that we disagree on that one.

> You are testing layouts here, not (only) bots.

Fair enough. I'm also testing bot against each others without humans. This gives an idea though.

  

I want to select a last criticism, which is that I've read I should "fix the current situation he [I] created" and that kind of stuff. I hope this comment explains why I think this PR is a step in doing that, and I think keeping this blocked is not going to help fixing the situation faster. So I intend to merge this soon if there are no more criticism.

### What are each commit doing and why it should help

Anyway, the commits:

#### *Bot: make mantis jump skill less strong*
This is skill cost tweaking

#### *Bot: add "small-attack-jump" skill*, and *Bot: add "strafe-attack" skill*
These are the only two commits that can favour aliens, as it makes alien bots
more interesting / less predictable by making e.g. dretch jump on attack, and
strafe only some times. The "strafe-attack" commit actually nerfs aliens when
off, because they will be easier to hit.

#### *Bot: tweak aim prediction to avoid predicting too far if aim is bad*
This is actually a balancing commit. I believe part of the reason human bots
get more trouble in 0.54 is that they do aim prediction "too well" now, instead
of doing it less strongly because of a bug, fixed in
https://github.com/Unvanquished/Unvanquished/pull/2281. This now caps the aim
prediction so that the human bots are not harmed by some overprediction.
(There are still weird things in that prediction code, but this PR doesn't try
to address it)

#### *Bot: refactor BotGetIdealAimLocation to include lag prediction*
This is a non-functional change, so it doesn't affect balancing considerations.

#### *Bot: introduce different aiming speed for teams and the "fast-aim" skill*
This helps the human team because the aliens got significantly better at aiming
because of changes in the aiming code. So this commit also tries to help the
balancing.

#### *Bot: remove aimShake and aimSlowness which were never implemented* and *Bot: refactor for shorter, less redundant names*
This is some simple cleanup, so it doesn't affect balancing considerations.

#### *Bot: Tweak skill tree point max accordingly accordingly*
This is to adapt to the changes in the tree skill in the previous commits

#### *Bot: Give a bit more skills at sk3 and sk5*
This doesn't affect sk5 bots, but helps making sk3 bots more interesting

#### *Bot: Give a bit less skill to aliens*
This is again a slight balancing tweak, to nerf aliens slightly

### Conclusion and other stuff
Anyway, I think this PR is a step in making the balance better again, and I intend to merge this soon if there are no more criticism. See this post for the rationale

This PR is to be merged after https://github.com/Unvanquished/Unvanquished/pull/2623 and is followed by https://github.com/Unvanquished/Unvanquished/pull/2624. See also https://forums.unvanquished.net/viewtopic.php?t=2406 (you need to log-in to the forum).

sweet235:
> Fair enough. I'm also testing bot against each others without humans. This gives an idea though.

But it is still like having self-driving cars race against each other on empty roads. I have no idea how we should measure bot balance, but I'd say it is not that easy.

bmorel:
>  I have no idea how we should measure bot balance, but I'd say it is not that easy.

It's not that easy no, but this at least gives some data. When you have a 75% win rate for a faction in player-less game, on a rather bot friendly map (most of official maps are quite bot friendly in their default configuration) this imo tells that there is a problem. Ofc, it might also show a problem in the game's balance itself, but I don't think it's the case in unvanquished (especially since bots don't build).

Issue #2443, opened 2023-02-10
Regression makes turrets useless


necessarily-equal:
Machine gun turrets (and maybe other things) are broken. This is a regression from 0.54.0 in master

Bisect says the regression happened between df7d097ffd3be5122132410b80d96aa311808ef9, e1b40e4db0b1880600f792f1cf8b51a6732f5bbb or a62fa9f470676234e5b0d95b2090603ddbd320c7. I suggest removing the new entity iterator for next (0.54.1) release to maybe reintroduce it later, when we'll have more testing time.

bmorel:
IMO the current situation shows that there is a real need for a stable branch:

* there are a lot of very important fixes for AI regressions in master
* but there is this refactoring that breaks the gameplay very badly

so, servers running 0.54 are badly broken because bots, but servers running master are badly broken because turrets (and yes, other ones are affected, IIRC rocketpods were).

bmorel:
This seems to be fixed in master.

PR #2444, opened 2023-02-10
Revert new entities iterator

necessarily-equal:
See https://github.com/Unvanquished/Unvanquished/issues/2443


DolceTriade:
LGTM

PR #2445, opened 2023-02-10
Rework poison to be configured by cvars and to be able to ignore armor

necessarily-equal:
Make poison effect configurable in cvars \+ tiny refactor commit

### DIFF
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -34,6 +34,26 @@ bool ClientInactivityTimer( gentity_t *ent, bool active );
 static Cvar::Cvar<float> g_devolveReturnRate(
 	"g_devolveReturnRate", "Evolution points per second returned after devolving", Cvar::NONE, 0.4);
 
+static Cvar::Cvar<bool> g_poisonIgnoreArmor(
+	"g_poisonIgnoreArmor",
+	"Make poison damage armored and armorless humans equally",
+	Cvar::NONE,
+	false);
+static Cvar::Range<Cvar::Cvar<int>> g_poisonDamage(
+	"g_poisonDamage",
+	"Damage per second dealt to a poisoned individual",
+	Cvar::NONE,
+	5,
+	0,
+	500);
+static Cvar::Range<Cvar::Cvar<int>> g_poisonDuration(
+	"g_poisonDuration",
+	"Number of seconds the poison effect lasts",
+	Cvar::NONE,
+	10,
+	0,
+	500);
+
 /*
 ===============
 P_DamageFeedback
@@ -1027,8 +1047,10 @@ static void ClientTimerActions( gentity_t *ent, int msec )
 		// deal poison damage
 		if ( client->ps.stats[ STAT_STATE ] & SS_POISONED )
 		{
-			ent->Damage(ALIEN_POISON_DMG, client->lastPoisonClient, Util::nullopt,
-			                    Util::nullopt, DAMAGE_NO_LOCDAMAGE, MOD_POISON);
+			int flags = g_poisonIgnoreArmor.Get() ? DAMAGE_PURE : DAMAGE_NO_LOCDAMAGE;
+			int dmg = g_poisonDamage.Get();
+			ent->Damage(dmg, client->lastPoisonClient, Util::nullopt,
+			                    Util::nullopt, flags, MOD_POISON);
 		}
 
 		// turn off life support when a team admits defeat
@@ -1955,7 +1977,7 @@ static void ClientThink_real( gentity_t *self )
 	}
 
 	if ( (client->ps.stats[ STAT_STATE ] & SS_POISONED) &&
-	     client->lastPoisonTime + ALIEN_POISON_TIME < level.time )
+	     client->lastPoisonTime + g_poisonDuration.Get()*1000 < level.time )
 	{
 		client->ps.stats[ STAT_STATE ] &= ~SS_POISONED;
 	}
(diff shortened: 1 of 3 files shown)


DolceTriade:
Needs gameplay testing

necessarily-equal:
Indeed. There's a dedicated server. Alternatively we could have a 
branch dedicated to gameplay changes for the next release. Like 
gameplay/testing that everyone can make PR or push directly to.

And then we would test every gameplay change at once

sweet235:
Why?

Gireen:
how about adding two cvars for it?
one about `poison ignore armor` and one for `poison damage`
since its all on server side the change should be easy and we can leave the old values as default until its tested enough

sweet235:
> how about adding two cvars for it?

That would be fine with me. We could also put it into a .cfg file.

necessarily-equal:
I've updated this to be configurable on cvars.

DolceTriade:
I would really prefer this to be tested on a server prior to merging, even if its cvar configurable. I don't think we can merge everything even if they can all be disabled.

sweet235:
I hesitate to test this on my server. Recently, I had to change some human bot settings to give humans a chance in bot games. In 0.54, humans do not stand a chance. This would reduce human's ability to push for a win later in the game.

DolceTriade:
Imo, let's schedule a dev game. Let's find a time that works for people and get some playing in and see how this fares.

sweet235:
I'll be there.

src/sgame/sg_active.cpp
```diff
@@ -34,6 +34,26 @@ bool ClientInactivityTimer( gentity_t *ent, bool active );
...
+	Cvar::NONE,
+	false);
+static Cvar::Range<Cvar::Cvar<int>> g_poisonDamage(
+	"g_poisonDamage",
+	"Damage per second dealt to a poisoned individual",
+	Cvar::NONE,
+	5,
+	0,
+	500);
+static Cvar::Range<Cvar::Cvar<int>> g_poisonDuration(
```
Gireen, line 49: i think g_poisonDuration should have a higher resolution, despite currently only being checked every second.
it would be nice to define times in a coherent manner, in code all other settings are in milliseconds 
for cvars it could be enough to have it as a float

src/sgame/sg_active.cpp
```diff
@@ -34,6 +34,26 @@ bool ClientInactivityTimer( gentity_t *ent, bool active );
...
+	"g_poisonDamage",
```
Gireen, line 50: g_poisonDuration

src/sgame/sg_active.cpp
```diff
@@ -1060,14 +1082,7 @@ static void ClientTimerActions( gentity_t *ent, int msec )
...
 			// Give clients some credit periodically
-			if ( client->pers.team == TEAM_ALIENS )
-			{
-				G_AddCreditToClient( client, PLAYER_BASE_VALUE, true );
-			}
-			else if ( client->pers.team == TEAM_HUMANS )
-			{
-				G_AddCreditToClient( client, PLAYER_BASE_VALUE, true );
-			}
+			G_AddCreditToClient( client, PLAYER_BASE_VALUE, true );
```
Gireen, line 1085: what happens with spectators?
i think they would get credits with this

Gireen:
> I would really prefer this to be tested on a server prior to merging, even if its cvar configurable. I don't think we can merge everything even if they can all be disabled.

I would agree if there where large chunks of code added for an disabled feature, but here its only one new cvar and two hardcoded values turned to cvars.
The default settings stay the same and the game becomes a bit more configurable.
Independent from how well this might work on its own it add another easy to access setting for future gameplay experiments and changes.

src/sgame/sg_active.cpp
necessarily-equal: I'd say I don't want to fix this problem because it would affect several things such as the "pain" reaction.

Should we fix that some day, then we can turn the cvar into a float that day.

src/sgame/sg_active.cpp
necessarily-equal: good catch

src/sgame/sg_active.cpp
necessarily-equal: good point, I'll check and update later

bmorel: spectators can not be alive right? But I'd rather keep some kind of check still. This commit is also unrelated to the feature I guess?

bmorel (approved):
I do not think of any reason to not merge changes that makes the gameplay more configurable. The only thing that possibly change in this are the unrelated commit not checking that the team is playable (not a big deal imo, there's a check for alive time right above), and the flag used to qualify damages. It also makes more sense that poison would not be affected by armor. Maybe the damages could be float though, but the decimals would probably be lost at some point in time anyway.

src/sgame/sg_active.cpp
necessarily-equal: I've tested this works correctly and added an assertion.

src/sgame/sg_active.cpp
necessarily-equal: Fixed

necessarily-equal:
I intend to merge this in a few days if no one wants to block this

DolceTriade:
sure, lgtm for this particular change but I really don't want to see our gameplay devolve into a mass of cvars in the name of configurability :(

necessarily-equal:
I would agree with @DolceTriade that we shouldn't do a multiplication of cvars too much. For now I think I'll merge this in the current state.

PR #2448, opened 2023-02-13
True support for old trigger building

bmorel:
fixes all known bugs in trigger_buildable

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -531,21 +531,21 @@ SetMoverState
 ===============
 */
 
-			// most elevators are *not* func_door, so that broken.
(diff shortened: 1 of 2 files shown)


cu-kai:
Tested on bunker and confirmed working. This patch is required for the map `mission-control` to work properly.

DolceTriade (approved):

slipher:
Why are there reverted commits here? Just wasted time reviewing that

bmorel:
that's old stuff I was working on, before changing my mind, while still wanted to have it around for reference I think. Don't worry, I'll squash.

necessarily-equal (approved):
LGTM, haven't tested.

(For an explaination of why the code was more convoluted, you can see 09f92c95a73349692aacfb2e2437bf84f0b0fb44 where the automatic movers used to be handled differently)

PR #2449, opened 2023-02-13
Fix damage and destructible map entities

### LINKED ISSUE #843: Damage-triggered doors don't work.
Reproduce: Attack the grating between the pipe shafts and the side storage room on Parpax.

The reason is that damage is done in the CBSE-part of the gamelogic now and map entities do not yet have CBSE entities attached to them. The proper solution is to port all map entities to the CBSE logic, which is a tremendous task. (Hi @t4im, what's up!? :grinning: )

bmorel:
Fix #843 

This fix 3 types of movers:

* func_door with health
* func_button with health
* func_destructable

func_door_rotating and models are left in their current state. There is no way to test that a change to those would work, though, as no map of my knowledge use those with damages.

add non-cbse interface for damaging entities

doors can die

buttons can die

melee damages can kill movers

movers only open when they die

warn when exploding entities do not have radius

reset healh of damage movers when they return

ents with health obviously take damages if not dead

missiles deal damages to entities, too

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -574,7 +574,7 @@ bool IsDoor( const gentity_t *ent )
 // for now, but not sure how this should be handled.
 bool IsAutomaticMover( const gentity_t *ent )
 {
-	if ( ent->takedamage )
+	if ( ent->config.health )
 	{
 		return false;
 	}
@@ -1102,6 +1102,11 @@ void BinaryMover_act( gentity_t *ent, gentity_t *other, gentity_t *activator )
 		// open areaportal
 		if ( ent->groupMaster == ent || !ent->groupMaster )
 			trap_AdjustAreaPortalState( ent, true );
+
+		//HACK: this is really ugly, should have specialise code,
+		//but I'm both lazy, buzy to fix old mess, and annoyed by
+		//said years-old mess
+		ent->health = ent->config.health;
 	}
 	else if ( ent->moverState == MOVER_2TO1 )
 	{
@@ -1466,7 +1471,7 @@ static void func_door_block( gentity_t *self, gentity_t *other )
 
 	if ( self->damage )
 	{
-		other->entity->Damage((float)self->damage, self, Util::nullopt, Util::nullopt, 0, MOD_CRUSH);
+		other->Damage((float)self->damage, self, Util::nullopt, Util::nullopt, 0, MOD_CRUSH);
 	}
 
 	if ( self->spawnflags & 4 )
@@ -1566,12 +1571,18 @@ static void Think_SpawnNewDoorTrigger( gentity_t *self )
 	}
 }
 
+static void Think_MoverDeath( gentity_t *self, gentity_t*, gentity_t* attacker, int )
+{
+	if ( self->moverState == MOVER_POS1 )
+	{
+		BinaryMover_act( self, nullptr, attacker );
+	}
+}
+
 static void func_door_reset( gentity_t *self )
 {
 	G_ResetIntField(&self->health, true, self->config.health, self->eclass->config.health, 0);
 	G_ResetIntField(&self->damage, true, self->config.damage, self->eclass->config.damage, 2);
-
-	self->takedamage = !!self->health;
 
 	reset_moverspeed( self, 400 );
 }
@@ -1661,11 +1672,17 @@ void SP_func_door( gentity_t *self )
 	InitMover( self );
 
 	self->nextthink = level.time + FRAMETIME;
+	self->health = self->config.health;
 
 	if ( self->names[ 0 ] || self->config.health ) //FIXME wont work yet with class fallbacks
 	{
 		// non touch/shoot doors
 		self->think = Think_MatchGroup;
+		if ( self->config.health )
+		{
+			self->health = self->config.health;
+			self->die = Think_MoverDeath;
+		}
 	}
 	else
 	{
@@ -1676,8 +1693,6 @@ void SP_func_door( gentity_t *self )
 static void func_door_rotating_reset( gentity_t *self )
 {
 	G_ResetIntField(&self->health, true, self->config.health, self->eclass->config.health, 0);
-
-	self->takedamage = !!self->health;
 
 	reset_rotatorspeed( self, 120 );
 }
@@ -1801,8 +1816,6 @@ static void func_door_model_reset( gentity_t *self )
 {
 	G_ResetFloatField(&self->speed, true, self->config.speed, self->eclass->config.speed, 200);
 	G_ResetIntField(&self->health, true, self->config.health, self->eclass->config.health, 0);
-
-	self->takedamage = !!self->health;
 
 	self->s.torsoAnim = self->s.weapon * ( 1000.0f / self->speed ); //framerate
 	if ( self->s.torsoAnim <= 0 )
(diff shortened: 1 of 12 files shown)


illwieckz:
I confirm it works! Good job! 👍️👍️👍️

src/sgame/sg_utils.cpp
```diff
@@ -945,3 +945,31 @@ glm::vec3 G_CalcMuzzlePoint( const gentity_t *self, const glm::vec3 &forward )
 	// Meh. I doubt it saves much. Casting to short ints might have, though. (copypaste)
 	return glm::floor( muzzlePoint + 0.5f );
 }
+
+//Dunno which ones are really useful, CBSE is not even self-contained.
+#include "sg_local.h"
```
slipher, line 950: Don't include headers in the middle of the file

src/sgame/sg_utils.cpp
```diff
@@ -945,3 +945,31 @@ glm::vec3 G_CalcMuzzlePoint( const gentity_t *self, const glm::vec3 &forward )
...
+	if ( pain )
+	{
+		pain( this, source, amount );
+	}
+	health -= amount;
+	if ( health <= 0 && die )
+	{
+		die( this, nullptr, source, 0 );
+	}
+	return true; // seriously, not sure what's this about
```
slipher, line 974: When you send a message a 'true' return means the entity had a handler for that message

src/sgame/sg_spawn_mover.cpp
```diff
@@ -1102,6 +1102,11 @@ void BinaryMover_act( gentity_t *ent, gentity_t *other, gentity_t *activator )
 		// open areaportal
 		if ( ent->groupMaster == ent || !ent->groupMaster )
 			trap_AdjustAreaPortalState( ent, true );
+
+		//HACK: this is really ugly, should have specialise code,
```
slipher, line 1106: Is the hack that it's set here instead of when spawning the entity? Why is it hard to set it upon spawning?

slipher (commented):
Some weapons are not going to work because they rely on searching for HealthComponent's. It's OK as is as a hack to allow achieving map objectives I guess

bmorel:
> Some weapons are not going to work because they rely on searching for HealthComponent's

I tested all weapons, and this only does not work with ground fires, zap, and ofc poison.

src/sgame/sg_spawn_mover.cpp
bmorel: No. It's more related to the fact that I think damage-entities should probably have their very own functions, instead of reusing this one. Here, the affectation is to *reset* when the door/button goes back to it's original position.

> Why is it hard to set it upon spawning?

It's already done on spawning, this one is a reset, not a set. Or maybe I do not understand what you mean?

src/sgame/sg_utils.cpp
bmorel: So, should return false? What are the implications, exactly?
IIRC, the return value is used exactly one time, for rocketpods. This is almost like a "return ent == rocketpod".
But do you prefer me to return false here?
There is no documentation around, so i basically tried to mimic what appeared to be previous behavior (should be explicit with that comment) , as I wanted this 8 years old regression fixed. Originally I didn't really intended to upstream it though.

src/sgame/sg_utils.cpp
bmorel: original intent was to try to get rid those, but i'm feeling too lazy to dig in that hole. I'll move them on top.

DolceTriade:
We should add a TODO to add a HealthComponent to movers too, but am fine with the hack for now.

bmorel:
> We should add a TODO to add a HealthComponent to movers too, but am fine with the hack for now.

This is TODO since 8 years, breaking several maps. It even broke previous versions of Parpax, an official unvanquished map which was changed to no longer use this.
At this point it seems ridiculous to me to continue to hope that, maybe, someone will kick in and finish the CBSE work, especially considering the fact it has extremely poor documentation and extremely convoluted architecture.

src/sgame/sg_utils.cpp
slipher: I suppose it should return true iff the mover has health.

src/sgame/sg_utils.cpp
```diff
@@ -945,3 +945,31 @@ glm::vec3 G_CalcMuzzlePoint( const gentity_t *self, const glm::vec3 &forward )
...
 	return glm::floor( muzzlePoint + 0.5f );
 }
+
+//Dunno which ones are really useful, CBSE is not even self-contained.
+#include "sg_local.h"
+#include "Entities.h"
+#include "CBSE.h"
+#include "sg_cm_world.h"
+
+bool gentity_t::Damage( float amount, gentity_t* source,
```
slipher, line 955: Entities.cpp might be a better place for this. You could have `Entities::Damage` similarly to `Entities::Kill` etc. (optional)

src/sgame/sg_spawn_mover.cpp
slipher: Sorry the 2nd question was just an incorrect guess. Where does the idea that the health is supposed to reset under some circumstances come from though?

src/sgame/sg_utils.cpp
bmorel: I'll be honest: I do not want to touch the CBSE more than strictly required. The patch currently works. It took me a full day of dev (more, if one considers a normal day of work). I don't want to rework it a lot just for that over-engineered thing that initially introduced the regression by being merged unfinished.

src/sgame/sg_spawn_mover.cpp
bmorel: If I do not reset the health, then doors will still have negative health when closing anew. This means that the 50HP (for example) value will only be valid the 1st time they're hit.

src/sgame/sg_utils.cpp
bmorel: Sounds reasonable. I'll return true if the mover have "config->health != 0", that is, maxHealth.

src/sgame/sg_utils.cpp
slipher: You don't have to touch CBSE any more that. It's just renaming the function basically

bmorel: Well, you said "optional", and I'll take it. I'll let people interested at this kind of nit-picking do the meaningless change themselves.

bmorel:
Today cu-kai gave me test maps for rotating entities, so I added that fix to this PR. He also spot a problem in their crushing code. Since this is a 1-line change, that was both found and tested because of this PR (and a test-map made for it), I added the commit to this PR too.

sweet235:
I tried this with the damage buttons on p4x and rotcannon-garbage, lgtm.

illwieckz:
As a notable thing I tested it with old Parpax map (more precisely, Parpax-mini, but that part is not expected to be modified), as recent Parpax modified the way to open grates to not rely on damage, likely to workaround the issue.

This the use case given in https://github.com/Unvanquished/Unvanquished/issues/843 to reproduce the bug:

> Reproduce: Attack the grating between the pipe shafts and the side storage room on Parpax.

And it works.

cu-kai (approved):
LGTM. Fixes a long standing bug I reported on IRC more than half a decade ago... (#843)

In future maybe it can be re-worked "properly" to use CBSE, but nobody seems interested in this right now so this is fine.

bmorel:
I'll merge this in few hours if nobody else have questions.

PR #2450, opened 2023-02-13
Hack ignore small movers

bmorel:
This fixes bots being unable to move through the electric fence in nova map, when electric fence is not disabled.

In practice, the electric fence is a door, which is closed most of the time and opens to "give" access to an area with trigger_hurt. This door generates an obstacle, preventing bots to navigate. This change prevents obstacle addition on "negligible" doors.

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -39,6 +39,12 @@ Maryland 20850 USA.
 
 #define DEFAULT_FUNC_TRAIN_SPEED 100
 
+static Cvar::Range<Cvar::Cvar<int>> g_bot_ignoreSmallObstacles(
+		"g_bot_ignoreSmallObstacles",
+		"this allows bots to ignore \"negligible\" navmesh blockers."
+		"0 is disabled. Should only be enabled on maps with known problem."
+		"Maximum is 18, which is bots' stepsize.", Cvar::NONE, 0, 0, 18 );
+
 /*
 ===============================================================================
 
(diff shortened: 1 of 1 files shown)


bmorel:
This hack is disabled by default, so that server owner can enable on a per-map basis.

src/sgame/sg_spawn_mover.cpp
```diff
@@ -598,19 +604,28 @@ bool IsAutomaticMover( const gentity_t *ent )
...
+	// this is a hack that allows to ignore "irrelevant" doors,
+	// implemented for nova's electric fence support
 	switch ( moverState )
 	{
 		case MOVER_POS1:
 		case MOVER_POS2:
 		case ROTATOR_POS1:
 		case ROTATOR_POS2:
-			if ( IsDoor( ent ) )
+			if ( false
```
DolceTriade, line 621: what's the if false for?

bmorel: Oh, it's nothing in itself, just a trick of mine.
Basically, using false on the if's line, and putting all OR'd condition (or true for AND'd) on their own line allows text editors to automatically align the code (I only do that for long lines).
It also easily allows to comment one part of the conditions for debug purposes.
Forgot to remove it before pushing to this repo.

src/sgame/sg_spawn_mover.cpp
```diff
@@ -39,6 +39,12 @@ Maryland 20850 USA.
 
 #define DEFAULT_FUNC_TRAIN_SPEED 100
 
+static Cvar::Range<Cvar::Cvar<int>> g_bot_ignoreSmallObstacles(
+		"g_bot_ignoreSmallObstacles",
+		"this allows bots to ignore \"negligible\" navmesh blockers."
+		"0 is disabled. Should only be enabled on maps with known problem."
+		"Maximum is 18, which is bots' stepsize.", Cvar::NONE, 0, 18, 0 );
```
slipher, line 46: The range is backward

slipher (commented):
I'm never happy to see code used by only one map :\

DolceTriade:
lgtm

bmorel:
> I'm never happy to see code used by only one map :\

Only one _known_ map. There are other cases such as this one (that is, fixing things for a single known map), for example trigger_building was used by only one known map when I fixed it. Same applies to some cases in trigger_equipment.
But I agree this one is rather hackish, and I would rather see a proper solution, which is: to generate navmeshes above movers, when they're in their starting *and* final position, to then try to block only a single one of those navmeshes depending on where the mover is.
This will even very likely to have support for movers that are meant to move players toward a destination.

In any case, should this be merged or not? If not I'll resend a PR with only the 1st commit, as it only does minor code cleanup.

DolceTriade:
I'm ok with merging this as is

bmorel:
I'll merge tomorrow if build passes and nobody else is against

necessarily-equal:
This makes code harder to read, but I'm still ok with merging this to help map support.

necessarily-equal (approved):

Issue #2454, opened 2023-02-15
New func_door and func_button properties


bmorel:
While fixing #843 and even before that, we noticed (see #2453) that `+activate` can be used to open damage doors. This could very easily be promoted as a feature, only the interface would need to be defined. Notably, about the USE_DISTANCE problem.

Other points are, it seems NetRadiant's documentation states that the "health" property is kind of boolean, while I implemented it exactly like it would work a human. From IRC discussion, 2 ideas emerged:

* those could only receive damages that have above a certain number. This could be done either as "count only if higher than N", or as if "reduce incoming damages by N"
* it would make sense to have a regeneration speed

Ping @cu-kai as you're the only active mapper around those days. What do you think about those, would those have an interest to you?
From a technical point of view, they all seem fairly easy to implement to me, and since the code is still fresh in my mind, this is likely the ideal moment for that.

bmorel:
One notable problem with the `+activate` problem is that this will depend on the difference between `USE_RANGE` and size of the entity to be used reliably. This is the only thing that might be a bit annoying to implement (mostly because of the talking though) as this constant is also used by the armoury, reactor and drill for human resupplies.
My opinion on this problem is, `USE_RANGE` should be a build-time constant, and each usable entity should have a use_distance property (specified by the .bsp or by configuration files for buildings). This use_distance property should be added to the "average radius" of the entity (that is: `half_width + sqrt ( half_width^2+half_width^2+ ) / 2` ) and would not be able to be more than `USE_RANGE`.
This would, I think, give a rather uniform feeling about the building reacting when close enough to the *perceived* building (perceived != bbox, especially since many stuff are cylinders) while giving decent performances (a distance is easy to calculate, and `USE_RANGE` would make sure that no ridiculous value would break performances).

cu-kai:
I don't remember how the `health` property worked in tremulous and whether you needed to damage by X health, or simply deal any damage. You might do to look at the code for that.

I think it would be interesting to see a `+use` property for func_door, func_button, etc and I think it should _probably_ be implemented as a spawnflag to be toggled on/off (if on, "health" would be ignored, too).

bmorel:
> (if on, "health" would be ignored, too).

I see no reason for such a limitation. I think it should be possible (and easier!) to open a door either gently, or by sending it a fully charged lucifer blast. They do that all the time on television :)

bmorel:
[...]

PR #2455, opened 2023-02-15
+activate no longer affects doors and buttons

### LINKED ISSUE #2453: `+activate` can be used to open damage entities
How to reproduce: move to door 7 (as close as possible) of map-test-doors, and press the `buy` key as human, the `evolve` key as alien (needs to be big enough, dretches have hard time doing that) or use the command `+activate` : the door will open, while it should not.

This is a behavior that is known about since a long while (since @cu-kai made the test map for damage entities actually), and was used on some very specific instances of damage entities to work around the regression. This regression is no longer alive, and this allows to cheat, so this should be repaired as well.

bmorel:
This fix #2453 but should probably not be merged as is, since I believe this disables something that would be a useful feature if it was handled in a better way and had a good interface. See #2454

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -1608,7 +1608,6 @@ void SP_func_door( gentity_t *self )
 
 	self->blocked = func_door_block;
 	self->reset = func_door_reset;
-	self->use = func_door_use;
 
 	// default wait of 2 seconds
 	if ( !self->config.wait.time )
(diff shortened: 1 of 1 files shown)


bmorel:
ping @cu-kai , @sweet235

bmorel:
OTOH the code is only disabled, so implementing the feature behind won't generate much conflicts.

DolceTriade:
Can we add a check for damage entities and return early in that case?

bmorel:
superseeded by #2453

bmorel:
Closing because #2469 have testing facilities and has been proven working, so there is no more point in this PR.

Issue #2456, opened 2023-02-16
Killing yourself makes you lose credits/evos


VReaperV:
Dying to your own weapons (e. g. lucifer cannon, grenade, firebomb, flamethrower, dragoon snipe) makes you loose credits/evos upon death. You already lose all of your equipment/alien class on death, further penalizing the player because they just happened to die to their own weapon in combat makes 0 sense.

VReaperV:
> Ah, we should also check if suiciding with `/kill` (sometime handy when stuck in some maps) doesn't remove credits either, losing equipment is bad enough.

`/kill` doesn't seem to take away any credits, just killing yourself with a weapon you have.

sweet235:
Easy to fix, here: https://github.com/Unvanquished/Unvanquished/blob/4d93a66cd77ce96fab329ef1191b16b4009b8546/src/sgame/sg_combat.cpp#L427

This line is a little suspicious, since

- `attacker == self` seems to imply `G_OnSameTeam( self, attacker )`
- there is a redudant pair of parentheses, since 38272099077f7f50e8d7d3b98b3cdb825c92e00f (over 10 years ago)

But I'm not entirely sure I understand what was going on there.

sweet235:
In fact, it has been like this since 98b3a1480522dbd90660ee4c6bc91f6d80885286, suggesting that GPP already had this. I think we can change it.

PR #2459, opened 2023-02-17
No team kill penalty for suicide

sweet235:
This is to fix https://github.com/Unvanquished/Unvanquished/issues/2456

Team kill penalty for suicide has been the case forever. I'm not sure why, but there might have been a reason.

One reason I can think of is that we might want to penalize teleporting back to base. However, there is no penalty for suicide by kill zones on the map. In any case, suicide will lead to loss of purchased equipment/class.

### DIFF
--- a/src/sgame/sg_combat.cpp
+++ b/src/sgame/sg_combat.cpp
@@ -426,7 +426,7 @@ void G_PlayerDie( gentity_t *self, gentity_t *inflictor, gentity_t *attacker, in
 	{
 		if ( G_OnSameTeam( self, attacker ) )
 		{
-			// punish team kills
+			// punish team kills and suicides
 			if ( attacker->client->pers.team == TEAM_ALIENS )
 			{
 				// don't punish suicides with reduced funds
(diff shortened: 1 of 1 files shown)


DolceTriade:
lgtm

slipher:
I think it would make more sense to treat the suicide by weapon case the same as the suicide by map case (i.e. score penalty but no credit penalty).

sweet235:
In https://github.com/Unvanquished/Unvanquished/issues/2456, I wrote
> `attacker == self` seems to imply `G_OnSameTeam( self, attacker )`

Obviously, this does not hold if the player switched teams since firing the weapon (it might be firebomb or flamer). I don't see a big problem in this. But there must have been a reason to specify `attacker == self` in GPP.

necessarily-equal (approved):
Sounds good, haven't tested in-game.

necessarily-equal:
About this thing,

> attacker == self seems to imply G_OnSameTeam( self, attacker )

I would argue that this may not be true if you switched team to spectators, as G_OnSameTeam returns false for two spectators, but then you probably shouldn't trigger this anyway.

sweet235:
I'm still confused. But I think that switchting to spec cannot gain you any advantage. Maybe I'm wrong.

PR #2460, opened 2023-02-18
Remove ground fires from disconnected players

slipher:
Leaving them could cause bugs because the damage is then attributed do a different or nonexistent player.

### DIFF
--- a/src/sgame/sg_team.cpp
+++ b/src/sgame/sg_team.cpp
@@ -251,7 +251,8 @@ void G_LeaveTeam( gentity_t *self )
 			// reset damage dealt to the other clients
 			ent->credits[ self->client->ps.clientNum ] = { 0.0f, 0, TEAM_NONE };
 		}
-		else if ( ent->s.eType == entityType_t::ET_MISSILE && ent->r.ownerNum == self->s.number )
+		else if ( ( ent->s.eType == entityType_t::ET_MISSILE || ent->s.eType == entityType_t::ET_FIRE )
+		          && ent->r.ownerNum == self->s.number )
 		{
 			G_FreeEntity( ent );
 		}


illwieckz:
Is there an entity that can inherit them? Like worldspawn? At worst it would show people being killed by world or something like that?

slipher:
Maybe. I just did the same thing we do for missiles.

illwieckz:
OKOK, that's consistent then.

DolceTriade:
lgtm

cu-kai:
I think it's different with ground flames (vs missiles) as they have a lasting effect; this is arguably a gameplay change. Could we not work around this by using `messageSuicides` for the obituary on the client (in this case, _"X was burned to death"_)? That way the server still tracks the kill(s) and there's no undefined behaviour on the client.

I can look at a making PR to do that if you like.

slipher:
To me the situation is really the same. Obviously the stuff should not be removed from a realism standpoint, but it is removed to avoid bugs. The longer lifetime of flames just increases the possibility of the bug triggering because there is more time for players to get burned or for someone to join in the same client slot.

If you want to test illwieckz's proposal to change the ownernum for abandoned missiles/fires, that would be a welcome improvement. But I don't think using suicide messages is a good idea because that just gives false information.

bmorel:
I agree that I would consider a more severe problem ground flames disappearing when someone's client crash or is disconnected than the current rare situation (is it rare because nobody really uses flamer anymore, though?).
I'm also wondering if it really makes this much sense to remove all missiles, even if most missiles are pretty fast (and thus limited impact on gameplay).

I would rather see missiles and ground flames re-parented, not sure to which entity though... Maybe this would need to reserve "ghosts" entities for this situation, one for each team? I remember people talking about "teamkills" because someone switched team while having killed someone because of ground flames (might also apply with poison?): re-parenting damage entities to such an team_ghost entity could possibly make the problem easier to solve.

sweet235:
The situation is not that rare, at least not on my server. I have maps where there is almost always at least one human bot with fire on the gound. This is because I let prefer bots firebombs over grenades. Joining humans will lead to fire disappearing. I can already hear the complaints.

bmorel:
> Joining humans will lead to fire disappearing. I can already hear the complaints.

AFAIK this shuold not happen. I was only mentionning that a few months ago, there was a problem about changing team could make you loose credits if a current friend dies because of something you did in previous team.

> But I'll choose the fire removal over a potential crash any day.

Can the unnamed bug lead to a crash, actually?

sweet235:
> Can the unnamed bug lead to a crash, actually?

I don't know. But I know that the server writes the name of the already disconnected bot into the console, if someone dies on its fire. That is a little scary.

sweet235:
If the memory holding the bot's name is heap-allocated, reading it later can certainly crash. I don't know if it is heap-allocated.

slipher:
I'll merge this now and then later we can implement the ownerNum change for fires and non-grenade missiles.

I remembered there being some suicide messages with "itself" etc. but apparently those were removed in 2014 so that's convenient.

Issue #2462, opened 2023-02-18
sgame: disconnected players can still gain kills


cu-kai:
The issue
--
While not inherently a problem in itself, this has the potential to cause a crash because as far as I can tell, the player entity is freed as soon as they disconnect. If that memory is re-allocated, bad things might happen. While I've not been able to produce a crash under testing conditions, it should not be ruled out.

Below is an extract from @sweet235's server log showing the incident we observed. In this instance, the bot `[bot] Dr. Henry Armitage` had dropped a firebomb whose flames persisted after the bot was kicked from the game (when Sweet joined the human team). We only investigated this bug when we saw a bugged obituary on the client (also below).
```
  8:02 ChangeTeam: 0 human: Sweet.genius switched teams
  8:02 Die: 10 16 MOD_BURN 4 2: LB killed [bot] C'thalpa; [bot] Dr. Henry Armitage assisted
  8:02 ClientDisconnect: 4 [] (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) "[bot] Dr. Henry Armitage"
  8:02 Construct: 12 470 mgturret: illwieckz [granger] is building Machinegun Turret
  8:02 Deconstruct: 3 444 mgturret MOD_LEVEL1_CLAW: Machinegun Turret destroyed by [bot] Shabbith-Ka
  8:02 Die: 10 15 MOD_BURN: LB killed [bot] Aiueb Gnshal
  8:03 Die: 4 17 MOD_BURN: [bot] Dr. Henry Armitage killed [bot] Nyctelios
```
![image](https://user-images.githubusercontent.com/13281185/219874186-a40ed9bf-1ec8-40f0-a010-7f42ba12ee5c.png)

The solution
--
#2460 attempts to fix this by removing ground flames when players disconnect. I disagree with this as a solution as it's technically gameplay altering. A better solution would be (and this is proposed in #2460) to remove the owner of the flames to avoid any potential for crashes. That would give the world the kill, and the obituary message could reuse `messageSuicides` as it's generic enough to be believable and would not use the ugly "noname" method we currently use.

Ideally, instead of _"●X was burned by ●noname's flames"_, the obituary would simply state that _"●X burned to death"_

As I mentioned in #2460, grenades should still be removed for disconnected players (prior to explosion) for the sake of ease of administration (catching some troll throwing grenades in the base, for example).

sweet235:
[...]

PR #2463, opened 2023-02-18
CG_Obituary improvements

cu-kai:
This PR improves obituaries in several ways:
The horrible "noname killed X" message should be seen less often, and instead, where there's no valid attacker (usually caused by the attacker disconnecting just before the event takes place, see #2460 and #2462), the suicide message will be used.

Obituaries now show teamkills (and where a teammate assists a kill) more obviously, without having to look at the team tag.

Emoticon obituaries are now fixed, adding MOD_SLAP, MOD_SPIKER and fixing a few other things such as missing emoticons / incorrect emoticons. It also fixes the arrangement of the obituary text when a player assists a world object in killing. Further details are in the commit message.

MOD_SLAP uses the red "Cross" emoticon.
MOD_SPIKER uses the legacy "Hovel" emoticon until we can find something better.

Now some screenshots of the new teamkill messages (the !! & !!! identifies a teamkill in Emoticon only mode)
Screenshot_20230218_192409
Screenshot_20230218_192443
Screenshot_20230218_192517
Screenshot_20230218_192545

Rocket pods are now identified as follows, until we get a suitable asset:
image

In future, it would do to move the Emoticon obituaries out of the chat, make them larger, and use the icons from the circle menu as they are more distinguishable, but that is outside the scope of this PR. Then we could have several modes for Emoticon obituaries (disabled, include in chat, or have them shown separately).

### DIFF
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -78,7 +78,7 @@ static const struct {
 	{ "[basilisk]",    false, true,  TEAM_ALIENS },
 	{ "[dragoon]",     false, true,  TEAM_ALIENS },
 	{ "[dragoon]",     false, true,  TEAM_ALIENS }, // pounce
-	{ "[advdragoon]",  false, true,  TEAM_ALIENS },
+	{ "[advdragoon]",  false, true,  TEAM_ALIENS }, // barbs
 	{ "[marauder]",    false, true,  TEAM_ALIENS },
 	{ "[advmarauder]", false, true,  TEAM_ALIENS }, // zap
 	{ "[tyrant]",      false, true,  TEAM_ALIENS },
@@ -89,17 +89,17 @@ static const struct {
 	{ "[booster]",     false, true,  TEAM_ALIENS }, // poison
 	{ "[hive]",        true,  true,  TEAM_ALIENS },
 
-	{ LONGFORM,        true,  false, TEAM_HUMANS }, // H spawn
-	{ "^dR[turret]",   true,  true,  TEAM_HUMANS }, // rocket pod FIXME: we need a rocket pod emoticon
+	{ LONGFORM,        true,  false, TEAM_HUMANS }, // H spawn (human building explosion)
+	{ "^dR[turret]",   true,  true,  TEAM_HUMANS }, // rocket pod. FIXME: we need a rocket pod emoticon
 	{ "[turret]",      true,  true,  TEAM_HUMANS },
 	{ "[reactor]",     true,  true,  TEAM_HUMANS },
 
-	{ LONGFORM,        true,  false, TEAM_ALIENS }, // A spawn
+	{ LONGFORM,        true,  false, TEAM_ALIENS }, // A spawn (alien building explosion)
 	{ "[acidtube]",    true,  true,  TEAM_ALIENS },
-	{ "[hovel]",       true,  true,  TEAM_ALIENS }, // spiker. FIXME: we need a hovel emoticon
+	{ "[hovel]",       true,  true,  TEAM_ALIENS }, // spiker. FIXME: we need a spiker emoticon
 	{ "[overmind]",    true,  false, TEAM_ALIENS },
-	{ "",              true,  false, TEAM_NONE },
-	{ "",              true,  false, TEAM_NONE },
+	{ "",              true,  false, TEAM_NONE },   // (MOD_DECONSTRUCT)
+	{ "",              true,  false, TEAM_NONE },   // (MOD_REPLACE)
 };
 
 static void CG_Obituary( entityState_t *ent )
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -61,7 +61,7 @@ static const struct {
 	{ "[fire]",        false, true,  TEAM_HUMANS }, // burn
 	{ "[grenade]",     false, true,  TEAM_HUMANS },
 	{ "[firebomb]",    false, true,  TEAM_HUMANS },
-	{ "crushed",       false, true,  TEAM_HUMANS }, // crushed by human player. FIXME: emoticon
+	{ "crushed",       false, true,  TEAM_HUMANS }, // crushed by human player. FIXME: emoticon?
 	{ LONGFORM,        true,  false, TEAM_NONE   }, // water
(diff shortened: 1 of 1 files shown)


src/cgame/cg_event.cpp
```diff
@@ -134,7 +134,7 @@ static void CG_Obituary( entityState_t *ent )
 	ci = &cgs.clientinfo[ target ];
 	gender = ci->gender;
 
-	if ( attacker < 0 || attacker >= MAX_CLIENTS )
```
slipher, line 136: This bounds checking shouldn't be removed

slipher (commented):
How do you use emoticon kill messages?

src/cgame/cg_event.cpp
```diff
@@ -58,15 +58,16 @@ static const struct {
 	{ "[lcannon]",     false, true,  TEAM_HUMANS }, // splash
 	{ "[flamer]",      false, true,  TEAM_HUMANS },
 	{ "[flamer]",      false, true,  TEAM_HUMANS }, // splash
-	{ "[flamer]",      false, true,  TEAM_HUMANS }, // burn
+	{ "[fire]",        false, true,  TEAM_HUMANS }, // burn
 	{ "[grenade]",     false, true,  TEAM_HUMANS },
 	{ "[firebomb]",    false, true,  TEAM_HUMANS },
-	{ "crushed",       true,  false, TEAM_NONE   }, // weight (H) // FIXME
```
DolceTriade, line 64: Why is this team_humans now?

src/cgame/cg_event.cpp
```diff
@@ -45,7 +45,7 @@ static const struct {
 	team_t   team;
 } meansOfDeath[] = {
 	// Icon            Envkill Assist? (Team)
-	{ "☠",             false, false, TEAM_HUMANS },
```
DolceTriade, line 48: Why is this team_humans?

I'm guessing this should  never happen?

DolceTriade (commented):
The downside of this change is that we lose deaths due to secondary fire right?

Can we make the kill messages for barb and zap different than the slash/bite attacks?

src/cgame/cg_event.cpp
```diff
@@ -212,26 +212,47 @@ static void CG_Obituary( entityState_t *ent )
...
 			else
 			{
 				if ( meansOfDeath[ mod ].showAssist && assistantInfo )
 				{
-					Log::Notice( "%s%s^* (+ %s%s^*) %s %s%s", teamTag[ attackerTeam ], attackerName, teamTag[ assistantTeam ], assistantName, meansOfDeath[ mod ].icon, teamTag[ ci->team ], targetName );
+					Log::Notice( "%s%s%s^* (+ %s%s%s^*) %s %s%s", 
+					             teamTag[ attackerTeam ], attackerName, 
+								 ( attackerTeam == ci->team ? " ^1!!!^*" : "" ),
+								 teamTag[ assistantTeam ], assistantName, 
+								 ( assistantTeam == ci->team ? " ^a!!" : "" ),
```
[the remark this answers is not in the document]
cu-kai: They're a different colour for assist teamkills.

src/cgame/cg_event.cpp
cu-kai: Because this is for `MOD_WEIGHT_H` aka crushed by human player. See line 68 for `MOD_CRUSH` and line 86 for `MOD_WEIGHT_A`.

src/cgame/cg_event.cpp
cu-kai: Should probably also be `NO_TEAM`. I only added the comment. I can add comments to all of them if you like, but they correspond to the list of `modNames` in https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_combat.cpp

src/cgame/cg_event.cpp
cu-kai: I'll fix that.

cu-kai:
> The downside of this change is that we lose deaths due to secondary fire right?
No?

> Can we make the kill messages for barb and zap different than the slash/bite attacks?
I already did, changing [dragoon] to [advdragoon] and [marauder] to [advmarauder]. Not much else I can do with the limited emoticons we have, and these ones aren't exactly legible. I'm only fixing the current implementation.

See my comment at the end of my OP for what I would like to see in future.

src/cgame/cg_event.cpp
cu-kai: Fixed in 400ae8a56d0be5f8b5d6cf7c4805c4971b9a4310

src/cgame/cg_event.cpp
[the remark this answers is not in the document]
cu-kai: > How do you use emoticon kill messages?

For what it's worth, I never answered this. Most people won't even know they exist because I don't see an option to enable it. The cvar is `cg_emoticonsInMessages`. I only discovered it when looking to fix another obituary bug yesterday...

There should probably be a PR for this to add the option to the menu.

src/cgame/cg_event.cpp
cu-kai: Fixed in e036bf3c1cb449fd5fdb6be3b247cea403f6775a

DolceTriade:
lgtm for now. We can fix the disambiguation between secondary attack kills later as this isn't a regression

DolceTriade (approved):

cu-kai:
I'll echo my thoughts on IRC here:

I think in future yes it would be good to have real emoticons. I am no artist & IANAL so I'm not getting involved in the asset side of this - if you can find some good, royalty-free assets we can use for generally expanding our emoticon library, the ambiguous killicons can be addressed in another PR.

Regarding missing emoticons, I have opened https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/issues/173

src/cgame/cg_event.cpp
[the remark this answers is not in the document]
illwieckz: We can do better in the future if we think we can do better. This is only a detail anyway this should not block the merge.

illwieckz:
@cu-kai the PR is now unlocked for merging (someone had to mark a remaining conversation _resolved_).

illwieckz (approved):
LGTM

illwieckz:
Someone with ideas to improve this can now simply submit a PR targetting `master`.

PR #2464, opened 2023-02-19
Modernize InformUnlockableStatusChanges

necessarily-equal:
`std::string` is just so much more readable here.

### DIFF
--- a/src/shared/bg_teamprogress.cpp
+++ b/src/shared/bg_teamprogress.cpp
@@ -95,15 +95,12 @@ static bool Disabled( unlockable_t *unlockable )
 #ifdef BUILD_CGAME
 static void InformUnlockableStatusChanges( int *statusChanges, int count )
 {
-	char         text[ MAX_STRING_CHARS ];
-	char         *textptr = text;
-	int          unlockableNum;
+	std::string text;
 	bool     firstPass = true, unlocked = true;
-	unlockable_t *unlockable;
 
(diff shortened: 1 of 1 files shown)


src/shared/bg_teamprogress.cpp
```diff
@@ -156,7 +147,7 @@ static void InformUnlockableStatusChanges( int *statusChanges, int count )
 			break;
 	}
 
-	CG_CenterPrint( text, 1.0f );
+	CG_CenterPrint( text.c_str(), 1.0f );
```
DolceTriade, line 150: Do we need a check for max str len or does CG_CenterPrint fail gracefuly? Maybe we should warn?

Otherwise LGTM

necessarily-equal: I would expect the size to be bounded by the number of equipment anyway. I currently is readable on several lines without problem.

I don't think we'll reach some limits in text length in RmlUi anyway.

DolceTriade: Figure it can't hurt. We've been bitten by string length before and while it's probably not going to be a factor now, better to add a check than get bitten by it later...

DolceTriade (approved):

src/shared/bg_teamprogress.cpp
necessarily-equal: `CG_CenterPrint` does `Q_strncpyz( cg.centerPrint, str, sizeof( cg.centerPrint ) );` so this is here already

PR #2465, opened 2023-02-20
Add bot action follow

sweet235:
Action `follow(500)` will make a bot roam in radius 500 around another player. `/tactic` saves the user's client number in the bot's mind. `/bot behavior` can optionally take an argument to specify who to follow. If it is not specified who to follow, the action fails.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -5791,7 +5791,7 @@ static void BotUsage( gentity_t *ent )
 	                                        "            bot del (<name> | all)\n"
 	                                        "            bot names (aliens | humans) <names>…\n"
 	                                        "            bot names (clear | list)\n"
-	                                        "            bot behavior (<name> | <slot#>) <behavior> [<x> <y> <z>]\n"
+	                                        "            bot behavior (<name> | <slot#>) <behavior> [<name> | <slot#> | <x> <y> <z>]\n"
 	                                        "            bot skill <skill level> [<team>]" ) );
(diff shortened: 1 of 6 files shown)


DolceTriade (commented):
https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/tree/master/bots

consider committing the the bts for your actions here

DolceTriade (approved):

src/sgame/sg_admin.cpp
```diff
@@ -5791,7 +5791,7 @@ static void BotUsage( gentity_t *ent )
 	                                        "            bot del (<name> | all)\n"
 	                                        "            bot names (aliens | humans) <names>…\n"
 	                                        "            bot names (clear | list)\n"
-	                                        "            bot behavior (<name> | <slot#>) <behavior> [<x> <y> <z>]\n"
+	                                        "            bot behavior (<name> | <slot#>) <behavior> [<name> | <slot#> | <x> <y> <z>]\n"
```
necessarily-equal, line 5794: maybe you can specify that it's the _followed_ name and _followed_ slot number, because if I didn't know already what it was because of the context I would have a hard time guessing

sweet235: It is not really only the followed name/number. This is a general mechanism to pass arguments to .bt files, albeit restricted to a single vector or a single client number. Other .bt files might do something completly different with it.

I know, I know we should have a mechanism capable of passing more arguments. But this is the minimal interface I need.

PR #2467, opened 2023-02-25
cg_event.cpp: fix colour leak with rocketpod indicator

cu-kai:
this fixes the blue colour of the rocket pod indicator leaking into the rest of the kill message (this only affected emoticon mode).


bmorel (approved):
lgtm, not tested.

necessarily-equal (approved):
LGTM, haven't tested

PR #2469, opened 2023-02-27
Activate ent

bmorel:
This allows mappers to make doors or buttons that can react to `+activate`, if the spawnflag 0x80 is set. Needs a way to test it actually work, though. @cu-kai you might be interested.

### DIFF
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -2181,9 +2181,17 @@ static void ClientThink_real( gentity_t *self )
 		gentity_t *ent;
 
 		// look for object infront of player
+
+		glm::vec3 viewpoint = VEC2GLM( client->ps.origin );
+		viewpoint[2] += client->ps.viewheight;
+		// should take target's radius instead, but let's try with that for now
+		glm::vec3 mins, maxs;
+		BG_BoundingBox( static_cast<class_t>( client->ps.stats[STAT_CLASS] ), &mins, &maxs, nullptr, nullptr, nullptr );
+		auto range1 = ENTITY_USE_RANGE + RadiusFromBounds( &mins[0], &maxs[0] );
+
 		AngleVectors( client->ps.viewangles, view, nullptr, nullptr );
-		VectorMA( client->ps.origin, ENTITY_USE_RANGE, view, point );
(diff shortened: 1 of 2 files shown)


slipher:
Don't use bit fields please. As you have already observed on IRC they have confusing semantics. You will notice that there are currently no bit fields in the Daemon and Unvanquished code bases.

bmorel:
Well, I'm ok to not use bitfields, but the current situation regarding with spawnflags is worst imo, there is currently no way to easily know which flags are taken or what a value is for. Bitfields improve on this situation.
Regarding the problem you're mentioning, yes, I was surprised that compiler said the int could be either -1 or 0, but now it kind of make sense. Using an unsigned fixes this problem, too, and bits are really about 0 or 1 now.
I guess I could use constants instead, but that's still meh imo.

In any case,  I don't intend to merge this until I have a way to test things actually work as expected, so this gives some more time to find a solution that would be easier to maintain than current situation, without the problems of bitfields? Tbh I think it would be better to avoid using the spawnflags at all, and use something else.
In the feature this PR adds, I could for example use a distance. If 0 or negative, disable the `+activate` feature, otherwise use `std::min( distance, ENTITY_USE_RANGE )` for distance. I think this might become needed when movers are too big or too small or misplaced a tiny bit. To understand what I mean, just try the `+activate` trick on various damage entities. Sometimes it works, sometimes it does not, and I'm pretty certain it's because of something along those lines.

slipher:
I don't get what's wrong with some named constants like every other flags thing in the codebase.

bmorel:
@slipher I removed the bitfield.

bmorel:
I think this is ready to merge.
I am not aware of any situation where this does not work, if problems are found later (and I'm expecting it) then we'll fix them later?
I would like either this or #2455 merged soon (too much open PRs is just calling for problems), ideally this one as it superseeds #2455.

cu-kai (approved):
I have tested, though we cannot test func_door_model right now.

Companion commit for NetRadiant's doc: https://github.com/Unvanquished/unvanquished-mapeditor-support/commit/9030c282847be94b731bf14d7f3ad7f16cddc21a

PR #2474, opened 2023-03-06
gameplay: reward players that attack bases

bmorel:
Most played servers, that is, @cu-kai and @sweet235 uses this setting since long. When I implemented it, my servers also used it. So far I've read no complaint about this, so I open this PR to make this the official gameplay.

### DIFF
--- a/src/sgame/sg_combat.cpp
+++ b/src/sgame/sg_combat.cpp
@@ -28,7 +28,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "Entities.h"
 #include "CBSE.h"
 
-Cvar::Cvar<float> g_rewardDestruction( "g_rewardDestruction", "Reward players when they destroy a building by momentum * g_rewardDestruction", Cvar::NONE, 0.f );
+Cvar::Cvar<float> g_rewardDestruction( "g_rewardDestruction", "Reward players when they destroy a building by momentum * g_rewardDestruction", Cvar::NONE, 10.f );
 // damage region data
 damageRegion_t g_damageRegions[ PCL_NUM_CLASSES ][ MAX_DAMAGE_REGIONS ];
 int            g_numDamageRegions[ PCL_NUM_CLASSES ];


DolceTriade:
Lgtm assuming this is the value used on those servers

cu-kai (approved):
Can confirm I've had this set at 10 since the feature was implemented.

slipher:
The argument against this is that it discourages building forward bases. A failed attempt in addition to losing momentum will also feed the enemy players.

cu-kai:
Agreed with @VReaperV 

Human forward bases are already so fragile most players don't bother building them, as you are almost guaranteed to end up losing BP.

bmorel:
> The argument against this is that it discourages building forward bases. A failed attempt in addition to losing momentum will also feed the enemy players

Well, there is already the momentum gain that makes this a very bad idea. And this setting allows to reward people who do not camp, making the game more dynamic. I think it's especially valuable when one plays aliens, as human bases are very sturdy and dangerous, they often kills the attackers, unlike alien bases.

> I imagine the perception of forward bases is skewed by playing too often against bots, with their omniscience of buildings

I disagree. When there are bots, people traditionally do *not* lack money to attack bases. Until very recently, it was impossible to ask bots to do something, and even like this, it's easy to lure "defenders" out of their base to feed on them.
When it's players that decide to camp, the game basically becomes "wait 10 minutes to get 1000¤ and be able to attack anew" because buildings don't reward.

DolceTriade:
Even when playing humans, building forward bases doesn't work well for Humans. Aliens probably only build minor forward bases for a forward booster and that hasn't been super discouraged by this change.

As it stands, this change has been tested quite extensively and there seems to be a consensus that this definitely makes people more likely to rush (which makes the game more fun...).

necessarily-equal (approved):
Sounds good to me.

bmorel:
@slipher : do you accept the consensus, or should this be discussed further? Otherwise I'll merge this week end.

sweet235:
lgtm

Btw, when losing an fwd base: you lose the BP, the enemy gains momentum and now some funds. I think the fund gain is the smallest impact of the three.

bmorel:
> Btw, when losing an fwd base: you lose the BP, the enemy gains momentum and now some funds. I think the fund gain is the smallest impact of the three

You forgot to mention that the team loosing the building also looses momentum, but this one is usually a rather minor annoyance, except when you're alien and just unlocked the advanced granger.

PR #2478, opened 2023-03-10
Fix a navgen error handling bug

slipher:
Due to the wrong return there the error message was not logged and the cached failure file not written.


DolceTriade:
lgtm

PR #2481, opened 2023-03-10
g_teamImbalanceWarnings is a boolean, not a duration

bmorel:


slipher:
I probably broke it by putting the wrong type when migrating it to new-style cvars.

bmorel:
It would be easy enough to make this an integer anew, but anyway is this really useful? I don't think this warning is that useful. The "balance" is just about having the same number of players in both teams, which is obviously wrong, as this does not considers the player's skills at all.

DolceTriade:
Fair enough. LGTM, I don't think making it less spammy is really important. If it is, we can change it back :shrug:

slipher:
This may be closed in favor of #2487

bmorel:
Sure.

PR #2482, opened 2023-03-10
Disabable g_teamImbalanceWarnings

### LINKED ISSUE #2468: "Teams are imbalanced" message is spammy
The "Teams are imbalanced. Humans/Aliens have mroe players" message gets sent 3 times when someone joins a team with more players, caused by: https://github.com/Unvanquished/Unvanquished/blob/6e6ec8b26536da0801fac1ebc8389a6ce730974d/src/sgame/sg_team.cpp#L576-L601 There's no need to spam it, especially since, to my knowledge, no other message gets duplicated like that. Furthermore, it only gets sent when there's more than 2 players difference, which I think should be changed to more than 1.

bmorel:
This warning is of dubious utility considering the fact that server owners who really care about balance should enable `g_teamForceBalance` instead
Fix #2468


DolceTriade:
lgtm

bmorel:
Closed in favor of #2487

PR #2483, opened 2023-03-11
Disable bot flamer by default

### LINKED ISSUE #2411: consider disabling bots to use flamer
Previously, flamer was usable in 1 vs 1 fight (even if this would often end up badly for the flamer user), this is no longer the case.
I think @illwieckz said that the current flamer is a specialized weapon to attack bases.

Bots are unable to plan a proper attack on a base (they're unable to plan anything anyway), to obey to a plan (in vanilla, I know some mods implement that), or even to move in group (again, I know some mods implement changes in that area), hence I think bots should be prevented to use the flamer at all, as this makes them easy, high value preys, breaking AI balance (alien bots do not use weapons they don't know how to use properly).

[This](https://github.com/Unvanquished/Unvanquished/blob/b957b985d5450a5e028fd2bfe048fa2546534847/src/sgame/sg_main.cpp#L257) is the only line that would need to be changed.
I would recommend server owners to at least disable it though configuration files as long as this is not done upstream. This might notably interest @sweet235  and @cu-kai

bmorel:
Currently bots with flamer will only feed the enemy. This change prevents them to buy flamer in the default configuration, it's still very trivial for a server owner to enable this weapon for them.

Fix #2411


DolceTriade:
lgtm

bmorel:
@illwieckz what do you think about this? Asking you specifically because I know you like the current flamer.

necessarily-equal (approved):
LGTM.

(Note I updated the PR's title to reflect what the PR does and not to explain a problem)

PR #2485, opened 2023-03-11
entities without health can't be attacked

### LINKED ISSUE #2470: dretches emits some noise when close enough to a func_door or func_button
I believe this is linked to the changes to re-introduce damage entities. This does not happens on 0.54 servers, only on servers that are close enough to master.
To reproduce with an official map, go as dretch on one of spacetracks' elevators doors. You'll hear the dretch doing some noise.

bmorel:
fix #2470

### DIFF
--- a/src/sgame/sg_weapon.cpp
+++ b/src/sgame/sg_weapon.cpp
@@ -40,7 +40,7 @@ static bool TakesDamages( gentity_t const* ent )
 {
 	return false
 		|| Entities::IsAlive( ent )
-		|| ( ent && ent->s.eType == entityType_t::ET_MOVER )
+		|| ( ent && ent->s.eType == entityType_t::ET_MOVER && ent->health > 0 )
 		;
 }
 


bmorel:
@cu-kai said lgtm on IRC

src/sgame/sg_weapon.cpp
```diff
@@ -40,7 +40,7 @@ static bool TakesDamages( gentity_t const* ent )
 {
 	return false
 		|| Entities::IsAlive( ent )
-		|| ( ent && ent->s.eType == entityType_t::ET_MOVER )
+		|| ( ent && ent->s.eType == entityType_t::ET_MOVER && ent->health > 0 )
```
DolceTriade, line 43: while I think this particular style is an antipattern in the code, I won't require that you change it.

DolceTriade, line 43: Would `https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/Entities.cpp#L82` be a better check?

bmorel: No, because movers are *not* in the CBSE, and so do not have components

bmorel: wdym?

DolceTriade: the 
```
false 
|| cond1
|| cond2
;
```

bmorel: Well, I'm ok to change it, but I would appreciate if you could elaborate on why it's an anti-pattern. According from wikipedia, an anti-pattern is something that: "has more bad consequences than good ones". Tell me, what are the bad consequences of this?

DolceTriade: I guess what I meant by anti pattern is that it's not really used anywhere in the code and I don't think it really provides anything over something like:
```
cond1 || cond2
```
or 
```
cond1
|| cond2
|| cond3;
```

but as I said, I don't want to make this a big enough deal to change, especially when we don't have a codified style to enforce :(

DolceTriade:
lgtm

PR #2487, opened 2023-03-11
Fix g_teamImbalanceWarnings

slipher:
Was broken by putting type bool instead of int upon migration to new-style cvars.


DolceTriade:
Should we disable this by default? I don't think it's a very useful message at all

src/sgame/sg_main.cpp
```diff
@@ -164,7 +164,7 @@ Cvar::Cvar<bool> g_alienAllowBuilding(
 
 Cvar::Cvar<float> g_alienOffCreepRegenHalfLife("g_alienOffCreepRegenHalfLife", "half-life in seconds for decay of creep's healing bonus", Cvar::NONE, 0);
 
-Cvar::Cvar<bool> g_teamImbalanceWarnings("g_teamImbalanceWarnings", "send 'Teams are imbalanced' messages?", Cvar::NONE, true);
+Cvar::Cvar<int> g_teamImbalanceWarnings("g_teamImbalanceWarnings", "send 'Teams are imbalanced' messages every x seconds", Cvar::NONE, 30);
```
DolceTriade, line 167: Also note 0 should disable it?

slipher: Done.

slipher:
> Should we disable this by default? I don't think it's a very useful message at all

I like it :shrug:

DolceTriade:
lgtm

necessarily-equal (approved):
LGTM, haven't tested. I must say I also wanted to do that when I saw the issue.

Fixes https://github.com/Unvanquished/Unvanquished/issues/2468.

PR #2488, opened 2023-03-11
Fix random substitution of particles, fix #1509

### LINKED ISSUE #654: Scoreboard is empty or shows bad content.
Sometimes the game end scoreboard is empty and shows no players even though they are on the server.

It might have happened because I was holding the shift key to pull up the scoreboard when the game ended or it might be unrelated.

### LINKED ISSUE #1509: non-flamer projectile with flame
This is super rare, and I think I never managed yet to take a screenshot, but at least two times since 0.52 release I git my weapon projectile surrounded by flames (maybe it is when using chaingun).

illwieckz:
Fix random substitution of particles, fix:

- https://github.com/Unvanquished/Unvanquished/issues/1509

I noticed there was no error message for `MAX_PARTICLE_SYSTEMS` being hit. I noticed there was an error message for `MAX_TRAIL_SYSTEMS` being hit.

So I wanted to add such error message to debug #1509.

So I decided to copy-paste the `MAX_TRAIL_SYSTEMS` hit message and implement it for `MAX_PARTICLE_SYSTEMS` the cargo cult way.

So I discovered the particle code was not only not printing anything when `MAX_PARTICLE_SYSTEMS` was hit, but was also returning the particle system anyway, so the previous one before `MAX_PARTICLE_SYSTEMS` is hit.

After that we may want to increase `MAX_PARTICLE_SYSTEMS` anyway because with maps like dretchstorm with grenade spam this log message would be spammed a lot.

---

_Edit:_ In fact the code did not return an unitialized particle system or unitialized particle when `MAX_PARTICLE_SYSTEMS` or `MAX_PARTICLES`  was hit, but returned the last known particle system or particle instead. That explains 1. why the particles systems and particles were valid ones but were not the expected ones and 2. we did not experienced any crash.

### DIFF
--- a/src/cgame/cg_particles.cpp
+++ b/src/cgame/cg_particles.cpp
@@ -365,11 +365,12 @@ static particle_t *CG_SpawnNewParticle( baseParticle_t *bp, particleEjector_t *p
 				}
 			}
 
-			break;
+			return p;
 		}
 	}
 
-	return p;
+	Log::Notice( "MAX_PARTICLES hit" );
+	return nullptr;
 }
 
 /*
@@ -497,11 +498,12 @@ static particleEjector_t *CG_SpawnNewParticleEjector( baseParticleEjector_t *bpe
 				Log::Debug( "PE %s created", ps->class_->name );
 			}
 
-			break;
+			return pe;
 		}
 	}
 
-	return pe;
+	Log::Notice( "MAX_PARTICLE_EJECTORS hit" );
+	return nullptr;
 }
 
 /*
@@ -551,11 +553,12 @@ particleSystem_t *CG_SpawnNewParticleSystem( qhandle_t psHandle )
 				Log::Debug( "PS %s created", bps->name );
 			}
 
-			break;
+			return ps;
 		}
 	}
 
-	return ps;
+	Log::Notice( "MAX_PARTICLE_SYSTEMS hit" );
+	return nullptr;
 }
 
 /*


illwieckz:
I also fixed another bug that was hidden by the first one. The code also returned an unitialized particle when `MAX_PARTICLES` was hit.

So, first commit fixes this bug (particle used instead of another particle):

[uninitialized particles](https://dl.illwieckz.net/b/unvanquished/bugs/uninitialized-particles/unvanquished_2023-03-11_224218_000.jpg)

[uninitialized particles](https://dl.illwieckz.net/b/unvanquished/bugs/uninitialized-particles/unvanquished_2023-03-11_225337_000.jpg)

[uninitialized particles](https://dl.illwieckz.net/b/unvanquished/bugs/uninitialized-particles/unvanquished_2023-03-11_225346_000.jpg)

Second commit fixes this bug (dynamic light from unseen particle):

[uninitialized particles](https://dl.illwieckz.net/b/unvanquished/bugs/uninitialized-particles/unvanquished_2023-03-11_231831_000.jpg)

[uninitialized particles](https://dl.illwieckz.net/b/unvanquished/bugs/uninitialized-particles/unvanquished_2023-03-11_232703_000.jpg)

[uninitialized particles](https://dl.illwieckz.net/b/unvanquished/bugs/uninitialized-particles/unvanquished_2023-03-11_233128_000.jpg)

illwieckz:
We may want to run the logger from Notice to Warning because this would be spammy crowded game.

Even if we increase the maxs, we would still hit them, unless we really increase them _a lot_, like _a very lot_.

illwieckz:
Well, the second bug (lights from unseen particle) can reapper if we increase some maxs without understanding how they relate together, as long as maxs is reached, so that second bug don't really comes or not only comes from uninitialized particles. But I'm not surprised this kind of bug happens if all the particles are not properly processed…

With current values I see no issue.

If we increase the values, we may think about it, because in the past (in 2003), Tim reduced them and said his patch saved more than an hundred of MB of ram, and the old values are still not enough for a firebomb spammy game in dretchstorm. We can say that 20 years later we may increase the ram usage, why not, but how much is needed? I don't know how those numbers translate to memory usage.

- https://github.com/darklegion/tremulous/commit/988e67ce2a9c9eea540f6907b6fede1b70a3f3ae

```diff
commit 988e67ce2a9c9eea540f6907b6fede1b70a3f3ae
Author: Tim Angus <***@***>
Date:   Wed Dec 24 03:08:49 2003 +0000

    * Added a creep projection under a locked human
    * Removed some debug messages
    * Severely cut down on static memory usage (130Mb)

diff --git a/src/cgame/cg_local.h b/src/cgame/cg_local.h
index 72ed253f..6d062b1d 100644
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -123,17 +123,17 @@ typedef enum
 } jetPackState_t;
 
 //particle system stuff
-#define MAX_SHADER_FRAMES         64
+#define MAX_SHADER_FRAMES         32
 #define MAX_EJECTORS_PER_SYSTEM   4
 #define MAX_PARTICLES_PER_EJECTOR 4
 
-#define MAX_BASEPARTICLE_SYSTEMS  512
+#define MAX_BASEPARTICLE_SYSTEMS  192
 #define MAX_BASEPARTICLE_EJECTORS MAX_BASEPARTICLE_SYSTEMS*MAX_EJECTORS_PER
_SYSTEM
 #define MAX_BASEPARTICLES         MAX_BASEPARTICLE_EJECTORS*MAX_PARTICLES_P
ER_EJECTOR
 
-#define MAX_PARTICLE_SYSTEMS      64
+#define MAX_PARTICLE_SYSTEMS      48
 #define MAX_PARTICLE_EJECTORS     MAX_PARTICLE_SYSTEMS*MAX_EJECTORS_PER_SYS
TEM
-#define MAX_PARTICLES             MAX_PARTICLE_EJECTORS*8
+#define MAX_PARTICLES             MAX_PARTICLE_EJECTORS*5
 
 #define PARTICLES_INFINITE        -1
 #define PARTICLES_SAME_AS_INITIAL -2

[…]
```

illwieckz:
I tried that:

```diff
diff --git a/src/cgame/cg_local.h b/src/cgame/cg_local.h
index 602d61c8a..a28983a34 100644
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -205,9 +205,9 @@ struct attachment_t
 #define MAX_BASEPARTICLE_EJECTORS (MAX_BASEPARTICLE_SYSTEMS * MAX_EJECTORS_PER_SYSTEM)
 #define MAX_BASEPARTICLES         (MAX_BASEPARTICLE_EJECTORS * MAX_PARTICLES_PER_EJECTOR)
 
-#define MAX_PARTICLE_SYSTEMS      48
+#define MAX_PARTICLE_SYSTEMS      192
 #define MAX_PARTICLE_EJECTORS     (MAX_PARTICLE_SYSTEMS * MAX_EJECTORS_PER_SYSTEM)
-#define MAX_PARTICLES             (MAX_PARTICLE_EJECTORS * 5)
+#define MAX_PARTICLES             (MAX_PARTICLE_EJECTORS * 128)
 
 #define PARTICLES_INFINITE        -1
 #define PARTICLES_SAME_AS_INITIAL -2
```

`MAX_PARTICLE_SYSTEMS 192` is enough to avoid `MAX_PARTICLE_SYSTEMS hit` and dropping particle systems on my firebomb spammy dretchstorm demo (`128` is not enough).

But `MAX_PARTICLES (MAX_PARTICLE_EJECTORS * 128)` is _not enough_ to avoid `MAX_PARTICLES hit` and dropping particles, this makes `MAX_PARTICLES` to `98304` and it's **not enough**. 😱️

illwieckz:
In fact the code did not return an unitialized particle system or unitialized particle when `MAX_PARTICLE_SYSTEMS` or `MAX_PARTICLES`  was hit, but returned the last known particle system or particle instead. That explains why the particles were valid but not the wrong ones and we did not experienced any crash.

illwieckz:
I turned the `MAX_PARTICLE_SYSTEMS hit` and `MAX_PARTICLES hit` messages into Warning instead of notices because there is no way we could 100% avoid them unless we increase those numbers _**a lot**_.

illwieckz:
It would be cool to have a `Log::WarnOnce()` command that would display a given warning only once per game.

DolceTriade:
Nice find

illwieckz:
I found the same bug affecting particle ejectors! 😅️

slipher:
Nice. We might want to think about removing the particle thing closest to expiration instead of not creating a new one, but it's certainly an improvement over the status quo.

> I turned the `MAX_PARTICLE_SYSTEMS hit` and `MAX_PARTICLES hit` messages into Warning instead of notices because there is no way we could 100% avoid them unless we increase those numbers _**a lot**_.

That's the opposite of what you want to do if a message is going to be spammy... using Warn makes it impossible to turn off.

> we increase the values, we may think about it, because in the past (in 2003), timangus reduced them and said his patch saved more than an hundred of MB of ram, and the old values are still not enough for a firebomb spammy game in dretchstorm.

The largest one of those arrays currently is 1.2M so I suspect Timbo was full of crap there with 130MB.

illwieckz:
> Nice. We might want to think about removing the particle thing closest to expiration instead of not creating a new one, but it's certainly an improvement over the status quo.

Hmm, I don't understand the code enough to know what that means but if you know how to add improvements to this code that would be welcome.

> > I turned the `MAX_PARTICLE_SYSTEMS hit` and `MAX_PARTICLES hit` messages into Warning instead of notices because there is no way we could 100% avoid them unless we increase those numbers _**a lot**_.
> 
> That's the opposite of what you want to do if a message is going to be spammy... using Warn makes it impossible to turn off.

Hmm, to me the perfect behaviour would be to only display this kind of message once per game anyway, whatever the log level. Recording and replaying demos can reproduce such issues so only displaying those messages once is enough to debug things and test values. Such proposed mechanism to only display some message once per game would also benefit the MAX messages from:

- https://github.com/DaemonEngine/Daemon/pull/312

> > we increase the values, we may think about it, because in the past (in 2003), timangus reduced them and said his patch saved more than an hundred of MB of ram, and the old values are still not enough for a firebomb spammy game in dretchstorm.
> 
> The largest one of those arrays currently is 1.2M so I suspect Timbo was full of crap there with 130MB.

Ah, that's better then. Anyway I tried a `MAX_PARTICLES` value close to 100 000 and it was not enough. Outside of memory size, there are multiple functions just doing `for (i = 0; i < MAX_PARTICLES; i++)` so multiplying `MAX_PARTICLES` by `100` may be costly. Maybe we need another storage layout but I don't understand that code enough to start a redesign.

illwieckz:
I squashed the 3 commits and turned the warning back to a notice. But I'm concerned about log spam, if log suppression is enabled, those notices will disappear whatever they are notice or warning, but if log suppression is disabled, players will get endless messages even with warning disabled.

DolceTriade:
Log suppression was made default for a reason. Maybe we can make a particle logger if you're worried.

illwieckz:
That can be done later. Anything against the merge?

slipher:
LGTM

DolceTriade:
lgtm

illwieckz:
I don't hit `MAX_PARTICLES` in with my dretchstorm demos and https://github.com/Unvanquished/Unvanquished/pull/2503

Maybe the reason why I was still hitting `MAX_PARTICLES` despite using higher numbers before it was I had only identified 2 bugs on 3 at the time, so maybe some code was still feeding lots of extra particles?

PR #2489, opened 2023-03-12
merge all flee skills into a single one

bmorel:
Impact of this change:

* simpler code maintenance
* easier configuration
* easier balancing
* easier to read skills
* less granularity

Also, I think it's a bit ridiculous to have a bot able to pounce as marauder to flee faster, but no more when using the dragoon class (or vice versa). IMO skills should be made as generic as possible to keep the maintenance cost sane.

### DIFF
--- a/src/sgame/sg_bot_local.h
+++ b/src/sgame/sg_bot_local.h
@@ -113,7 +113,8 @@ enum bot_skill
 
 	// situation awareness and survival
 	BOT_B_PAIN, // basic awareness: notice an enemy if it bites you, or shoots at you
-	BOT_FAST_FLEE,
+	BOT_A_FAST_FLEE,
+	BOT_H_FAST_FLEE,
 	BOT_A_SAFE_BARBS, // don't barb yourself as adv goon
 	BOT_H_BUY_MODERN_ARMOR, // if this is disabled, bot will buy the second-to-last unlocked armor, and will never buy battlesuit
 	BOT_H_PREFER_ARMOR, // prefer to buy armor rather than guns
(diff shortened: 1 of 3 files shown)


bmorel:
ping @sweet235 and @necessarily-equal

necessarily-equal:
I actually think this makes some sense, and I may have added too many flee skills. I think it would be a good idea to do this at least partially: it makes sense that a bot good at fleeing with a class would often also be good at fleeing with another. I don't think we should merge the human and alien skills though; because they have different value to the different teams.

Even if you update the max skill points accordingly, I guess the same remark as in https://github.com/Unvanquished/Unvanquished/pull/2442#issuecomment-1463031280 applies. On the other hand, once you update the max skill points accordingly, this should make alien bots weaker so I think this is a step in the right direction.

What I would recommend going forward, is to wait for Sweet's (and cu-kai's ?) approval on https://github.com/Unvanquished/Unvanquished/pull/2442, then
1. separate again the human and alien skills
2. update the alien skill cost (I would suggest somewhere from 7 to 9 — it's quite powerful)
3. tweak the total skillsetBudgetAliens accordingly
4. preferably do some testing :) In the current state this gives every skill to sk5 aliens which breaks balance even more

bmorel:
> In the current state this gives every skill to sk5 aliens which breaks balance even more

In previous release, all bots had fast flee at SK5 and considering how naive is the implementation I don't think it led to balance breaks.
Especially for humans, because you must keep in mind that human bots walk and run slower than normal humans, for a reason I'm unaware of.

sweet235:
Maybe it would be better to have only one `fast-flee` for both teams, and cvars to enable/disable for each team.

bmorel:
I've split the fast-flee skill anew, even if I consider this is a mistake, it's still better than the current mess.

bmorel:
bump.

bmorel:
any reason to not merge?

DolceTriade:
lgtm if sweet/perturbed approve

sweet235:
lgtm

PR #2493, opened 2023-03-12
correctly handles bot obstacles for func_spawn

bmorel:


cu-kai (approved):
LGTM. (have tested this locally on a map)

DolceTriade (commented):
I thought `self->num()` and `self.s.clientNum` are the same thing?

Can we add more info to the commit message as to why this correctly handles bot obstacles?

slipher:
func_spawn is not a client so clientNum is not set (or is set to -1 or something)

necessarily-equal:
Interesting. When I auto-replaced some of those to use `->num()` in https://github.com/Unvanquished/Unvanquished/commit/a5ad752371a471e38bd628b30710cb249206a910, auto-replace didn't catch these because I replaced the `ent - g_entities` and checked manually the `ps.clientNum` — but not the `s.clientNum` it seems.

necessarily-equal (approved):

necessarily-equal:
Also it would be nice to include an idea of testing procedure for those. This would help the overall testing. Here, I didn't start looking for a map with `func_spawn` to check this is correct. This is still a LGTM from me, because the code can't be less correct than before, but more testing is always nice.

bmorel:
for all the PRs I do which concern map entities, I have a map to easily reproduce and verify the solutions.

Said map is not yet in the test repo though.

PR #2496, opened 2023-03-13
bots basic handling of func_destructable

bmorel:
Adds and removes bot obstacle when a func_destructable is created or destroyed.

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -2787,6 +2787,7 @@ void SP_func_spawn( gentity_t *self )
 
 static void func_destructable_die( gentity_t *self, gentity_t*, gentity_t *attacker, int )
 {
+	G_BotRemoveObstacle( self->num() );
 	trap_UnlinkEntity( self );
 
 	G_RadiusDamage( self->restingPosition, attacker, self->splashDamage, self->splashRadius, self,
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

cu-kai (approved):
LGTM, including the additional change to fix the `func_destructable` not firing its targets.

I haven't tested the obstacle change directly, but this exact fix worked for func_spawn as per #2493.

bmorel:
Sorry @DolceTriade I accidentally added a commit on this PR, that triggers a func_destructable targets. Wanted to do another branch, but got a brain bug I guess.

DolceTriade:
I guess the logic is that func destructible ents should perform their action upon destruction?

LGTM still.

cu-kai:
> I guess the logic is that func destructible ents should perform their action upon destruction?
> 
> LGTM still.

`func_destructable` is supposed to trigger its targets when it is destroyed, yes. This is as per the netradiant documentation:
```
------ PROPERTIES ------
[...]
onDie : Name of the target fired when destroyed. (string (target))
[...]
```

The "onDie" key will need changing in the netradiant docs but that can be done later (also it references "target" at the end, rofl. what a mess).

bmorel:
Apparently (according to @cu-kai ) func_destructable can have targets, which are triggered when the func_destructable (and each time I see this I want to fix the damn word!) is destroyed. This patch (the 2nd commit) implements that.

necessarily-equal:
As in https://github.com/Unvanquished/Unvanquished/pull/2493#issuecomment-1467665834, I think it would be good to give hints on how to test this. At least a map name and place. I'll try to test a bit later if I find some time and a map.

bmorel:
for all the PRs I do which concern map entities, I have a map to easily reproduce and verify the solutions.

PR #2497, opened 2023-03-14
rewrite particle code to make it less vulnerable to human errors

illwieckz:
**cg_particles: define loop iterators in loops**
    
This makes less likely to return an iterated pointer from outside a loop.

The random particle substitution (#1509, #2488) bug would never have been introduced if this principle was implemented as the compiler would have refused to build. Doing this rewrite made me find the third occurrence of that bug, while I initially only found two occurrences of the bug and thought the fix was ready. Without doing this rewrite I would have left one occurrence of that bug.

**cg_particles: improve variable locality**
    
Define variables the closest possible to their usage.

This makes also more explicit when a variable is reused outside of loops on purpose, and more explicit when multiple loops using the same variable name in different parts of the code are actually not using the same variable.

I also added an extra commit:

**cg_particles: unroll a vector component loop**

This removes an iterator, and makes it more obvious for the reader that this code can benefit from advanced vector functions or operators when porting this code to GLM in the future.

### DIFF
--- a/src/cgame/cg_particles.cpp
+++ b/src/cgame/cg_particles.cpp
@@ -144,16 +144,14 @@ Introduce a new particle into the world
 */
 static particle_t *CG_SpawnNewParticle( baseParticle_t *bp, particleEjector_t *parent )
 {
-	int               i;
-	particle_t        *p = nullptr;
 	particleEjector_t *pe = parent;
 	particleSystem_t  *ps = parent->parent;
 	vec3_t            attachmentPoint, attachmentVelocity;
 	vec3_t            transform[ 3 ];
 
-	for ( i = 0; i < MAX_PARTICLES; i++ )
+	for ( int i = 0; i < MAX_PARTICLES; i++ )
 	{
-		p = &particles[ i ];
+		particle_t *p = &particles[ i ];
 
 		//FIXME: the + 1 may be unnecessary
 		if ( !p->valid && cg.clientFrame > p->frameWhenInvalidated + 1 )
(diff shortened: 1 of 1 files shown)


src/cgame/cg_particles.cpp
```diff
@@ -1842,17 +1836,20 @@ void CG_LoadParticleSystems()
...
+	for ( int i = 0; i < numFiles; i++ )
 	{
-		fileLen = strlen( filePtr );
+		int fileLen = strlen( filePtr );
+		char fileName[ MAX_QPATH ];
 		Q_strncpyz( fileName, "scripts/", sizeof fileName );
 		Q_strcat( fileName, sizeof fileName, filePtr );
 		// Log::Notice(_( "...loading '%s'"), fileName );
 		CG_ParseParticleFile( fileName );
+		filePtr += fileLen + 1;
```
illwieckz, line 1852: This change is less obvious.

src/cgame/cg_particles.cpp
```diff
@@ -2099,8 +2095,8 @@ static void CG_GarbageCollectParticleSystems()
 
 		//check systems where the parent cent has left the PVS
 		//( local player entity is always valid )
-		if ( ( centNum = CG_AttachmentCentNum( &ps->attachment ) ) >= 0 &&
-		     centNum != cg.snap->ps.clientNum )
+		int centNum = CG_AttachmentCentNum( &ps->attachment );
+		if ( centNum >= 0 && centNum != cg.snap->ps.clientNum )
```
illwieckz, line 2099: This change is less obvious.

illwieckz (commented):
Because most of it is a lot of noise of very obvious changes, I marked two less obvious change to help reviewing, those less-obvious changes are only part of `improve variable locality` commit.

illwieckz:
For every commit I modified function after function and tested the game function after function with prerecorded demo with lots of particles and I seen no regression.

src/cgame/cg_particles.cpp
```diff
@@ -664,6 +660,7 @@ static void CG_ParseValueAndVariance( const char *token, float *value, float *va
 
 		localValue = atof_neg( valueBuffer, allowNegative );
 
+		char *varEndPointer = nullptr;
 		varEndPointer = strchr( variancePtr, '%' );
```
necessarily-equal, line 664: Could be one line instead of 2 I guess.

necessarily-equal (approved):
LGTM

src/cgame/cg_particles.cpp
illwieckz: Right, I've collapsed many of them but I missed this one! Done.

PR #2498, opened 2023-03-14
Add RmlUi blurall document action

necessarily-equal:
This is needed for https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/174.

### DIFF
--- a/src/cgame/cg_rocket_events.cpp
+++ b/src/cgame/cg_rocket_events.cpp
@@ -55,6 +55,11 @@ static void CG_Rocket_EventShow()
 	Rocket_DocumentAction( CG_Argv( 1 ), "show" );
 }
 
+static void CG_Rocket_EventBlurAll()
+{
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

src/cgame/cg_rocket_events.cpp
```diff
@@ -55,6 +55,11 @@ static void CG_Rocket_EventShow()
 	Rocket_DocumentAction( CG_Argv( 1 ), "show" );
 }
 
+static void CG_Rocket_EventBlurAll()
+{
+	Rocket_DocumentAction( CG_Argv( 1 ), "blurall" );
```
slipher, line 60: blurall doesn't take an argument. The empty string is usually used for the first parameter

necessarily-equal: I followed the convention in the rest of the file I guess.

Fixed.

slipher:
LGTM

PR #2501, opened 2023-03-14
cg_particles: add a dedicated logger for particles

illwieckz:
Add a dedicated logger for particles

Trails already have a dedicated logger, Tremulous had a cvar to disable particle logs (and another one for trail logs).

### DIFF
--- a/src/cgame/cg_particles.cpp
+++ b/src/cgame/cg_particles.cpp
@@ -28,6 +28,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "common/FileSystem.h"
 #include "cg_local.h"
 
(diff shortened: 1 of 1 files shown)


illwieckz:
Question, what is the difference between:

```c++
static Log::Logger logger("cgame.particles", "[Particle Systems]");
```

and:

```c++
static Log::Logger logger = Log::Logger("cgame.particles", "[Particle Systems]");
```

Various parts of the code base use either first wording or second wording.

necessarily-equal:
I would say the first wording is preferable because it would use only the constructor and not both the constructor and the `operator=` function. It's notvreally important though

necessarily-equal (approved):
Looked quickly without testing, LGTM

slipher:
LGTM

PR #2503, opened 2023-03-14
cg_local: increase `MAX_PARTICLE_SYSTEMS` and `MAX_PARTICLES`

illwieckz:
Increase `MAX_PARTICLE_SYSTEMS` and `MAX_PARTICLES`.

⚠️ Merging this code without shipping optimized particles would result on huge performance loss.

See optimized particles:

- https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/15

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -193,21 +193,41 @@ struct attachment_t
 
 //======================================================================
 
-//particle system stuff
+// Maximum .particle files to load.
 #define MAX_PARTICLE_FILES        128
 
 #define MAX_PS_SHADER_FRAMES      32
 #define MAX_PS_MODELS             8
+
+// Maximum ejector definitions in a particle system definition.
 #define MAX_EJECTORS_PER_SYSTEM   4
+// Maximum particle definitions in an ejector definition.
 #define MAX_PARTICLES_PER_EJECTOR 4
 
-#define MAX_BASEPARTICLE_SYSTEMS  192
+/* Maximum particle systems, ejectors and particle definitions to load from .particle files.
+
+One particle file may contain more than one particle system definitions. Making particle system
+definition maximum greater than particle file maximum makes sure it's not obviously low. */
+#define MAX_BASEPARTICLE_SYSTEMS  (MAX_PARTICLE_FILES + 564)
 #define MAX_BASEPARTICLE_EJECTORS (MAX_BASEPARTICLE_SYSTEMS * MAX_EJECTORS_PER_SYSTEM)
 #define MAX_BASEPARTICLES         (MAX_BASEPARTICLE_EJECTORS * MAX_PARTICLES_PER_EJECTOR)
 
-#define MAX_PARTICLE_SYSTEMS      48
+/* Maximum particle systems, ejectors and particle objects to process and render.
+
+Not all particle system, ejector or particle definitions will instantiate particle
+systems, ejectors or particles at a given time.
+
+A single particle system definition can instantiate many particle systems
+and then many ejectors, and a single ejector with a single particle definition
+can instantiate many particles.
+
+The chosen numeric values for those maximums are arbitrary, but making them greater
+than or equal to the maximum of their definitions makes sure they're not obviously low. */
+#define MAX_PARTICLE_SYSTEMS      MAX_BASEPARTICLE_SYSTEMS
 #define MAX_PARTICLE_EJECTORS     (MAX_PARTICLE_SYSTEMS * MAX_EJECTORS_PER_SYSTEM)
-#define MAX_PARTICLES             (MAX_PARTICLE_EJECTORS * 5)
+/* FIXME: Something in the game is preventing to render more than a thousand particle textures.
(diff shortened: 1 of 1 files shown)


illwieckz:
I don't hit `MAX_PARTICLES` in with my dretchstorm demos.

illwieckz:
But I still get `MAX_PARTICLE_SYSTEMS`…

illwieckz:
If I increase `MAX_PARTICLE_SYSTEMS` to avoid it, then I hit `MAX_PARTICLES`…

illwieckz:
Well, I cannot fix this. Increasing `MAX_PARTICLE_SYSTEMS` hurts performance a lot.

illwieckz:
Bumping `MAX_PARTICLE_SYSTEMS` from `48` to `128` makes me going from 95fps to 30fps…

illwieckz:
Interestingly, this code has huge impact on performances. If I reduce `MAX_PARTICLE_SYSTEMS` from `48` to `16` I bump from 95fps to 120fps…

illwieckz:
With `MAX_PARTICLE_SYSTEMS 16` I hit the maximum with a single firebomb. With `24` I cannot throw a firebomb and fire with the rifle without hitting the maximum. In fact there is some case where a single firebomb spawns more than `24` particle systems: just load `plat23` map, spawn as human strafe left to the armoury, buy and throw a firebomb… BAM, more than `24` particle systems…

illwieckz:
So let's take `25` particle systems per firebomb. The maximum credit an human player can get by maximizing its money then selling everything it is already carrying is `3250` credits, let's round this to `3200`. It means a single human can throw `16` firebombs. A single human can then spawn `400` particle systems just by buying and throwing firebomb, I'm not counting the other particles he can spawn with his free rifle.

I tried that, I've set `MAX_PARTICLE_SYSTEMS` to `400` and thrown `16` firebombs , performances went from 178fps to 47fps…

slipher:
I imagine it's not the particle systems per se that eat performance, but rather the rendering effects which may be attached to them such as dynamic lights.

illwieckz:
Yeah, but then, something is wrong, we may spawn 24 particle system per firebomb while the current code only allows for 48 particle systems… that's only 2 firebombs.

Anyway I found this in `src/sgame/sg_weapon.cpp`:

```c++
#define FIREBOMB_SUBMISSILE_COUNT 15
```

So for every grenade, we spawn at least 15 fires, plus some random ignite I guess through fire propagation or something like that…

illwieckz:
> I imagine it's not the particle systems per se that eat performance, but rather the rendering effects which may be attached to them such as dynamic lights.

I'm currently looking for particle systems because of that maximum being so quickly reached, but about performances, every ground fire has 4 particle ejectors: heat haze, smoke, flame, widespread flame…

illwieckz:
By reducing `FIREBOMB_SUBMISSILE_COUNT` to `8` I get ~14 particle systems…

illwieckz:
This looks ready to me.

⚠️ Merging this code without shipping optimized particles would result on huge performance loss.

See optimized particles:

- https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/15

illwieckz:
I've set `#define MAX_PARTICLE_SYSTEMS 192` because this allows someone to throw 16 firebombs and get everything rendered.

If I allow more, I hit some other limitations of the engine or the game: some particles are not rendered while not getting any maximum hit message. I guess there are other limits in other places of the game without any message for when those other maximums are hit.

That looks to be a good accommodation. Being able to throw 16 firebombs while not getting any visual problem is already a good achievement.

illwieckz:
I arbitrarily limited `MAX_PARTICLES` to `1024` because something unknown in the game or the engine is actually preventing to render more than a thousand particle textures.

If I don't do that, once there are more than a thousand particles, only the dynamic lights from a lot of them are rendered, not the textures.

I failed to find what was implementing such limit. Allowing more particles would display buggy things, like particles with dynamic lights but without textures:

[max particles](https://dl.illwieckz.net/b/unvanquished/bugs/max-particles/unvanquished_2023-03-22_035555_000.jpg)

With unoptimized fire (see https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/15 ), a single human was able to reach this limit with 10 simultaneous firebombs (2000 credits), now 20 simultaneous firebombs are needed to reach this limit.

The PR looks ready to me.

bmorel:
I am not particularly fond of the new flame sprites to be honest, but the previous ones were cheap as well, and if this can give us a performance improvement it's very welcome.

bmorel:
Does this require a daemon update to not suffer bad performances?

illwieckz:
> Does this require a daemon update to not suffer bad performances?

Only data changes. The original groundfire particle system was just too heavy for the game.

DolceTriade:
For fire specifially, I wonder if we can use the clustering code to not spawn fires if there are fires in the vicinity to save on particle systems...

bmorel:
I do not think there is any good reason to delay this.
The performance problem *already* exists, and is not directly related to this.

bmorel (approved):

PR #2504, opened 2023-03-15
cherry picks of some betterai BT changes

bmorel:
This PR cherry picks 3 commits allowing bots to go grab ammo and poison when they need to. Needs to be used by BT to have any impact on gameplay.

### DIFF
--- a/src/sgame/sg_bot_parse.cpp
+++ b/src/sgame/sg_bot_parse.cpp
@@ -282,6 +282,11 @@ static AIValue_t randomChance( gentity_t*, const AIValue_t* )
 	return AIBoxFloat( random() );
 }
 
+static AIValue_t resupplyScore( gentity_t* self, const AIValue_t* )
+{
+	return AIBoxFloat( BotGetResupplyScore( self ) );
+}
+
 // Read a cvar, no matter its type
 static AIValue_t cvar( gentity_t*, const AIValue_t *params )
 {
@@ -361,6 +366,7 @@ static const struct AIConditionMap_s
 	{ "percentAmmo",       percentAmmo,       0 },
 	{ "percentHealth",     percentHealth,     1 },
 	{ "random",            randomChance,      0 },
+	{ "resupplyScore",     resupplyScore,     0 },
 	{ "skill",             botSkill,          0 },
 	{ "stuckTime",         stuckTime,         0 },
 	{ "team",              botTeam,           0 },
--- a/src/sgame/sg_bot_parse.cpp
+++ b/src/sgame/sg_bot_parse.cpp
@@ -148,9 +148,36 @@ static AIValue_t haveUpgrade( gentity_t *self, const AIValue_t *params )
 		&& BG_InventoryContainsUpgrade( upgrade, self->client->ps.stats ) );
 }
 
-static AIValue_t percentAmmo( gentity_t *self, const AIValue_t* )
+// This function returns the ratio of 2 ammo values (ammo/clips)
+// for a specific weapon.
+// Only used to reduce duplicated code.
+static AIValue_t Ratio( int current, int max, bool infiniteAmmo )
 {
-	return AIBoxFloat( PercentAmmoRemaining( BG_PrimaryWeapon( self->client->ps.stats ), &self->client->ps ) );
+	if ( max == 0 || infiniteAmmo )
+	{
+		return AIBoxFloat( 1.f );
+	}
+	return AIBoxFloat( current / static_cast<float>( max ) );
+}
+
+// Return emptyness ammo ratio of currently loaded clip
+static AIValue_t percentAmmoClip( gentity_t *self, const AIValue_t* )
+{
+	ASSERT( self && self->client );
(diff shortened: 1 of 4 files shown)


bmorel:
Note that this PR does **not** include a BT action that make bots reload, as I remember that this action required some heavier modifications to work reliably.

bmorel:
I removed the commit that implements `action resupply` as it's already possible to do that with the current API, using `changeGoal` and `moveToGoal` actions.

src/sgame/sg_bot_util.cpp
```diff
@@ -365,7 +365,7 @@ float BotGetResupplyScore( gentity_t *self )
 			ASSERT_UNREACHABLE();
 	}
 	float timeDist = dist / GetMaximalSpeed( self );
-	return ( 1 + 5 * self->botMind->botSkill.level ) * ( 1 - percentAmmo ) / sqrt( timeDist );
+	return ( 1 + 5 * self->botMind->botSkill.level ) * ( needPoison ? 1 : ( 1 - percentAmmo ) ) / sqrt( timeDist );
```
slipher, line 368: The poison thing could be kind of broken for large maps; you may not be able to get to the human base before the poison runs out.

bmorel: Good point.
But this is easy to counter in the BT itself by checking distance to BOOSTER before checking for the score. I believe this is something that should be done more often anyway, notably for the actions `heal` and `equip`.

IIRC the logic here is the same as in `heal` for the distance (or was when I initially implemented this).

src/sgame/sg_bot_parse.cpp
```diff
@@ -148,9 +148,48 @@ static AIValue_t haveUpgrade( gentity_t *self, const AIValue_t *params )
...
+
+// Return emptyness ammo ratio of currently loaded clip
+// TODO: implement for non-primary weapon (such as barbs or future secondary weapons)
+static AIValue_t percentAmmoClip( gentity_t *self, const AIValue_t* )
+{
+	ASSERT( self && self->client );
+	playerState_t const& ps = self->client->ps;
+	weaponAttributes_t const* wpa = BG_Weapon( BG_PrimaryWeapon( ps.stats ) );
+
+	if ( wpa->maxAmmo == 0 && !wpa->infiniteAmmo )
```
slipher, line 172: The warning seems unnecessary as you already check these conditions in `Ratio` and handle them correctly. And it would print every frame here, so if you really want a warning it should be in the weapon parser code.

slipher (commented):
I see it's been simplified a bit by removing goon barb support. Reducing the scope of the resupply score to a single responsibility (only poison rather than poison + barbs) resolves my concern about getting poison not always being feasible. LGTM

src/sgame/sg_bot_parse.cpp
bmorel: Fair. I'll remove this. I guess I was just going paranoid here, or maybe it's simply older than the implementation of `Ratio`. I don't remember history details behind this.

bmorel:
> I see it's been simplified a bit by removing goon barb support. Reducing the scope of the resupply score to a single responsibility (only poison rather than poison + barbs) resolves my concern about getting poison not always being feasible.

No, there is still barb support: if the bot have no need for poison (because it already have poison), then it still might go (or rather stay) at booster for ammo, this is what `bool needAmmo = !( weapon->infiniteAmmo && weapon->maxAmmo == 0 );` is for: there is a need for ammo if either condition is met: no infinite ammo, or maxAmmo != 0. This means that melee attacks won't set needAmmo, but barbs (the only weapon which have regenerating ammo) will.

I think those commits represent almost all the resupply API I used in betterai, with the only exception of the `action reload` which triggers bugs in the memory system. Those bugs i've fixed too, but the changeset is *huge* and I would not consider it ready for merge. It works without known problems, but it does not completely fix bot's memory issues so I'm not happy with it and won't try to push that for now (but maybe I could extract few changes from there that fix some actual bugs in enemy detection? I'll think about it.).

bmorel:
I updated the comments as discussed on IRC, and added support for energy resupply near drills, as when I wrote this drills were not able to resupply energy weapons.

src/sgame/sg_bot_parse.cpp
```diff
@@ -148,9 +148,39 @@ static AIValue_t haveUpgrade( gentity_t *self, const AIValue_t *params )
...
-	return AIBoxFloat( PercentAmmoRemaining( BG_PrimaryWeapon( self->client->ps.stats ), &self->client->ps ) );
+	if ( max == 0 || infiniteAmmo )
+	{
+		return AIBoxFloat( 1.f );
+	}
+	return AIBoxFloat( current / static_cast<float>( max ) );
+}
+
+// Return emptyness ammo ratio of currently loaded clip
+// TODO: implement for non-primary weapon (such as barbs or future secondary weapons)
```
slipher, line 165: Oh there's two more comments here claiming it doesn't work for barbs.

slipher:
LGTM

PR #2508, opened 2023-03-16
fix inverted condition in g_bot_ignoreSmallObstacles

bmorel:
Apparently I inverted a condition when I added the 2 dimensions X and Y, and didn't tested appropriately.


bmorel:
Nevermind, the original patch works, I was fooled by the fact commands in command-line are executed *after* the map started, and thus obstacles are already there (and are thus not moved anymore)

PR #2530, opened 2023-03-20
sgame: listbots list more important info before others

### LINKED ISSUE #2405: `botlist` is unreadable
While the /botlist command seems to have some potential, I find the `botlistold` much more useful on @sweet235 mod. I would suggest upstream to use it. The problem with the vanilla version is that there are much too many info, and those are not presented nicely at all.

bmorel:
fix #2405 

Well, I would not say it makes it readable, but it's a quite big improvement still. ping @sweet235 @necessarily-equal


sweet235:
This is growing like Unix' `ps`. I say we need arguments to control the verbosity. Like `/listbots` ommits the skillset and `/listbots skillset` shows it.

necessarily-equal:
A bit late, but LGTM also

PR #2532, opened 2023-03-21
Introduce FontAwesome-based emoticons

slipher:
Companion: https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/177

Add the possibility to define emoticons based on the icon font glyphs (as opposed to images). Works toward solving https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/issues/173.

TODO: define more emoticons (I'm hoping someone else will do it :laughing:) As with Unicode there is the possibility to generate names based on a data file of official names, although these names may be overly verbose.

Ignore the fire in the screenshot as that is one of our pre-existing emoticons :stuck_out_tongue: 

fa emoticons

### DIFF
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -45,7 +45,7 @@ static const struct {
 	team_t   team;
 } meansOfDeath[] = {
 	// Icon            Envkill Assist? (Team)
-	{ "☠",             false, false, TEAM_NONE   }, // unknown
+	{ "[crossbones]",  false, false, TEAM_NONE   }, // unknown
 	{ "[shotgun]",     false, true,  TEAM_HUMANS },
 	{ "[blaster]",     false, true,  TEAM_HUMANS },
 	{ "[painsaw]",     false, true,  TEAM_HUMANS },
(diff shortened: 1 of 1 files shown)


src/shared/bg_misc.cpp
```diff
@@ -2416,7 +2416,18 @@ weapon_t BG_PrimaryWeapon( int const stats[] )
 	return WP_NONE;
 }
 
-static std::unordered_map<std::string, emoticonData_t, Str::IHash, Str::IEqual> emoticons;
+static emoticonData_t Icon( int codePoint )
+{
+	return { codePoint, {} };
+}
```
DolceTriade, line 2422: Let's put this in a separate file

src/shared/bg_misc.cpp
```diff
@@ -2416,7 +2416,18 @@ weapon_t BG_PrimaryWeapon( int const stats[] )
...
+static std::unordered_map<std::string, emoticonData_t, Str::IHash, Str::IEqual> emoticons {
```
DolceTriade, line 2423: Should write a script to generate from this: https://fontawesome.com/v4/cheatsheet/

src/shared/bg_public.h
```diff
@@ -1777,10 +1777,14 @@ voiceTrack_t *BG_VoiceTrackFind( voiceTrack_t *head, int team,
                                  int pClass, int weapon,
                                  int enthusiasm, int *trackNum );
 
+// Exactly one of the fields is nonzero/nonempty. codePoint if the emoticon is displayed with the
```
DolceTriade, line 1780: Why are emoticons in bg anyways...

PR #2535, opened 2023-03-22
Lesser autojump

bmorel:
Port the "lesser auto-jump" feature from my betterai branch.

It has to be noted that this PR will enable more than just automatically jumping around geometry: it will *also* let bots jump over enemy structures, which includes the barricade.

Speaking about the barricade, I do have commits that properly handle this case, *but* this is out of scope. Simply put, the barricade needs to be considered an obstacle even for friends if it's high enough, otherwise bots will stuck on it. A simple way to see that is to put barricade in plat23's bases, on top of the slopes. This should make tyrant tyrant bots stuck (and overall should annoy any tyrant player a lot obviously).

It has to be said that there is a risk of seeing mantis (notably) doing their deadly head-dance with this change. To prevent this, if it is really desired to, requires to reinject the team test, or at least only ignore enemies.
Myself, I think the deadly head dancing is not a problem. If it is a problem, then players should also be prevented to do so.

### DIFF
--- a/src/shared/navgen/nav.cpp
+++ b/src/shared/navgen/nav.cpp
@@ -916,7 +916,8 @@ static NavgenStatus rasterizeTileLayers( Geometry& geo, rcContext &context, int 
 
 void NavmeshGenerator::StartGeneration( class_t species )
 {
-	LOG.Notice( "Generating navmesh for %s", BG_Class( species )->name );
+	classAttributes_t const& agent = *BG_Class( species );
+	LOG.Notice( "Generating navmesh for %s", agent.name );
 	d_.reset( new PerClassData );
 	d_->species = species;
 	d_->status = initStatus_;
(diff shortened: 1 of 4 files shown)


sweet235:
We have to find a way to force a fresh navgen when `sg_botAutojump` changes.

sweet235:
I use this on my server now, with `set g_botAutojump 0.5`.

slipher:
The setting needs to be added to NavgenConfig for invalidation.

If you want to have specific per-map values then #2419 is the way to go. However I don't think this really plays well with a cvar; it would be unclear which setting is supposed to take precedence. If per-map configurability and a configurable default value are both desired; maybe I should make it possible to write a default config file in #2419 (the per-map config would be applied on top of that). Thoughts?

sweet235:
> The setting needs to be added to NavgenConfig for invalidation.

I agree. I had to do some dirty bit magic on the version number to invalidate.

bmorel:
the cvar thing is only about having some way to configure, I also think a proper config system for navgen would be ideal

sweet235:
Crash is gone, I run this again. The cvar is useful to have, we might want to tweak it after the navgen is finished.

bmorel:
I added a commit for trivial ladder support (requires an offMesh connection manually added), as this feature requires one of the commits of lesser autojump.
My original code to do this was more complex, maybe for no reason. I have tested it on parpax, with the ladder in the "twin-elevators room".

bmorel:
removed the ladder commit.

src/shared/navgen/nav.cpp
```diff
@@ -943,11 +945,27 @@ void NavmeshGenerator::StartGeneration( class_t species )
 	d_->tw = ( gw + ts - 1 ) / ts;
 	d_->th = ( gh + ts - 1 ) / ts;
 
+	float climb = config_.stepSize;
+	if ( sg_botAutojump.Get() > 0.f )
+	{
+		glm::vec3 mins, maxs;
+		BG_BoundingBox( species, &mins, &maxs, nullptr, nullptr, nullptr );
+		float safety = ( maxs.z - mins.z ) * sg_botAutojump.Get();
```
slipher, line 953: Why is the ratio relative to the player's height? I would have thought making it relative to the jump height makes more sense. Dretch has a higher jumpMagnitude than tyrant.

src/shared/navgen/nav.cpp
```diff
@@ -943,11 +945,27 @@ void NavmeshGenerator::StartGeneration( class_t species )
...
+		//FIXME use g_gravity
+		float jump = Square( agent.jumpMagnitude ) / 1600; //( g_gravity.Get() * 2 );
+
+		climb = std::min( safety, climb + jump );
+
+		// keep the STEPSIZE as minimum, just in case
+		climb = std::max( config_.stepSize, climb );
+	}
+
+	Log::Notice( "generating agent %s with stepsize of %d", agent.name, climb );
```
slipher, line 963: Use the logger

src/shared/navgen/nav.cpp
```diff
@@ -943,11 +945,27 @@ void NavmeshGenerator::StartGeneration( class_t species )
 	d_->tw = ( gw + ts - 1 ) / ts;
 	d_->th = ( gh + ts - 1 ) / ts;
 
+	float climb = config_.stepSize;
+	if ( sg_botAutojump.Get() > 0.f )
```
slipher, line 949: The config file is merged so you can store the value of this in the config and customize it per map

bmorel: done.

src/shared/navgen/nav.cpp
bmorel: this is a limitation of recast, stepsize  should not be bigger than agent's height. It has to be said that using climb height to generate jumping connections is a hack, but the alternative is to manually create offmesh connections for jumps, which is unrealistic giving the number of navmeshes we'd need to maintain ( number of maps * number of classes, so ~2500 ).

slipher:
Also NAVMESHSET_VERSION should be bumped since one of the header structs is changed

bmorel:
> Also NAVMESHSET_VERSION should be bumped since one of the header structs is changed

done

sweet235:
lgtm

bmorel:
If no more questions, 'ill merge this tomorrow

src/shared/navgen/nav.cpp
bmorel: Done.

PR #2536, opened 2023-03-22
Shuffle entity state bits to fix base beacons

### LINKED ISSUE #2390: Bases' beacons ("home" and "tents") are wrongly all "home" icons
The effect are different on a team or as a spectator, but it's still wrong in both cases.

A good map to test this is mission-mantis because the default layout is huge. (You can find the map on Sweet's server for example)

slipher:
EF_BC_BASE_OUTPOST was BIT(24) which is out of range for the eFlags netcode.

Fixes #2390. Though I must say it doesn't look so great -- the tent icon has terrible jaggies.

### DIFF
--- a/src/cgame/cg_players.cpp
+++ b/src/cgame/cg_players.cpp
@@ -2505,7 +2505,7 @@ static void CG_PlayerUpgrades( centity_t *cent, refEntity_t *torso )
 	}
 
 	// creep below bloblocked players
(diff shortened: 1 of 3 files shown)


DolceTriade (approved):
I wish there were a static check or something we could have to detect this.

necessarily-equal (approved):

PR #2539, opened 2023-03-24
Ai fight range

bmorel:
This PR notably fixes:

* the painsaw dance, when bots detect a building or an alien on the ceiling while equiped with a painsaw. This also happens with flamer, even if it's rarer for obvious reasons (longer range)
* bots literally cutting their way through friends (and killing said friends for no reason, this is very frustrating to be psawed by a bot while all aliens are far away)
* bots waiting to be at range of their target to be able to damage it, despite blaster being available (and an actual weapon)
* bots unable to jump over or steer around blockers to reach their target. This is not perfect though, if no path can be found they'll still walk endlessly against the obstacle, but this is still a net improvement over current situation

Note that the range handling is not perfect, bots could (and likely should) swap weapons a bit sooner, because weapon swapping takes time, and there is the aforementioned bad obstacle management. I still think this is a huge improvement.

The PR was tested against `eggs-master` 1.1 because this map made those problems very obvious and easy to reproduce.
I think it's ready for reviewing, but I won't merge it until I have run more extensive tests. There might even be some possible cleanup in there.

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -799,6 +799,7 @@ AINodeStatus_t BotActionSay( gentity_t *self, AIGenericNode_t *node )
 // TODO: Move decision making out of these actions and into the rest of the behavior tree
 AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 {
+	bool canSprint = true;
 	team_t myTeam = ( team_t ) self->client->pers.team;
 	botMemory_t* mind = self->botMind;
 
@@ -886,7 +887,8 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 	BotAimAtEnemy( self );
 	glm::vec3 dummy;
 	//TODO when failing to handle obstacle, a different target should be taken
-	BotAvoidObstacles( self, dummy, false );
+	//  for now, at least avoid wasting stamina
+	canSprint = !BotAvoidObstacles( self, dummy, false );
 	BotMoveInDir( self, MOVE_FORWARD );
 
 	if ( inAttackRange )
@@ -932,7 +934,7 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 		BotStandStill( self );
 	}
 
-	BotSprint( self, true );
+	BotSprint( self, canSprint );
 
 	return STATUS_RUNNING;
 }
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -837,7 +837,8 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 	}
 
 	//aliens have radar so they will always 'see' the enemy if they are in radar range
-	if ( myTeam == TEAM_ALIENS && DistanceToGoalSquared( self ) <= Square( g_bot_aliensenseRange.Get() ) )
+	float goalDist = DistanceToGoalSquared( self );
+	if ( myTeam == TEAM_ALIENS && goalDist <= Square( g_bot_aliensenseRange.Get() ) )
 	{
 		mind->enemyLastSeen = level.time;
 	}
@@ -895,21 +896,21 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 	// We are human and we either are at fire range, or have
 	// a direct path to goal
 
-	if ( mind->skillLevel >= 3 && DistanceToGoalSquared( self ) < Square( MAX_HUMAN_DANCE_DIST )
-	        && ( DistanceToGoalSquared( self ) > Square( MIN_HUMAN_DANCE_DIST ) || mind->skillLevel < 5 )
+	if ( mind->skillLevel >= 3 && goalDist < Square( MAX_HUMAN_DANCE_DIST )
+	        && ( goalDist > Square( MIN_HUMAN_DANCE_DIST ) || mind->skillLevel < 5 )
 	        && self->client->ps.weapon != WP_PAIN_SAW && self->client->ps.weapon != WP_FLAMER )
 	{
 		BotMoveInDir( self, MOVE_BACKWARD );
 	}
-	else if ( DistanceToGoalSquared( self ) <= Square( MIN_HUMAN_DANCE_DIST ) ) //we wont hit this if skill < 5
+	else if ( goalDist <= Square( MIN_HUMAN_DANCE_DIST ) ) //we wont hit this if skill < 5
 	{
 		// We will be moving toward enemy, strafing to
 		// the result: we go around the enemy
 		BotAlternateStrafe( self );
 	}
-	else if ( DistanceToGoalSquared( self ) >= Square( MAX_HUMAN_DANCE_DIST ) && self->client->ps.weapon != WP_PAIN_SAW )
+	else if ( goalDist >= Square( MAX_HUMAN_DANCE_DIST ) && self->client->ps.weapon != WP_PAIN_SAW )
 	{
-		if ( DistanceToGoalSquared( self ) - Square( MAX_HUMAN_DANCE_DIST ) < 100 )
+		if ( goalDist - Square( MAX_HUMAN_DANCE_DIST ) < 100 )
 		{
 			BotStandStill( self );
 		}
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -826,12 +826,15 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 		return STATUS_FAILURE;
 	}
 
-	if ( WeaponIsEmpty( BG_GetPlayerWeapon( &self->client->ps ), &self->client->ps ) && myTeam == TEAM_HUMANS )
+	playerState_t const* ps = &self->client->ps;
+	weapon_t primaryWP = static_cast<weapon_t>( ps->weapon );
+	weapon_t equipedWP = BG_GetPlayerWeapon( ps );
+	if ( WeaponIsEmpty( equipedWP, ps ) && myTeam == TEAM_HUMANS )
 	{
 		G_ForceWeaponChange( self, WP_BLASTER );
 	}
 
-	if ( BG_GetPlayerWeapon( &self->client->ps ) == WP_HBUILD )
+	if ( equipedWP == WP_HBUILD )
 	{
 		G_ForceWeaponChange( self, WP_BLASTER );
 	}
@@ -898,7 +901,7 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 
 	if ( mind->skillLevel >= 3 && goalDist < Square( MAX_HUMAN_DANCE_DIST )
 	        && ( goalDist > Square( MIN_HUMAN_DANCE_DIST ) || mind->skillLevel < 5 )
-	        && self->client->ps.weapon != WP_PAIN_SAW && self->client->ps.weapon != WP_FLAMER )
+	        && primaryWP != WP_PAIN_SAW && primaryWP != WP_FLAMER )
 	{
 		BotMoveInDir( self, MOVE_BACKWARD );
 	}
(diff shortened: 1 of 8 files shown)


bmorel:
I just noticed that I forgot to edit titles of commits, mea culpa. I'll edit them before merging, or while fixing stuff the reviewers see.

src/sgame/sg_bot_util.h
```diff
@@ -62,7 +62,7 @@ void  BotAimAtLocation( gentity_t *self, const glm::vec3 &target );
 bool BotEntityIsValidTarget( const gentity_t *ent );
 bool BotEntityIsValidEnemyTarget( const gentity_t *self, const gentity_t *enemy );
 bool BotTargetIsVisible( const gentity_t *self, botTarget_t target, int mask );
-bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target );
+bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target, weapon_t wp );
```
slipher, line 65: What's the point of "AI: BotTargetInAttackRange allows to choose weapon" ?

src/sgame/sg_weapon.cpp
```diff
@@ -46,18 +46,32 @@ static bool TakesDamages( gentity_t const* ent )
 
 void G_ForceWeaponChange( gentity_t *ent, weapon_t weapon )
 {
+	if ( G_Team( ent ) != TEAM_HUMANS ) // only humans have several weapons
+	{
+		return;
+	}
 	playerState_t *ps = &ent->client->ps;
 
+	weapon_t prevWeapon = static_cast<weapon_t>( ps->persistant[ PERS_NEWWEAPON ] );
```
slipher, line 55: I believe you should use BG_GetPlayerWeapon to get the prev weapon

src/sgame/sg_bot_ai.cpp
```diff
@@ -888,7 +885,7 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 
 	if ( mind->botSkill.level >= 3 && goalDist < Square( MAX_HUMAN_DANCE_DIST )
 	        && ( goalDist > Square( MIN_HUMAN_DANCE_DIST ) || mind->botSkill.level < 5 )
-	        && primaryWP != WP_PAIN_SAW && primaryWP != WP_FLAMER )
+	        && primaryInRange )
```
slipher, line 893: Is moving backward with the painsaw desired?

src/sgame/sg_bot_ai.cpp
```diff
@@ -824,7 +824,7 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 
 	//aliens have radar so they will always 'see' the enemy if they are in radar range
 	float goalDist = DistanceToGoalSquared( self );
-	if ( ( myTeam == TEAM_ALIENS && goalDist <= Square( ALIENSENSE_RANGE ) )
+	if ( ( myTeam == TEAM_ALIENS && goalDist <= Square( g_bot_aliensenseRange.Get() ) ) // cf ALIENSENSE_RANGE
```
slipher, line 827: No merge commits please

src/sgame/sg_bot_ai.cpp
```diff
@@ -871,7 +871,7 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 	BotAimAtEnemy( self );
 	BotMoveInDir( self, MOVE_FORWARD );
 
-	if ( inAttackRange || self->client->ps.weapon == WP_PAIN_SAW )
+	if ( inAttackRange )
```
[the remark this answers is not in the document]
bmorel: Dunno. Bots already have aim speed dependent on their skill. But I agree they could use a random modifier to decide wether to shoot or not. This could apply to *any* weapon though. This change notably does not impacts their efficiency against enemy, but it makes it less frustrating for allies as no more stupid death because cut by "friend".
To have watched this patch in action, it honestly is not that aimbotty, as they tend to keep blaster for too long (since they shoot with blaster, they can't swap weapon before cooldown time). It might even make this better.
Need to think more on that topic.

src/sgame/sg_weapon.cpp
bmorel: I used this originally, and results felt really bogus. But maybe this was because of other changes, I don't remember for certain (there's more than a few improvements in that PR).

src/sgame/sg_bot_ai.cpp
bmorel: Yes and no. I guess what would be desired is to go backward as long as your range is superior to your opponent's .

src/sgame/sg_bot_ai.cpp
bmorel: that was just to fix a conflict quickly. I don't like those merge/empty commits neither and will rebase instead.

bmorel:
See [this video](https://user-images.githubusercontent.com/1316300/229504467-faaf71b1-7c94-4544-8b52-34516945000d.mp4) to see this PR in action. As you can see there, it does not makes bots particurlarly "aimbotty", they even uselessly switch to blaster between 2 eggs. This is something that should probably be fixed, later.

The suggested change from `static_cast<weapon_t>( ps->persistant[ PERS_NEWWEAPON ] );` to `BG_GetPlayerWeapon` changes nothing in that situation.
I actually think this function is never called with the conditions required for it to return ps->weapon. The case where `ps->persistant[ PERS_NEWWEAPON ] ` is WP_NONE is handled by this PR so I do not think there is any problem currently.
I would appreciate though to have a better understanding of how the weapon switching is done in practice, there are too many variables involved in this, so I'm a bit confused about it.

src/sgame/sg_weapon.cpp
slipher: The new weapon is set to 0 upon spawning so it will be wrong then at the very least.

src/sgame/sg_bot_ai.cpp
```diff
@@ -869,37 +868,41 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
...
-	if ( inAttackRange || self->client->ps.weapon == WP_PAIN_SAW )
+	if ( equipedInRange )
 	{
 		BotFireWeaponAI( self );
 	}
 
 	if ( myTeam == TEAM_ALIENS )
 	{
-		BotClassMovement( self, inAttackRange );
+		BotClassMovement( self, primaryWP );
```
slipher, line 884: The 2nd argument is supposed to be a bool

src/sgame/sg_bot_ai.cpp
slipher: I doubt that the painsaw range being lower than the opponents is even possible... but then that means the special case is not necessary, so I guess it's OK.

src/sgame/sg_bot_ai.cpp
bmorel: indeed, way a copy/pasta error

src/sgame/sg_bot_ai.cpp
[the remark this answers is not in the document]
bmorel: I honestly think this arbitrary `2 * PAINSAW_RANGE` idea would not bring much.

bmorel:
Current code contains a regression: bots often switch to blaster without apparent reason. This does not makes them unable to fight, but this is still not ready for merging. I'll hunt this bug and mark this as ready for review anew when fixed.

bmorel:
I remove commits that tried to improve range handling, that is, switch to blaster if target is too far, as this can not be done in a clean way until weapon ranges are not put in weaponAttributes_t, which means in configuration files.
Since I already tried in the past such thing but it was rejected, better to just remove the commits requiring that.

src/sgame/sg_bot_ai.cpp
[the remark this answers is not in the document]
bmorel: Guess you're right on that. I'll see to change that.

src/shared/bg_public.h
```diff
@@ -1519,7 +1519,7 @@ int BG_GetPlayerPrice( playerState_t &ps );
 int BG_GetPlayerValue( playerState_t &ps );
 
 bool                    BG_PlayerCanChangeWeapon( playerState_t *ps );
-weapon_t                    BG_GetPlayerWeapon( playerState_t *ps );
+weapon_t                    BG_GetPlayerWeapon( playerState_t const *ps );
 bool                    BG_PlayerLowAmmo( const playerState_t *ps, bool *energy );
```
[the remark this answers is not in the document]
bmorel: nitpick: I think we should avoid creating noise [just for nitpicks](https://www.commitstrip.com/fr/2019/05/29/its-an-art/)

bmorel:
I'm tempted to orphan/drop this, as imo the requested changes changes does not makes the situation worst than it is is, plus the bug I noticed required deeper refactoring. I'll see in few weeks if I change my mind.

slipher:
> requested changes changes does not makes the situation worst than it is is

What does this mean?

bmorel:
I means that I disagree with you that the trigger-bot is worst that current situation.

bmorel:
to be exact, I do have some ideas about how to correctly fix that, and this, for *all* weapons, but the skill_t structure was removed, which I used as base in my betterai improvements taht would have allowed to do that easily

necessarily-equal:
If you want to revert parts of that cleanup commit to bring the struct back, I don't mind

Le 29 avril 2023 23:53:18 GMT+02:00, "Morel Bérenger" ***@***.***> a écrit :

bmorel:
No, I'm tired to cope with that kind of mistakes. This PR works and improve things. It's been discussed since months and now have conflicts. I don't have the strength to constantly struggle to get improvements done while fighting unmaintained regressions in both functionalities and architecture.

bmorel:
Last word on this, what do you guys and gals want for the project? Improvements, or stalling and regressions?
It has been said that psaw bots are now trigger-bots and it's true. It also brings a bug where bots switch too often to blaster because of unrelated bug.
Do you want this as is because can be considered improvements, or not?
I don't intend to touch this code anymore, I might only do the conflict fix and push on "merge" or "close" buttons, by now.

slipher:
I think the switching to blaster logic is something that only exists in your mod, so it's not an obstacle to merging.

The painsaw range thing can be fixed with a 1-line change in BotTargetInAttackRange.

bmorel:
call this mental fatigue if you want, but I no longer feel motivated to maintain what looks to me as improvements.
At this point, the logical course of action to me is to work on my thing (in case of AIs, a tactical FPS AI lib), and include it in my my own fork. This would allow me to only have a diff of 3-5 lines with upstream (the entry point) and would remove all endless and mostly useless discussions.
I personally think that progressive improvements is better than doing things the perfect way on 1st shot, especially if they're improvements (it's ok to block for regressions, though, like those that happened with skillset and navgen... which were not blocked) but my mindset obviously differs from most unv's devs.
The trigger-bot problem as you say, is still a huge improvement over the noisy TK-bot that vanilla currently features and that will try to psaw an alien building that is out of of it's weapons reach.
I also prefer generic code, and adding an exception for the painsaw is directly against this.

To have "efficient" (as in, easy to maintain, that scale with gameplay changes) bots, that are easy to maintain, one thing I "learned" (actually was obvious to me from the start) is that hard-coded symbols such as PSAW_RANGE (or alike) must disappear entirely from the code.
When I tried moves toward that, there was again friction because hard-code is better.

This led me to the point that even changing single lines can sometimes disgust me.

slipher:
Fine I'll do it. #2676

slipher:
All right this LGTM, just needs conflicts fixed

bmorel:
I still think this PR is buggy tbh.

bmorel:
I'm going to host this for a while.
I suspect regressions in that code, maybe I'm wrong, seeing it "for real" is going to be helpful

bmorel:
BTW, I changed the code so that it's no longer *specific* to painsaw, as I think this is non-sense: there are other cases when the "triggerbot" feeling goes, especially when playing human.
So I used a random modifier, which depends on the global skill.

This still remains to be tested, though.

src/sgame/sg_bot_util.cpp
```diff
@@ -1292,7 +1293,7 @@ bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target, weapon_t
 			break;
 		case WP_PAIN_SAW:
 			// Start running the saw when an alien is closeby, not literally in range
-			range = PAINSAW_RANGE + 60;
+			range = PAINSAW_RANGE;
```
slipher, line 1296: Undo this change

src/sgame/sg_bot_util.cpp
```diff
@@ -1376,6 +1377,13 @@ bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target, weapon_t
 	float dist = glm::distance( muzzle, VEC2GLM( trace.endpos ) );
 	range = std::max( range, secondaryRange );
 
+	float blurrDist = ( SkillModifier( self->botMind->skillLevel ) * g_bot_distanceAccuracy.Get() ) / 2;
+	if ( random() <= 0.5 )
+	{
+		blurrDist = -blurrDist;
+	}
+	range = blurrDist;
+
```
slipher, line 1386: A lot of stuff doesn't make sense here.
- The range can be negative (!)
- The cvar is supposed to be a percentage but doesn't work like that 
- Re-rolling a random number every frame would result in starting and stopping shooting more frequently, not less.

How about saving this for another PR so we can merge the months-old stuff already?

PR #2542, opened 2023-03-24
AI: less stubborn bots

bmorel:
This PR allows bots to change their target if they detect a matching target closer than the original one.
There is a default setting of 5s between rescans.

To test this:

* place a telenode somewhere near a wall (and remove other ones)
* place an armory on the other side of the wall, so that the distance is very short. A navmesh between node and armoury is, ofc, required
* place a 2nd armory so that it's on the path between 1st arm and node, but have a "ghost distance"  higher than for the 1st armory
* add a human bot: it will try to go at first armory, but will change it's mind to use the closer one on the road

This PR is still a draft because of lack of documentation and insufficiently testing. Development tests were done on `eggs-master` 1.1 since this map have a perfect shape to demonstrate the feature.


bmorel:
Just realised that there is a major risk of bots looping endlessly between 2 targets if they are "well placed".

bmorel:
This PR makes it too easy imo to make bots circle endlessly. Needs a lot more work, and considering the currently broken bots I don't have will to work on this.

PR #2544, opened 2023-03-26
maprotation: add the "between" operator

bmorel:
Currently, map rotation only accepts operators <, = and >, which makes it annoying to allow a map only if the number of clients or players is between a specific range.
This implements this feature.

Example use:
if numClients between 4 10 plat23


slipher:
A more general alternative: #2549

PR #2545, opened 2023-03-26
remove useless \n

bmorel:


src/shared/bg_parse.cpp
```diff
@@ -2328,10 +2320,6 @@ void BG_ParseMissileDisplayFile( const char *filename, missileAttributes_t *ma )
 #endif
 			defined |= IMPACT_FLESH_SND;
 		}
-		/*else
-		{
-			Log::Warn( "%s: unknown token '%s'\n", filename, token );
```
illwieckz, line 2333: I wonder why those warnings are disabled in the first place…

bmorel: so, shuold I let this dead code around?

illwieckz (approved):
LGTM.

PR #2547, opened 2023-03-26
Glm pmove

bmorel:
migrate platerState_t, pmove_t and pm_t to glm.

This does not completely migrates the use of those to glm though, the goal is only to migrate the higher level interface, next changes will be made simpler this way.

Edit: actually migrates all shared and sgame structs to glm, for vec3_t (vec2_t is already done). There might be other old vec api around but I don't thing so.
Again: I have *not* migrated the functions, this should come later in a more progressive pace.


bmorel:
I've disabled the DEPRECATE code because it's only effect is to make migration harder.
There is already enough work as it is.

bmorel:
Switching this to "ready for review", but I know there are still bugs, notably for some reason the commit `glm: playerState_t` let me go out of map as dretch while wallwalking randomly on plat23.

I'll probably split everything in smaller chunks to help the review process, and maybe we should even merge it in smaller chunks, too.

bmorel:
Current status: each commit builds individually, so that bisecting is doable.

There is a known bug in `643e95d13 glm: playerState_t::velocity` which lets players move out of geometry when wallwalking.

I have not tested further, since that commit is buggy, it would be a bit pointless. I have no idea yet why it is buggy (obviously).
I suspect a floating point precision error, but that still have to be proven, it might also be an incorrect assertion from my part that a daemon's function does the same as it's glm "equivalent".
It could also be related to the previous commit "grapplepoint" but I could not get the problem when testing the grapplebpoint build.

src/shared/bg_pmove.cpp
```diff
@@ -303,7 +303,12 @@ PM_ClipVelocity
 Slide off of the impacting surface
 ==================
 */
-void PM_ClipVelocity( const vec3_t in, const vec3_t normal, vec3_t out )
+static glm::vec3 PM_ClipVelocity( const glm::vec3& in, const glm::vec3& normal )
+{
+	return in - glm::dot( in, normal );
```
necessarily-equal, line 308: I think this should be:
```suggestion
	return in - glm::dot( in, normal ) * normal;
```

This could be the reason why you have seen a player going through the ground

necessarily-equal (dismissed):
I think I may have found your bug. Here's a review:

src/sgame/sg_admin.cpp
```diff
@@ -5170,7 +5170,7 @@ bool G_admin_builder( gentity_t *ent )
 	glm::vec3 end = start + 1000.f * forward;
 
 	//TODO there are many occurrences of trap_Trace without mins/maxs, a specialized API is likely needed
```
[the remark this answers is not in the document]
bmorel: it's not done.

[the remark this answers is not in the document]
necessarily-equal: If the change you did at the line below is not what you had in mind, then maybe you might want to edit the comment to make it more clear then. Anyway, this isn't a problem

src/shared/bg_misc.cpp
```diff
@@ -2618,12 +2619,12 @@ static const NetcodeTable playerStateFields =
...
-	{ PSF( origin[ 2 ]       ), 0                , 0 },
-	{ PSF( velocity[ 0 ]     ), 0                , 0 },
-	{ PSF( velocity[ 1 ]     ), 0                , 0 },
-	{ PSF( velocity[ 2 ]     ), 0                , 0 },
+	{ STR( origin[ 0 ] ), offsetof( playerState_t, origin ) + offsetof( glm::vec3, x ), 0, 0 },
+	{ STR( origin[ 1 ] ), offsetof( playerState_t, origin ) + offsetof( glm::vec3, y ), 0, 0 },
+	{ STR( origin[ 2 ] ), offsetof( playerState_t, origin ) + offsetof( glm::vec3, z ), 0, 0 },
+	{ STR( velocity[ 0 ] ), offsetof( playerState_t, velocity ) + offsetof( glm::vec3, x ), 0, 0 },
+	{ STR( velocity[ 1 ] ), offsetof( playerState_t, velocity ) + offsetof( glm::vec3, y ), 0, 0 },
+	{ STR( velocity[ 2 ] ), offsetof( playerState_t, velocity ) + offsetof( glm::vec3, z ), 0, 0 },
```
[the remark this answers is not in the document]
bmorel: if I could avoid using a different macro I would have done so.

src/shared/bg_pmove.cpp
bmorel: no, because this is in a different commit, and the bug starts with 643e95d13020836451e271e60f4f74bc74f9383c

necessarily-equal: I didn't check if it cause The Big Bug™ or not, but I'm pretty sure it's a big bug anyway.

src/shared/bg_misc.cpp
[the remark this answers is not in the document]
necessarily-equal: I confirm that this is working: https://github.com/necessarily-equal/Unvanquished/commit/7d7e750ec0c974b87d72a31bba3eaa6cc7d4edd1#diff-4bf829addb5131de8b85e3b9c7cdf651b2f9a698ccec51011d5cf89072f29c18

I know that the array operator (e.g. `origin[0]`) doesn't work, but that's not what I'm suggesting here

src/shared/bg_pmove.cpp
bmorel: have you read the code of the original PM_ClipVelocity? This is merely a direct translation of it.

slipher: The new code looks wrong. Subtracting a scalar from a vector is almost certainly not what you want.

bmorel: indeed, afontain was right on this one, looks like I forgot to add the "* normal" part

bmorel:
reordered commits so that the one introducing the bug is the 2nd one. Considering what the 1st is, it means the problem obviously come from this specific commit.
I hope I missed a buggy change there, because otherwise it means that either:

* glm is buggy;
* daemon's math API is buggy;
* we have another `-ffast-math` problem?

In any of those cases, it means there are bugs in other parts of the code, no idea how severe they could be. I really hope it's linked to a change I did...

bmorel:
So, apparently the problems reveals in those lines:

```
    if ( trace.fraction > 0 )
    {    
      // actually covered some distance
      VectorCopy( trace.endpos, pm->ps->origin );
    }    
```

And happens 100% times at the `glm: playerState_t::velocity` commit, the moment one starts wallwalking. This makes it rather easy to debug, so I should have this fixed soon.

slipher:
Our `VectorNormalize` returns the 0 vector for a 0 input, whereas I imagine that's undefined in GLM. Maybe that's the difference?

bmorel:
> Our VectorNormalize returns the 0 vector for a 0 input, whereas I imagine that's undefined in GLM. Maybe that's the difference?

Thanks for the hint!
I didn't thought about that, yes, it's likely the problem: obviously, if the vector have a length of 0 it can not be normalized, and that would explain the NaN situation which itself leads to being moved "far away".
Well, I didn't knew neither than daemon's would default to 0 in those cases, which is something I would easily classify as a bug.
GLM's and GLSL docs do not explicitly say what the code does in that case. I can not easily check right now, but I will when I'm back at home.

This also means that since several calls to `VectorNormalize` were moved to `glm::normalize`, we might have other weird issues on that topic, this should probably be checked.

PR #2549, opened 2023-03-27
Allow complex logical expressions in map rotation conditions

slipher:
This will allow more complex conditions like `numPlayers >= 6 && lastWin Aliens` in a map rotation.

Also I added arithmetic operators. There's probably no use for them now but it's just really easy to add more operators so I added them. If we added a 16-bit random number or something then we could use modulo for probabilities other than 50%.

### DIFF
--- a/src/sgame/sg_maprotation.cpp
+++ b/src/sgame/sg_maprotation.cpp
@@ -753,7 +753,6 @@ Print the parsed map rotations
 void G_PrintRotations()
 {
 	int i, j;
-	int size = sizeof( mapRotations );
 
 	Log::Notice( "Map rotations as parsed:" );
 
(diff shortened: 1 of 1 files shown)


slipher:
Just realized that using `COM_Parse2` will probably parse a goto in an if like `#mainloop` as 2 tokens so I need to work around that somehow.

bmorel:
nice, because the current rotation stuff is barely useful at all. Also, <= and >= operators will be much less annoying to use than < and > when you want a min an max $something, as no need to remember to add or remove 1 to the number.

slipher:
I fixed the issue with `#` being parsed as one token.

src/sgame/sg_maprotation.cpp
```diff
@@ -268,126 +247,367 @@ static bool G_ParseMapCommandSection( mrNode_t *node, const char **text_p )
...
+
+using mrValue_t = float;
 
-	if ( !*token )
+static bool ParseValue( const char *token, mrValue_t &result )
+{
+	if ( Str::ToFloat( token, result ) )
 	{
-		return false;
+		// yay
```
DolceTriade, line 291: nit: just return true

DolceTriade (approved):

src/sgame/sg_maprotation.cpp
slipher: Done

PR #2556, opened 2023-03-30
Spiker rework

### LINKED ISSUE #1722: spiker shoot too soon
Spikers try to hit enemies even when the likelihood to hit is low: 9 misses for 1 hit on the following screenshot, while the human is *not* moving:

2022-01-29-132030_1362x748_scrot

bmorel:
This PR is a crude spiker rework.

There are 2 ideas in here:

1. only fire spikes that might hit an enemy, instead of spikes that might not hit a friend. This removes A LOT of friendly fire
2. number of fired spikes impact the cooldown time: the less spikes sent, the faster the spiker can shoot anew

It probably needs more work, notably a real spike count, but this should be enough to test if the idea is worth working on.

Fix #1722

### DIFF
--- a/src/sgame/components/SpikerComponent.cpp
+++ b/src/sgame/components/SpikerComponent.cpp
@@ -168,7 +168,7 @@ bool SpikerComponent::SafeToShoot(glm::vec3 direction) {
 		trap_Trace( &trace, entity.oldEnt->s.origin, mins, maxs, &end[0], entity.oldEnt->num(), ma->clipmask, 0 );
 		gentity_t* hit = &g_entities[trace.entityNum];
 
-		if (hit && ( not use_experiments && G_OnSameTeam( entity.oldEnt, hit ) ) || ( use_experiments && G_Team( hit ) == TEAM_HUMANS ) ) {
+		if (hit && ( ( not use_experiments && G_OnSameTeam( entity.oldEnt, hit ) ) || ( use_experiments && G_Team( hit ) == TEAM_HUMANS ) ) ) {
 			return true;
 		}
 	}
--- a/src/sgame/components/SpikerComponent.cpp
+++ b/src/sgame/components/SpikerComponent.cpp
@@ -10,7 +10,7 @@ constexpr int   COOLDOWN = 5000;
 constexpr float DAMAGE_THRESHOLD = 60.0f;
 /** Total number of missiles to shoot. Some might be skipped if they endanger allies.
  *  Actual values is +/- MISSILEROWS. */
-constexpr int   MISSILES = 26;
+constexpr int   MISSILES = 39;
 /** Number of rows from which to launch spikes. */
 constexpr int   MISSILEROWS = 4;
 /** An estimate on how far away spikes are still relevant, used to upper bound computations. */
@@ -166,12 +166,12 @@ bool SpikerComponent::SafeToShoot(glm::vec3 direction) {
 		trap_Trace( &trace, entity.oldEnt->s.origin, mins, maxs, &end[0], entity.oldEnt->num(), ma->clipmask, 0 );
 		gentity_t* hit = &g_entities[trace.entityNum];
 
-		if (hit && G_OnSameTeam(entity.oldEnt, hit)) {
-			return false;
+		if (hit && G_Team( hit ) == TEAM_HUMANS) {
+			return true;
 		}
 	}
 
-	return true;
+	return false;
 }
 
 bool SpikerComponent::Fire() {
@@ -211,6 +211,7 @@ bool SpikerComponent::Fire() {
 	PerpendicularVector(rotAxis, zenith);
 
 	// Distribute and launch missiles.
+	int launched = 0;
 	for (int row = 0; row < MISSILEROWS; row++) {
 		// Set the base altitude and get the perimeter for the current row.
 		float rowAltitude = (((float)row + 0.5f) * M_PI_2) / (float)MISSILEROWS;
@@ -242,6 +243,7 @@ bool SpikerComponent::Fire() {
 			if (!fire) {
 				continue;
 			}
+			++launched;
 
 			G_SpawnMissile(
 				MIS_SPIKER, self, self->s.origin, dir, nullptr, G_FreeEntity,
@@ -249,7 +251,7 @@ bool SpikerComponent::Fire() {
 		}
 	}
 
-	restUntil = level.time + COOLDOWN;
+	restUntil = level.time + COOLDOWN * launched * MISSILES;
 	RegisterSlowThinker();
 
 	return true;
--- a/src/sgame/components/SpikerComponent.cpp
+++ b/src/sgame/components/SpikerComponent.cpp
@@ -4,13 +4,14 @@
 
 static Log::Logger logger("sgame.spiker");
 
+static Cvar::Cvar<bool> sg_spiker_experiments( "sg_spiker_experiments", "wether to use experimental reloading for spikers or not", Cvar::NONE, true );
 /** Delay between shots. */
 constexpr int   COOLDOWN = 5000;
 /** Shoot prematurely whenever this much damage is expected. */
 constexpr float DAMAGE_THRESHOLD = 60.0f;
 /** Total number of missiles to shoot. Some might be skipped if they endanger allies.
  *  Actual values is +/- MISSILEROWS. */
-constexpr int   MISSILES = 39;
+static Cvar::Cvar<int> sg_spiker_missiles( "sg_spiker_missiles", "number of missiles spikers have", Cvar::NONE, 39 );
 /** Number of rows from which to launch spikes. */
 constexpr int   MISSILEROWS = 4;
 /** An estimate on how far away spikes are still relevant, used to upper bound computations. */
@@ -96,7 +97,7 @@ void SpikerComponent::Think(int /*timeDelta*/) {
 		float hitEdge      = bboxEdge + ((1.0f / M_ROOT3) * ma->size); // Add half missile edge.
 		float hitArea      = hitEdge * hitEdge; // Approximate area resulting in a hit.
 		float effectArea   = 2.0f * M_PI * distance * distance; // Area of a half sphere.
-		float damage       = (hitArea / effectArea) * (float)MISSILES * spikeDamage;
+		float damage       = (hitArea / effectArea) * (float)sg_spiker_missiles.Get() * spikeDamage;
 
 		// Sum up expected damage for all targets, regardless of whether they are in sense range.
 		expectedDamage += damage;
@@ -158,6 +159,7 @@ bool SpikerComponent::SafeToShoot(glm::vec3 direction) {
 	trace_t trace;
 	vec3_t mins, maxs;
 	glm::vec3 end = VEC2GLM( entity.oldEnt->s.origin ) + (SPIKE_RANGE * direction);
+	bool use_experiments = sg_spiker_experiments.Get();
 
 	// Test once with normal and once with inflated missile bounding box.
 	for (float traceSize : {missileSize, missileSize * SAFETY_TRACE_INFLATION}) {
@@ -166,7 +168,7 @@ bool SpikerComponent::SafeToShoot(glm::vec3 direction) {
 		trap_Trace( &trace, entity.oldEnt->s.origin, mins, maxs, &end[0], entity.oldEnt->num(), ma->clipmask, 0 );
 		gentity_t* hit = &g_entities[trace.entityNum];
 
-		if (hit && G_Team( hit ) == TEAM_HUMANS) {
+		if (hit && ( not use_experiments && G_OnSameTeam( entity.oldEnt, hit ) ) || ( use_experiments && G_Team( hit ) == TEAM_HUMANS ) ) {
 			return true;
 		}
 	}
@@ -218,7 +220,7 @@ bool SpikerComponent::Fire() {
 		float rowPerimeter = 2.0f * M_PI * cosf(rowAltitude);
 
 		// Attempt to distribute spikes with equal expected angular distance on all rows.
-		int spikes = (int)round(((float)MISSILES * rowPerimeter) / totalPerimeter);
+		int spikes = (int)round(((float)sg_spiker_missiles.Get() * rowPerimeter) / totalPerimeter);
 
 		// Launch missiles in the current row.
 		for (int spike = 0; spike < spikes; spike++) {
@@ -250,8 +252,7 @@ bool SpikerComponent::Fire() {
 				level.time + (int)(1000.0f * SPIKE_RANGE / (float)BG_Missile(MIS_SPIKER)->speed));
 		}
 	}
-
-	restUntil = level.time + COOLDOWN * launched * MISSILES;
+	restUntil = level.time + COOLDOWN * ( sg_spiker_experiments.Get() ? launched * sg_spiker_missiles.Get() : 1 );
 	RegisterSlowThinker();
 
 	return true;


illwieckz:
> number of fired spikes impact the cooldown time: the less spikes sent, the faster the spiker can shoot anew

I like the idea that the spiker attack is less previsible, and then less mechanical, it's expected to be a living organism!

Viech:
Does it look good if the spiker animation shows a volley but only few spikes are actually launched? If not, maybe launch additional spikes, nudged away from friends and/or towards enemies, until some minimum is reached. The mechanic in theory sounds good but it would be great if it remains immersive, i.e. matches the animation.

I personally think it's better to test this change and eventually make it the only mechanic than to have multiple "experimental" mechanics with a switch cvar. Your experiment might become the default mechanic for many years and then we'd have dead code and a weirdly named cvar left over.

bmorel:
> Does it look good if the spiker animation shows a volley but only few spikes are actually launched? 

I do not remember for certain, but IIRC one barely notices that detail, if at all. And certainly not the human which is targeted, since better to not idle in front of that spiker now, contrary to current situation.

> I personally think it's better to test this change and eventually make it the only mechanic

Agree.

src/sgame/components/SpikerComponent.cpp
```diff
@@ -4,13 +4,14 @@
...
 
+static Cvar::Cvar<bool> sg_spiker_experiments( "sg_spiker_experiments", "wether to use experimental reloading for spikers or not", Cvar::NONE, true );
 /** Delay between shots. */
 constexpr int   COOLDOWN = 5000;
 /** Shoot prematurely whenever this much damage is expected. */
 constexpr float DAMAGE_THRESHOLD = 60.0f;
 /** Total number of missiles to shoot. Some might be skipped if they endanger allies.
  *  Actual values is +/- MISSILEROWS. */
-constexpr int   MISSILES = 26;
+static Cvar::Cvar<int> sg_spiker_missiles( "sg_spiker_missiles", "number of missiles spikers have", Cvar::NONE, 39 );
```
Viech, line 14: If this was kept, then it should be `g_spikerMissiles` as opposed to `sg_spiker_missiles`.

I am wary of allowing server operators to fine tune **combat** balance values without packaging a mod. This runs the risk of giving everyone, including us developers, an unclear picture of the current state of the game. For players in particular it runs the risk of producing "false muscle memory" (not being able to intuitively judge your own and the enemies strength). This is also not done for any other setting that immediately affetcs damage, afair. So I think this should either remain hardcoded or be a field in a config file so that at least one can tell that they are playing on an arbitrarily modded server.

src/sgame/components/SpikerComponent.cpp
```diff
@@ -242,14 +245,14 @@ bool SpikerComponent::Fire() {
...
+			++launched;
 
 			G_SpawnMissile(
 				MIS_SPIKER, self, self->s.origin, dir, nullptr, G_FreeEntity,
 				level.time + (int)(1000.0f * SPIKE_RANGE / (float)BG_Missile(MIS_SPIKER)->speed));
 		}
 	}
-
-	restUntil = level.time + COOLDOWN;
+	restUntil = level.time + COOLDOWN * ( sg_spiker_experiments.Get() ? launched * sg_spiker_missiles.Get() : 1 );
```
Viech, line 255: It seems this PR is not up to date with whatever changes you have tested locally. Could you update to your latest, functional version?

Viech (changes requested):

src/sgame/components/SpikerComponent.cpp
bmorel: done. The problem was basically about the PR being months old, and in the meantime, we merged the data submodule. Which means I'd be able to put the other changes I initially did inside this PR, actually.

src/sgame/components/SpikerComponent.cpp
bmorel: > If this was kept, then it should be g_spikerMissiles as opposed to sg_spiker_missiles.

That's a nitpicking, but sure.

> I am wary of allowing server operators to fine tune combat balance values without packaging a mod. This runs the risk of giving everyone, including us developers, an unclear picture of the current state of the game. For players in particular it runs the risk of producing "false muscle memory" (not being able to intuitively judge your own and the enemies strength). This is also not done for any other setting that immediately affetcs damage, afair. So I think this should either remain hardcoded or be a field in a config file so that at least one can tell that they are playing on an arbitrarily modded server.

You are wrong here.
This code being on sgame means there is no need for a mod owner to share the change. Using a cvar for this is even the only way to communicate about the change in a reliable manner.
We also have seen the effect of that conservatism politics: no changes were done for years on 1st thrown values such as the ground flame stuff.
To balance things more easily, we need to enable server owners to change them without recompiling anything.

The "muscular memory" you mention is a decent argument, though, but I do know rather old games such as red-eclipse who indicate how much a server derives from the official settings. We could, and likely *should* do that, too, possibly in a better way, but this is out of topic.

src/sgame/components/SpikerComponent.cpp
bmorel: actually only 3 weeks old, but the initial patch is from february

bmorel:
@Viech could you see if your requested changes still applies? I think dismissing those by PR authors is kind of rude, so won't do that. I think this is ready for *testing* but so far only one server have it, and it's not played much (I don't put much efforts in my server, have to say) and so could not get any serious feedback. Should not be merged before getting some.

bmorel:
@Viech I'd appreciate if you could look at the changes you requested

DolceTriade:
Before we do code review, can we get this tested in a game? It's hard to determine the effects a change can have without a gameplay test? WDYT @bmorel

bmorel:
It can already be tested on deadbeef since few days (weeks?)

Viech:
I don't see that you have made changes apart from possibly rebasing the commits on master?

bmorel:
Ah, well, I think it's what I have on deadbeef, but since I know embed the patches directly it's easy to check, gimme one minute.

bmorel:
```
From 2656984598d7e124344073f15b4962da0e412b5a Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Morel=20B=C3=A9renger?= <berengermorel76@gmail.com>
Date: Sun, 5 Mar 2023 16:22:46 +0100
Subject: [PATCH 16/44] gameplay: make spiker experiments configurable

---
 src/sgame/components/SpikerComponent.cpp | 13 +++++++------
 1 file changed, 7 insertions(+), 6 deletions(-)

diff --git a/src/sgame/components/SpikerComponent.cpp b/src/sgame/components/SpikerComponent.cpp
index 519784f77..eff326d90 100644
--- a/src/sgame/components/SpikerComponent.cpp
+++ b/src/sgame/components/SpikerComponent.cpp
@@ -4,13 +4,14 @@
 
 static Log::Logger logger("sgame.spiker");
 
+static Cvar::Cvar<bool> sg_spiker_experiments( "sg_spiker_experiments", "wether to use experimental reloading for spikers or not", Cvar::NONE, true );
 /** Delay between shots. */
 constexpr int   COOLDOWN = 5000;
 /** Shoot prematurely whenever this much damage is expected. */
 constexpr float DAMAGE_THRESHOLD = 60.0f;
 /** Total number of missiles to shoot. Some might be skipped if they endanger allies.
  *  Actual values is +/- MISSILEROWS. */
-constexpr int   MISSILES = 39;
+static Cvar::Cvar<int> sg_spiker_missiles( "sg_spiker_missiles", "number of missiles spikers have", Cvar::NONE, 39 );
 /** Number of rows from which to launch spikes. */
 constexpr int   MISSILEROWS = 4;
 /** An estimate on how far away spikes are still relevant, used to upper bound computations. */
@@ -96,7 +97,7 @@ void SpikerComponent::Think(int /*timeDelta*/) {
 		float hitEdge      = bboxEdge + ((1.0f / M_ROOT3) * ma->size); // Add half missile edge.
 		float hitArea      = hitEdge * hitEdge; // Approximate area resulting in a hit.
 		float effectArea   = 2.0f * M_PI * distance * distance; // Area of a half sphere.
-		float damage       = (hitArea / effectArea) * (float)MISSILES * spikeDamage;
+		float damage       = (hitArea / effectArea) * (float)sg_spiker_missiles.Get() * spikeDamage;
 
 		// Sum up expected damage for all targets, regardless of whether they are in sense range.
 		expectedDamage += damage;
@@ -158,6 +159,7 @@ bool SpikerComponent::SafeToShoot(glm::vec3 direction) {
 	trace_t trace;
 	vec3_t mins, maxs;
 	glm::vec3 end = VEC2GLM( entity.oldEnt->s.origin ) + (SPIKE_RANGE * direction);
+	bool use_experiments = sg_spiker_experiments.Get();
 
 	// Test once with normal and once with inflated missile bounding box.
 	for (float traceSize : {missileSize, missileSize * SAFETY_TRACE_INFLATION}) {
@@ -166,7 +168,7 @@ bool SpikerComponent::SafeToShoot(glm::vec3 direction) {
 		trap_Trace( &trace, entity.oldEnt->s.origin, mins, maxs, &end[0], entity.oldEnt->num(), ma->clipmask, 0 );
 		gentity_t* hit = &g_entities[trace.entityNum];
 
-		if (hit && G_Team( hit ) == TEAM_HUMANS) {
+		if (hit && ( not use_experiments && G_OnSameTeam( entity.oldEnt, hit ) ) || ( use_experiments && G_Team( hit ) == TEAM_HUMANS ) ) {
 			return true;
 		}
 	}
@@ -218,7 +220,7 @@ bool SpikerComponent::Fire() {
 		float rowPerimeter = 2.0f * M_PI * cosf(rowAltitude);
 
 		// Attempt to distribute spikes with equal expected angular distance on all rows.
-		int spikes = (int)round(((float)MISSILES * rowPerimeter) / totalPerimeter);
+		int spikes = (int)round(((float)sg_spiker_missiles.Get() * rowPerimeter) / totalPerimeter);
 
 		// Launch missiles in the current row.
 		for (int spike = 0; spike < spikes; spike++) {
@@ -250,8 +252,7 @@ bool SpikerComponent::Fire() {
 				level.time + (int)(1000.0f * SPIKE_RANGE / (float)BG_Missile(MIS_SPIKER)->speed));
 		}
 	}
-
-	restUntil = level.time + COOLDOWN * launched * MISSILES;
+	restUntil = level.time + COOLDOWN * ( sg_spiker_experiments.Get() ? launched * sg_spiker_missiles.Get() : 1 );
 	RegisterSlowThinker();
 
 	return true;
-- 
2.30.2
```

and

```
From 2656984598d7e124344073f15b4962da0e412b5a Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Morel=20B=C3=A9renger?= <berengermorel76@gmail.com>
Date: Sun, 5 Mar 2023 16:22:46 +0100
Subject: [PATCH 16/44] gameplay: make spiker experiments configurable

---
 src/sgame/components/SpikerComponent.cpp | 13 +++++++------
 1 file changed, 7 insertions(+), 6 deletions(-)

diff --git a/src/sgame/components/SpikerComponent.cpp b/src/sgame/components/SpikerComponent.cpp
index 519784f77..eff326d90 100644
--- a/src/sgame/components/SpikerComponent.cpp
+++ b/src/sgame/components/SpikerComponent.cpp
@@ -4,13 +4,14 @@
 
 static Log::Logger logger("sgame.spiker");
 
+static Cvar::Cvar<bool> sg_spiker_experiments( "sg_spiker_experiments", "wether to use experimental reloading for spikers or not", Cvar::NONE, true );
 /** Delay between shots. */
 constexpr int   COOLDOWN = 5000;
 /** Shoot prematurely whenever this much damage is expected. */
 constexpr float DAMAGE_THRESHOLD = 60.0f;
 /** Total number of missiles to shoot. Some might be skipped if they endanger allies.
  *  Actual values is +/- MISSILEROWS. */
-constexpr int   MISSILES = 39;
+static Cvar::Cvar<int> sg_spiker_missiles( "sg_spiker_missiles", "number of missiles spikers have", Cvar::NONE, 39 );
 /** Number of rows from which to launch spikes. */
 constexpr int   MISSILEROWS = 4;
 /** An estimate on how far away spikes are still relevant, used to upper bound computations. */
@@ -96,7 +97,7 @@ void SpikerComponent::Think(int /*timeDelta*/) {
 		float hitEdge      = bboxEdge + ((1.0f / M_ROOT3) * ma->size); // Add half missile edge.
 		float hitArea      = hitEdge * hitEdge; // Approximate area resulting in a hit.
 		float effectArea   = 2.0f * M_PI * distance * distance; // Area of a half sphere.
-		float damage       = (hitArea / effectArea) * (float)MISSILES * spikeDamage;
+		float damage       = (hitArea / effectArea) * (float)sg_spiker_missiles.Get() * spikeDamage;
 
 		// Sum up expected damage for all targets, regardless of whether they are in sense range.
 		expectedDamage += damage;
@@ -158,6 +159,7 @@ bool SpikerComponent::SafeToShoot(glm::vec3 direction) {
 	trace_t trace;
 	vec3_t mins, maxs;
 	glm::vec3 end = VEC2GLM( entity.oldEnt->s.origin ) + (SPIKE_RANGE * direction);
+	bool use_experiments = sg_spiker_experiments.Get();
 
 	// Test once with normal and once with inflated missile bounding box.
 	for (float traceSize : {missileSize, missileSize * SAFETY_TRACE_INFLATION}) {
@@ -166,7 +168,7 @@ bool SpikerComponent::SafeToShoot(glm::vec3 direction) {
 		trap_Trace( &trace, entity.oldEnt->s.origin, mins, maxs, &end[0], entity.oldEnt->num(), ma->clipmask, 0 );
 		gentity_t* hit = &g_entities[trace.entityNum];
 
-		if (hit && G_Team( hit ) == TEAM_HUMANS) {
+		if (hit && ( not use_experiments && G_OnSameTeam( entity.oldEnt, hit ) ) || ( use_experiments && G_Team( hit ) == TEAM_HUMANS ) ) {
 			return true;
 		}
 	}
@@ -218,7 +220,7 @@ bool SpikerComponent::Fire() {
 		float rowPerimeter = 2.0f * M_PI * cosf(rowAltitude);
 
 		// Attempt to distribute spikes with equal expected angular distance on all rows.
-		int spikes = (int)round(((float)MISSILES * rowPerimeter) / totalPerimeter);
+		int spikes = (int)round(((float)sg_spiker_missiles.Get() * rowPerimeter) / totalPerimeter);
 
 		// Launch missiles in the current row.
 		for (int spike = 0; spike < spikes; spike++) {
@@ -250,8 +252,7 @@ bool SpikerComponent::Fire() {
 				level.time + (int)(1000.0f * SPIKE_RANGE / (float)BG_Missile(MIS_SPIKER)->speed));
 		}
 	}
-
-	restUntil = level.time + COOLDOWN * launched * MISSILES;
+	restUntil = level.time + COOLDOWN * ( sg_spiker_experiments.Get() ? launched * sg_spiker_missiles.Get() : 1 );
 	RegisterSlowThinker();
 
 	return true;
-- 
2.30.2


```

bmorel:
So, I just compared the 3 commits of the PR and the 3 commits embedded in my mod, and they seems strictly identical. I would appreciate it if you gave some more details about what other changes you think were introduced?

Viech:
So, I had the following three points in my first review:

1. `sg_spiker_missiles` should be `g_spikerMissiles`,
2. `g_spikerMissiles` controls directly the damage dealt by the Spiker, so it should not be a cvar but either hardcoded or config-file controlled,
3. this PR cannot possibly contain the changes you tested, because this PR makes it such that the Spiker fires once and then takes minutes to reload.

To clarify point (3):

```
restUntil = level.time + COOLDOWN * ( sg_spiker_experiments.Get() ? launched * sg_spiker_missiles.Get() : 1 );
```

This multiplies the cooldown by the number of spikes fired (this is intended) **and by the total number of spikes that can be fired** (this was almost surely not intended by you). You should to divide these numbers, not multiply them.

I understand that you have a conflicting opinion on point (2) but it does not look like you have made any changes with respect to points (1) and (3) either. I'm judging this based on the `Files changed` tab of this PR.

Viech:
(Apart from my immediate review of this PR, I think this change is a good idea gameplay-wise but I would insist that we don't regress on the visual communication. I don't think one can make this change without accounting for animations. As far as I remember, the Spiker has a single fire animation where the Spikes take five seconds to regrow? We'd need to do _something_ about this, otherwise you'd just cancel an animation early and shoot spikes which have not regrown. It's also not ideal that the animation shoots all spikes while your proposal shoots just a few, although I did not find this too problematic in my tests.)

bmorel:
No you're right that I have not applied any change beside the rebase, but I think you're wrong when you say this is not what I tested locally :)

1) is nitpick, I'll address it in due time
2) is, as you said, a conflictual point. I don't want to hard-code it, because hard-coding just makes it a pain. Config files could be ok, but not for prototyping which can require live changes, and there's an on-going discussion on forums. I'll base the final decision on the consensus there. For now, I have the feeling that most people are rather in favor of cvars than config files, and I'm totally against hard-coding which has proved to encourage immobilism (i.e. fire parameters which were never tweaked)
3) you're probably right, I'll investigate when I'll be at home, and I have no idea when this will be.

necessarily-equal:
I re-enabled the testing server because it was rebased (and thus no longer broken). You can find the server in the server list

bmorel:
if there is no feedback in few weeks, I'll close this PR

PR #2559, opened 2023-03-31
cg_rocket_draw: support emoticons in map and author names

illwieckz:
I first thought it was a bug, but maybe we prefer to keep map and author names text only.

The author name can be used elsewhere than a map a map loading screen, etc.

What's your thought about it?

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -3594,12 +3594,12 @@ static void CG_Rocket_DrawLoadingText()
 
 static void CG_Rocket_DrawLevelAuthors()
 {
-	Rocket_SetInnerRML( cg.mapAuthors.c_str(), RP_QUAKE );
+	Rocket_SetInnerRML( cg.mapAuthors.c_str(), RP_QUAKE | RP_EMOTICONS );
 }
 
 static void CG_Rocket_DrawLevelName()
 {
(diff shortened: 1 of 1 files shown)


illwieckz:
Since I was the one implementing that map loading screen and that I haven't set the emoticons flag at first, I'm confident my opinion at the time was it's better to not put emoticons in names map / author names.

illwieckz:
Maybe we can also define some guidelines to write author names without clan, for example. Maps likely survive clans.

slipher:
> Do we want to support emoticons in map and author names?

Not really

DolceTriade:
Why not?

bmorel:
That's the problem: emoticons in names would make it sooo 2022, and we're already in 2023!

illwieckz:
> It's 2022 man!

That's so 2022 !

slipher:
Maybe we want to kill some obsolete emoticons like the hovel and do so without breaking backwards compatibility with maps :stuck_out_tongue:

bmorel:
> But even if we enable the emoticon parsing, we may still want to recommend to not use them

Then just don't?

illwieckz:
Yeah, I'm not sure we should strictly forbid this, though I'm in favor of writing recommendations. It's good to have somewhat portable metadata so it's good to provide best practices for those who are interested in this.

PR #2560, opened 2023-03-31
Loop entities (almost) once in `CG_AddPacketEntities()`

illwieckz:
cg_ents: process entities only once in `CG_AddPacketEntities()`

1. Iterate all valid entities and process them.
2. Iterate all entities skipping already processed ones.

cgame: use dynamic `cg_maxEntities` when iterating entities

Instead of iterating entities to `MAX_GENTITIES`, iterate to `cg_maxEntities` which is the the limit above the highest entity in current scene.

cgame: use dynamic `cg_highestActiveEntity` when iterating entities

Instead of iterating entities to `MAX_GENTITIES,` iterate to `cg_highestActiveEntity` (included) which is the highest entity number from the previous frame.

cg_players: loop clients to `MAX_CLIENTS`

---

_Original message:_

By running the game on Orbit profiler after having tweaked the engine to run on 0 frametime and not be polluted by any idling code, and running the game with lowest graphics profile, this code was the functions reported to eat CPU cycles after the actual rendering by the OpenGL driver and the RmlUi code..

We recently increased `MAX_GENTITIES` if I'm right. This code iterated the entities four time, and two time up to the max.

It now iterates only once, to the max.

~~This is WIP, I kept the separate rewrite steps I've done if that can help reviewers to understand what I am doing, and verify step after step I'm not doing any mistakes.~~

### DIFF
--- a/src/cgame/cg_ents.cpp
+++ b/src/cgame/cg_ents.cpp
@@ -28,6 +28,12 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 #include "cg_local.h"
 
+/* It is dynamically adjusted in CG_AddPacketEntities() to avoid looping
+over unused entities. It can be zero before the first frame starts adding
+entities, as none is added yet and none has to be removed yet. This should
+never reach the MAX_GENTITIES value. */
+int cg_highestActiveEntity = 0;
+
 static Cvar::Range<Cvar::Cvar<float>>
 	cg_drawDebugDistance("cg_drawDebugDistance", "draw a sphere around "
 			"yourself to give an idea of distances. Set this "
@@ -1318,10 +1324,17 @@ void CG_AddPacketEntities()
 
 	bool done[ MAX_GENTITIES ] = {};
 
+	int highest = 0;
+
 	// add each entity sent over by the server
 	for ( const entityState_t &ent : cg.snap->entities )
 	{
-		const unsigned num = ent.number;
+		const int num = ent.number;
+
+		/* This number may become cg_highestActiveEntity
+		to loop over the cg_entities array. */
+		ASSERT( num < MAX_GENTITIES );
+
 		centity_t *cent = &cg_entities[ num ];
 
 		if ( !cent->valid )
@@ -1334,9 +1347,22 @@ void CG_AddPacketEntities()
 		CG_AddCEntity( cent );
 
 		done[ num ] = true;
+
+		if ( num > highest )
+		{
+			highest = num;
+		}
 	}
-	
-	for ( unsigned num = 0; num < MAX_GENTITIES; num++ )
+
+	/* We assume cg_highestActiveEntity is never greater
+	than MAX_GENTITIES. cg_highestActiveEntity carries the
+	highest entity number used by an entity in the previous game frame.
+	This is not the amount of entities from the previous game frame.
+	A short lifetime entity with a small number will have to be removed
+	before a long lifetime entity with a high number is removed, so we
+	need to iterate all entities (valid or not) up to the previous highest
+	entity to be able to remove entities that should be removed. */
+	for ( int num = 0; num <= cg_highestActiveEntity; num++ )
 	{
 		if ( done[ num ] )
 		{
@@ -1352,6 +1378,20 @@ void CG_AddPacketEntities()
 
 		cent->valid = false;
 	}
+
+	if ( cg_debugPVS.Get() )
+	{
+		if ( highest > cg_highestActiveEntity )
+		{
+			Log::Debug("Highest entity in PVS went up: %u", highest);
+		}
+		else if ( highest < cg_highestActiveEntity )
+		{
+			Log::Debug("Highest entity in PVS went down: %u", highest);
+		}
+	}
+
+	cg_highestActiveEntity = highest;
 
 	//make an attempt at drawing bounding boxes of selected entity types
 	if ( cg_drawBBOX.Get() > 0 )
--- a/src/cgame/cg_ents.cpp
+++ b/src/cgame/cg_ents.cpp
@@ -1271,7 +1271,6 @@ CG_AddPacketEntities
 */
 void CG_AddPacketEntities()
 {
-	centity_t     *cent;
 	playerState_t *ps;
 
 	// set cg.frameInterpolation
@@ -1317,39 +1316,41 @@ void CG_AddPacketEntities()
 	// lerp the non-predicted value for lightning gun origins
 	CG_CalcEntityLerpPositions( &cg_entities[ cg.snap->ps.clientNum ] );
 
-	for ( unsigned num = 0; num < MAX_GENTITIES; num++ )
-	{
-		cg_entities[ num ].valid = false;
-	}
+	bool done[ MAX_GENTITIES ] = {};
 
 	// add each entity sent over by the server
-	for ( unsigned num = 0; num < cg.snap->entities.size(); num++ )
+	for ( const entityState_t &ent : cg.snap->entities )
 	{
-		cent = &cg_entities[ cg.snap->entities[ num ].number ];
-		cent->valid = true;
-	}
+		const unsigned num = ent.number;
+		centity_t *cent = &cg_entities[ num ];
 
-	for ( unsigned num = 0; num < MAX_GENTITIES; num++ )
-	{
-		cent = &cg_entities[ num ];
-
-		if ( cent->valid && !cent->oldValid )
+		if ( !cent->valid )
 		{
 			CG_CEntityPVSEnter( cent );
 		}
-		else if ( !cent->valid && cent->oldValid )
+
+		cent->valid = true;
+
+		CG_AddCEntity( cent );
+
+		done[ num ] = true;
+	}
+	
+	for ( unsigned num = 0; num < MAX_GENTITIES; num++ )
+	{
+		if ( done[ num ] )
+		{
+			continue;
+		}
+
+		centity_t *cent = &cg_entities[ num ];
+
+		if ( cent->valid )
 		{
 			CG_CEntityPVSLeave( cent );
 		}
 
-		cent->oldValid = cent->valid;
-	}
-
-	// add each entity sent over by the server
-	for ( unsigned num = 0; num < cg.snap->entities.size(); num++ )
-	{
-		cent = &cg_entities[ cg.snap->entities[ num ].number ];
-		CG_AddCEntity( cent );
+		cent->valid = false;
 	}
 
 	//make an attempt at drawing bounding boxes of selected entity types
@@ -1360,7 +1361,7 @@ void CG_AddPacketEntities()
 			vec3_t        mins, maxs;
 			entityState_t *es;
 
-			cent = &cg_entities[ cg.snap->entities[ num ].number ];
+			centity_t *cent = &cg_entities[ cg.snap->entities[ num ].number ];
 			es = &cent->currentState;
 
 			switch ( es->eType )
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -1782,6 +1782,7 @@ enum altShader_t
 extern  cgs_t               cgs;
 extern  cg_t                cg;
 extern  centity_t           cg_entities[ MAX_GENTITIES ];
+extern int cg_highestActiveEntity;
 
 extern  weaponInfo_t        cg_weapons[ 32 ];
 extern  upgradeInfo_t       cg_upgrades[ 32 ];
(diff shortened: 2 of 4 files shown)


illwieckz:
In fact the entities are iterated a fifth time, but only if `cg_drawBBOX.Get()` is true, but given this one is a debug cheat, we can keep it outside of the main loop, especially because when not debugging, only one test is done (the cvar) instead of doing the test for every entities if we would collapse that fifth loop. Keeping this extra loop makes non-debug game faster at the expense of keeping debug games slower.

illwieckz:
I split the loop into two half-loops (entities are still iterated only once) to remove a test.

slipher:
What you could do is to keep a high-water mark of the highest entity number seen and only iterate up to there. That's what sgame does to avoid paying for unused space.

illwieckz:
> What you could do is to keep a high-water mark of the highest entity number seen and only iterate up to there. That's what sgame does to avoid paying for unused space.

You mean something like my latest commit?

Note: this one is just a PoC to make sure that's the kind of thing you're talking about, this may be completely rewritten, especially yo be usable by other functions in the code that are also looping to max. Also I wonder if we can have a totally dynamic “previous max” that would be computed by decrementing it when finding old valid in invalid list and incrementing it when finding old invalid in valid list, this way the floating max would go down after a pike.

For reference, here is the weight in CPU time of this function by simply rendering a spawned human view on vega, recording 20s with Orbit:

 |original|single loop| single loop + floating max
---|---|---|---
cpu time|3.13%|1.67%|0.16%|

Original code with 4 loops:

[loop entities once](https://dl.illwieckz.net/b/unvanquished/wip/loop-entities-once/20230401-111915-000.orbit-unvanquished-loop-entities-once.png)

Single loop:

[loop entities once](https://dl.illwieckz.net/b/unvanquished/wip/loop-entities-once/20230401-112152-000.orbit-unvanquished-loop-entities-once.png)

Single loop with floating max:

[loop entities once](https://dl.illwieckz.net/b/unvanquished/wip/loop-entities-once/20230401-112430-000.orbit-unvanquished-loop-entities-once.png)

illwieckz:
The latest commit (_dynamicMax_) totally breaks particles.

slipher:
> > What you could do is to keep a high-water mark of the highest entity number seen and only iterate up to there. That's what sgame does to avoid paying for unused space.
> 
> You mean something like my latest commit?

No. I mean the highest entity num which has been seen in any snapshot during the whole game.

illwieckz:
> > > What you could do is to keep a high-water mark of the highest entity number seen and only iterate up to there. That's what sgame does to avoid paying for unused space.
> > 
> > 
> > You mean something like my latest commit?
> 
> No. I mean the highest entity num which has been seen in any snapshot during the whole game.

Hmm, I don't know how to do it… I removed the broken commit.

What is already done doubles the performance of that function anyway.

slipher:
> What is already done doubles the performance of that function anyway.

It's wrong though. The snapshot can be any subset of entities; you can't assume that it's just 0 through N

illwieckz:
Ah, it looks like I make some confusions between `cg_entities` and `cg.snap->entities`… Those pointers of pointers of pointers are annoying!

illwieckz:
@slipher I made a new implementation. This _almost_ loops only once.

What's done is that all active entities are looped first, then all entities are looped but those who were just processed before are skipped.

Since we need to know if an entity is already processed or not I've set a simple array of booleans that is zeroed before running the loops. Basically, the initial loop setting `cent.valid` to false was the equivalent of this, except zeroing an array of booleans is believed to be way faster than zeroing a struct component through multiple pointer dereference without probably no way to vectorize the zeroing.

src/cgame/cg_ents.cpp
```diff
@@ -1317,39 +1316,43 @@ void CG_AddPacketEntities()
...
 
-	for ( unsigned num = 0; num < MAX_GENTITIES; num++ )
-	{
-		cg_entities[ num ].valid = false;
-	}
+	bool done[ MAX_GENTITIES ] = {};
 
 	// add each entity sent over by the server
-	for ( unsigned num = 0; num < cg.snap->entities.size(); num++ )
+	for ( unsigned n = 0; n < cg.snap->entities.size(); n++ )
```
slipher, line 1322: You can use `for(const entityState_t &ent : cg.snap->entities)` for efficient loop iteration

slipher (commented):
OK now it makes sense. You can delete the `oldValid` field

src/cgame/cg_ents.cpp
illwieckz: Done.

illwieckz:
> OK now it makes sense. You can delete the `oldValid` field

Done.

src/cgame/cg_ents.cpp
```diff
@@ -1318,6 +1318,8 @@ void CG_AddPacketEntities()
 
 	bool done[ MAX_GENTITIES ] = {};
 
+	unsigned highest = 0;
```
slipher, line 1321: Use `int` for everything in this commit. Since math does not work as expected if you go into negatives with unsigned numbers they should only be used if there is a good reason

src/cgame/cg_local.h
```diff
@@ -1761,6 +1761,7 @@ enum altShader_t
 extern  cgs_t               cgs;
 extern  cg_t                cg;
 extern  centity_t           cg_entities[ MAX_GENTITIES ];
+extern unsigned cg_maxEntities;
```
slipher, line 1764: How about `cg_numEntities` for more consistency with sgame `num_entities`

src/cgame/cg_ents.cpp
```diff
@@ -1334,9 +1336,20 @@ void CG_AddPacketEntities()
...
+		}
 	}
-	
-	for ( unsigned num = 0; num < MAX_GENTITIES; num++ )
+
+	// On frame 0 there is nothing to remove.
+	static unsigned oldHighest = 0;
+
+	// We assume maxCgEntities is never greater than MAX_GENTITIES.
+	unsigned cg_maxEntities = oldHighest + 1;
```
slipher, line 1350: You probably didn't want to make a local variable shadowing the global one

src/cgame/cg_local.h
illwieckz: I was also thinking about `cg_maxEntitiesInPVS` to be more explicit about the purpose, but that can also be `cg_numEntitiesInPVS` and then we iterate with `<=` test instead of `<` and that may fix the variable shadowing.

slipher: The PVS part is probably not true anyway because of beacons

illwieckz: Wait, `num_entities` is not the amount of entities, it is the highest number used by an entity in the previous game frame. If you spawn 40 drills and remove the 39 first ones, that number will be 40 anyway, with the 39 earlier entities being iterated but skipped because invalid (removed).

The code doesn't iterate the amount of entities, but it iterates the entities up to the one with the highest number among previously active entities, so we can iterate removed entities to actually remove them.

So I suggest `cg_highestPreviousEntityNumberInPVS`.

illwieckz:
I updated the second commit:

> cgame: use dynamic `cg_highestPreviousEntityNumberInPVS` when iterating entities
>
> Instead of iterating entities to `MAX_GENTITIES,` iterate to `cg_highestPreviousEntityNumberInPVS` (included) which is the highest entity number from the previous frame.

src/cgame/cg_ents.cpp
```diff
@@ -28,6 +28,12 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 #include "cg_local.h"
 
+/* It is dynamically adjusted in CG_AddPacketEntities() to avoid looping
+over unused entities. It can be zero before the first frame starts adding
+entities, as none is added yet and none has to be removed yet. This should
+never reach the MAX_GENTITIES value. */
+int cg_highestPreviousEntityNumberInPVS = 0;
```
slipher, line 35: The PVS wording should be removed since it includes all entities known to the cgame including beacons which are not in the PVS. I think "previous" is not good either since it's current for everywhere except CG_AddPacketEntities.

Maybe `cg_highestActiveEntity`

src/cgame/cg_players.cpp
```diff
@@ -1279,12 +1279,13 @@ static void CG_LoadClientInfo( clientInfo_t *ci )
 	// frames for this new model
 	clientNum = ci - cgs.clientinfo;
 
-	for ( i = 0; i < MAX_GENTITIES; i++ )
+	for ( int num = 0; num <= cg_highestPreviousEntityNumberInPVS; num++ )
```
slipher, line 1282: Huh, why is this even going above MAX_CLIENTS?

bmorel:
I took the freedom to rebase this on master so that CI can be useful anew.

src/cgame/cg_ents.cpp
illwieckz: Done.

src/cgame/cg_ents.cpp
illwieckz: Done.

src/cgame/cg_ents.cpp
illwieckz: Done.

src/cgame/cg_players.cpp
illwieckz: I added a commit to replace the upper limit by `MAX_CLIENTS`.

src/cgame/cg_ents.cpp
```diff
@@ -1353,6 +1379,20 @@ void CG_AddPacketEntities()
 		cent->valid = false;
 	}
 
+	if ( cg_debugPVS.Get() )
+	{
+		if ( highest > cg_highestActiveEntity )
```
slipher, line 1384: Why would someone care about what the highest entity number known to the cgame is? Doesn't seem very relevant.

bmorel (approved):
No bug noticed.

I tested it after a rebase though, as without rebase I'd have a bug about sound preventing the game the launch. It works.

Testing on plat23:

* from 0.54, official, no fire: around 115 FpS I guess
* from 0.54, official, spam nade: around 95 FpS
* from master, debug, no fire: between 70 and 120 FpS
* from master, debug, spam nade: around  55 FpS
* from master, release, no fire: around 115 FpS
* from master, release, spam nade: around 115 FpS

src/cgame/cg_ents.cpp
illwieckz: > Why would someone care about what the highest entity number known to the cgame is? Doesn't seem very relevant.

Hu… It's not what that code does… 🤨️ It's a debug print to debug the ballooning mechanism. The important part of the debug log is “`went up`” and “`went down`”.

This debug print is hidden behind two closed doors, you need to _both_ enable debug logs and enable `cg_debugPVS` to get this debug print.

In that file there are already existing very verbose debug prints that print a debug message every time an entity enters or leave PVS, and no one complained about it, ever:

```c++
static void CG_CEntityPVSEnter( centity_t *cent )
{
	entityState_t &es = cent->currentState;

	if ( cg_debugPVS.Get() )
	{
		Log::Debug( "Entity %d entered PVS", cent->currentState.number );
	}
```

```c++
static void CG_CEntityPVSLeave( centity_t *cent )
{
	int           i;

	if ( cg_debugPVS.Get() )
	{
		Log::Debug( "Entity %d left PVS", cent->currentState.number );
	}
```

bmorel: > In that file there are already existing very verbose debug prints that print a debug message every time an entity enters or leave PVS, and no one complained about them, ever (but they were very useful to me so I'm glad they were there)

I don't think removing debug messages is a good idea, if they don't bring a big performance cost. If they do though, maybe we can think of a solution to only have that cost on debug builds?

src/cgame/cg_local.h
illwieckz: I did 'cg_highestActiveEntity'.

src/cgame/cg_players.cpp
```diff
@@ -1279,7 +1279,7 @@ static void CG_LoadClientInfo( clientInfo_t *ci )
 	// frames for this new model
 	clientNum = ci - cgs.clientinfo;
 
-	for ( int num = 0; num <= cg_highestActiveEntity; num++ )
```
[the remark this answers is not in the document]
illwieckz: I'm don't know what you mean, @slipher if you know a better implementation, please make a PR once this one is merged.

src/cgame/cg_ents.cpp
illwieckz: Yeah, same. I don't believe this is blocking.

PR #2566, opened 2023-04-01
very trivial ladder support (upper direction only)

bmorel:
This allows bots to use ladders when there is a navcon.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -586,9 +586,17 @@ static bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const glm:
 	//check if jumping will clear us of entity
 	trap_Trace( &trace, origin, playerMins, playerMaxs, end, self->s.number, MASK_SHOT, 0 );
 
+	classAttributes_t const* pcl = BG_Class( pClass );
+	bool ladder = ( pcl->abilities & SCA_CANUSELADDERS )
+		&& trace.entityNum == ENTITYNUM_WORLD
+		&& trace.surfaceFlags & SURF_LADDER
+		;
+
 	//if we can jump over it, then jump
 	//note that we also test for a blocking barricade because barricades will collapse to let us through
-	return blocker->s.modelindex == BA_A_BARRICADE || trace.fraction == 1.0f;
+	return blocker->s.modelindex == BA_A_BARRICADE
+		|| trace.fraction == 1.0f
+		|| ladder;
 }
 
 // Determine a new direction, which may be unchanged (forward) or sideway. It
--- a/src/sgame/botlib/bot_load.cpp
+++ b/src/sgame/botlib/bot_load.cpp
@@ -131,6 +131,7 @@ static void BotLoadOffMeshConnections( const char *species, OffMeshConnections &
 
 	if ( !f )
 	{
+		Log::Debug("Failed to load \"maps/%s-%s.navcon\"", mapname.c_str(), species );
 		return;
 	}
 
@@ -139,11 +140,10 @@ static void BotLoadOffMeshConnections( const char *species, OffMeshConnections &
 
 	version = LittleLong( version );
 
+	Log::Debug("Loading format version %d navcon file %s", version, filePath );
 	// Old binary format.
 	if ( version == 2 )
 	{
-		Log::Debug("Loading format version %d navcon file %s", version, filePath );
-
 		int conCount;
 		trap_FS_Read( &conCount, sizeof( conCount ), f );
 
@@ -229,8 +229,6 @@ static void BotLoadOffMeshConnections( const char *species, OffMeshConnections &
 		Log::Warn("Unknown format version %d for navcon file %s", version, filePath );
 		return;
 	}
-
-	Log::Debug("Loading format version %d navcon file %s", version, filePath );
 
 	con.offMeshConCount = 0;
 
--- a/src/shared/navgen/nav.cpp
+++ b/src/shared/navgen/nav.cpp
@@ -26,6 +26,11 @@
 #include <iostream>
 #include <vector>
 #include <queue>
+#include <map>
+#include <unordered_set>
+
+#include <glm/gtx/string_cast.hpp>
+#include <glm/gtx/hash.hpp>
 
 #include "engine/qcommon/qcommon.h"
 #include "shared/CommonProxies.h"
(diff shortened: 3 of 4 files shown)


bmorel:
I have C++ code for a 3rd party tool that can extract ladder positions from a `.map` file, tested on both parpax and yocto with a 100% success rate, but I'm not sure how to integrate it into unvanquished. I could do it into a `contrib/` folder I guess. The tool is still rough on the edges too, and is not strictly required (navcons can be built by hand, and the tool only give ladder positions for now).

src/sgame/sg_bot_nav.cpp
```diff
@@ -586,9 +586,17 @@ static bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const glm:
 	//check if jumping will clear us of entity
 	trap_Trace( &trace, origin, playerMins, playerMaxs, end, self->s.number, MASK_SHOT, 0 );
 
+	classAttributes_t const* pcl = BG_Class( pClass );
+	bool ladder = ( pcl->abilities & SCA_CANUSELADDERS )
```
slipher, line 590: The trace here goes forward 20 units but PM_CheckLadder only looks forward 1 unit. Maybe there should be a condition on the trace distance being small enough?

bmorel: I'm afraid that using a shorter trace would make the bot trigger avoidance maneuvers, and thus would prevent it to climb at ladder.

slipher: I'm not saying to use a shorter trace. Rather to check how far the hit point is using trace fraction or whatever

bmorel: I don't get you... why would that be useful? Is there is a known problem with this PR?

slipher: What stops the bots from uselessly jumping and wasting stamina when they aren't close enough to climb?

bmorel: have you seen this with this PR merged? Condisering unv's slippy physics, there are very, very little chances for this to happen.
The code is certainly nor perfect, but I can't do perfect handling with the current status, because code handling AI moves is just a pile of hacks.
I will need further refactoring to get wallwalking done, and it won't be a 5-line patch, it will take weeks of discussions and talking to be merged. I know it in advance.
In the meantime, I need this merged because this will allow to check for regressions, and because in the meantime it will allow the bots to use ladders.

In any case, this requires off-mesh connections to be placed, so no, there is no "loss" of stamina to expect.

cu-kai: All things considered I think we could probably merge this if there's no technical reason other than "bots might waste stamina" - this will require testing and I am willing to merge it into bunker once it's upstream (ping @sweet235 for his tactic server too).

If worst comes to the worst, we can revert this, but it looks like @bmorel is laying the groundwork for more improvements.

It certainly cannot break things more than the revamped navmesh generator did in 0.54.0 ...

slipher:
I tried to test out the jumping thing but had trouble getting the code to trigger since I don't have any navcons (I got bots to run into the ladder but then it still didn't work because the z component of the desired direction was 0). Maybe the stamina loss is not important but I'm still curious -- do bots jump when they approach the ladder?

Meanwhile I noticed a bug affecting the ladder under the Parpax elevator. It has only `CONTENTS_PLAYERCLIP | CONTENTS_TRANSLUCENT` which is not detected by the obstacle trace using MASK_SHOT. I have a commit in #2575 fixing these to use MASK_PLAYERSOLID.

bmorel:
> It has only CONTENTS_PLAYERCLIP | CONTENTS_TRANSLUCENT which is not detected by the obstacle trace using MASK_SHOT. I have a commit in #2575 fixing these to use MASK_PLAYERSOLID.

Apparently I did testing with a different codebase that the one in this PR, which did not suffered this problem. So your PR (or at least the mask commit) is actually required for this to work.

The code now prints (with log::notice) what should be placed in the navcon file, I'll do something cleaner later. Well, you need to remove the prefix "ladder: " but that should not be a big deal.
The hard-coded radius of 30 in the pseudo-navcon-generator is arbitrary, and might need adjustments.

This works.... more or less, for the simpler cases. But cases like yocto's pit will not.

bmorel:
> Meanwhile I noticed a bug affecting the ladder under the Parpax elevator. It has only CONTENTS_PLAYERCLIP | CONTENTS_TRANSLUCENT which is not detected by the obstacle trace using MASK_SHOT. I have a commit in #2575 fixing these to use MASK_PLAYERSOLID.

All this is weird, because I've always been doing my tests with parpax, and especially this ladder, and I have had videos showing it to work. The old cobebase for those vids did not modified the trace masking, and I'm pretty certain I tested this before doing a PR.
I tried rebasing at the time of this PR just in case, but the ladder code was broken still.

This is all quite mysterious to me.
Also (from a very old commit of mine in abandoned branch):

```
  // MASK_PLAYERSOLID ( CONTENTS_SOLID | CONTENTS_PLAYERCLIP | CONTENTS_BODY )
  // MASK_SHOT        ( CONTENTS_SOLID | CONTENTS_BODY )
```

And from daemon:
```
daemon/src/engine/qcommon/SurfaceFlags.h:#define CONTENTS_SOLID              BIT( 0 )  // 0x1
daemon/src/engine/qcommon/SurfaceFlags.h:#define CONTENTS_PLAYERCLIP         BIT( 16 ) // 0x10000
daemon/src/engine/qcommon/SurfaceFlags.h:#define CONTENTS_BODY               BIT( 25 ) // 0x2000000  // not used by brushes (must not), used by game entities
```

If I read this correctly, but maybe I'm wrong, there's no clear definition of those stuff AFAIK... it means that the current traces won't collide with other brushes that are player clips, that is, area which are invisible but mapper decided to make collide?

In any case, the trace does find an obstacle, but for some reason, without the mask fix, it has a `trace.surfaceflags = 0`, which is why this PR fails in that place. And again I'm pretty sure I had it working at some point in time, on this same version of the same map!

sweet235:
I run this on my server now, seems to work well enough.

> I tried to test out the jumping thing but had trouble getting the code to trigger since I don't have any navcons

I set up a quick and dirty repository for the navcons I use for parpax, here: https://github.com/sweet235/parpax-navcon

bmorel:
For this to work, navcons need a "landing area" in front of the top of ladder. Also, you _probably_ don't want to have single ladder with multiple floors.
The emitted notices will help, but if ladder goes higher than top floor you need to adjust. Same if ground is tricky. Means that yocto won't work, for example, for all of those reasons. Yocto's pit is a good example to sum up what this can NOT support (yet)

sweet235:
I just saw a battlesuit get stuck at the bottom of the ladder. Not sure why.

sweet235:
lgtm

sweet235:
> I just saw a battlesuit get stuck at the bottom of the ladder. Not sure why.

The reason is: it was alterted to an enemy. This PR does not change what they do when alerted to an enemy at all, so this is an improvement without a drawback.

bmorel:
>  do bots jump when they approach the ladder?

No.

PR #2567, opened 2023-04-01
Crouching bots

### LINKED ISSUE #1862: allow bots to crouch
Current bots can't crouch.
I have a workaround in navmesh generation which allows this: generate navmeshes using CrouchMaxs instead of Maxs for BBoxs (and be sure aliens's Maxs are equal to CrouchMaxs!). This also rely on improved obstacle detection and hence have a toll on server's performance, since traces needs to be emitted when bot is blocked.

bmorel:
Allows bots to crouch when it would allow them to pass an obstacle.

This is placed before other tests since it's notably more reliable than jumping.
Note that all this part of the code needs a serious refactoring, that I'll do *after* all PRs concerning bot movements are merged. And after said refactoring, I shuold be able to bring back the wallwalking experiments, too.

Fix #1862

### DIFF
--- a/src/shared/bot_nav_shared.h
+++ b/src/shared/bot_nav_shared.h
@@ -79,16 +79,32 @@ enum navPolyAreas
 
 // Part of header which must contain 4-byte members only
 struct NavgenConfig {
-	float requestedCellHeight;
-	float stepSize;
-	int excludeCaulk; // boolean - exclude caulk surfaces
-	int excludeSky; // boolean - exclude surfaces with surfaceparm sky from navmesh generation
-	int filterGaps; // boolean - add new walkable spans so bots can walk over small gaps
-	int generatePatchTris; // boolean - generate triangles from the BSP's patches
-	float autojumpSecurity; // percentage - allow to use part of jump magnitude (with default gravity of 800) as stepsize. The result can not excess the agent's height, except if STEPSIZE is already doing it (then STEPSIZE will be used)
-	bool crouchSupport = true;
+	float requestedCellHeight = 2.f;
+	float stepSize = STEPSIZE;
+	float autojumpSecurity = 0.5f; // percentage - allow to use part of jump magnitude (with default gravity of 800) as stepsize. The result can not excess the agent's height, except if STEPSIZE is already doing it (then STEPSIZE will be used)
+	bool excludeCaulk = 1; // exclude caulk surfaces
+	bool excludeSky = 1; // exclude surfaces with surfaceparm sky from navmesh generation
+	bool filterGaps = 1; // add new walkable spans so bots can walk over small gaps
+	bool generatePatchTris = 1; // generate triangles from the BSP's patches
+	bool crouchSupport = true; // use crouchMaxs.z value instead of Maxs.z to compute paths
 
-	static NavgenConfig Default() { return { 2.0f, STEPSIZE, 1, 1, 1, 1, 0.5f }; }
+	bool operator==( NavgenConfig const& other ) const
+	{
+		return requestedCellHeight == other.requestedCellHeight
+			&& stepSize == other.stepSize
+			&& excludeCaulk == other.excludeCaulk
+			&& excludeSky == other.excludeSky
+			&& filterGaps == other.filterGaps
+			&& generatePatchTris == other.generatePatchTris
+			&& autojumpSecurity == other.autojumpSecurity
+			&& crouchSupport == other.crouchSupport
+			;
+	}
+
+	bool operator!=( NavgenConfig const& other ) const
+	{
+		return not ( *this == other );
+	}
 };
 
 struct NavgenMapIdentification
--- a/src/shared/bot_nav_shared.h
+++ b/src/shared/bot_nav_shared.h
@@ -86,6 +86,7 @@ struct NavgenConfig {
 	int filterGaps; // boolean - add new walkable spans so bots can walk over small gaps
 	int generatePatchTris; // boolean - generate triangles from the BSP's patches
 	float autojumpSecurity; // percentage - allow to use part of jump magnitude (with default gravity of 800) as stepsize. The result can not excess the agent's height, except if STEPSIZE is already doing it (then STEPSIZE will be used)
+	bool crouchSupport = true;
 
 	static NavgenConfig Default() { return { 2.0f, STEPSIZE, 1, 1, 1, 1, 0.5f }; }
 };
--- a/src/shared/bot_nav_shared.h
+++ b/src/shared/bot_nav_shared.h
@@ -49,7 +49,7 @@ ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 #define MIN_WALK_NORMAL 0.7f
 
 static const int NAVMESHSET_MAGIC = 'M'<<24 | 'S'<<16 | 'E'<<8 | 'T'; //'MSET';
-static const int NAVMESHSET_VERSION = 9; // Increment when navgen algorithm or data format changes
+static const int NAVMESHSET_VERSION = 10; // Increment when navgen algorithm or data format changes
 
 enum navPolyFlags
 {
--- a/src/shared/bot_nav_shared.h
+++ b/src/shared/bot_nav_shared.h
@@ -86,7 +86,7 @@ struct NavgenConfig {
 	bool excludeSky = 1; // exclude surfaces with surfaceparm sky from navmesh generation
 	bool filterGaps = 1; // add new walkable spans so bots can walk over small gaps
 	bool generatePatchTris = 1; // generate triangles from the BSP's patches
-	bool crouchSupport = true; // use crouchMaxs.z value instead of Maxs.z to compute paths
+	bool crouchSupport = false; // use crouchMaxs.z value instead of Maxs.z to compute paths
 
 	bool operator==( NavgenConfig const& other ) const
 	{
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -741,6 +741,27 @@ static bool BotAvoidObstacles( gentity_t *self, glm::vec3 &dir, bool ignoreGeome
 		return false;
 	}
 
+	// check for crouching
+	{
+		glm::vec3 playerMins;
+		glm::vec3 playerCMaxs;
+		trace_t trace;
+
+		class_t pClass = static_cast<class_t>( self->client->ps.stats[STAT_CLASS] );
+		BG_BoundingBox( pClass, &playerMins, nullptr, &playerCMaxs, nullptr, nullptr );
+
+		glm::vec3 origin = VEC2GLM( self->s.origin );
+		glm::vec3 end = origin + BOT_OBSTACLE_AVOID_RANGE * dir;
+
+		//make sure we are moving into a block
+		trap_Trace( &trace, origin, playerMins, playerCMaxs, end, self->s.number, MASK_PLAYERSOLID, 0 );
+		if ( trace.fraction >= 1.0f )
+		{
+			self->botMind->cmdBuffer.upmove = -127;
+			return false;
+		}
+	}
+
 	if ( BotShouldJump( self, blocker, dir ) )
 	{
 		BotJump( self );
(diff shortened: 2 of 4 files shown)


bmorel:
Btw, best map to test is parpax, since it provides this neat underground that let no other choice for a bot spawning there than to crouch to reach other places. Works best combined with ladder support, but it's a different PR.

bmorel:
I made this optional as I know in some places crouch + autojump can lead to stuck bots. Places such as under those staircases:



There, I remember regularly having marauders stuck under. Can also happen on cruz, near H default, and likely in other places. Disabling and fine tuning crouch/autojump on those maps will prevent those problems, at the cost of less features. Maybe we'll get those done correctly later though.

bmorel:
FWIW I think I know understand why the bug I had happened for humans at least: take a bot that can't jump because no stamina: it will crouch there, instead of taking the correct path.
This was a bug in my experiments, but also in the current navigation system, and it won't be easy to fix with the current codebase status. In any case, this PR allows to disable crouch support for maps where this happens, so this shuold be safe.

src/shared/bot_nav_shared.h
```diff
@@ -86,6 +86,7 @@ struct NavgenConfig {
 	int filterGaps; // boolean - add new walkable spans so bots can walk over small gaps
 	int generatePatchTris; // boolean - generate triangles from the BSP's patches
 	float autojumpSecurity; // percentage - allow to use part of jump magnitude (with default gravity of 800) as stepsize. The result can not excess the agent's height, except if STEPSIZE is already doing it (then STEPSIZE will be used)
+	bool crouchSupport = true;
```
slipher: `// Part of header which must contain 4-byte members only`

src/shared/bot_nav_shared.h
```diff
@@ -86,6 +86,7 @@ struct NavgenConfig {
...
 
 	static NavgenConfig Default() { return { 2.0f, STEPSIZE, 1, 1, 1, 1, 0.5f }; }
```
slipher: Add a default value for crouch

src/shared/bot_nav_shared.h
```diff
@@ -79,16 +79,32 @@ enum navPolyAreas
...
+	bool excludeSky = 1; // exclude surfaces with surfaceparm sky from navmesh generation
+	bool filterGaps = 1; // add new walkable spans so bots can walk over small gaps
+	bool generatePatchTris = 1; // generate triangles from the BSP's patches
+	bool crouchSupport = true; // use crouchMaxs.z value instead of Maxs.z to compute paths
+
+	bool operator==( NavgenConfig const& other ) const
+	{
+		return requestedCellHeight == other.requestedCellHeight
+			&& stepSize == other.stepSize
+			&& excludeCaulk == other.excludeCaulk
```
slipher, line 95: Why should we have to write this boilerplate?

src/shared/bot_nav_shared.h
bmorel: imo default values should use the C++11 feature of init in the declaration. This is what I did.

bmorel:
I'm tempted to orphan/drop this, as imo the requested changes changes does not makes the situation worst than it is is. I'll see in few weeks if I change my mind.

src/shared/bot_nav_shared.h
bmorel: just FYI I disagree with both of you here. This code is safer because less fragile from padding.
This whole struct is very fragile anyway since there is no way to check the size of the header. To me, it's poorly engineered, and I'd rather have boilerplate than loosing hours wondering why code does not work because I introduced a boolean or a short, which triggers paddings and fucks the memcmp.

src/shared/bot_nav_shared.h
illwieckz: @bmorel is this thread solved?

src/shared/bot_nav_shared.h
illwieckz: @bmorel is this thread solved?

bmorel: this comment was always wrong, the default was always here, but using C++ style instead of C style.

src/shared/bot_nav_shared.h
illwieckz: Does using `int` instead of `bool` would solve the thread?

src/shared/bot_nav_shared.h
illwieckz: Can a commit squash close solve thread?

src/shared/bot_nav_shared.h
illwieckz: @slipher is the code correct even if that's not the way you would do it?

slipher: > @slipher is the code correct even if that's not the way you would do it?

No because it breaks the endianness swapping stuff. As I said elsewhere in the thread, we could consider getting rid of that stuff, but that should be done intentionally, rather than by letting it fall into disrepair.

slipher:
I'd make a fixed version if anyone vouches that the crouching navgen option is something they plan to use

bmorel:
endianness of a file that is destined to be generated and used on the same computer seems a rather ridiculous thing.
Ok, got enough of this.

illwieckz:
I totally don't mind if navmesh is totally incompatible when booting the same Power computer as ppc or ppcle. All I want is that the game actually detect invalid files if they are invalid, so it can regenerate them.

> No because it breaks the endianness swapping stuff. As I said elsewhere in the thread, we could consider getting rid of that stuff, but that should be done intentionally, rather than by letting it fall into disrepair.

I would just had a comment "we totally assume we don't care about endianness on this file".

> But even without that, `bool` is a bad idea to store to disk since not all possible bytes are valid values for it, so an invalid one could cause undefined behavior.

Let's just change that `bool` into a `int` or byte, and that part is fixed. Files are stored as bytes anyway.

slipher:
> I would just add a comment saying "we totally assume we don't care about endianness on this file".

Sure but if you do that you should actually remove the endianness swapping code. Not let it fall into disrepair.

illwieckz:
Where is that endianness swapping code, is it used on something else?

bmorel:
endianness support *was* important before the navgen merge, I'd agree with that without any second of doubt.
Now, it's totally useless, and forcing the use for the only reason that nothing else needs it is... I won't end this phrase.
Also, I would consider it's the engine's job to care about endianness, not the gamelogic's. Is not the engine some kind of VM after all?

bmorel:
Just so that it's noted: I closed this PR, which means I'm not the one which will maintain it anymore. The code is open, you can do everything with it, it's the essence of FOSS.
On my side, I'll probably start seriously working on the pathfinding stuff. We overuse obstacles for buildings, and the stuck code is quite bad. Our support for navcons is so trivial it's almost absent (only @sweet235 worked on it, which shows that even trivial stuff can change things a lot).
I'll use, ofc, my version of code, maybe the refactors will be easy to port, maybe they won't. I said it long ago I wanted to merge as much as possible before doing the wallwalking (well, @sweet235 found how to make my commit working, so that even better) but this part of the code still needs lot of work (jetpacks, submarine tunnels. Those should not be that hard to "manually implement", with navcons, but will make the code more of a mess, especially since there's the fighting problem). 
Map accessibility by bots is very important, and PvE gaming feels less about learning bots tricks to exploit them, and this effort should probably be continued, so that it would be impossible to blaster an overmind while alien bots move around. 

This very PR *will* increase the number of cases of humans stuck under stairs (this already happens with marauders on nexus6 notably), because the code is too trivial.
It's the only reason why it's optional at all actually, and I may know a fix (got the idea few days ago) for that (this is a problem that happens with a combination of autojump and crouching, actually): checking the dir.z: if the bot wants to go higher, crouching is probably a bad idea, unlike jumping. If can't jump, then should likely steer.
In any case, jumping tiles and crouching tiles should have a travel cost (very low for aliens since it's faster than walking, but higher for humans for the stamina cost). This need to be investigated.
And navcons must carry informations, and as much as possible be made fully automatic, like with ladders (since the detection code works correctly, it means we can get various checks for most classes to see if they can go through by jumping, pouncing, or wallwalking. This needs ladders dimensions to be kept in memory until navmeshes are generated, then for each class, check by tracing if the ladder can actually be used) or the pushers that are linked to a "reception" entity (don't remember the name, but yes, that exists) if not too far (distance in number of navmeshes is a limitation of recast though).

In short: too much  work to do to waste time on quarrels about endianess for a file that is generated and used on the same machine, with a game that only runs on intel and arm. I thought intel had fixed endianness, unlike arm which can choose, but it seems [I was wrong](https://en.wikipedia.org/wiki/Endianness) and both can do both endianness:

> Bi-endianness
>
Some architectures (including [ARM](https://en.wikipedia.org/wiki/ARM_architecture) versions 3 and above, [PowerPC](https://en.wikipedia.org/wiki/PowerPC), [Alpha](https://en.wikipedia.org/wiki/DEC_Alpha), [SPARC](https://en.wikipedia.org/wiki/SPARC) V9, [MIPS](https://en.wikipedia.org/wiki/MIPS_architecture), [Intel i860](https://en.wikipedia.org/wiki/Intel_i860), [PA-RISC](https://en.wikipedia.org/wiki/PA-RISC), [SuperH SH-4](https://en.wikipedia.org/wiki/SuperH) and [IA-64](https://en.wikipedia.org/wiki/IA-64)) feature a setting which allows for switchable endianness in data fetches and stores, instruction fetches, or both. This feature can improve performance or simplify the logic of networking devices and software. The word bi-endian, when said of hardware, denotes the capability of the machine to compute or pass data in either endian format.
>
Many of these architectures can be switched via software to default to a specific endian format (usually done when the computer starts up); however, on some systems, the default endianness is selected by hardware on the motherboard and cannot be changed via software (e.g. the Alpha, which runs only in big-endian mode on the [Cray T3E](https://en.wikipedia.org/wiki/Cray_T3E)).

Still, before being bi-endian, arm was little-endian, like intel's defaults, so it's very unlikely to get problem of that genre with the systems we currently *can* function on. Especially for a file that is used by the same machine that generates it.

bmorel:
I did 2 changes on @sweet235 request, to make the navgen opt-in and the behavior change opt-out, this one controlled by a cvar.

bmorel:
I accidentally sneaked in the skillset removal commit on which I base all my work on top, nowadays.

So, this PR introduces (again) crouch code, which can be disabled in 2 ways:

1: in navmesh generation, defaults to off
2: in obstacle handling, there's a cvar which defaults to on, so that bots can crouch under turrets & the like

Code works, and I consider the current "requirements" ridiculous: there is no need to bother about endianness for files which are generated locally. If the endianness point is just so that the endianness support code is still useful, then this support should be removed as it's in practice useless.

slipher:
> I accidentally sneaked in the skillset removal commit on which I base all my work on top, nowadays.
> 
> So, this PR introduces (again) crouch code, which can be disabled in 2 ways:
> 
> 1: in navmesh generation, defaults to off 2: in obstacle handling, there's a cvar which defaults to on, so that bots can crouch under turrets & the like
> 
> Code works, and I consider the current "requirements" ridiculous: there is no need to bother about endianness for files which are generated locally. If the endianness point is just so that the endianness support code is still useful, then this support should be removed as it's in practice useless.

I'm tired of explaining the difference between properly removing a feature vs. deliberately breaking it. So would you like me to make a fixed version of this PR and merge it?

cu-kai:
I don't think the endianness debate should block this PR, that could be addressed by another once the feature is implemented. I've seen crouching bots work fine on @sweet235's server.

bmorel:
> I'm tired of explaining the difference between properly removing a feature vs. deliberately breaking it. So would you like me to make a fixed version of this PR and merge it?

And I'm tired of your constant refusal to let bot crouching for bad reason.
2 years ago, it was because you "forgot" to merge daemonmap's PR before porting navgen into gamelogic, and now it's for some weird endianness reason, on a game that only runs on intel or arm, which are both little-endian (ok, technically, those archs support both endianness, but I still have to see an OS doing big-endian on them).

Also, gamelogic should not have to bother about endianness: this is the engine's job.

This code work and is deployed on several servers, and I won't spend more unpaid time (like everyone in this project, I am not paid to contribute) to do changes on working code that make no sense.
Upstream is free to reject or not this code. On my end, I'll simply keep running it, as I have no reason not to.

Also, because I'm, (like you, but on another topic that only concerns people who play) **tired** of bots constantly being stuck because the obstacle code is an ugly pile of patches, I'm going to build on top of this.
The resulting commits will ofc have git conflict with this, but it's not my fault if upstream, for __the second time__ refuses to merge improvements.

slipher:
All right I merged this code.
b6f1d0a898f20ab44b93875cb69c4cdf82e4eb9f
22a4fc999a43e8e64e056f019e430bf3e9a37745
7e18d379d61942dbc5cbf007c9da053d68e8aaae
635cdf1b2870b4108605e39ab374c920848252d0

PR #2570, opened 2023-04-03
teach bots how to use luci

bmorel:
cherry-picked from my betterai mod

This allows bots to handle the powerful lucifer cannon correctly, instead of wasting energy and time. It does not makes the lcanon usage less previsible though, but it does make it deadlier when bots know how to aim, which is controlled by a bot's skill (this means: this does not makes low skill bots too much stronger by itself).

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -2100,30 +2100,30 @@ void BotFireWeaponAI( gentity_t *self )
 				int selfHP = self->client->ps.stats[ STAT_HEALTH ];
 
 				gentity_t const* botTarget = self->botMind->goal.getTargetedEntity();
-				float targetHP;
+				float targetMaxHP;
 				if ( botTarget->client )
 				{
 					class_t targetType = static_cast<class_t>( botTarget->client->ps.stats[ STAT_CLASS ] );
 					const classAttributes_t * targetAttr = BG_Class( targetType );
 					// Ignore armor because aliens don't have armors or damage reductions anyway.
-					targetHP = static_cast<float>( targetAttr->health );
+					targetMaxHP = static_cast<float>( targetAttr->health );
 				}
 				else //assume we're fighting a building, but it could also be a map entity, and this would then fail
 				{
 					auto targetType = static_cast<buildable_t>( botTarget->s.modelindex );
 					const buildableAttributes_t * targetAttr = BG_Buildable( targetType );
-					targetHP = static_cast<float>( targetAttr->health );
+					targetMaxHP = static_cast<float>( targetAttr->health );
 				}
 				// TODO: use charging is there are more than one enemies around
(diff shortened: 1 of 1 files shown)


src/sgame/sg_bot_util.cpp
```diff
@@ -2095,9 +2095,40 @@ void BotFireWeaponAI( gentity_t *self )
...
+				// TODO: use charging is there are more than one enemies around
+				if ( targetHP <= LCANNON_SECONDARY_DAMAGE || selfHP <= CRITICAL_HEALTH )
+				{
+					BotFireWeapon( WPM_SECONDARY, botCmdBuffer );
+				}
+				else
+				{
+					float dmgRatio = targetHP / LCANNON_DAMAGE;
+					float chargeMax = static_cast<float>( LCANNON_CHARGE_TIME_MAX );
+					float charge = chargeMax * std::min( dmgRatio, 0.95f );;
```
slipher, line 2126: extra `;`

src/sgame/sg_bot_util.cpp
```diff
@@ -2095,9 +2095,40 @@ void BotFireWeaponAI( gentity_t *self )
...
 			break;
 		case WP_LUCIFER_CANNON:
-			if ( self->client->ps.weaponCharge < LCANNON_CHARGE_TIME_MAX * Math::Clamp( random(), 0.5f, 1.0f ) )
 			{
-				BotFireWeapon( WPM_PRIMARY, botCmdBuffer );
+				const int CRITICAL_HEALTH = 30; //TODO do not use a constant
+				int selfHP = self->client->ps.stats[ STAT_HEALTH ];
+
+				gentity_t const* botTarget = self->botMind->goal.getTargetedEntity();
+				float targetHP;
```
slipher, line 2103: Why is this set to the max health rather than the current health?

bmorel: Because i do not like cheating bots. Using actual health means that even if they catch a fleeing enemy (without hearing it taking a shot) they'll still know exactly how much to load their weapon, which is basically cheating.

slipher: It should be named `targetMaxHP` then.

So ignoring the self HP condition, this amounts to use secondary fire iff the target is a dretch. That seems quite different from the strategy most human players use which is to shoot the floor with the primary attack, targeting dretches with splash damage. Maybe it could work if the bot's aim is really good

bmorel: it's actually very good. And I think players use splash because this allows to hit stuff around, at the expense of requiring more time to load. Not sure. Myself I also use the 2ndary when fighting dretches... because it's faster to reuse, too.
I guess I can just make bots use primary all the time if you prefer

slipher:
I tried it out and it does work well. I was underestimating how much splash damage the secondary fire does. LGTM

bmorel:
>  was underestimating how much splash damage the secondary fire does. LGTM

I think most of us does, really. Maybe I should avoid make bots using it, as it could teach players how to improve :D

PR #2573, opened 2023-04-03
increase MAX_CBEACONS to 1024

bmorel:
50 is not even enough to hold all players, since the max client value is 64. This is thus obviously not enough for PvE layouts or any game where builders put enough buildings around.

1024 can probably be "not enough", or it might be "too much", I just choose this value randomly.

Easy way to test is to load a PvE map, I did the test with eggs-master.

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -630,7 +630,7 @@ struct buildableCache_t
 
 //=================================================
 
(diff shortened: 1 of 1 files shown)


slipher:
This is just an array of pointers so the memory cost is small. The beacon struct is in centity_t so we're already paying for the memory bloat of a full 8000 beacons

LGTM

PR #2574, opened 2023-04-03
allow to disable obituaries

bmorel:
This allows to actually read the chat, instead of it being constantly flushed by the "foo killed bar" spam. Of course, it's disabled by default as people might enjoy it.

### DIFF
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -102,8 +102,13 @@ static const struct {
 	{ "",              true,  false, TEAM_NONE },   // (MOD_REPLACE)
 };
 
+static Cvar::Cvar<bool> cg_showObituaries( "cg_showObituaries", "show obituaries in chat", Cvar::NONE, true );
 static void CG_Obituary( entityState_t *ent )
 {
+	if ( not cg_showObituaries.Get() )
+	{
+		return;
+	}
 	int          mod;
 	int          target, attacker, assistant;
 	int          attackerClass = -1;
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -105,7 +105,7 @@ static const struct {
 static Cvar::Cvar<bool> cg_showObituaries( "cg_showObituaries", "show obituaries in chat", Cvar::NONE, true );
 static void CG_Obituary( entityState_t *ent )
 {
-	if ( not cg_showObituaries.Get() )
+	if ( !cg_showObituaries.Get() )
 	{
 		return;
 	}


slipher:
Can we name it "showObituaries" instead (avoiding NOT conditions in names)

bmorel:
Sure. I'm actually thinking I should add this some granularity, like, filtering out only enemy players, filtering all enemies, etc.

bmorel:
> Sure. I'm actually thinking I should add this some granularity, like, filtering out only enemy players, filtering all enemies, etc.

But I feel too lazy to do that

src/cgame/cg_event.cpp
```diff
@@ -102,8 +102,13 @@ static const struct {
 	{ "",              true,  false, TEAM_NONE },   // (MOD_REPLACE)
 };
 
+static Cvar::Cvar<bool> cg_showObituaries( "cg_showObituaries", "show obituaries in chat", Cvar::NONE, true );
```
Viech, line 105: Nitpick: Maybe use a different wording like "death messages" in the description only as the word obituary may not be self-explanatory to non-english speakers.

Viech:
This is a good PR as-is (with the logic fix) but the option to show obituaries hat affect you (you made the kill or assisted, you got killed) would be a really nice middle ground to completely disabling it.

bmorel:
> This is a good PR as-is (with the logic fix) 

The initial name was about allowing to disable, hence the "logic fix".

> but the option to [...]

Yes, it would be nice to have higher granularity on this, but:

> But I feel too lazy to do that


What this PR tries to do is to reduce the constant noise generated by kills, which is one of the reasons communication is barely possible at all. This is obviously a surface fix, for a problem which needs a lot more work to be properly done.

This PR was also opened before the submodule merge, so now what I should really do is to add a GUI to configure this.

src/cgame/cg_event.cpp
bmorel: This problem would also apply to the variable's name. Not too fond of the wording "death messages" neither. And discovering cvars is hardly possible anyway... Either you know they exist, or you don't.

This whole obituaries thing needs proper work, and should not even be difficult to do, same as the whole chat situation, but this requires someone to tackle it for real. Which I don't think will be me, as I am barely able to add checkboxes without breaking formatting in rmlUI.

Viech:
> > This is a good PR as-is (with the logic fix)
> 
> The initial name was about allowing to disable, hence the "logic fix".

I believe you have inverted the name/syntax, but not the logic/semantic yet. See my non-nitpick code comment above.

src/cgame/cg_event.cpp
Viech: My point was just to reduce redundancy in cvar name and description to make the latter potentially more helpful as an explanation. But this is a nitpick, so if you disagree then that resolves it.

bmorel:
Yes indeed, that was just to explain.

bmorel:
i'll squash commits when approved, but should be ready now

slipher:
LGTM

PR #2580, opened 2023-04-05
Firebomb cooldown

sweet235:
Add a cvar `g_firebombCooldown`, that is a time span in milliseconds. Clients can only buy a firebomb after at least this much time has passed since the previous buy. I set this to 2 seconds on my server right now.

I only did this after i have been convinced by multiple persons that we currently have a problem with firebomb spam. Some people (@illwieckz) can regularly be observed standing on armouries and throwing 10 firebombs in a fraction of a second. That makes my poor machine slow and my blind bots burn.

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -3310,6 +3310,7 @@ static bool Cmd_Sell_conflictingUpgrades( gentity_t *ent, upgrade_t upgrade )
 	return sold;
 }
 
+Cvar::Cvar<int> g_firebombCooldown("g_firebombCooldown", "clients can buy firebombs every <g_firebombCooldown> milliseconds.", Cvar::NONE, 0);
 
 static bool Cmd_Buy_internal( gentity_t *ent, const char *s, bool sellConflicting, bool quiet )
 {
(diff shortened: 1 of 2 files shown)


bmorel:
being able to spam grenades is indeed a problem, and can easily make computers run *very* slowly.

slipher:
I disapprove. There shouldn't be a random purchasing restriction on one item at the armoury that doesn't apply to any of the other items.

https://github.com/Unvanquished/gameplay/issues/58 would be a better way to address this

illwieckz:
Instead of a cooldown on buying, I would set a cooldown on firing as, “no one can fire a second grenade before a given time after a first grenade”. That would be the first objective and reaching it would be enough to counteract the grenade spam hack. I believe there is a consensus on the objective “prevent to spam many grenades in a row without a cooldown” whatever where is the cooldown.

There is also a second objective, but that can be done later, discussed independently, or done at the same time if we agree and this is easier to do both at once. We may even prevent usage of any weapon for the cooldown time… We may actually hide the first person weapon while the grenade countdown is counting (with hope that in the future we would have a first person grenade throw animation to display instead).

Viech:
I agree with the cooldown on firing, unless this needs a new/dedicated netcode field to work; then the server-side solution of a buy timer seems a better fit/less intrusive solution.

PR #2582, opened 2023-04-05
proper barricade handling

### LINKED ISSUE #1863: proper handling of barricade shrinking for bots
Bots don't really know how to jump over barricades.
If they see a barricade and are aliens, they jump. If they're not fast fleeing.
Barricades have a constant indicating their shrink size, it should be used by obstacle handling.
Or better, navmesh for aliens should be updated so that they understand the barricade's actual size. Humans should not, because they're not meant to know about it yet.
Not an easy topic, despite how it looks.

bmorel:
current barricade handling by bots is suboptimal, because the actual size of shrunk barricade is never considered. This commit solves the problem.
Of course it also means that bots will try to steer around, which is likely useless as barricades are usually in narrow corridors. What should happen is that the action MoveToGoal shuold then fail, but this problem is for another day to solve.

Fix #1863

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -592,11 +592,23 @@ static bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const glm:
 		&& trace.surfaceFlags & SURF_LADDER
 		;
 
+	if ( blocker->s.eType == entityType_t::ET_BUILDABLE
+			&& blocker->s.modelindex == BA_A_BARRICADE
+			&& G_OnSameTeam( self, blocker ) )
+	{
+		glm::vec3 mins, maxs;
+		BG_BoundingBox( static_cast<buildable_t>( blocker->s.modelindex ), &mins, &maxs );
+		maxs.z *= BARRICADE_SHRINKPROP;
+		glm::vec3 pos = VEC2GLM( blocker->s.origin );
+		BG_BoundingBox( pClass, &playerMins, &playerMaxs, nullptr, nullptr, nullptr );
+		if ( origin.z + playerMins.z >= pos.z + maxs.z )
+		{
+			return true;
+		}
+	}
+
 	//if we can jump over it, then jump
-	//note that we also test for a blocking barricade because barricades will collapse to let us through
-	return blocker->s.modelindex == BA_A_BARRICADE
-		|| trace.fraction == 1.0f
-		|| ladder;
+	return trace.fraction == 1.0f || ladder;
 }
 
 // Determine a new direction, which may be unchanged (forward) or sideway. It


bmorel:
Fix #1863

src/sgame/sg_bot_nav.cpp
```diff
@@ -586,9 +586,24 @@ static bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const glm:
...
+	{
+		glm::vec3 mins, maxs;
+		BG_BoundingBox( static_cast<buildable_t>( blocker->s.modelindex ), &mins, &maxs );
+		if ( G_Team( self ) == TEAM_ALIENS )
+		{
+			maxs.z *= BARRICADE_SHRINKPROP;
+		}
+		glm::vec3 pos = VEC2GLM( blocker->s.origin );
+		BG_BoundingBox( pClass, &playerMins, &playerMaxs, nullptr, nullptr, nullptr );
+		if ( origin.z + playerMins.z >= pos.z + maxs.z )
```
slipher, line 599: This tests whether the player's feet are above the top of the (hypothetically) shrunken barricade. Surely that's not what you want

src/sgame/sg_bot_nav.cpp
```diff
@@ -586,9 +586,24 @@ static bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const glm:
...
+		{
+			return true;
+		}
+	}
+
 	//if we can jump over it, then jump
-	//note that we also test for a blocking barricade because barricades will collapse to let us through
-	return blocker->s.modelindex == BA_A_BARRICADE || trace.fraction == 1.0f;
+	return trace.fraction == 1.0f;
```
slipher, line 606: What if you just changed this to `return (blocker->s.modelindex == BA_A_BARRICADE && G_Team(self) == TEAM_ALIENS ) || trace.fraction == 1.0f;` ?

bmorel: You mean keeping the old code, that breaks in many situations, like when barricade is placed so high that even shrunk, a tyrant can't pass it?

src/sgame/sg_bot_nav.cpp
```diff
@@ -592,11 +592,24 @@ static bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const glm:
 		&& trace.surfaceFlags & SURF_LADDER
 		;
 
+	if ( blocker->s.modelindex == BA_A_BARRICADE )
```
slipher, line 595: I think we should check for ET_BUILDABLE since `modelindex` could have a different meaning for other entities.

Also this block ought to be skipped for humans since the normal trace already works correctly for them.

src/sgame/sg_bot_nav.cpp
bmorel: mins are basically the "half-cube" of entities, starting from origin and going downside. Yes, it's not always very readable, I agree. Especially when you think that in fine, the only useful info there is the negative height, as the half-extents can easily be derived from maxs.
But that's not a refactoring I'll do soon for obvious reasons :D

bmorel: yes, they're the only ones allowed to. It's used for *most* models I think. I'm not even sure they're symetrical once.

As I've said on IRC though, I want this as functional as possible before a real refactoring that will be necessary to implement wallwalking.
I hope that having as much features as possible, even if in a PoC state, will make it easier to test for regressions and for reviewers to get what happens. But from my memory, wallwalking is way too fragile to be implemented in such a hackish way as the autojump, crouching or ladder code.

src/sgame/sg_bot_nav.cpp
bmorel: done

slipher:
LGTM

PR #2584, opened 2023-04-06
giving poison to aliens when healing 8x

bmorel:
grabbed @sweet235 commit to give poison to aliens from afar, and added a commit to make this opt-in.

I just tried to play with vanilla's mechanism (that is, have to touch booster) and definitely think it's better to *not* have to touch booster, even if I think it would be nice to have smaller range, but then, smaller range would likely need a visualisation system and I do not know how to implement that.


slipher:
> I just tried to play with vanilla's mechanism (that is, have to touch booster) and definitely think it's better to _not_ have to touch booster

Why?

> even if I think it would be nice to have smaller range, but then, smaller range would likely need a visualisation system and I do not know how to implement that.

There is already a HUD element showing when you have poison. What more would you need?

bmorel:
> Why?

it can be annoying to grab poison when a tyrant or dragoon camps the booster, for example.

> There is already a HUD element showing when you have poison. What more would you need?

We could say the same about the healing mod, and yet I think the sphere around booster is quite useful. I would rather avoid adding new "hiden" mechanics that makes the game harder to learn for beginners.

slipher:
Sphere around the booster, you mean the one drawn for grangers? Grangers have little need for poison so I don't think that matters.

bmorel:
oh, right, I thought it was also shown to other classes.

sweet235:
> Why?

That is a hack to make bots pick up poison. But it also helps if someone is sitting on the only booster.

I'm not quite sure if this is a good idea. You can get poison through walls for example.

bmorel:
> My main motivation for running this right now is bot balance

we currently have code to make bots go touch booster when they move close enough (resupplyscore) but you need to combine it with a distance check in BT, otherwise they'll camp.
I still think this is more comfortable for players too, though, but:

> I'm not quite sure if this is a good idea. You can get poison through walls for example.

That's why I turned this into an opt-in cvar.

sweet235:
I'm not sure if the 8x healing through walls is a good idea either.

bmorel:
> But most sensibly placed boosters are hard for current bots to touch.

Indeed.

Viech:
I like the change to remote healing in general but I think this solution is not great. Instead of hiding it behind a default-off cvar, I would prefer if it was improved to a point where most people can agree that this is strictly an improvment over the old state. I think the following would be sufficient for this:

1. Do not fill the poison meter immediately but over the course of maybe one or two seconds.
2. Let the booster emit spores also when providing poison.
3. _Optional/to-be-considered-later: Distinguish healing and poison-replenishment with different colored clouds, possibly only provide one at a time._

With these changes, the (visual) communication of acquiring poison would improve strictly, the whole process would be smoother than before, and it would be harder to argue that this makes acquiring poison too easy or too fast compared to before.

bmorel:
> Do not fill the poison meter immediately but over the course of maybe one or two seconds.

Good idea, but out of the scope. This PR is a simple one, which is off by default because it does imply gameplay changes *but* seems a globally accepted feature.
It's also used by the most played mod, and I think it's impact is too low to justify not helping sweet to maintain his mod.

> Let the booster emit spores also when providing poison.

IIRC enemies do see the booster's shader effect when a player is using it, so that should be doable for poison too.

bmorel:
if the consensus is to not allow this option, let's close this in few days.

bmorel:
will close this PR this week end.

bmorel:
@Viech if this solution is not great, ask yourself if it's better than current situation instead of requiring much more complicated changes.

cu-kai:
To be honest, this LGTM. We've been using it on @sweet235's server for long enough (months at this point) and I don't think it unbalances games.

I think one change that could be made is: give it a smaller range. But that should not be considered a blocker to merging.

illwieckz:
That looks fine to me too.

Viech:
> @Viech if this solution is not great, ask yourself if it's better than current situation instead of requiring much more complicated changes.

I have suggested but not _required_ changes (I will use the review feature when I do). I have—in response to a lack of enthusiasm on this PR—proposed an improvement that would render this feature a strong move forward to the point where we would not need to support two modes, both of which are reasonable but not convincing without further polishing.

Was this feature done in a more thorough way (including visual communication and a quick-but-smooth filling of the poison meter)—as opposed to a quick and functional fix to bots not being able to pick up poison—then people in here would very likely just love this change and there would be not much of a discussion. My honest opinion is thus that this hasn't moved not despite but _because_ it is such a tiny change.

So my verdict is: LNDTM – Looks non-detrimental to me. I'm not against it if others want this as it is.

Issue #2564, opened 2023-04-01
allow to mark enemy building faster


bmorel:
Currently, you need to look at them for several seconds, which as alien to do so from afar means you're defenseless

PR #2588, opened 2023-04-06
make tagging duration configurable and default to 0.5s

bmorel:
Current duration of 1s feels way too much when you're staring at a rocketpod.

### DIFF
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -33,6 +33,7 @@ bool ClientInactivityTimer( gentity_t *ent, bool active );
 
 static Cvar::Cvar<float> g_devolveReturnRate(
 	"g_devolveReturnRate", "Evolution points per second returned after devolving", Cvar::NONE, 0.4);
+static Cvar::Cvar<int> g_tagDelay( "g_tagDelay", "duration in ms one has to stare at enemy to place a beacon", Cvar::NONE, 500 );
 
 /*
 ===============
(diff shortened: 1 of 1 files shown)


bmorel:
fix #2564

bmorel:
Actually, I think this improves on, but does not makes the situation much better: the beacon is still only put if a precise trace hits, which is unlikely when from afar, especially if there are other buildings nearby as BBoxes can "hide" the real target.

Viech:
> Actually, I think this improves on, but does not makes the situation much better: the beacon is still only put if a precise trace hits, which is unlikely when from afar, especially if there are other buildings nearby as BBoxes can "hide" the real target.

I did a quick test and it seems that you can tag both by

1. looking in the general direction if there is a clear line of sight to the target's origin
2. looking directly at an enemy building's bounding box whose origin is not visible

I think this is rather sensible. If you make tagging more aggressive than this, you might show the player buildings hidden around a corner whose bounding box sticks out.

I suggest to merge this PR as is (nitpick: "stare at _an_ enemy").

bmorel:
> I think this is rather sensible. If you make tagging more aggressive than this, you might show the player buildings hidden around a corner whose bounding box sticks out.

I agree it's sensible for automatic tagging. And more improvements on this would likely be more complex, affect gameplay more, and thus brings their own problems. Totally out of the scope of this.

Viech (approved):

PR #2593, opened 2023-04-07
not allowed is not unknown

### LINKED ISSUE #2517: trying to spawn with a ckit when it's disabled gives wrong message
It gives: `Unknown starting item`.

For a start, the option should be disabled one way or another, but also the error message should say the item is forbidden, not unknown.
I had this on map `siege`

bmorel:
fix #2517


src/sgame/sg_cmds.cpp
```diff
@@ -2758,19 +2758,8 @@ static bool Cmd_Class_spawn_internal( gentity_t *ent, const char *s, bool report
...
-			weapon = WP_MACHINEGUN;
-		}
-		else if ( !Q_stricmp( s, BG_Weapon( WP_HBUILD )->name ) &&
-		          !BG_WeaponDisabled( WP_HBUILD ) )
-		{
-			weapon = WP_HBUILD;
-		}
-		else
+		weapon_t weapon = BG_WeaponNumberByName( s );
+		if ( weapon == WP_NONE )
```
slipher, line 2762: This allows e.g. `/class lcannon`

bmorel: oh right. Fixed.

bmorel:
I think this works, I should probably squash the commits though, if someone review?

src/sgame/sg_cmds.cpp
```diff
@@ -2779,6 +2771,12 @@ static bool Cmd_Class_spawn_internal( gentity_t *ent, const char *s, bool report
 			return false;
 		}
 
+		if ( BG_WeaponDisabled( weapon ) || end == std::find( start, end, weapon ) )
+		{
+			G_TriggerMenu( ent->client->ps.clientNum, MN_H_SPAWNITEMNOTALLOWED );
```
slipher, line 2776: Needs to respect the `report` option

src/sgame/sg_cmds.cpp
```diff
@@ -2758,19 +2758,11 @@ static bool Cmd_Class_spawn_internal( gentity_t *ent, const char *s, bool report
...
-			weapon = WP_MACHINEGUN;
-		}
-		else if ( !Q_stricmp( s, BG_Weapon( WP_HBUILD )->name ) &&
-		          !BG_WeaponDisabled( WP_HBUILD ) )
-		{
-			weapon = WP_HBUILD;
-		}
-		else
+		weapon_t birthWeapons[] = { WP_HBUILD, WP_MACHINEGUN };
+		auto start = std::begin( birthWeapons );
```
slipher, line 2762: Nit: I think it would be more readable without the `start` and `end` variables

slipher:
LGTM

PR #2597, opened 2023-04-07
Cancel votes

### LINKED ISSUE #1940: allow admins to start a vote without voting
Currently, when one starts a vote, it is considered this person if in favor for the change. No way to be neutral.

### LINKED ISSUE #1941: allow to change one's vote
Currently, it is not possible to change one's opinion, which is annoying after discussions.

bmorel:
This PR does 3 things:

1. it makes the G_Vote function less convoluted
2. fixing the vote cancellation system while we're at it
3. allows players to change their mind (i.e. because of a typo or because they were talked into a different point of view)

fix #1941 
fix #1940 as "good enough" since then admins can immediately change their mind (if they allowed that beforehand, but if they're admins...)

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -47,6 +47,7 @@ Cvar::Range<Cvar::Cvar<int>> g_killDelay(
 		"how many seconds a player needs to wait before the suicide (/kill) command take effect",
 		Cvar::NONE,
 		20, 0, 120 );
+static Cvar::Cvar<bool> g_voteAllowChange( "g_voteAllowChange", "allows players to change their mind (they might spam)", Cvar::NONE, false );
 
 /*
 ==================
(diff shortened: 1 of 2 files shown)


bmorel:
The code is still quite convoluted after this, but it's better, and refactoring this is out of the scope of my PR. There's enough stuff to review as there are, right?

sweet235:
Note: this is not about `/cancelvote`.

bmorel:
> Note: this is not about /cancelvote (I think).

It is not.

slipher:
First commit fixing the config string updating looks good.

The user interface for changing votes may be confusing. If you click "yes", then "no", your vote will be canceled but then if you click "no" a second time, the vote changes to no.

This needs some kind of rate limiting. Currently you can vote arbitrarily often, spamming the users with the vote sound.

bmorel:
> The user interface for changing votes may be confusing. If you click "yes", then "no", your vote will be canceled but then if you click "no" a second time, the vote changes to no.

Agree, but to be honest, I was not expecting to have to do that many changes just for such a basic feature, so I focused on making the cancel vote actually working, since it previously was not.

> Currently you can vote arbitrarily often, spamming the users with the vote sound.

I am aware of this, and this is the reason why I don't propose it as opt-out, but as opt-in. I am not too sure how to do a rate limit that would integrate nicely with the other mechanics like the chat spamming, since this is exactly this that we're talking about.
I also wonder if the problem is not simply that voting emits sound... what does this sound brings, exactly? Why would one care about the fact someone else voted?

slipher:
I took the liberty of pushing the vote counter bug fix commit `fix vote cancelling (and make code readable)` 538a96466e70783b67e21897fe11d4ede9c45c79.

bmorel:
I think you're right, the interface is confusing. But I don't have the strength to work on improving it further, so I'll close this in few days.

bmorel:
will close this this week end

bmorel:
So, before this week end arrives, and it will arrive quickly, let's ask this: is this preferable to current situation or not, considering the fact it is optional?

bmorel:
As promised.

PR #2599, opened 2023-04-07
Implement `common/outside`

illwieckz:
Implement `common/outside` material.

See PR on data side for more explanations:

- https://github.com/UnvanquishedAssets/tex-common_src.dpkdir/pull/6

To not draw navmesh on the surface, the navgen is expected to actually recognize the `textures/common/outside` name the same way it already recognizes `textures/common/caulk`, which was probably an early attempt to exclude some surfaces that are usually not seen within maps. I see no reason to waste a precious surfaceparm for that.

I also created special `SkipContent` and `SkipSurface` functions to be usable on both brushes and patches without omitting stuff. Even if some skip conditions doesn't make much sense (like a caulked patch), a mapper can still do nonsensical things (like a caulked patch) so we better be sure we filter everything.

### DIFF
--- a/src/shared/navgen/nav.cpp
+++ b/src/shared/navgen/nav.cpp
@@ -228,16 +228,30 @@ static void AddTri( std::vector<float> &verts, std::vector<int> &tris, vec3_t v1
 	AddVert( verts, tris, v3 );
 }
 
+static bool SkipContent( const dshader_t *shader ) {
+	const int solidFlags = CONTENTS_SOLID | CONTENTS_PLAYERCLIP;
+
+	return !( shader->contentFlags & solidFlags );
+}
+
+static bool SkipSurface( const NavgenConfig *config, const dshader_t *shader ) {
+	if ( config->excludeSky && shader->surfaceFlags & SURF_SKY )
+	{
+		return true;
+	}
+
+	if ( config->excludeCaulk && !Q_stricmp( shader->shader, "textures/common/caulk" ) )
+	{
+		return true;
+	}
+
+	return false;
+}
+
 // TODO: Can this stuff be done using the already-loaded CM data structures
 // (e.g. cbrush_t instead of dbrush_t) instead of reopening the BSP?
 void NavmeshGenerator::LoadTris( std::vector<float> &verts, std::vector<int> &tris ) {
 	std::map<size_t,std::unordered_set<glm::vec3>> laddersVerts;
-	int solidFlags = CONTENTS_SOLID | CONTENTS_PLAYERCLIP;
-	int surfaceSkip = 0;
-
-	if ( config_.excludeSky ) {
-		surfaceSkip = SURF_SKY;
-	}
 
 	const byte* const cmod_base = reinterpret_cast<const byte*>(mapData_.data());
 	auto& header = *reinterpret_cast<const dheader_t*>(mapData_.data());
@@ -270,9 +284,11 @@ void NavmeshGenerator::LoadTris( std::vector<float> &verts, std::vector<int> &tr
 		int firstSide = bspBrushes[i].firstSide;
 		const dshader_t* brushShader = &bspShaders[bspBrushes[i].shaderNum];
 
-		if ( !( brushShader->contentFlags & solidFlags ) ) {
+		if ( SkipContent( brushShader ) )
+		{
 			continue;
 		}
+
 		/* walk the list of brush sides */
 		for ( int p = 0; p < numSides; p++ )
 		{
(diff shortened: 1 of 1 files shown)


src/shared/navgen/nav.cpp
```diff
@@ -415,8 +440,10 @@ void NavmeshGenerator::LoadTris( std::vector<float> &verts, std::vector<int> &tr
 	for ( int k = model->firstSurface, n = 0; n < model->numSurfaces; k++, n++ )
 	{
 		const dsurface_t *surface = &bspSurfaces[ k ];
+		const dshader_t *shader = &bspShaders[surface->shaderNum];
 
-		if ( !( bspShaders[surface->shaderNum].contentFlags & solidFlags ) ) {
```
slipher, line 419: Why is the content flags check removed here?

src/shared/navgen/nav.cpp
```diff
@@ -231,7 +231,7 @@ static void AddTri( std::vector<float> &verts, std::vector<int> &tris, vec3_t v1
 static bool SkipContent( const dshader_t *shader ) {
 	const int solidFlags = CONTENTS_SOLID | CONTENTS_PLAYERCLIP;
 
-	if ( !( shader->contentFlags & solidFlags ) )
+	if ( !( shader->contentFlags & solidFlags ) == true )
```
slipher, line 234: Why add `== true`? It looks weird

src/shared/navgen/nav.cpp
illwieckz: I've mistakenly wrote the content check in the wrong block, now moved back to its expected place.

src/shared/navgen/nav.cpp
illwieckz: This was a joke trolling @bmorel, it's now removed.

illwieckz:
Anything extra comment on this? This looks ready to me.

bmorel:
IIRC I was fine with it, except for the trolling. One might also be concerned by the fact this is not used at all yet. Could we have some easy way to test it? I'd give my approval after that, can't test without example case after all.

illwieckz:
I can make a preview of chasm using it. I plan to use it on chasm, plat23, thunder and other maps with accessible sky (citadel, hamunaptra if we rebuild it…).

illwieckz:
For information that is meant to replace the various hacks that were used until today, like a slick brush with a playerclip surface to blend playerclip brush content parm with slick surfaceparm (or the other way I don't remember) by leveraging the fact only one surface from a brush contributes to contentparm, crossing the finger and hoping the map compiler doesn't actually actually doesn't read the same surface for both content and surface on the side facing the playable area, plus adding an extra nobuild brush. 😝️

illwieckz:
@bmorel:

- https://dl.illwieckz.net/b/unvanquished/wip/common-outside/map-chasm_1.2.3-20230407-132752-80a797b.dpk
- https://dl.illwieckz.net/b/unvanquished/wip/common-outside/tex-common_2.3-20210414-125356-4ce4107.dpk

src/shared/navgen/nav.cpp
```diff
@@ -415,8 +428,17 @@ void NavmeshGenerator::LoadTris( std::vector<float> &verts, std::vector<int> &tr
 	for ( int k = model->firstSurface, n = 0; n < model->numSurfaces; k++, n++ )
 	{
 		const dsurface_t *surface = &bspSurfaces[ k ];
+		const dshader_t *shader = &bspShaders[surface->shaderNum];
+
+		/* Patches don't have any content but their shader may
+		have contentparm. */
+		if ( SkipContent( brushShader ) )
```
slipher: The first commit would not compile as it uses an out-of-scope variable `brushShader`.

slipher:
LGTM

src/shared/navgen/nav.cpp
illwieckz: Ah right, it should be named `shader` and this change went into the second patch. I'm fixing this.

bmorel:
how urgent is that? I'd like to give it a try before lgtm it, but I'm a bit sick those days

slipher:
> how urgent is that? I'd like to give it a try before lgtm it, but I'm a bit sick those days

Note that `common/outside` is not used in any existing map, so that won't affect you for the moment. However there is a small change in BSP patch handling.

illwieckz:
Well actually the sample provided above is named `tex-common_2.3-20210414-125356-4ce4107.dpk` with date string being `20210414` (yeah, year 2021) because that's the commit date of my patch on material side. 2021 is the  time I first implemented the feature, but this feature was on hold waiting for navgen to be merged. So one could say a few days more would not make any difference, but on the other hand I really want it to be behind me. And I actually verified this works in 2021 and in 2023.

PR #2609, opened 2023-04-09
UI: stop spurious cvar setting by sliders

slipher:
For example if you set /g_freeFundsPeriod 4000 and then open the 'Server setup' menu, g_freeFundsPeriod will no longer be set to 3600.


bmorel (approved):
tried. It works well.

DolceTriade (approved):

PR #2610, opened 2023-04-09
Translation: split command line messages into another .pot

slipher:
The command-line messages are numerous and translating them is of questionable utility, so it's nice to have them in a separate file so that translators can prioritize the other, more important messages.

### DIFF
--- a/src/utils/generate_pot.sh
+++ b/src/utils/generate_pot.sh
@@ -30,7 +30,8 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 repo_dir="$(realpath "${script_dir}/../..")"
 dpk_dir="${repo_dir}/pkg/unvanquished_src.dpkdir"
 pot_dir="${dpk_dir}/translation"
-pot_file="${pot_dir}/game.pot"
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

DolceTriade (commented):
Do we need more changes to try to load a "commands.po" file?

illwieckz:
> Do we need more changes to try to load a "commands.po" file?

It looks like tinygettext iterates all file in search for po files:

https://github.com/Unvanquished/Unvanquished/blob/e4e94de219d2578cde8223c2d224a1db05396fe2/libs/tinygettext/src/dictionary_manager.cpp#L115-L145

Assuming we rely in this file-search code, this PR would be enough.

slipher:
I just noticed that vote strings are here, which would be a good thing to translate... but #2235 would fix that :)

illwieckz:
I wonder if those are admin stuff or game stuff:

```pot
#: src/sgame/sg_maprotation.cpp:1007
msgid "^3listrotation:^* there is no active map rotation on this server"
msgstr ""

#: src/sgame/sg_maprotation.cpp:1013
msgid "^3listrotation:^* there are no maps in the active map rotation"
msgstr ""
```

PR #2614, opened 2023-04-09
improve ladder support

bmorel:
This allows to support chasm's ladder.
To test, remove all H buildings, place one near the alien-side ladder (the only normal ladder there), build an armoury on the plateau, add a bot and use this navcon:

```
navcon 3
-1216 150 2282 -1216 260 2240 20 1 63 0
```

Without the patch, the bot won't successfully go inside the vents. With the path, it will. Of course it might have troubles doing that, as they currently have, but this is still an improvement, and the change is rather obvious.


slipher:
So it apparently enables detection of a ladder whose top is below the player's jump height. Makes sense. LGTM

PR #2619, opened 2023-04-10
bots: Delete useless warning

### LINKED ISSUE #1812: "Warn: no more slots for bot" is spammed all the time

DolceTriade:
It's spammy. There are two calls to G_BotAdd(): One in sg_admin.cpp and another when running the bot fill code. The sg_admin codepath already warns when a bot cannot be added. The bot fill code path shouldn't care if a bot can't be added.

Fixes #1812

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -281,7 +281,6 @@ bool G_BotAdd( const char *name, team_t team, int skill, const char *behavior, b
 
 	if ( clientNum < 0 )
 	{
-		Log::Warn( "no more slots for bot" );
 		return false;
 	}
 	gentity_t *bot = &g_entities[ clientNum ];


necessarily-equal:
We might prefer to use https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/180 as a solution. I'm ok either way but I think I would rather keep the warning

bmorel:
> Then in my experience you often get unbalanced teams. 

This has been fixed a while ago. Now, bot fill adds bots in each team in alternance.

I think it's a good idea to keep the warning to give  a hint to admins that the limit has been reached. What is annoying is that the warning is spammed every 2 seconds, though, instead of only when there is an actual change in number of players, which would make more sense (as the filler function is only useful to call in that condition)

DolceTriade:
Agreed. It's hard to determine max bots though at fill set time.

bmorel:
Not sure (from daemon):

```
/*
==================
SV_BotAllocateClient
==================
*/
int SV_BotAllocateClient()
{
  int i;
  for (i = std::max(1, sv_privateClients.Get()); i < sv_maxclients->integer; i++) {
    if (svs.clients[i].state == clientState_t::CS_FREE) {
      break;
    }   
  }

  if (i >= sv_maxclients->integer) {
    return -1; 
  }

  client_t* cl = svs.clients + i;
  cl->gentity = SV_GentityNum(i);
  cl->gentity->s.number = i;
  cl->state = clientState_t::CS_ACTIVE;
  cl->lastPacketTime = svs.time;
  cl->netchan.remoteAddress.type = netadrtype_t::NA_BOT;
  cl->rate = 16384;

  return i;
}
```

So basically `sv_maxclients->integer - sv_privateClients.Get()` might do it. Not sure if gamelogic can access that though.

PR #2620, opened 2023-04-11
RecastLayers: set RC_MAX_LAYERS and RC_MAX_NEIS as optional defines

illwieckz:
Custom values are set as CXXFLAGS to remain build system agnostic.


illwieckz:
The related recast patch has been merged upstream:

- https://github.com/recastnavigation/recastnavigation/pull/624

We may want to update Recast to benefit from this, and check that the Recast update still works with our code base.

PR #2621, opened 2023-04-11
translation: improve translatable strings

illwieckz:
- remove ending space or new line
- move some color codes out of the string
- remove endline from some log

Also:

- exclude maprotation command log from game pot

### DIFF
--- a/src/cgame/cg_tutorial.cpp
+++ b/src/cgame/cg_tutorial.cpp
@@ -111,7 +111,7 @@ static const char *CG_KeyNameForCommand( const char *command )
 			{
 				if ( !keyNames.empty() )
 				{
-					keyNames += _(" or ");
+					keyNames += va( " %s ", _("or") );
 				}
 				keyNames += CG_KeyDisplayName( key );
 			}
@@ -124,7 +124,7 @@ static const char *CG_KeyNameForCommand( const char *command )
 		}
 	}
 
-	return "(⚠ BUG)"; // shouldn't happen: if it does, BUG
+	return "(⚠  BUG)"; // shouldn't happen: if it does, BUG
 }
 
 #define MAX_TUTORIAL_TEXT 4096
@@ -169,17 +169,21 @@ static void CG_BuilderText( char *text, playerState_t *ps )
 	{
 		const char *item = _( BG_Buildable( buildable )->humanName );
 		Q_strcat( text, MAX_TUTORIAL_TEXT,
-		          va( _( "Press %s to place the %s\n"),
+		          va( _( "Press %s to place the %s"),
 		              CG_KeyNameForCommand( "+attack" ), item ) );
+		Q_strcat( text, MAX_TUTORIAL_TEXT, "\n" );
+
 		Q_strcat( text, MAX_TUTORIAL_TEXT,
-		          va( _( "Press %s to cancel placing the %s\n" ),
+		          va( _( "Press %s to cancel placing the %s" ),
 		              CG_KeyNameForCommand( "+attack2" ), item ) );
+		Q_strcat( text, MAX_TUTORIAL_TEXT, "\n" );
 	}
 	else
 	{
 		Q_strcat( text, MAX_TUTORIAL_TEXT,
-		          va( _( "Press %s to build a structure\n" ),
+		          va( _( "Press %s to build a structure" ),
 		              CG_KeyNameForCommand( "+attack" ) ) );
+		Q_strcat( text, MAX_TUTORIAL_TEXT, "\n" );
 	}
 
 	if ( ( es = CG_BuildableInRange( ps ) ) )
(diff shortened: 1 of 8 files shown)


src/cgame/cg_tutorial.cpp
```diff
@@ -441,55 +476,62 @@ static void CG_HumanText( char *text, playerState_t *ps )
...
+		          va( _( "Press %s to use the %s" ),
 		              CG_KeyNameForCommand( "weapnext" ),
 		              _( BG_Weapon( nextWeapon )->humanName ) ) );
+		Q_strcat( text, MAX_TUTORIAL_TEXT, "\n" );
 	}
 
 	if ( BG_InventoryContainsUpgrade( UP_JETPACK, ps->stats ) )
 	{
-		Q_strcat( text, MAX_TUTORIAL_TEXT, va( _( "Hold %s to use the %s\n" ),
+		Q_strcat( text, MAX_TUTORIAL_TEXT, va( _( "Hold %s to use the %s" ),
```
slipher, line 487: We would get better translation if strings like this said `Hold %s to use the jetpack` rather than `Hold %s to use the %s`.

illwieckz: Well, this hardcoding is only doable for two or three things, the jetpack, the medkit, and the armoury, probably no more. If we consider those names not common names but proper nouns (like if we name a branded product), I guess this would work in most languages. Languages that would be badly translated are the ones with declensions (like Latin) because those names would lack the proper declension (here the accusative one), but then only those 2 or 3 sentences among all the tutorial ones would be correct. I would take care of those 2 or 3 strings the day we want to make all strings from tutorial correct in such languages. We may even chose another solution, where declension languages would provide the accusative variation for the word and the function would pick the accusative variation if it exists. So to me it's too early to decide on that.

src/cgame/cg_tutorial.cpp
```diff
@@ -124,7 +124,7 @@ static const char *CG_KeyNameForCommand( const char *command )
 		}
 	}
 
-	return "(⚠ BUG)"; // shouldn't happen: if it does, BUG
+	return va( "(⚠  %s )", _("BUG") ); // shouldn't happen: if it does, BUG
```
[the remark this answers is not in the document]
illwieckz: Ok, I reverted that.

src/cgame/cg_tutorial.cpp
illwieckz: I finally did what's suggested (hardcoding those names), as there is already some sentences doing that, for example:

```sh
_( "Find an Armoury, Drill or Reactor for more ammo." ) );
_( "Find an Armoury for more ammo." ) );
```

But I would not oppose to doing the absolute opposite and modifying those other sentences instead, so they rely on entity names programmatically.

illwieckz:
> One thing it maybe time to decide if we want to change that, is that if we want to end those tutorial sentences with a dot.
> 
> In fact we already have two of them, when there are two sentences in the same line:
> 
> ```c++
> va( _( "Press %s to spit. Spits can slow humans and remove fire." ),
> ```
> 
> _Note:_ I just noticed we may even split those translatable sentences, but that changes nothing to the decision about dot-ended sentences.

I added ending dots to sentences, and split that one into two translatable strings.

necessarily-equal (approved):
Didn't test in game, but LGTM

slipher:
> I added ending dots to sentences, and split that one into two translatable strings.

Why split it? More finely chopped strings = less context, hence worse translations.

illwieckz:
> > I added ending dots to sentences, and split that one into two translatable strings.
> 
> Why split it? More finely chopped strings = less context, hence worse translations.

OK then, I liked the small sentences. Reverted.

illwieckz:
Question about “Spits can slow humans and remove fire.”, is “Spits can slow down humans and remove fire.” better or not?

illwieckz:
Once this is merged I generate `game.po` to make it discoverable by Weblate, and I submit Unvanquished to Weblate for evaluation for free open source plan.

slipher:
> Question about “`Spits can slow humans and remove fire.`”, is “`Spits can slow down humans and remove fire.`” better or not?

They're both acceptable; for tutorial text I'd keep the former since it should be concise.

`Spits` is wrong though; spit is an uncountable noun. Should be `Split can slow`...

illwieckz:
> `Spits` is wrong though; spit is an uncountable noun. Should be `Split can slow`...

Thanks, it's now fixed.

bmorel:
lgtm.

src/utils/generate_pot.sh
```diff
@@ -49,7 +49,10 @@ cd "${repo_dir}"
 	>> "${temp_pot_file}"
 
 find 'src/cgame' 'src/sgame' 'src/shared' \
-	-name '*.cpp' -a ! -name sg_admin.cpp -a ! -name sg_cmds.cpp \
+	-name '*.cpp' \
+	-a ! -name sg_admin.cpp \
+	-a ! -name sg_cmds.cpp \
+	-a ! -name sg_maprotation.cpp \
```
slipher, line 55: You have to add this to the xgettext command at the end

illwieckz: Good catch! Fixed.

Issue #2207, opened 2022-09-17
Commit to a single code style in the code base and mandate it with automated tool


DolceTriade:
We have a lot of code styles in our code base. Let's decide on a single style we. Subsequent commits will be required to be formatted using clang-format.

I propose this code style: https://gist.github.com/DolceTriade/94250674f24bd4465c6d00ff2e0c99aa, but that's just because that's the legacy q3 style. If people prefer other styles, then please provide an appropriate clang-format file. I don't especially care which style we use as long as we decide on one and we have appropriate tooling to mandate it.

We can use a file versioned in the repo to ignore the bulk formatting commits so we don't lose history: https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame

I see two approaches here:

 1. Do it gradually. Any code that is touched can be formatted with the new style. If it touches a function using a different style, the whole function should be reformatted (or the entire logical code block if it is a class or something...).
 2. Do a one time reformat of the code base.  This will need to be done with care as any outstanding PRs will conflict. Therefore we will need to have to merge all outstanding PRs and ensure all forks of the project are aware of our plans.

Let's decide on this and move forward. Having many different styles in the engine sucks and it sucks that we spend a shit load of time in PR reviews bikeshedding on code style issues. Doing this once and mandating this style with a CI check will basically eliminate all of these discussions and allow us to focus on the design and concepts of the code.

slipher:
https://wiki.unvanquished.net/wiki/Reformatting_the_source some notes on the subject

DolceTriade:
Yes, but the wiki is not a mandate. Hopefully we can come to a decision here and actually enforce it.

DolceTriade:
I propose we move forward with @slipher's suggestion.

let's make sure any large outstanding PRs are merged and then let's just do this.

I propose we use this style: https://gist.githubusercontent.com/DolceTriade/32b1c00498035f96afc848524f5ec459/raw/aea34e78ed9bd4599723fac6538014ea0f07e748/.clang-format

Open to other suggestions.

PR #2622, opened 2023-04-11
clang-formt: Add a .clang-format

DolceTriade:


slipher:
Wouldn't it make more sense to put this in Daemon so we can use the same formatting for both?

DolceTriade:
Put in daemon, then symlink to here or duplicate?

slipher:
Oh I was just thinking that Unvanquished can rely on the existence of Daemon but not vice-versa. But if there's some stuff where it automatically finds the config file in certain locations I don't know how that works

DolceTriade:
For example, with this file here, we can do stuff like `git clang-format HEAD~1` or you can do `git add file1 file2 && git clang-format` which will format only the diffs.

Viech:
How would this interfere with existing files, open branches/PRs, and `git blame`? I like autoformatting but I'm worried about history.

DolceTriade:
@Viech That is a bigger discussion (https://github.com/Unvanquished/Unvanquished/issues/2207)

This PR is just the current q3 style that we seem to have fallen back on over the years. It doesn't affect existing code. If you touch existing code, that code will be reformatted to match the existing style.

I believe we should solve the greater problem (by discussing it in the relevant issue)

TL;DR of the issue is:
 - reformat code using agreed upon style
 - add commits to .gitignorerevs or similar file to not show up in git blame to preserve git blame/history
 - ???
 - Profit

For existing PRs, reformatting the PRs (git clang-format origin/master) should help avoid conflicts.

.clang-format
```diff
@@ -0,0 +1,80 @@
...
+  SplitEmptyNamespace: false
+BreakBeforeBraces: Custom
+ColumnLimit: 100
+PointerAlignment: Right
+SpacesInParentheses: true
+SpacesInSquareBrackets: true
+TabWidth: 4
+UseTab: AlignWithSpaces
+CommentPragmas:  '^ keep:'
+IncludeCategories:
```
slipher, line 33: It might be better to avoid include sorting for now; I bet this causes a lot of compile errors because we have non-self-contained headers.

slipher (commented):
Can we disable formatting of comments? It's not very beneficial and seems to be overly aggressive with often incorrect reflows.

One difference I noticed with our current style: it wants to have 2 spaces before a `//` comment on the same line as code but we usually use 1.

.clang-format
```diff
@@ -0,0 +1,80 @@
...
+  - Regex:           '^<.*'
+    Priority:        2
+    SortPriority:    0
+    CaseSensitive:   false
+  - Regex:           '.*'
+    Priority:        3
+    SortPriority:    0
+    CaseSensitive:   false
+StatementMacros:
+  - GETTER
```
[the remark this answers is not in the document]
DolceTriade: It's for the Lua code

.clang-format
```diff
@@ -0,0 +1,80 @@
...
+  SplitEmptyFunction: false
+  SplitEmptyRecord: false
+  SplitEmptyNamespace: false
+BreakBeforeBraces: Custom
+ColumnLimit: 100
+PointerAlignment: Right
+SpacesInParentheses: true
+SpacesInSquareBrackets: true
+TabWidth: 4
+UseTab: AlignWithSpaces
```
illwieckz, line 31: > `UseTab: AlignWithSpaces`

Does that mean every tab will be replaced with spaces?

.clang-format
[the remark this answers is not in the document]
slipher: Maybe you could just put a (technically unnecessary) semicolon after these in your code.

.clang-format
DolceTriade: It means it will use tabs to indent, but then will use spaces to align outside these tabs. This will only look correct if your tab width is 4.

This could be our opportunity to settle tabs vs spaces.

Personally, I used to be all tabs, but now I'm down with spaces.

illwieckz: Well, “tab for indentation, space for ASCII art” is what I expect. The information the tab provides to the user is so precious to destroy it with spaces, so I would have been against a “replace tabs with space” rule. There can't be any any spacing issue to fear if tabs are only used for indentation and spaces only used for ASCII art. So if that's what this option does, that's good.

bmorel:
> Wouldn't it make more sense to put this in Daemon so we can use the same formatting for both?

No, because daemon and unvanquished in practice use a different coding style.

DolceTriade:
Feel free to add comments if there are aspects of the code style you want to change.

illwieckz:
To sum up the purpose of this change, here are some selected quotes from IRC:

```
<illwieckz> Ishq, is this clang format only used for committed changes?
<Ishq> illwieckz: yes
<Ishq> only affects new code
```

This restyle stuff is currently only for properly styling new code.

```
<perturbed> it's the old quake-derived style that Ishq is proposing
```

This restyle stuff should not changes things a lot, just avoid things to differ from the usual existing style.

```
<freem> changing a mostly used codestyle only brings problems, that's my opinion
<Ishq> reformatting the codebase has implications for branches and forks
<Ishq> so we shouldn't do that lightly
<illwieckz> yes, strong restyles are fork makers
<Ishq> enforcing this style on newly committed code seems a much lower barrier of entry
<perturbed> yes, in the wiki page i wrote about this i said we should have a branch rewriting script
<perturbed> if we go for reformatting
```

There is a kind of consensus we should not do a full restyle without minding it deeply, so that's not the purpose of that PR.

Viech:
This PR should probably add to README.md instructions to use this: the command to run or how to setup a pre-commit hook, etc.

PR #2625, opened 2023-04-12
Undoing `unvanquished_src.dpkdir` submodule

illwieckz:
This removes the `pkg/unvanquished_src.dpkdir` and reimports the files to be stored directly in this repository as a `pkg/unvanquished_src.dpkdir` subfolder.

All history since the split was replayed. Only one file had to be fixed after the rebase because of a rebase mistake.

The `pkg/unvanquished_src.dpldir` folder is now 100% equals to the submodule.

The PR only affects files in `pkg/unvanquished_src.dpldir`, so other files are also 100% equal with the ones from `Unvanquished` folder from before the reimport.


DolceTriade:
How will this affect things like urcheon and creating assets?

illwieckz:
> How will this affect things like urcheon and creating assets?

For now one would have to run Urcheon two times, one `UnvanquishedAssets/src/*.dpkdir` and one `Unvanquished/pkg/unvanquished_src.dpkdir`.

Also Urcheon already has options to output the produced dpk to a single build folder even if the source dpkdirs are in a different directory.

That extra work that only happens as release time is acceptable compared to the extra work done on every PR that requires to change stuff on both code and data.

illwieckz:
Before pulling the `master` branch you would have to make sure there is no uncommitted change in `pkg/unvanquished_src.dpkdir` and do:

```sh
git submodule deinit pkg/unvanquished_src.dpkdir
```

Then you can pull:

```sh
git pull
```

PR #2626, opened 2023-04-12
gameplay: Allow lgun to zoom

DolceTriade:
Tested on sweet's server for a while.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/configs/weapon/lgun.attr.cfg
+++ b/pkg/unvanquished_src.dpkdir/configs/weapon/lgun.attr.cfg
@@ -17,6 +17,8 @@ reloadTime 2000
 
 isPurchasable
 isLongRanged
+canZoom
+zoomFov 40
 
 // Attributes
 damage 9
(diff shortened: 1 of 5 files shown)


sweet235:
It works fine indeed. But I noticed an aesthetic problem on my server. Lasgun and mdriver use different fields of view, but both show 8x as zoom magnitude. Compare:

a
b

Also, I just noticed that the lasgun model is visible on my screenshot, but the mdriver model is not. I have no idea why.

sweet235:
Being a lazy programmer, I thought about simply removing the 8x. But on the other hand, I do not want to do change the rather nice artwork out of laziness.

DolceTriade:
Technically yes, but I was also rather lazy :smile: 

I wouldn't consider this a blocker I hope...

Viech:
Would maybe be good to duplicate the shader and refer to distinct shaders in the code? Then, one can later add a new shader/materials for this without touching the code again. Not a blocker; just a possible improvment.

sweet235:
> I wouldn't consider this a blocker I hope...

I do not consider this to be blocking. I'm just reporting what multiple users told me when I introduced this on my server. Have to admit: I did not notice this myself.

sweet235:
lgtm

bmorel (approved):

PR #2629, opened 2023-04-12
change datagrid background colour

### LINKED ISSUE #1989: server list background uses #333333 which is used by unv's color codes
#333333 is the color code "Dark Grey", `^0`. I think it should be replaced with something not in [this](https://wiki.unvanquished.net/wiki/Colours) list (except for the free-style colors, obviously).

bmorel:
fix #1989 server list background uses #333333 which is used by unv's color codes

The colour is probably a bit too bright, but this would solve the aforementioned issue.


DolceTriade:
screenie?

DolceTriade (approved):

PR #2630, opened 2023-04-12
scoreboard shows number of players per team

### LINKED ISSUE #1816: add number of player in a team in scoreboard
Currently, scoreboard only shows 3 areas: alien players, human players, spectators. I think it would be nice if, in front of each team name, there was the number of players, and, why not, the number of bots.

bmorel:
fix #1816

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/ingame_scoreboard.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/ingame_scoreboard.rml
@@ -157,6 +157,7 @@
 	<half>
 		<teamname><translate>Aliens</translate></teamname>
 		<teambox class="alien">
+			<playerscount team="aliens"/>
 			<datagrid source="playerList.aliens" id="aliens">
 				<col fields="num" formatter="GearOrReady"></col>
 				<col fields="num" formatter="PlayerName">Name</col>
@@ -170,6 +171,7 @@
 	<half>
 		<teamname><translate>Humans</translate></teamname>
 		<teambox class="human">
+			<playerscount team="humans"/>
 			<datagrid source="playerList.humans" id="humans">
 				<col fields="num" formatter="GearOrReady"></col>
 				<col fields="num" formatter="PlayerName">Name</col>
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -2442,6 +2442,35 @@ private:
 	int spawns_;
 };
 
+class PlayersCountElement : public TextHudElement
+{
+public:
+	PlayersCountElement( const Rml::String& tag ) :
+			TextHudElement( tag, ELEMENT_GAME )
+			{}
+
+	void DoOnUpdate() override
+	{
+		//there is probably a better way to do that, maybe reusing value
+		//cached in some struct?
+		std::string teamname = GetAttribute< Rml::String >( "team", "" );
+
+		team_t team = BG_PlayableTeamFromString( teamname.c_str() );
+		size_t count = 0;
+
+		for ( int i = 0; i < cg.numScores; ++i )
+		{
+			score_t const& score = cg.scores[ i ];
+
+			if ( score.team == team )
+			{
+				++count;
+			}
+		}
+
+		SetText( va( "Player count: %zu", count ) );
+	}
+};
 
 static void CG_Rocket_DrawPlayerHealth()
 {
@@ -3715,4 +3744,5 @@ void CG_Rocket_RegisterElements()
 	RegisterElement<TranslateElement>( "translate" );
 	RegisterElement<SpawnQueueElement>( "spawnPos" );
 	RegisterElement<NumSpawnsElement>( "numSpawns" );
+	RegisterElement<PlayersCountElement>( "playerscount" );
 }


bmorel:
Help wanted, because the player counts could really need some margin, but I could not get this done. Feel free to update this branch if you get fixes for that.

slipher:
Why do we need that? You can just see how long the list is.

illwieckz:
If we want this to be displayed, I would display this at the bottom on the list, especially since it would be more useful when the lists are longer than what the scoreboard shows.

illwieckz:
Or better, next to team names in parenthesis.

Aliens (1)|Humans (1)
--|--
Dave|Roger

bmorel:
> Why do we need that? You can just see how long the list is.

Yes and no.
When there are enough players or bots, the list is full and have unusable scrollbars. This is also not very convenient. Well, it is "enough" to see if 2 teams have same number of players, since you could only check that the last lines match.

> Why do we need that? You can just see how long the list is.

I agree it would be better (and likely also easier, even!) but I was not sure.

pkg/unvanquished_src.dpkdir/ui/ingame_scoreboard.rml
```diff
@@ -170,6 +171,7 @@
 	<half>
 		<teamname><translate>Humans</translate></teamname>
 		<teambox class="human">
+			<playerscount team="humans"/>
```
DolceTriade, line 174: nit: playercount here and in the element.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2442,6 +2442,35 @@ class NumSpawnsElement : public TextHudElement
...
+		{
+			score_t const& score = cg.scores[ i ];
+
+			if ( score.team == team )
+			{
+				++count;
+			}
+		}
+
+		SetText( va( "Player count: %zu", count ) );
```
DolceTriade, line 2471: Just set the count, not the other text, let the UI decide whether it wants to the text there or not. (It could want to style it differently or use an image for example).

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2442,6 +2442,35 @@ class NumSpawnsElement : public TextHudElement
...
+	void DoOnUpdate() override
+	{
+		//there is probably a better way to do that, maybe reusing value
+		//cached in some struct?
+		std::string teamname = GetAttribute< Rml::String >( "team", "" );
+
+		team_t team = BG_PlayableTeamFromString( teamname.c_str() );
+		size_t count = 0;
+
+		for ( int i = 0; i < cg.numScores; ++i )
```
DolceTriade, line 2461: For the cgame, probably not...

The only option might be `rocketInfo.data.playerCount[ team ]`

But this isn't guaranteed to be updated.

Also, this will be run for every element, which is bad. We should run this once, and use the value for all elements.

DolceTriade (commented):
You could add another subtable to the playerList datasource with the counts. Something like 
`Rocket_DSAddRow( "playerList", "alienCount", buf );`

And then use <datasource_single> to get player count.

slipher:
> Or better, next to team names in parenthesis.

+1

bmorel:
> You could add another subtable to the playerList datasource with the counts. Something like
Rocket_DSAddRow( "playerList", "alienCount", buf );
>
And then use <datasource_single> to get player count.

I don't know how to do that. rmlUI's documentation is as good as unvanquished's: not at all.

src/cgame/cg_rocket_draw.cpp
bmorel: wdym "for every element"? This code is not run only for the element I wrote a class for?

bmorel:
About the parens thing:

```
diff --git a/pkg/unvanquished_src.dpkdir/ui/ingame_scoreboard.rml b/pkg/unvanquished_src.dpkdir/ui/ingame_scoreboard.rml
index 5a7886ec5..4c8bde163 100644
--- a/pkg/unvanquished_src.dpkdir/ui/ingame_scoreboard.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/ingame_scoreboard.rml
@@ -155,9 +155,8 @@
        <inlinecvar cvar="ui_winner" class="winner" />
 
        <half>
-               <teamname><translate>Aliens</translate></teamname>
+               <teamname><translate>Aliens (<playerscount team="aliens"/> players)</translate></teamname>
                <teambox class="alien">
-                       <playerscount team="aliens"/>
                        <datagrid source="playerList.aliens" id="aliens">
                                <col fields="num" formatter="GearOrReady"></col>
                                <col fields="num" formatter="PlayerName">Name</col>
@@ -169,9 +168,8 @@
        </half>
 
        <half>
-               <teamname><translate>Humans</translate></teamname>
+               <teamname><translate>Humans (<playerscount team="humans"/> players)</translate></teamname>
                <teambox class="human">
-                       <playerscount team="humans"/>
                        <datagrid source="playerList.humans" id="humans">
                                <col fields="num" formatter="GearOrReady"></col>
                                <col fields="num" formatter="PlayerName">Name</col>
diff --git a/src/cgame/cg_rocket_draw.cpp b/src/cgame/cg_rocket_draw.cpp
index 9536be7a4..9dd934fe6 100644
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -2468,7 +2468,7 @@ public:
                        }
                }
 
-               SetText( va( "Player count: %zu", count ) );
+               SetText( va( "%zu", count ) );
        }
 };
```

leads to that kind of results:





Just in case, the "help wanted" tag is because I don't understand how html/css and rmlui works. I think the theoretical performance problem this brings is acceptable. There are not that many max players anyway, it's maximum 64.

Viech:
Maybe the `z` in `%zu` is causing this? I normally just use `%d` for integers, even if they are non-negative in practice.

Viech:
Makes sense; so maybe then somehow the numbers of the alien and human team are combined into one string?

bmorel:
If this feature is not considered useful, let's just close this PR. I dislike working with rmlui *a lot* after all. Will close in few days.

slipher:
I like it as long as it's done the `Aliens (5 players)` way so that it doesn't take away any screen real estate.

bmorel:
but apparently simply moving the markup generates weird bugs, and seriously rmlui is a calamity. I'm tired of fighting with it to do the most basic features

bmorel:
will abandon this in 2 weeks as I don't have the strenght to fight against a shitty lib that does not provides the basic features, is slow as fuck and buggy as hell. Or at least it's my feeling.

bmorel:
Actually, I'll abandon this right now, as I no longer want to waste time with CSS black magic.

src/cgame/cg_rocket_draw.cpp
illwieckz: I guess he means for every `TextHudElement`, and this is just a call to move the for loop outside of that RML c++ code and just use there a variable that is computed elsewhere and once for all per frame.

PR #2633, opened 2023-04-13
use menu background when chatting

### LINKED ISSUE #2578: give a visual clue that chat is open
I "often" fat finger chat, which leads to completely blocked input. This is bad, and my fault, but it would be nice if game would give an obvious visual clue that chat is open. Such as darkening/blurring the screen.

Note that I'm probably not the only person affected since random strings of wasd are not rarely seen.

This is "moved" from https://github.com/DaemonEngine/Daemon/issues/816 (wrong repo, my bad)

bmorel:
fix #2578 

Screen could probably be darkened too, but that's better than current situation I guess?

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/ingame_chat.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/ingame_chat.rml
@@ -2,7 +2,7 @@
 	<head>
 	<link type="text/rcss" href="/ui/shared/basics.rcss" />
 	<style>
-		body {
+		.chat {
 			position: absolute;
 			top: 50%;
 			width: 100%;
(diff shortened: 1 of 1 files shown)


bmorel:
Should probably darken background a bit too tbh

DolceTriade:
I think if your goal is to make it more obvious that the chat menu is open, I recommend a darker background on the chat box.

bmorel:
Depends what's the size of the chatbox though. I'll try to darken the whole screen a bit like menu does later, the circles stuff (taken from menu) is hardly enough I think.

bmorel:
Added the missing dark gradient. I think this is now coherent with the stuff where one can actually use the mouse, *and* very obvious that you're chatting.

DolceTriade:
lgtm

PR #2635, opened 2023-04-13
Add debug commands to make skilltree maintenance easier (6/?)

necessarily-equal:
I've split the first part of this PR into a separate one, https://github.com/Unvanquished/Unvanquished/pull/2722. This PR also depends on #2697. Only the last three commits of this PR are from this PR.

This introduces two new commands, `/devbotcountskillpoints` and `/devbotgraphskilltree`. The first just prints a message with the total cost of (reachable) skills in the tree for each team, and the second saves a `.dot` graph file that can be converted to an image. For example one can do: `dot -Tpng ~/.local/share/unvanquished/game/bot-skilltree-human.dot > result.png && eog result.png`.

This also refactors the skilltree implementation, making the above possible even though it makes it a bit less flexible (extra complexities like {having two conflicting skills that can't be selected at the same time} could be probably be brought back if desired, but for now we don't use those). This should bring no balancing change.

Previous PR in this series: https://github.com/Unvanquished/Unvanquished/pull/2697

### DIFF
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -470,6 +470,8 @@ static const struct cg_cmd_t
 	{ "destroy",          0,                       0                },
 	{ "destroyTestPS",    CG_DestroyTestPS_f,      0                },
 	{ "destroyTestTS",    CG_DestroyTestTS_f,      0                },
+	{ "devbotcountskillpoints", 0,                 0                },
+	{ "devbotgraphskilltree", 0,                   0                },
 	{ "devteam",          0,                       0                },
 	{ "follow",           0,                       CG_CompleteName  },
 	{ "follownext",       0,                       0                },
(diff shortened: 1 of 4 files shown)


necessarily-equal:
Note to self: this seems to be working fine, but I'd like to do a bit more checks before merging.

It is ready for review IMO

necessarily-equal:
I've also added colors, skill percentages and a legend to the graph if it is printed for a specific skill level.

necessarily-equal:
@slipher said in https://github.com/Unvanquished/Unvanquished/pull/2697#issuecomment-1565119033:
> The output [of the listing commands] is implemented wrong on those btw, they are registered as networked commands but only output to the server log/filesystem

I think it's ok because those are developer commands, protected by `CVAR_CHEAT` and those are not expected to be used in a networked setting. Other commands such as `/layoutsave` have a similar problem where they save the file on the server.

If you want me to fix that, how should I proceed ? I put those as sgame commands in sg_cmds.cpp because I don't think admin commands are appropriate because those are not meant for managing a server (even though it would fix the "only output to the server log" problem).

slipher:
I think admin commands make sense but if you don't want to deal with that, there's a third command list in sg_svcmds.cpp that can only be used by the server console.

Issue #1538, opened 2021-10-19
Crouching while in the air reduces weapon recoil.


necessarily-equal:
This can be exploited (for example) by players with a gattling gun to have less recoil and thus more precise fire, while the precision should definitely not be better while jumping than when standing.

Viech:
I will fix the title of this issue: Reducing recoil while in the air through the crouch key is certainly not intended (and this is the only thing the issue body talks about), but being able to reach higher places by using a combiantion of jump and crouch is an integral part of the movement and not an issue at all in my eyes.

bmorel:
I still question the assertion that it's possible to crouch while mid-air. None of my tests shows that this is possible at all.

slipher:
@bmorel In this screenshot from Obstacle Course Experimental Server, the barricade directly in front is an example of a place you can only jump onto if you crouch in midair:
![unvanquished_2023-05-02_230951_000](https://user-images.githubusercontent.com/7809431/235831474-46488955-0a78-4a67-b468-b16c08967cf1.jpg)

Another demonstration (which also reveals #1538 itself) is to set `/timescale 0.3` and then try jumping + firing chaingun vs. jumping + crouching + firing chaingun. At this timescale the difference is easy to see, unlike with timescale 1.

bmorel:
No, it does not. The easy test is this:

* devmap parpax
* god
* cg_thirdperson 1
* cg_drawbbox 1

go on huge elevator, jump in void and crouch: neither the animation nor the bbox shows crouching.

My bet is that with the situation you're talking to, the character crouches because it's feet touches something. I have yet to test the recoil problem: it's very possible that PMF_CROUCH is set but not applied correctly.
It's also well known that barricades can be climbed by humans, and sure, you can crouch on them.

PR #2636, opened 2023-04-13
PM_CheckJump cleanup

### LINKED ISSUE #312: Using the jetpack in crawl tunnels makes you use a 'standing' posture.  This posture clips through the ceiling.
jetpack_tunnel

Only possible in spaces where you have to crawl to fit, as pressing crouch while near normal ceilings just turns your jetpack off.

### LINKED ISSUE #1322: Touching a ladder with jetpack activated keeps it activated
Touching a ladder with jetpack activated keep it activated, if you release the jump key you'll be kept attached to the ladder, the jetpack fuel will get stuck to the amount it had when you touched the ladder (no refuel), and the jetpack smoke will continues to be emitted.

### LINKED ISSUE #1539: Recoil should be higher while jumping and/or precision should be lower
We have something in place for lower recoil: crouching does that. It should be easy enough to do the reverse while jumping.

I can help someone wanting to resolve this. Just hop in in the [chat](https://unvanquished.net/chat/) and ask.

bmorel:
fix #312

### DIFF
--- a/src/shared/bg_pmove.cpp
+++ b/src/shared/bg_pmove.cpp
@@ -1517,6 +1517,13 @@ static void PM_LandJetpack( bool force )
 
 static bool PM_CheckJump()
 {
+	// can't jump while crouched, prevents cheating in vents and
+	// displaying "jumping" model animation which goes outside of
+	// vents. See #312, #1538
+	if ( pm->ps->pm_flags & PMF_DUCKED )
+	{
+		return false;
+	}
 	// can't jump while in air
 	if ( pm->ps->groundEntityNum == ENTITYNUM_NONE )
 	{


bmorel:
Can't check this works as intended yet, will test ASAP

bmorel:
After checking, all but the last commit seems fine, #1322 is indeed fixed, other stuff are not (and situation is worst, I'm working on it right now)

bmorel:
The cause of the bug introduced was using `|` instead of `&`. Obviously, inverted conditions won't work.

This is now fixed, and all bugs mentioned are fixed by this PR.

slipher:
Aww but using jetpack in vents is fun :sob:

bmorel:
I honestly hesitate to fix this one :)

slipher:
Some commentaries

```
<freem> and yeah, fixing the vents bugs is something I hesitated to do, because otherwise vents are quite boring, and aliens have enough of the speed advantages already
<freem> yet, they are bugs, so...
<perturbed`> a bug is in the eye of the beholder
<freem> well, that was reported a long while ago and nobody seemed to disagree
<freem> I could split that PR to have only the ladder+jetpack fix, and we could promote the jump/fly in vents to features
<perturbed`> well the #312 op just says it looks bad, not that it shouldn't be allowed
<freem> true
<perturbed`> i have to admit that it doesn't make much sense with the unv jetpack mechanics
<perturbed`> whereas the magical perfectly stable tremulous jetpack would totally work fine in vents :)
<freem> could probably try to get the jump/fly give a speed gain without the ugly animation though
<freem> I'm starting to be a bit familiar with that bg_pmove.cpp file
<freem> with all the refactorings I made there...
<freem> it should be just a matter of sending the correct event to clients if the player is crouched, instead of preventing the action entirely
<perturbed`> anyway even if we disallow jetpack in events, no jump while crouching is too big of a hammer
<perturbed`> no jetpack while ducked would be better
<freem> pretty sure I can handle those separately
<freem> well, this is exactly what this code does
<freem> if ( pm->ps->pm_flags & PMF_DUCKED )
<freem> ￼ {
<freem> ￼  return false;
<freem> ah
<freem> no actually
<freem> it's jump
<freem> but one needs to jump to trigger jetpack
<freem> https://github.com/Unvanquished/Unvanquished/pull/2636/commits/8ca0928cbf593e3e2227e5967fa51a1c112331d8#diff-17afe9a22f5c9a9b6eb051857c46cf2afe58bdc7b1caab53145cc396b6eee77eR1622
<freem> I'll split the PR to at least get the no jetpack on ladder stuff mergeable without the vent stuff
<freem> since that one brings no interesting gameplay effects
```

Viech:
I think disabling jumps while crouched (but not the other way around!¹) is a sensible thing to do. Not being able to use the jepack while on a ladder also sounds fine. A different solution would be to jump away from the ladder when the jump key is pressed.

---
¹Issue #1538 should not be taken by its old title: Pulling your legs up while in the air has been an integral and perfectly fine aspect of movement. You would annoy people very hard for no good reason if this was changed. I'm saying this because your PR claims it fixes #1538: I do see that it actually doesn't remove the ability to crouch while in the air, and that's a good thing.

bmorel:
> Not being able to use the jepack while on a ladder also sounds fine.

The problem here is that on a ladder, you already do not use the jetpack: the fuel is simply not updated (neither increased nor decreased), and the flying animation is played.
This PR fixes that weird behavior, itself described in the issue.

> I think disabling jumps while crouched (but not the other way around!¹) is a sensible thing to do (although it does remove an arguably fun thing from the game in vent jumping/jetpacking).

In few words, I think the current cheaty situation is sensible for a fast paced game, but the fix would be "more realistic" (not really a good reason to change gameplay) and would fit slower paced games, which would arguably be more adapted to a game with RTS elements.

> ¹Issue https://github.com/Unvanquished/Unvanquished/issues/1538 should not be taken by its old title:

Indeed, I only looked at the issue's original title "One can crouch while jumping", and this PR does not change anything about the accuracy in various situations.

I also would like to tackle old bugs as much as possible, and consider that bugs which are several years old and not closed or without much discussions are indeed considered behaviors to fix. Even if fixing bugs is not the most fun thing in the world.

Viech:
Let me again stress that I'm OK with the changes so far, if you want them.

But,

> I also would like to tackle old bugs as much as possible, and consider that bugs which are several years old and not closed or without much discussions are indeed considered behaviors to fix.

here I disagree. The old bug #312 (which originally concerns only a bad visualization of the movement) may just be open for so long because

1. it is not wrong, so no one could close it,
2. it is of minor importance, so no one did fix it.

Me commenting in 2014 that we maybe do not need to have the ability to jetpack through tunnels at all, and no one countering this, does not make that an agreed-upon notion. The issue itself existing for long also does not mean it has to be fixed at a price/with a downside. Maybe it is sufficient by fixing it "directly" by avoiding the standing posture in this particular scenario.

bmorel:
well, in any case I'm ready to go in either way. I'd even tend to be in favor of just fixing the clipping problem, and this will allow us to get rid of the issue in both cases.
So in both cases, that will be one less long standing open issue, which is something I consider important.

cu-kai:
I would prefer it if the gameplay was not altered. This will make humans nearly useless in vents vs maras and extremely vulnerable to mantis/dretches.

Crawling through vents is tiresome, especially at endgame if you are chasing a lone dretch around the map or going on an egghunt.

VReaperV:
Agreed with @cu-kai there, vents without jetpack are basically a death sentence, especially with all-knowing bots swarming the vent if they can get to it. Traversing them without jetpack is too slow, which would make them diasadvantageous pretty much all of the time compared to any other route.

bmorel:
So, little update here since I merged the consensual commits.

It is impossible to crouch while mid-air, according to animations seen with `cg_thirdperson 1`, so I believe #1538 is not a bug, as it can't be done at all.

It's also impossible to jump while crouched: pressing both keys will just make the character stand up.

All this means, it's not possible to crouch while air to help climbing on a crate, which is, imo, problematic and counter-intuitive. This: `<perturbed`> anyway even if we disallow jetpack in events, no jump while crouching is too big of a hammer` is indeed a bug in regard of current situation as it's only possible to crouch+jump in vents. I would propose to just allow crouching all the time, including while mid-air.
This proposal would make #1538 true, but then it should be easy to fix: just check that the character is mid-air before applying the recoil reduction.

So, my proposal would be:

* do not play the jumping animation when in vents. Not sure how this would affect jetpacking yet.
* allow players to crouch mid-air, and to initiate a jump while crouched. This will make climbing over crates much easier and... well, more realistic, too.
* double check that chaingun recoil while mid-air is not reduced by crouching. Probably should still be reduced if wearing battlesuit, though.

Viech:
> So, my proposal would be:
> 
>     * do not play the jumping animation when in vents. Not sure how this would affect jetpacking yet.

In other words: don't stand the player up if there's no space for it? Sounds sensible. Jetpack anims (smoke and wings) *should* be separate from the standing posture, but not sure.

>     * allow players to crouch mid-air, and to initiate a jump while crouched. This will make climbing over crates much easier and... well, more realistic, too.

Crouch mid-air sounds good, I thought this was already possible. Not clear if jumping while crouched has uses but I see no reason against it. So far holding both keys just cancels them so that you end up standing. Both of these changes would need a bit of testing to see if movement feels normal.

>     * double check that chaingun recoil while mid-air is not reduced by crouching. Probably should still be reduced if wearing battlesuit, though.

Makes sense.

bmorel:
> Both of these changes would need a bit of testing to see if movement feels normal.

I totally agree on that. I think the movement mechanics could use some rethinking overall tbh.

bmorel:
It would be nice to actually engage a conversation on what the movement mechanics should be in vanilla. Instead of wasting people's time into endless github talking and pseudo-diplomacy.

slipher:
The majority view seems to be:
- Jetpack in vent should be kept
-  Crouching in midair should be kept
- #1538 should be fixed
- #312 should be fixed (in the narrow terms of fixing the animation)

bmorel:
#312 is not just about jetpack, it applies to jumping as well.

bmorel:
thinking about it, I don't think I want to spend more time on that. bg_pmove.cpp is a huge mess, and it takes way too much energy for very little results. I still think a discussion on what the moves in unvanquished should be would be beneficial, but the remaining commit in this PR is definitely something nobody wants.

PR #2638, opened 2023-04-13
spawn suicide when blocked by geometry

### LINKED ISSUE #2515: geometry do not kill spawns when they birth
To reproduce, load map `tranc3`: the game will flood `Warn: Spawn blocking entity has oldEnt == nullptr`

bmorel:
This can happen when loading layouts or maps which were built for a version where the spawner was smaller.

fix #2515

### DIFF
--- a/src/sgame/components/SpawnerComponent.cpp
+++ b/src/sgame/components/SpawnerComponent.cpp
@@ -40,6 +40,7 @@ void SpawnerComponent::Think(int timeDelta) {
 	if (blocker) {
 		if (!blocker->oldEnt) {
 			logger.Warn("Spawn blocking entity has oldEnt == nullptr");
+			Entities::Kill(entity, nullptr, MOD_SUICIDE);
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


bmorel:
I just verified that this works (could not sooner, no access to proper computer)

src/sgame/components/SpawnerComponent.cpp
```diff
@@ -40,6 +40,7 @@ void SpawnerComponent::Think(int timeDelta) {
 	if (blocker) {
 		if (!blocker->oldEnt) {
 			logger.Warn("Spawn blocking entity has oldEnt == nullptr");
```
slipher, line 42: When does this happen?

bmorel: I don't think so. The CBSE pass loose that information, and it might be many things, depending on what the code becomes.

slipher:
Alternative fix: #2639

bmorel:
superseeded by the alternative.

PR #2639, opened 2023-04-13
Create a distinct EmptyEntity for every non-CBSE entity

### LINKED ISSUE #2515: geometry do not kill spawns when they birth

slipher:
This will fix #2515 by making the code that checks if the blocker is ENTITYNUM_WORLD work correctly.


DolceTriade (approved):

bmorel (approved):
tested, it works.

PR #2640, opened 2023-04-13
some string rework for translation

illwieckz:
_Question:_ can we assume any “`(%s)`” will never have any translation variant? I know for example that English “`%s:`” is actually translated to “`%s :`” in French (French typography requires a non-breakable thin space before double-punctuation like “`;`” and “`:`”). If we assume “`(%s)`” will never have any translation variant we can exclude more strings or parts of strings to be translated.

I actually split things like “`Vote failed (reason1)`” and “`Vote failed (reason2)`” into “`Vote failed`”, “`(reason1)`”, “`(reason2)`”. I see no reason to translate the same string many times. And if we can exclude the parenthesis, we would also be able to simplify some “`reason`” string or even remove some from translatable strings.

_Note:_ I now consider that we can split such strings because Weblate displays by default strings that are in the same context, so there is no real context issue to fear about.

### DIFF
--- a/src/sgame/sg_client.cpp
+++ b/src/sgame/sg_client.cpp
@@ -1151,7 +1151,8 @@ const char *ClientConnect( int clientNum, bool firstTime )
 			if ( !G_ClientIsLagging( level.clients + i ) )
 			{
 				G_LogPrintf( "Duplicate GUID: %i", i );
-				trap_SendServerCommand( i, "cp \"Your GUID is not secure\"" );
+				trap_SendServerCommand( i, "cp_tr "
+					QQ( N_("Your GUID is not secure") ) );
 				G_LogPrintf( "FailedConnecting: %i [%s] (%s) \"%s^*\" \"%s^*\"",
 				             clientNum, client->pers.ip.str[0] ? client->pers.ip.str : "127.0.0.1", client->pers.guid,
 				             client->pers.netname,
--- a/src/sgame/sg_client.cpp
+++ b/src/sgame/sg_client.cpp
@@ -841,17 +841,19 @@ const char *ClientUserinfoChanged( int clientNum, bool forceName )
 		     level.time - client->pers.namelog->nameChangeTime <=
 		     g_minNameChangePeriod.Get() * 1000 )
 		{
-			trap_SendServerCommand( ent->num(), va(
-			                          "print_tr %s %g", QQ( N_("Name change spam protection (g_minNameChangePeriod = $1$)") ),
-			                          g_minNameChangePeriod.Get() ) );
+			trap_SendServerCommand( ent->num(), va("print_tr %s %s %g",
+				QQ( N_("Name change spam protection ($1$ = $2$)") ),
+				"g_minNameChangePeriod",
+				g_minNameChangePeriod.Get() ) );
 			revertName = true;
 		}
 		else if ( !forceName && g_maxNameChanges.Get() > 0 &&
 		          client->pers.namelog->nameChanges >= g_maxNameChanges.Get() )
 		{
-			trap_SendServerCommand( ent->num(), va(
-			                          "print_tr %s %d", QQ( N_("Maximum name changes reached (g_maxNameChanges = $1$)") ),
-			                          g_maxNameChanges.Get() ) );
+			trap_SendServerCommand( ent->num(), va("print_tr %s %s %d",
+				QQ( N_("Maximum name changes reached ($1$ = $2$)") ),
+				"g_maxNameChanges",
+				g_maxNameChanges.Get() ) );
 			revertName = true;
 		}
 		else if ( !forceName && client->pers.namelog->muted )
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -2356,9 +2356,13 @@ void G_RunFrame( int levelTime )
 			}
 		}
 
+		// We may de-hardcode this value.
 		if ( level.pausedTime > 120000 )
 		{
-			trap_SendServerCommand( -1, "print_tr \"" N_("Server: The game has been unpaused automatically (2 minute max)") "\"" );
+			// FIXME: This assumes plural.
+			trap_SendServerCommand( -1, va( "print_tr_p %s %d",
+				QQ( N_("Server: The game has been unpaused automatically ($1$ minutes max)") ),
+				2 ) );
 			trap_SendServerCommand( -1, "cp \"The game has been unpaused!\"" );
 			level.pausedTime = 0;
 		}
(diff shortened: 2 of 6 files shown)


src/sgame/sg_main.cpp
```diff
@@ -1974,7 +1974,7 @@ void CheckExitRules()
 	{
 		//humans win
 		level.lastWin = TEAM_HUMANS;
-		trap_SendServerCommand( -1, "print_tr \"" N_("Humans win") "\"" );
+		trap_SendServerCommand( -1, "print_tr " QQ( N_("Humans win") ) );
```
bmorel, line 1977: Just pointing at this example, but there's a bit more of that:

This could probably use `const char *BG_TeamName( int team );` or `const char *BG_TeamNamePlural( int team );` that is:

```
trap_SendServerCommand( -1, "print_tr " QQ( N_( va( "%s win", BG_TeamNamePlural( level.lastWin ) ) ) ) );
```

Which could lead to the removal of more lines of code and strings to translate. One have to check if BG_TeamNamePlural and BG_TeamName returns are translated or not, but I doubt they are.

bmorel (approved):
More dedup could be done, but that might be out of the score of PR. LGTM.

src/sgame/sg_main.cpp
illwieckz: Hu no you can't assume every language has subject before verb, neither every language doesn't have any declension for a word being subject or not.

I'm splitting between sentences, not within sentences, because every language can translate a sentence.

illwieckz:
I would appreciate an answer to this question:

> Question: can we assume any “(`%s`)” will never have any translation variant?

src/sgame/sg_main.cpp
```diff
@@ -2180,18 +2180,28 @@ void G_CheckVote( team_t team )
...
+			: QQ( N_("Team vote failed") " " N_("($1$ of $2$; quorum not reached)") ),
+			level.team[ team ].voteYes + level.team[ team ].voteNo,
+			level.team[ team ].numPlayers );
 	}
 	else if ( pass )
 	{
-		cmd = va( "print_tr %s %d %d", ( team == TEAM_NONE ) ? QQ( N_("Vote passed ($1$ — $2$)") ) : QQ( N_("Team vote passed ($1$ — $2$)") ),
-		            level.team[ team ].voteYes, level.team[ team ].voteNo );
+		cmd = va( "print_tr %s %d %d", ( team == TEAM_NONE )
+			? QQ( N_("Vote passed") " " N_("($1$ — $2$)") )
```
slipher, line 2196: It may be hard to tell what `($1$ — $2$)` represents in isolation

src/sgame/sg_client.cpp
```diff
@@ -841,17 +841,17 @@ const char *ClientUserinfoChanged( int clientNum, bool forceName )
 		     level.time - client->pers.namelog->nameChangeTime <=
 		     g_minNameChangePeriod.Get() * 1000 )
 		{
-			trap_SendServerCommand( ent->num(), va(
-			                          "print_tr %s %g", QQ( N_("Name change spam protection (g_minNameChangePeriod = $1$)") ),
-			                          g_minNameChangePeriod.Get() ) );
+			trap_SendServerCommand( ent->num(), va("print_tr %s %f",
```
slipher, line 844: Why did you change `%g` to `%f`?

slipher:
> Question: can we assume any “(`%s`)” will never have any translation variant?

I found some languages like Chinese where there is a slightly different Unicode character for parentheses, but none where it would be seriously wrong.

src/sgame/sg_main.cpp
illwieckz: We don't need to know what it represents, only that this typography is used. It's the same issue like translating english `%s: %s` to french `%s : %s`, the context and the words should have no importance. It's about translating typography, not translating words.

There may be problems right-to-left language (which one is read first, right? or left?), but that problem is not added by this change and already exists in the original string.

illwieckz:
So, doing more research, actually one can translate `($1$ — $2$)` to:

- `（$1$ — $2$）` (japanese and chinese full-width parenthesis)
- `﴾$1$ — $2$﴿` (arab ornated parenthesis)

And probably many other variants.

I'm among the ones who want `$1$: $2$` or `$1$!` to be translated as `$1$ : $2$` or `$1$ !` in French according to French typography rules on double punctuation, so I won't prevent other people to use their own typography with parenthesis.

That japanese full-width parenthesis is interesting, because it reminds me we can't assume some language will not remove the space we usually write in alphabetical languages before a parenthesis. In fact some languages may even not use space between words. It means some deduplication can't be done by moving parenthesis outside of a sentence. I'll rewrite my patches then.

src/sgame/sg_client.cpp
illwieckz: was a typo

src/sgame/sg_main.cpp
```diff
@@ -1943,13 +1943,16 @@ void CheckExitRules()
 		else if ( level.matchTime >= ( level.timelimit - 5 ) * 60000 &&
 		          level.timelimitWarning < TW_IMMINENT )
 		{
-			trap_SendServerCommand( -1, "cp \"5 minutes remaining!\"" );
+			trap_SendServerCommand( -1, va( "cp_tr %s %d",
+				QQ( N_("$1$ minutes remaining!" ) ),
```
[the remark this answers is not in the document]
illwieckz: I just added a `cp_tr_p` implementation and used it.

src/sgame/sg_team.cpp
```diff
@@ -583,15 +583,15 @@ void CheckTeamStatus()
 		if ( level.team[ TEAM_ALIENS ].numSpawns > 0 &&
 		     G_PlayerCountForBalance( TEAM_HUMANS ) - G_PlayerCountForBalance( TEAM_ALIENS ) > 2 )
 		{
-			trap_SendServerCommand( -1, "print_tr \"" N_("Teams are imbalanced. "
-			                        "Humans have more players.") "\"" );
+			trap_SendServerCommand( -1, "print_tr %s"
```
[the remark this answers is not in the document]
illwieckz: Hmm, that's a good question. At least we're not adding an extra space within a sentence there. Can we guarantee we don't add any space between sentences? I remove this commit for now, we can still decide later.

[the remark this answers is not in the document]
slipher: Could be wrong but it seems like Chinese/Japanese generally don't use spaces.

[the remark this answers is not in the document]
illwieckz: Yes. We may have other place of the code where we assume the language has spaces.

illwieckz:
I added a commit modifying a code in `cg_tutorial.cpp` to not assume the language has spaces.

illwieckz:
I also added a commit to rewrite the `x, x and x` author name string to not assume space is not translatable.

PR #2643, opened 2023-04-18
Remove trailing whitespace within src/.

Viech:
Trailing whitespace are annoying to me: If I let my IDE remove them, these changes will end up in commits where they don't belong, complicating the review. If I do not let my IDE remove them, I will add more of them, annoying more people in the long run. With #2622, there is hope that we won't add more of them in the future.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Remove trailing whitespace within src/.
+c2291f23124ac1267573f0315938b3463a31efa8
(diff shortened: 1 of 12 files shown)


bmorel:
Usually I just remove them while I edit lines with those, and avoid adding them: I can see those in git's CLI output.
Still, LGTM.

DolceTriade:
Should we add this to `.git-blame-ignore-revs`?

DolceTriade:
(LGTM either way)

illwieckz:
> .git-blame-ignore-revs

good idea

Viech:
File added and confirmed it works with `git blame --ignore-revs-file=.git-blame-ignore-revs`. Haven't added any other commits to it though.

illwieckz (approved):
LGTM! Make sure the commit id doesn't get rewritten when merging! 😁️

Viech:
> LGTM! Make sure the commit id doesn't get rewritten when merging! grin

If it was possible I'd disable all three "options" that GitHub offers for synchronizing PRs… :grin:

PR #2644, opened 2023-04-21
Jumpfix

### LINKED ISSUE #1322: Touching a ladder with jetpack activated keeps it activated

bmorel:
This PR contains most commits of #2636. Actually, all of those that do not affect gameplay.

fix #1322


slipher:
LGTM

Issue #776, opened 2015-07-25
missing texture but no warning message


illwieckz:
**Edit: the example above was a false positive. See [this comment](https://github.com/Unvanquished/Unvanquished/issues/776#issuecomment-151153560) to get real use cases instead.**

For history, this message was:

> Sometime, some textures are missings, like in this screenshot:
> 
> ![missing texture](http://dl.illwieckz.net/b/unvanquished/bugs/unvanquished_2015-07-26_004818_000.jpg)
> But in some case (like in this screenshot), uvnanquished does not complain at all.
> 
> I don't know yet where is the problem, perhaps it's a shader issue. I've experienced many similar issues in my attempt to port tremulous maps, many missing textures are not reported.
> 
> I will investigate more to find in which case it is happening.

illwieckz:
Hmm, this screenshot is a false positive, this one seems to be a lightmap issue. But I have sometime some real missing textures without message (perhaps textures called by shaders or models, not directly called by the bsp).

DolceTriade:
Are you sure it isn't the same issue. If it is, it would probably require a debugger to step through the loading process to see where it returns early.

illwieckz:
I'm not sure since I've not found the good lightmap, but I think this weird pattern is in a lightmap itself, it looks like a shadow and it does not appear when I load the map without lightmaps.

But I agree about the debugger, I have an issue on a map where Unvanquished complains about a missing shader that is not missing: http://forums.unvanquished.net/viewtopic.php?f=9&t=1782 . Sometime Unvanquished complains about missing textures that are not missing, and sometime it does not complains about missings textures that are missing. But I don"t know how to trace it properly.

Viech:
Is this a duplicate of #782?

illwieckz:
They are definitely not.

Just one thing, the missing texture in initial post was a false positive (the problem was a bad generated lightmap, so the problem was at build time, not at run time, see [this forum post](http://forums.unvanquished.net/viewtopic.php?f=9&t=1888#p16287)).

But this bug is about real issues that have wrong, meaningless or missing error message like.

You can see:

* #799 `“some pending renderer regressions introduced in alpha 38 and alpha 40”`
* #780 `“Some fire/smoke effect broken since alpha 30 (vertexpack merge)”`
* #806 `“R_FindShader looks unnecessarily for filenames that does not exist then abort”` (and the related [forum post](http://forums.unvanquished.net/viewtopic.php?f=9&t=1782))

[...]

PR #2648, opened 2023-04-25
print a warning on missing shader/model/sound/particle system, etc. and skip broken light flare entities

illwieckz:
There are multiple ways to have some stuff missing without getting any warning.

Right now I got a but, it's a regression in the nexus6 map after [I released 1.2.2](https://forums.unvanquished.net/viewtopic.php?p=19199#p19199). It happens that nexus6 1.2.2 is now built from source, and source had a mistake, a light flare entity in a wall (wrong, but not the cause of the regression) that does not have any shader set (the cause of the regression).

Surprinsingly, the game was drawing the fallback texture without printing any warning about the missing texture.

This PR adds warnings to many missing stuff.

It also prevents to produce the invalid light flare entity if such entity misses the required texture, meaning maps using broken light flare entity without texture would print a warning _and_ will render correctly.

### DIFF
--- a/src/sgame/sg_spawn_gfx.cpp
+++ b/src/sgame/sg_spawn_gfx.cpp
@@ -137,6 +137,12 @@ static void findEmptySpot( glm::vec3 const& origin, float radius, glm::vec3& spo
 
 void SP_gfx_light_flare( gentity_t *self )
 {
+	if ( !self->shaderKey )
+	{
+		Log::Warn( "Light flare entity %d at (%d, %d, %d) has no shader", self->num(), self->s.origin[0], self->s.origin[1], self->s.origin[2] );
+		return;
+	}
+
 	self->s.eType = entityType_t::ET_LIGHTFLARE;
 	self->s.modelindex = G_ShaderIndex( self->shaderKey );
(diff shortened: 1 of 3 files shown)


illwieckz:
This fixes some occurrences of:

- https://github.com/Unvanquished/Unvanquished/issues/776

but not all of them.

src/sgame/sg_spawn_gfx.cpp
```diff
@@ -137,6 +137,12 @@ static void findEmptySpot( glm::vec3 const& origin, float radius, glm::vec3& spo
 
 void SP_gfx_light_flare( gentity_t *self )
 {
+	if ( !self->shaderKey )
+	{
+		Log::Warn( "Light flare entity at (%d, %d, %d) has no shader", self->s.origin[0], self->s.origin[1], self->s.origin[2] );
```
DolceTriade, line 142: Consider including entity number as a better way of identifiying entities (in addition to the location).

DolceTriade (approved):

src/sgame/sg_spawn_gfx.cpp
illwieckz: Done.

PR #2649, opened 2023-04-25
fix an unused variable warning in `parse.cpp`

illwieckz:
Fix an unused variable warning in `parse.cpp`:

```
src/shared/parse.cpp:500:14: warning: variable 'i' set but not used [-Wunused-but-set-variable]
        int c, val, i;
                    ^
```

Also deobfuscate the code a bit, this file is incredibly convoluted!


slipher:
lgtm

PR #2650, opened 2023-04-26
Question: cg_main: do not apply MSVC workaround on not MSVC compilers

illwieckz:
This is mostly a question, I was toying with some compilers and I noticed this exact code in `cg_main.cpp` was crashing ICC, actually the two occurrences of the `new(&something) something{};` form:

```c++
	cgs.~cgs_t();
	new(&cgs) cgs_t{}; // Using {} instead of () to work around MSVC bug
	cg.~cg_t();
	new(&cg) cg_t{};
```

```
[ 81%] Building CXX object CMakeFiles/cgame-native-dll.dir/src/cgame/cg_main.cpp.o
icpc: command line warning #10148: option '-march=x86-64' not supported
"": internal error: ** The compiler has encountered an unexpected problem.
** Segmentation violation signal raised. **
Access violation or stack overflow. Please contact Intel Support for assistance.
icpc: error #10106: Fatal error in /opt/intel/oneapi/compiler/2023.1.0/linux/bin/intel64/../../bin/intel64/mcpcom, terminated by segmentation violation
compilation aborted for src/cgame/cg_main.cpp (code 1)
gmake[3]: *** [CMakeFiles/cgame-native-dll.dir/build.make:272 : CMakeFiles/cgame-native-dll.dir/src/cgame/cg_main.cpp.o] Error 1
```

Though even if ICC is the least of our concern, I'm surprised we apply an MSVC workaround to all non-MSVC compilers.

This MSVC workaround was implemented in faee5bae3576085e01775e61cb3b754e08fbc6ae

Interestingly there are two other occurrences in engine code in `cl_main.cpp` but ICC doesn't crash with them:

```c++
	cl.~clientActive_t();
	new(&cl) clientActive_t{}; // Using {} instead of () to work around MSVC bug
```

```c++
	clc.~clientConnection_t();
	new(&clc) clientConnection_t{}; // Using {} instead of () to work around MSVC bug
```

So basically, I find this curious…

- Can the ICC crash reveal a bug in our code somewhere else?
- Is ICC stupid? 
- Should the workaround be fine on any compiler?
- Should we implement a kind of macro we would allow us to do something like `ZERO(cgs, cgs_t)` and hide the magic, eventually using per-compiler code?

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1252,10 +1252,23 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 	// clear everything
 	// reset cgs in-place to avoid creating a huge struct on stack (caused a stack overflow)
 	// this is equivalent to cgs = cgs_t()
+#ifdef _MSC_VER
+	/* Use a C++11 braced initializer instead of a bracket initializer when
+	zeroing a struct. This works around a bug in how MSVC generates implicit
+	default constructors. */
 	cgs.~cgs_t();
-	new(&cgs) cgs_t{}; // Using {} instead of () to work around MSVC bug
+	new(&cgs) cgs_t{};
 	cg.~cg_t();
 	new(&cg) cg_t{};
+#else
+	/* The MSVC workaround is known to crash ICC, there is no reason to apply
+	MSVC workarounds on non-MSVC compilers. */
+	cgs.~cgs_t();
+	new(&cgs) cgs_t();
+	cg.~cg_t();
+	new(&cg) cg_t();
+#endif
+
 	for ( centity_t& ent : cg_entities) { ent = {}; }
 
 	CG_UpdateLoadingStep( LOAD_START );


illwieckz:
There are indeed many other occurrences if that `{}` wording without getting any ICC crash!

in Unvanquished:

```sh
$ rg 'new\(' src
src/cgame/cg_trails.cpp
1446:			new(ts) trailSystem_t{};

src/cgame/cg_beacon.cpp
53:	new(bc) beaconsConfig_t{};

src/sgame/sg_entities.cpp
196:	new(entity) gentity_t{};

src/cgame/cg_particles.cpp
522:			new(ps) particleSystem_t{};

src/cgame/cg_view.cpp
80:	new(&cg.testModelEntity) refEntity_t{};
83:	new(&cg.testModelBarrelEntity) refEntity_t{};

src/cgame/cg_main.cpp
1262:	new(&cgs) cgs_t{};
1264:	new(&cg) cg_t{};
1269:	new(&cgs) cgs_t();
1271:	new(&cg) cg_t();
```

In Dæmon:

```sh
$ rg 'new\(' src
src/common/System.cpp
412:void* operator new(size_t n)

src/common/Util.h
145:		new(&data) T(std::forward<Args>(args)...);

src/engine/client/cl_main.cpp
796:	new(&cl) clientActive_t{}; // Using {} instead of () to work around MSVC bug
860:	new(&clc) clientConnection_t{}; // Using {} instead of () to work around MSVC bug
3033:	new(&cls) clientStatic_t{};

src/engine/server/sv_init.cpp
751:	new(&svs) serverStatic_t{};

src/engine/renderer/tr_init.cpp
1375:		new(&tr) trGlobals_t{};
1378:		new(&backEnd) backEndState_t{};
1381:		new(&tess) shaderCommands_t{};
```

So I don't know what is happening…

illwieckz:
Actually it looks like one can give life to such convenient monstrosity:

```c++
#ifdef _MSC_VER
#define ZERO( S, T ) do { S.~T(); new(&S) T{}; } while (false)
#else
#define ZERO( S, T ) do { S.~T(); new(&S) T(); } while (false)
#endif

	ZERO(cgs, cgs_t);
	ZERO(cg, cg_t);
```

slipher:
> * Is ICC stupid?

Yes. But I believe GCC, Clang, and MSVC have all been seen to crash during compilation on either the `()` or `{}` forms of these constructors, so he's not alone.

We may just delete a lot of these (including the `cgs_t` one). The only time they would be useful is if you reused a DLL for more than one game, which presumably was supported in the past. But more recently we have done down the path of requiring DLL unloading to work correctly, so that we can have a truly fresh start on each game. (This is good because trying to reset every static variable in the whole codebase is too onerous.) Systems with broken DLL unloading now exhibit obvious problems and it is a non-goal to support that.

illwieckz:
> Yes. But I believe GCC, Clang, and MSVC have all been seen to crash during compilation on either the () or {} forms of these constructors, so he's not alone.

Right. Anyway what do you think of the proposal of a reusable feature like the one I show (this one or another, you may have better implementation ideas), especially for the engine?

illwieckz:
Here is a proposal for an implementation of reusable `ZERO_STRUCT( struct, type )` macro on engine side:

- https://github.com/DaemonEngine/Daemon/pull/841

illwieckz:
I will do a PR with `ResetStruct` instead for what's the remaining after:

- https://github.com/Unvanquished/Unvanquished/pull/2655

PR #2654, opened 2023-04-27
cgame,sgame: request and resend `pmoveParams` when client is ready

### LINKED ISSUE #2653: Performance loss after `vid_restart`
Moved from:

- https://github.com/DaemonEngine/Daemon/issues/845

This is better seen with low graphics profile, so the spent time is probably on CPU side. Here we can see how I jump from 1000fps to 500fps after a `vid_restart`:

Fresh start:

[vid restart performance loss](https://dl.illwieckz.net/b/daemon/bugs/vid-restart-performance-loss/unvanquished_2023-04-27_035941_000.jpg)

After `vid_restart`:

[vid restart performance loss](https://dl.illwieckz.net/b/daemon/bugs/vid-restart-performance-loss/unvanquished_2023-04-27_040035_000.jpg)

I'm running game code as native exe and not dll, so I doubt that is caused by dll memory leaking.

---

Also, what does mean `snc ms`?

illwieckz:
Previously, `clientParams` was only sent on client connect, but the client needs the server to resend them after reloading the game with `vid_restart`.

Fixes #2653:

- https://github.com/Unvanquished/Unvanquished/issues/2653

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1399,6 +1399,9 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 	CG_ShaderStateChanged();
 
 	trap_Cvar_Set( "ui_winner", "" ); // Clear the previous round's winner.
+
+	// Request server to resend pmoveParams.
+	trap_SendClientCommand( "client_ready" );
 }
 
 /*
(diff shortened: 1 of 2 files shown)


slipher:
Ideally we should use infostrings for this instead of inventing an ad-hoc synchronization mechanism, but infostring synchronization is suspected to be buggy so this is OK for now. LGTM

PR #2655, opened 2023-04-27
Skip unnecessary clearing of cgame globals

slipher:
As you can from the next line CG_UpdateLoadingStep, CG_Init is the stuff that runs in the loading screen which can happen only once during the cgame lifecycle. So there is no need to clear variables that are already clear upon the cgame module initialization.


illwieckz (approved):
LGTM.

PR #2657, opened 2023-04-28
rmlui: implement web custom tags and b, i, em and ul/li standard tags, make use of them

illwieckz:
- Introduce `web` tag for common Unvanquished urls, to be used this way: `Discuss news in the forums: <web forums />`, this will be displayed this way: `Discuss news in the forums: forums.unvanquished.net`. This makes the strings translatable without translating the URL itself. The day we implement hyperlinks, we would have nothing to translate. If we want to change an URL, we would have nothing to translate. And we don't have to put any styling code in the translatable string.

- Implement `b`, `i`, `em` and even `ul/li` standard tags.

Make use of all that niceness.

[rmlui](https://dl.illwieckz.net/b/unvanquished/wip/rmlui/unvanquished_2023-07-24_062320_000.png)

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/shared/basics.rcss
+++ b/pkg/unvanquished_src.dpkdir/ui/shared/basics.rcss
@@ -62,6 +62,10 @@ indent {
 	margin-left: 1.5em;
 }
 
+web {
+	color: #ffffff;
+}
+
 /* #### Links #### */
 /* Inline */
 ilink {
--- a/pkg/unvanquished_src.dpkdir/ui/shared/basics.rcss
+++ b/pkg/unvanquished_src.dpkdir/ui/shared/basics.rcss
@@ -45,6 +45,18 @@ h3 {
 	height: 1.3em; /* Same as input boxes */
 }
 
+b {
+	font-weight: bold;
+}
+
+i {
+	font-style: italic;
+}
+
+em {
+	font-style: italic;
+}
+
 indent {
 	display: block;
 	margin-left: 1.5em;
--- a/pkg/unvanquished_src.dpkdir/ui/shared/basics.rcss
+++ b/pkg/unvanquished_src.dpkdir/ui/shared/basics.rcss
@@ -57,6 +57,26 @@ em {
 	font-style: italic;
 }
 
+/* For now, render ordered list
+as unordered list as a fallback. */
+ul, ol {
+	display: table;
+	margin: 1em 0em;
+}
+
+li {
+	display: table-row;
+}
+
+li-bullet {
+	display: table-cell;
+	width: .6em;
+}
+
+li-content {
+	display: table-cell;
+}
+
 indent {
 	display: block;
 	margin-left: 1.5em;
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -75,6 +75,41 @@ void Rocket_UpdateLanguage()
 		hudContext->Update();
 	updateLanguage = false;
 }
+
+class WebElement : public Rml::Element
+{
+public:
+	WebElement(const Rml::String& tag) : Rml::Element(tag) {}
+
+private:
+	const std::unordered_map<std::string, std::string> webUrls = {
+		{ "bugs", "bugs.unvanquished.net" },
+		{ "chat", "unvanquished.net/chat" },
+		{ "forums", "forums.unvanquished.net" },
+		{ "wiki", "wiki.unvanquished.net" },
+	};
+
+	void OnAttributeChange( const Rml::ElementAttributes& changed_attributes ) override
+	{
+		Rml::Element::OnAttributeChange( changed_attributes );
+
+		if ( changed_attributes.find( "dest" ) != changed_attributes.end() )
+		{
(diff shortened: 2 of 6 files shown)


src/cgame/cg_rocket_draw.cpp
```diff
@@ -75,6 +75,65 @@ void Rocket_UpdateLanguage()
...
+	void OnUpdate() override
+	{
+		bool found = false;
+
+		for ( const std::pair<const std::string, std::string>& p : webUrls )
+		{
+			if ( HasAttribute( p.first ) )
+			{
+				found = true;
+				SetInnerRML( va( "%s", p.second.c_str() ) );
```
slipher, line 124: Useless invocation of `va`

src/cgame/cg_rocket_draw.cpp
```diff
@@ -75,6 +75,65 @@ void Rocket_UpdateLanguage()
 		hudContext->Update();
 	updateLanguage = false;
 }
+
+// Standard HTML
+
+class LiElement : public Rml::Element
```
slipher, line 81: I guess this is only intended to work for *unordered* lists?

pkg/unvanquished_src.dpkdir/ui/shared/basics.rcss
```diff
@@ -62,6 +62,11 @@ indent {
 	margin-left: 1.5em;
 }
 
+web {
+	color: #136d15; /* Green */
```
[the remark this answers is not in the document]
illwieckz: No it's not supposed to look like a link, because indeed you can't click on it. I just wanted to make these more obvious to spot in the page, but maybe that may be confusing if people believes it's a link.

src/cgame/cg_rocket_draw.cpp
illwieckz: Ah yes, in my first version of the patch I was adding some `<b>` and stuff like that. 😁️

src/cgame/cg_rocket_draw.cpp
illwieckz: Yes, I may rename accordingly, maybe `LuLiElement`?

necessarily-equal:
I said hi on the rmlui channel, here's a small suggestion:

> Nice, this is something I'm sure is useful to others too. Interesting to see you use table layout.
>
> If you are interested in feedback, one change I'd recommend is to use a custom node handler to create the inner elements and move the content (see e.g. XMLNodeHandlerTabSet). The get-/setinnerrml might not always be 100% reliable, and certainly not as performant.

It's certainly easier to avoid the hack, in any case.

I also wasn't sure if we needed a `vertical-align: top;`, but this is apparently the default value already

src/cgame/cg_rocket_draw.cpp
illwieckz: This only works for unordered list but checking if it's in an `ul` or `ol` would be overcomplicated for now, we better want to make all list unordered for now, treating `ol` as `ul` as a fallback, and write complex code the day we support `ol`.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -75,6 +75,42 @@ void Rocket_UpdateLanguage()
...
 	updateLanguage = false;
 }
+
+class WebElement : public Rml::Element
+{
+public:
+	WebElement(const Rml::String& tag) : Rml::Element(tag) {}
+
+private:
+	std::unordered_map<std::string, std::string> webUrls = {
```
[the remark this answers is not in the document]
illwieckz: I don't get why you would want an array there.

The compiler doesn't accept `static` for the `unordered_map`, but I added `const`.

pkg/unvanquished_src.dpkdir/ui/shared/basics.rcss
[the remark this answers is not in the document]
illwieckz: I've set it white for now. But even white can be confused with a link unless we color the main menu links like other in-game links.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -75,6 +75,42 @@ void Rocket_UpdateLanguage()
...
+		{ "bugs", "bugs.unvanquished.net" },
+		{ "chat", "unvanquished.net/chat" },
+		{ "forums", "forums.unvanquished.net" },
+		{ "wiki", "wiki.unvanquished.net" },
+	};
+
+	void OnUpdate() override
```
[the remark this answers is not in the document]
illwieckz: Done.

src/cgame/cg_rocket_draw.cpp
illwieckz: Fixed.

illwieckz:
The code works, if someone has ideas for improved c++ism, please make a PR to improve this code.

I don't want to delay more this as this brings translatable string updates, and people have translated already-obsolete strings for three months, while we already knew those strings were obsolete. This should not happen.

src/cgame/cg_rocket_draw.cpp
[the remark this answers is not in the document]
illwieckz: See https://github.com/Unvanquished/Unvanquished/pull/2657#issuecomment-1647190160

PR #2664, opened 2023-04-29
Do not try to print `nullptr` in error message

illwieckz:
Necessary change after:

- https://github.com/Unvanquished/Unvanquished/pull/2648

When loading chasm map I got this:

```
Warn: Assertion failure at src/common/String.cpp:37 (AssertOnTinyFormatError): "reason == """ expected: "tinyformat: Null char* pointer passed for string parameter", actual:  
Thread 39 "daemon" received signal SIGTRAP, Trace/breakpoint trap.
0x00007fffaa80e2bd in Str::AssertOnTinyFormatError (reason="\"tinyformat: Null char* pointer passed for string parameter\"") at src/common/String.cpp:37
37	            ASSERT_EQ(reason, "");
```

### DIFF
--- a/src/sgame/sg_utils.cpp
+++ b/src/sgame/sg_utils.cpp
@@ -150,7 +150,7 @@ int G_ParticleSystemIndex( const char *name )
 
 	if ( !i )
 	{
-		Log::Warn( "Missing particle system: %s", name );
+		Log::Warn( "Missing particle system: %s", name ? name : "<nullptr>" );
 	}
 
 	return i;
@@ -162,7 +162,7 @@ int G_ShaderIndex( const char *name )
 
 	if ( !i )
 	{
-		Log::Warn( "Missing shader: %s", name );
+		Log::Warn( "Missing shader: %s", name ? name : "<nullptr>" );
 	}
 
 	return i;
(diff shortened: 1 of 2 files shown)


src/sgame/sg_utils.cpp
```diff
@@ -162,7 +162,7 @@ int G_ShaderIndex( const char *name )
 
 	if ( !i )
 	{
-		Log::Warn( "Missing shader: %s", name );
```
slipher, line 165: "Missing shader nullptr" is not very actionable for the user...

illwieckz: yes 😅️

Well, this is mostly to not crash the game while leaving an useful message for the developper, I doubt an user-provided asset is expected to produce a nullptr by itself, a code has to do it.

illwieckz: Well, this PR fixes a regression in current master that can crash the game, I'll merge this as is. I don't think we need something “better”, because a more literary wording would just be less meaningful, and I see no reason to complicate more the code by adding more tests and alternate error messages. If this messages prints there is a nullptr, that's a message for the developer, and I want user provides logs that are explicit to developers.

PR #2667, opened 2023-04-30
cg_main: fix navgen messages by copying translated string

illwieckz:
Fix navgen messages by copying translated string, otherwise we may reuse a pointer that can contain other data.

Fixes #2666:

- https://github.com/Unvanquished/Unvanquished/issues/2666

PR #2668, opened 2023-04-30
reuse `CG_UpdateLoadingProgress` for navmesh generation

illwieckz:
Reuse `CG_UpdateLoadingProgress` for navmesh generation.

Include commit from #2667 because I don't want to deal with rebase conflicts. 😀️

We may port `CG_UpdateLoadingProgress` to use `std::string` but that's outside of the scope of this PR.


src/cgame/cg_main.cpp
```diff
@@ -1129,11 +1129,11 @@ static void GenerateNavmeshes()
 		return;
 	}
 
+	const std::string label = "Navmesh";
 	const std::string message = _("Generating bot navigation meshes");
 	cg.navmeshLoadingFraction = 0;
 	cg.loadingNavmesh = true;
-	Q_strncpyz( cg.loadingText, message.c_str(), sizeof(cg.loadingText) );
-	trap_UpdateScreen();
+	CG_UpdateLoadingProgress( LOAD_START, label.c_str(), message.c_str() );
```
slipher, line 1136: This commit seems like an awfully hacky way to make the code one line shorter. Setting a bogus value to cg.loadingFraction and emitting an  incorrect debug log message.

bmorel: Unrelated to this commit or comment, but could it be possible to start duplicating (yes, I know) interfaces to no longer use `char const*` but instead `std::string const&`? Thoses `.c_str()` that pollute the code are getting a bit annoying.

slipher: We have StringRef for such interfaces. The point of StringRef is to make function arguments that can accept either std::string or char* without any .c_str() or inefficient copying.

bmorel: well, there's no copy with const references. But, I should take a look at this myself, thanks for info.

slipher: Yes there is, if you put char* into a std::string argument then you are copying

bmorel: ofc

illwieckz: > This commit seems like an awfully hacky way to make the code one line shorter. Setting a bogus value to cg.loadingFraction and emitting an incorrect debug log message.

Ah yes, I totally forgot I removed the printing of the log message because i was asked to remove it.

When I revamped this function when reimplementing map load message, I initially revamped it in order to be reusable, but this reusability was removed before merging so now the log message is meaningless when reusing the function.

illwieckz:
Well, I merged in that PR some other changes I was planning to do on another PR. Initially I split it into 3, but reworking this would have made the other ones impossible to rebase. So there is now more work:

- now what happens is that only loading step uses the multi-variable loading fraction, it means `CG_UpdateLoadingProgress` only use `cg.loadingFraction` when not iterating a step. In the future I would like to see this used for GLSL compilation too.
- some sound stuff were done at different places, so I grouped them together.
- more `std::string` and other c++isms.
- now the navmesh fraction goes until the navmesh step, so basically the usual loading progress up to navmesh step, if there is navmesh to generate, it display the navmesh progression from 0 to navmesh step and continue from there with other steps.

illwieckz:
Also I reintroduced the loading message in debug log, even if some are funny ones, I really want to have the complete log in debug log in console.

src/cgame/cg_main.cpp
illwieckz: That code was totally rewritten so the thread is obsolete. New remarks should be done in another thread.

illwieckz:
> So you're getting rid of the 2nd progress bar for navmesh generation? I don't like it.

What second progress bar for navmesh generation?

The way it worked before my PR is that the navmesh progress bar replaced the global one, starting from zero and going to 100%, then the global one continued from where it was.

Now when the global progress bar reaches the navmesh point, if there is no navmesh to do it continues from there, otherwise the navmesh progress bar replaces the global one starting from zero and going to the navmesh point, and then continue with the global progress bar from there.

The choice to make the navmesh bar top at the navmesh point of the global bar, or the choice to make it use the whole screen, this is just a cosmetic choice which is only one line in the PR.

slipher:
OK, I guess I incorrectly guessed what it is supposed to do. I still don't understand though: what is the purpose of this PR? The "code reuse" doesn't seem to be buying us anything as the delta adds 6 lines on net.

illwieckz:
I don't undestand what means:

> as the delta adds 6 lines on net

slipher:
Sorry, I missed earlier that there was a Github comment explaining some changes. Anyway I think if we have a secondary progress bar, it should go 0 to 100, not end at some smaller number.

Then new code is confusing for me; I can't understand what the `bool loadingStep` variables are supposed to do.

illwieckz:
> if we have a secondary progress bar, it should go 0 to 100, not end at some smaller number.

Ok, this one is just a matter of taste, I'll change that.

> I can't understand what the `bool loadingStep` variables are supposed to do.

It is telling the rml code that we are iterating the global loading bar, so we need the complex progress number computation that aggregates multiple progress fractions.

I can rename it `loadingGlobal` or something like that if needed.

slipher:
I don't think it's a good refactoring. The proposed version is more confusing

PR #2669, opened 2023-04-30
Implement `entityLock` svcmd

cu-kai:
This can be used to lock any door or button entity and prevent it from being used. Ideally this would be placed in a map config.

It adds a "locked" field to gentity_t, which could in future be used for other purposes.

The player is warned with a "This door is locked!" message when they approach or try to use a locked entity. I chose this wording because the proposed use case is for doors, or elevators where the "door" wording still makes the most obvious sense.

This could be used for:
 - PvE maps, where the server admin wants players to follow a certain path.
 - Large maps, where the server admin wants to only allow access to a small part of a map for gameplay.
 - Bad maps, where the server admin wants to prevent access to campy / "win" rooms.

### DIFF
--- a/src/sgame/sg_svcmds.cpp
+++ b/src/sgame/sg_svcmds.cpp
@@ -291,13 +291,11 @@ static void Svcmd_EntityLock_f()
 
 	trap_Argv( 1, argument, sizeof( argument ) );
 
-	if ( !atoi( argument ) )
+	if ( !Str::ParseInt( e, argument ) )
 	{
 		Log::Notice( "entitylock: this is not numeric" );
 		return;
 	}
-
-	e = atoi( argument );
 
 	if ( e >= level.num_entities || e < MAX_CLIENTS )
 	{
--- a/src/sgame/sg_svcmds.cpp
+++ b/src/sgame/sg_svcmds.cpp
@@ -277,7 +277,7 @@ Svcmd_EntityLock_f
 
 static void Svcmd_EntityLock_f()
 {
-	char      argument[ 4 ];
+	char      argument[ 16 ];
 	int       e;
 	gentity_t *door;
 
--- a/src/sgame/sg_svcmds.cpp
+++ b/src/sgame/sg_svcmds.cpp
@@ -271,6 +271,68 @@ static gclient_t *ClientForString( char *s )
 
 /*
 ===================
+Svcmd_EntityLock_f
+===================
+*/
+
+static void Svcmd_EntityLock_f()
+{
+	char      argument[ 4 ];
+	int       e;
+	gentity_t *door;
+	vec3_t    mins, maxs;
+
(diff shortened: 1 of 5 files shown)


src/sgame/sg_svcmds.cpp
```diff
@@ -269,6 +269,68 @@ static gclient_t *ClientForString( char *s )
...
+===================
+Svcmd_EntityLock_f
+===================
+*/
+
+static void Svcmd_EntityLock_f()
+{
+	char      argument[ 4 ];
+	int       e;
+	gentity_t *door;
```
bmorel, line 282: it's easier to read when variables are declared as close as where they're used/init (nitpick)

src/sgame/sg_svcmds.cpp
```diff
@@ -269,6 +269,68 @@ static gclient_t *ClientForString( char *s )
...
+	vec3_t    mins, maxs;
```
bmorel, line 283: please use `glm::vec3` instead

src/sgame/sg_svcmds.cpp
```diff
@@ -269,6 +269,68 @@ static gclient_t *ClientForString( char *s )
...
+	if ( trap_Argc() < 1 )
+	{
+		Log::Notice( "usage: entitylock <entityNum> (tip: use " 
+		             "'cg_drawEntityInfo on' on your client to find this)" );
+		return;
+	}
+
+	trap_Argv( 1, argument, sizeof( argument ) );
+
+	if ( !atoi( argument ) )
```
bmorel, line 294: you could store the value into `e` first and then check that `e == 0` it would avoid useless computations.

bmorel (commented):
Just a few nitpicks that don't have to be changed (but would be appreciated), but otherwise it seems good. I have not tried to the code myself so can't approve, will do when I'll can.

sweet235:
lgtm

src/sgame/sg_svcmds.cpp
sweet235: I moved `mins` and `maxs` to a smaller scope. Let's resolve this.

src/sgame/sg_svcmds.cpp
```diff
@@ -269,6 +269,65 @@ static gclient_t *ClientForString( char *s )
...
+		Log::Notice( "usage: entitylock <entityNum> (tip: use " 
+		             "'cg_drawEntityInfo on' on your client to find this)" );
+		return;
+	}
+
+	trap_Argv( 1, argument, sizeof( argument ) );
+
+	if ( !Str::ParseInt( e, argument ) )
+	{
+		Log::Notice( "entitylock: this is not numeric" );
```
DolceTriade, line 295: print the argument in the warning. Also make this an error or warning

DolceTriade (approved):

PR #2670, opened 2023-05-01
remove useless crouch armor variations

bmorel:


slipher:
Why do you say it's dead code? Superficially checking, the crouch tag appears to be used in `ArmorComponent::GetLocationalDamageMod`

bmorel:
it's used, but have the same values as non crouched

slipher:
Still, location damage will be broken by having no matching block for any crouching player and thus always returning 1.0.

Issue #1861, opened 2022-07-06
limit number of bots wearing the same weapon


bmorel:
I implemented a feature to limit the number of bots using the same weapon/class. This was initially done to reduce number of tyrants and lucifer cannons which are rather specialised weapons that need supports.
This have various cons and pros at lower momentum level, notably, alien bots might take normal dragoon instead of adv marauder, in current status.

PR #2672, opened 2023-05-03
Make bots sometimes evolve to advanced marauder

sweet235:
Currently, bots do not use the advanced marauder class at all. That is a shame, as they are able to use its zap quite well.

This PR uses random to make a bot choose between advanced marauder and dragoon, such that it will pick the advanced marauder 30% of the time.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -597,6 +597,8 @@ float PercentAmmoRemaining( weapon_t weapon, playerState_t const* ps )
 	}
 }
 
+static Cvar::Cvar<float> g_bot_probLevel2upg("g_bot_probLevel2upg", "probability for a bot to evolve to advanced marauder", Cvar::NONE, 0.3);
+
 AINodeStatus_t BotActionEvolve( gentity_t *self, AIGenericNode_t* )
 {
 	class_t currentClass = static_cast<class_t>( self->client->ps.stats[ STAT_CLASS ] );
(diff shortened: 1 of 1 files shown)


illwieckz:
While I'm not very happy with randomness in general, at least this time it's just about selecting a class, so that doesn't make the bot randomly behaving once the class is selected so the idea looks fine to me. I haven't reviewed the code.

bmorel:
There was already code for that, what happened to it? Sure, it did not had a cvar, but I clearly remember that there was randomness between choosing dragoon or adv marauder when I looked at this part of the code months ago.

sweet235:
> There was already code for that

When was this, roughly?

Viech:
LGTM, although I had preferred a general approach (to pick at random between same-cost (or even similar-cost) classes) as opposed to a special case for the Advanced Marauder/Dragoon pair. This would be more future-robust: think of price changes or new classes.

bmorel:
> When was this, roughly?

I'm not too sure, I thin it was always there, it's part of the AI code I barely touched because alien's gameplay offers so little options. It was not controlled by a cvar though, so that at least is an improvement. And if the feature was restored, it's good too.

sweet235:
For good reasons, this is not universally appreciated. As this part of the code base is among the more controversial ones, a more robust implementation does not seem feasible at the moment. I will close this.

bmorel:
Just for the sake of the linking, this would have participated into improving #1861

PR #2673, opened 2023-05-03
 scoreboard shows number of players per team (improved)

illwieckz:
Improvement over:

- https://github.com/Unvanquished/Unvanquished/pull/2630

The main change is that it reuse an existing loop to compute player count, to only do it once per frame.

It also lists spectator count.

I renamed the tag `playerCount` because I thought the wording was more consistent with other things.

unvanquished_2023-05-03_144853_000

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/ingame_scoreboard.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/ingame_scoreboard.rml
@@ -155,7 +155,7 @@
 	<inlinecvar cvar="ui_winner" class="winner" />
 
 	<half>
-		<teamname><translate>Aliens</translate> (<playerscount team="aliens"/>)</teamname>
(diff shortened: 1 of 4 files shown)


Viech:
Do you think it would be better to show only real players or `($real+$bots)` if there are bots? I personally find the real player count most significant for balancing (and we did have some cases where teams where 3v1 or worse but filled up with bots).

Otherwise (including when you disagree) LGTM.

illwieckz:
> Do you think it would be better to show only real players or `($real+$bots)` if there are bots?

Yes and no, I would like to know the real players indeed, but I don't like the (X+Y) display.

Anyway I don't want to make this PR wait for something else, anything considered better can be done over that later.

sweet235:
LGTM, but didn't test the code.

> we did have some cases recently where teams were 3v1 or worse

My secret goal is to make the bots useful enough for this to be irrelevant.

PR #2677, opened 2023-05-05
Remove a lot of p_* vars

slipher:
Some of them could be used to cheat with binds to automate player actions:
- p_hp
- p_ammo
- p_clips
- p_credits

Others are just not very useful:
- p_maxhp
- p_score

Finally p_availableBuildings was obviously meant to be parsed by some sort of UI code. Maybe that's why some other ones existed too.

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -185,9 +185,6 @@ these should refer only to playerstates that belong to the client, not the follo
 static void CG_SetPVars()
 {
 	playerState_t *ps;
-	char          buffer[ MAX_CVAR_VALUE_STRING ];
-	int           i;
-	bool      first;
 
 	if ( !cg.snap )
 	{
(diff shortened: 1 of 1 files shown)


illwieckz:
This may also speed-up the game:

[slow trap](https://dl.illwieckz.net/b/unvanquished/bugs/slow-trap/20230505-141022-000.orbit-unvanquished.png)

illwieckz (approved):
LGTM.

DolceTriade:
What can a user do with this?

slipher:
> What can a user do with this?

If you're referring to the deleted ones, then you could use p_hp to cheat by making a script that automatically uses the medkit when you have low health.

If you're referring to the ones that are kept, p_classname could be used to have different binds depending on what kind of alien you are.

DolceTriade:
honestly I'm not sure automedkit binds are not super abusable. The maxhp cvar was so people could make binds like

"I have $p_hp$ out of $p_maxhp$"

but if you feel strongly, then lgtm

slipher:
> The maxhp cvar was so people could make binds like
> 
> `"I have $p_hp$ out of $p_maxhp$"`

Right, and now it's useless because there is no longer `p_hp` :)

illwieckz:
> If you're referring to the deleted ones, then you could use p_hp to cheat by making a script that automatically uses the medkit when you have low health.

I guess one could do the same to drop the grenade automatically before dying…

sweet235:
I have a bind to tell my team what weapon I am carrying, using `p_weapon`. I have another to tell them my current health, using `p_hp`. Even though I rarely use these binds, I would not want to lose them.

More on topic:

**Theorem.** The console is a turing-complete language.

**Proof.** First, notice that we can use `set` to create an arbitrary number of cvars, each holding an arbitrary string (probably up to some maximal length). We can use `concat` to chain these cvars together, creating longer strings. We can use `if` to do this conditionally. Finally, we can use `vstr` to execute the string a cvar holds. An example is given in a comment in https://github.com/Unvanquished/Unvanquished/pull/2782. If we can find no way to work around a maximal lentgh limit, we always increase this limit and recompile.

To complete the proof, notice that we only need to escape some quotes in order to embed the lambda calculus. Alternatively, create two stacks of cvars `a0`, `a1`, ...; and `b0`, `b1`, ...; thereby constructing a turing machine. □

Thus removing these cvars will not prohibit any of the more nasty cheats I can think of. I propose we accept this and move on.

slipher:
Yeah, obviously the console scripting language has more than enough ways to be Turing-complete. I took that as a given when writing this PR. The point was to avoid giving the script useful data that would help it to automate decisions. It doesn't matter how powerful the script is as long as it doesn't have any inputs.

sweet235:
> It doesn't matter how powerful the script is as long as it doesn't have any inputs

That is not correct. Many nasty things can be done by using only constants.

sweet235:
Why was this merged? I count one thumbs up, three thumbs down, and one "but if you feel strongly, then lgtm".

slipher:
> Why was this merged? I count one thumbs up, three thumbs down, and one "but if you feel strongly, then lgtm".

It would seem these "reactions" were added after the PR was merged. Next time I will consult my crystal ball before merging to see if there are any objections in the future.

sweet235:
> It would seem these "reactions" were added after the PR was merged. Next time I will consult my crystal ball before merging to see if there are any objections in the future.

Easy: if the word cheat occurs in your description, it is probably a bad idea for an open source game.

sweet235:
> Easy: if the word cheat occurs in your description, it is probably a bad idea for an open source game.

Once the Open Source Game has 1000+ regular users, I might change my opinion.

PR #2679, opened 2023-05-06
cg_players: implement `trap_R_BatchInPVS` to test if a bunch of entities are in PVS in one call

illwieckz:
Use `trap_R_inPVSArray` to test array of entities in PVS in one call.

Game sidecar of:

- https://github.com/DaemonEngine/Daemon/pull/852

### DIFF
--- a/src/cgame/cg_players.cpp
+++ b/src/cgame/cg_players.cpp
@@ -3488,7 +3488,7 @@ void CG_PlayerDisconnect( vec3_t org )
 	}
 }
 
-centity_t *CG_GetLocation( vec3_t origin )
+centity_t *CG_GetPlayerLocation()
 {
 	std::vector<int> validEntities;
 	std::vector<std::array<float, 3>> posEntities;
@@ -3507,6 +3507,8 @@ centity_t *CG_GetLocation( vec3_t origin )
 		posEntities.push_back( posEntity );
 		validEntities.push_back( num );
 	}
+
+	const vec3_t& origin = cg.predictedPlayerState.origin;
 
 	std::vector<bool> inPVS = trap_R_BatchInPVS( origin, posEntities );
 
@@ -3535,14 +3537,6 @@ centity_t *CG_GetLocation( vec3_t origin )
 	}
 
 	return best;
-}
-
-centity_t *CG_GetPlayerLocation()
-{
-	vec3_t    origin;
-
-	VectorCopy( cg.predictedPlayerState.origin, origin );
-	return CG_GetLocation( origin );
 }
 
 void CG_InitClasses()
--- a/src/cgame/cg_players.cpp
+++ b/src/cgame/cg_players.cpp
@@ -3490,10 +3490,8 @@ void CG_PlayerDisconnect( vec3_t org )
 
 centity_t *CG_GetLocation( vec3_t origin )
 {
-	float     bestlen, len;
-
-	centity_t *best = nullptr;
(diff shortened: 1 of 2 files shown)


src/cgame/cg_ents.cpp
```diff
@@ -32,7 +32,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 over unused entities. It can be zero before the first frame starts adding
 entities, as none is added yet and none has to be removed yet. This should
 never reach the MAX_GENTITIES value. */
-int cg_highestActiveEntity = 0;
+size_t cg_highestActiveEntity = 0;
```
slipher, line 35: Let's keep this int as entity numbers are always stored as ints.

src/cgame/cg_players.cpp
```diff
@@ -3488,31 +3488,72 @@ void CG_PlayerDisconnect( vec3_t org )
 
 centity_t *CG_GetLocation( vec3_t origin )
 {
-	float     bestlen, len;
+	size_t activeCount = cg_highestActiveEntity - MAX_CLIENTS;
```
slipher, line 3491: This is off by 1 since cg_highestActiveEntity is an inclusive bound. But anyway the variable is not really needed; I suggest the following rewrite:

```
std::vector<std::array<float, 3>> positions;
std::vector<int> entities;

for ( size_t i = MAX_CLIENTS; i <= cg_highestActiveEntity; i++ )
{
    centity_t *eloc = &cg_entities[ num ];

    if ( !eloc->valid || eloc->currentState.eType != entityType_t::ET_LOCATION )
    {
        continue;
    }

    std::array<float, 3> pos;
    VectorCopy( eloc->lerpOrigin, pos );
    positions.push_back( pos );
    entities.push_back( i );
}
```

If you care about the vectors not being reallocated you can call `reserve(MAX_LOCATIONS)` on them.

src/cgame/cg_ents.cpp
illwieckz: Done.

src/cgame/cg_players.cpp
illwieckz: Done.

illwieckz:
I added a commit removing an useless vec3 copy and an useless function call:

Sometime I find things that are _very meh_:

20240510-112028-000 unvanquished-orbit

20240510-112030-000 unvanquished-orbit

src/cgame/cg_players.cpp
```diff
@@ -3488,10 +3488,14 @@ void CG_PlayerDisconnect( vec3_t org )
...
 centity_t *CG_GetLocation( vec3_t origin )
 {
-	float     bestlen, len;
+	static std::vector<int> validEntities;
+	validEntities.reserve( MAX_LOCATIONS );
+	validEntities.resize( 0 );
 
-	centity_t *best = nullptr;
-	bestlen = 3.0f * 8192.0f * 8192.0f;
+	// Shared with the engine.
```
slipher, line 3495: I find this comment confusing. It sounds like it means it is in shared memory or something

BTW why did you make the vectors `static`?

illwieckz: Cleaned-up.

illwieckz:
Las change just modified some comment/spacing and moved a comment to a separate commit (no functional change).

slipher:
LGTM

Issue #2663, opened 2023-04-29
Creep performance cost


illwieckz:
Rendering the creep costs more CPU time than rendering the 3D models.

Let's use this scene, which is very good for testing model or alien buildable rendering performance:

[![creep performance cost](https://dl.illwieckz.net/b/unvanquished/bugs/creep-performance-cost/unvanquished_2023-04-29_072259_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/creep-performance-cost/unvanquished_2023-04-29_072259_000.jpg)

You don't see it, but in fact from that point of view the renderer is rendering all alien buildables from the alien base:

[![creep performance cost](https://dl.illwieckz.net/b/unvanquished/bugs/creep-performance-cost/unvanquished_2023-04-29_072315_000.jpg)](https://dl.illwieckz.net/b/unvanquished/bugs/creep-performance-cost/unvanquished_2023-04-29_072315_000.jpg)

So, here is some screenshot of Orbit profiling Unvanquished while rendering this scene, more time is spent in `CG_Creep` (9.48%) than in `Tess_SurfaceIQM` (7.36%):

[![creep performance cost](https://dl.illwieckz.net/b/unvanquished/bugs/creep-performance-cost/20230429-072217-000.orbit-unvanquished-creep.png)](https://dl.illwieckz.net/b/unvanquished/bugs/creep-performance-cost/20230429-072217-000.orbit-unvanquished-creep.png)

In `CG_Creep` a lot of time is spent in `trap_R_AddPolyToScene` and `trap_CM_MarkFragments`.

slipher:
Measuring OpenGL functions in a CPU profiler tends to result in garbage data. The implementation apparently buffers commands and sends batches to the GPU at random times so the time spent in an OpenGL command has no correspondence to how much stuff you did in that command.

illwieckz:
That time spent has nothing to do with OpenGL. I should have not used the word “rendering” in my original message.

That time spent is, as usual, the time spent in doing a `trap` call.

illwieckz:
See my comment there about the cost of `trap_CM_MarkFragments` and `trap_R_AddPolyToScene` and how we may save time:

- https://github.com/DaemonEngine/Daemon/issues/846#issuecomment-1537131791

illwieckz:
[...]

PR #2681, opened 2023-05-06
Add option to disable marks, disable marks in `lowest` graphics profile

illwieckz:
- Add option to disable marks
- Disable marks in `lowest` graphics profile

Marks are very costly:

- https://github.com/Unvanquished/Unvanquished/issues/2663

Now I'm looking for a proper title for part where both particles and marks can be tweaked.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
@@ -224,8 +224,12 @@
 					<input cvar="r_ssao" type="checkbox" />
 					<p><translate>Simulate dark corners.</translate>&nbsp;<translate>Experimental.</translate></p>
 				</row>
-
-				<h2><translate>Particles</translate></h2>
+				<row>
+					<h3><translate>Marks</translate></h3>
+					<input cvar="cg_marks" type="checkbox" />
+					<p><translate>Draw marks (such as bullet holes or creep) on surfaces
+</translate></p>
+				</row>
 				<row>
 					<h3><translate>Particle physics</translate></h3>
 					<input cvar="cg_bounceParticles" type="checkbox" />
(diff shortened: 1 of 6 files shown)


slipher:
Are they really that costly if you have an actually bad GPU? Slow IPC operations can be a bottleneck if you have a fast GPU but with a slow GPU maybe rendering actually takes up most of the time and the IPC is not much slower on a weaker system.

illwieckz:
If you have a bad GPU the models may be animated on CPU instead of GPU so anything saving CPU time is welcome.

I only disabled that in `lowest`, lowest already has things like the disablement of sky, etc. That's the very last resort.

Those traps done for setting up marks are very costly in CPU time anyway, so I may disable marks myself even if I enable the ultra preset. If you mix marks + sky + firebomb spam the game can be very laggy (typical dretchstorm game). I have a good Ryzen and the game can jump from 800fps to 100 just by spawning some firebombs, even after having optimized the particles, so I would not want to be one using an entry level laptop.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
```diff
@@ -226,6 +226,11 @@
 				</row>
 
 				<h2><translate>Particles</translate></h2>
+				<row>
+					<h3><translate>Marks</translate></h3>
+					<input cvar="cg_marks" type="checkbox" />
+					<p><translate>Allow marks to be painted on surfaces.</translate></p>
```
slipher, line 232: How about examples to make this clearer. `Draw marks (such as bullet holes or creep) on surfaces`

slipher (commented):
Looks OK to me but it's blocked by #2682 aparently

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
illwieckz: Done.

Issue #2641, opened 2023-04-13
Make bots use beacons to seek enemy buildings


slipher:
Bots' omniscience of enemy buildings is quite annoying because as soon as you build a new structure that's closer to the enemy base than existing structures, bots will immediately swarm it. It would be nice if they used beacons instead when seeking rush targets etc., so that bots are a bit less cheaty and more fun to build against.

Two possible implementation strategies:

1. (Quick and dirty) When populating a bot's `closestBuildings` array, filter out enemy buildings for which there is no beacon. This would prevent bots from knowing about newly constructed structures, but they *would* still magically know when an enemy building has died.
2. (Correct but potentially difficult) Populate the entries for enemy buildings in `closestBuildings` with the beacons themselves. In this case bots would have no magical knowledge of enemy buildings. However, it might be difficult to implement as we would need to make sure bots can handle a beacon as a target everywhere which is likely to break some assumptions.

I lean towards implementation (1) since I believe it would give us 80% of the benefit (in terms of user experience) for 20% of the work.

PR #2683, opened 2023-05-08
Bots use beacons to locate enemy buildables

slipher:
Implement #2641 (the easy alternative <aoeu>#</aoeu>1).

Specifically, this is just for the `closestBuildings` array. There are other ways bots can locate buildables such as BotFindClosestEnemy (which requires being within sense range) that don't depend on this.

I noticed that due to their lower mobility, human bots can take several minutes to find the enemy base on plat23. So I imagine that on larger maps, it could be quite a long time before bots find the enemy base via random roaming. In case this is a problem, what if we give each team one freebie beacon at the start of the game pointing to the enemy's main buildable? It could be useful for flesh-and-blood players too to get oriented a bit quicker in an unfamiliar map. What do you think?

### DIFF
--- a/src/sgame/Beacon.cpp
+++ b/src/sgame/Beacon.cpp
@@ -49,6 +49,10 @@ along with Daemon.  If not, see <http://www.gnu.org/licenses/>.
 #define bc_etime time2
 #define bc_mtime apos.trTime
 
+static Cvar::Cvar<bool> g_mainBuildableStartingBeacons(
+	"g_mainBuildableStartingBeacons", "get beacon for enemy's main structure at match start",
+	Cvar::NONE, true);
+
 namespace Beacon //this should eventually become a class
 {
 	/**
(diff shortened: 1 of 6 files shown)


DolceTriade:
I think the default base beacon SGTM. We could cvar block it if its undesirable in some situations (mods/scenarios/custom layouts/etc)

sweet235:
Using this, we can actually beat the enemy by staring at them.

> what if we give each team one freebie beacon at the start of the game pointing to the enemy's main buildable?

This will fail as soon as the enemy base is moved.

slipher:
Added the starting beacons for enemy main buildables.

necessarily-equal (approved):
LGTM, haven't tested

slipher:
Is this OK to merge?

sweet235:
This gives us the cartesian product of possible bot bugs and beacon bugs to deal with if a bot fails to attack. I feel like there is at least one bug concerning missing beacons for enemy users/bots, but cannot reproduce it yet.

Bot attacks can be a decisive factor in 1v1 or 2v2. So I hesitate to approve this without scrutiny.

I am running this now.

slipher:
> I have yet to find out how to observe beacons without my observation influencing the beacons.


All right, I made it so that you can use noclip to look at beacons without affecting them.

sweet235:
LGTM

PR #2686, opened 2023-05-09
New entity iteration, replacing ForEntities (take 2)

slipher:
Replay of #2253.

`Avoid two Entity's with same number at one time` and `Add assertions against 2 Entity's with same ID` are new commits; the rest is the same.

Draft pull request to remind me to push the CBSE submodule before merging (no changes there).


src/sgame/sg_client.cpp
```diff
@@ -1396,6 +1396,7 @@ static void ClientSpawnCBSE(gentity_t *ent, bool evolving) {
 	Entity *oldEntity = ent->entity;
 	gclient_t *client = oldEntity->Get<ClientComponent>()->GetClientData();
 	class_t pcl = client->pers.classSelection;
+	delete oldEntity;
```
slipher, line 1399: This is not right

PR #2687, opened 2023-05-11
fix draw votes using the wrong cvar for time

cu-kai:
This fixes the vote handler using the `g_drawVotesPercent` instead of `g_drawVotesAfter` when calling the "draw" vote.


sweet235 (approved):

DolceTriade (approved):
oops, thanks

PR #2688, opened 2023-05-12
cg_main: do not capture mouse on loading screen

illwieckz:
There may be better ways to do this (why it is captured to begin with?).

There may also be other places where to do something like this (server connection screen, map download screen…).

One may also want to capture the mouse when the game is fullscreen as a hack to hide the system cursor when loading map in fullscreen, but that should be done after this.

The current code captures the mouse when loading maps to hide the system cursor, but that also captures the mouse move, and centers the mouse pointer… which is very bad and give birth to a full load of bugs when the game is windowed and especially not in foreground.

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1201,6 +1201,8 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 {
 	const char *s;
 
+	Rocket_SetActiveContext( KEYCATCH_UI );
+
 	CG_UpdateLoadingStep( LOAD_START );
 
 	// get the rendering configuration from the client system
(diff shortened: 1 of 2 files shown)


slipher:
This does not fix https://github.com/DaemonEngine/Daemon/issues/628.

illwieckz:
> There may also be other places where to do something like this (server connection screen, map download screen…).

It looks like this is already correct: it uses game pointer, so system pointer is hidden if the game is focused, but it looks like I can move the pointer outside of the game window.

src/cgame/cg_main.cpp
```diff
@@ -1201,6 +1201,8 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 {
 	const char *s;
 
+	Rocket_SetActiveContext( KEYCATCH_UI );
```
slipher, line 1204: The new version using Rocket_SetActiveContext doesn't work right. The call here at the beginning of CG_Init doesn't actually end up calling trap_SetMouseMode.

illwieckz:
Replaced by:

- https://github.com/Unvanquished/Unvanquished/pull/2695

PR #2690, opened 2023-05-12
fix several other voting issues following refactoring

cu-kai:
Companion to #2687

This fixes the vote handler using the `g_extendVotestime` instead of `g_extendVotesPercent` when calling the "extend" vote.

I have also fixed fillbots_{humans|aliens} not working, and made the vote text more descriptive.


slipher:
LGTM

PR #2693, opened 2023-05-13
Fix crash with TurrentComponent targeting

### LINKED ISSUE #2692: Server crashes
For reasons I do not yet know, current master leads to frequent server crashes. I do not good and bad commits, these are: good e5d57bc2e, bad 3145e8e78.

slipher:
This reportedly crashes when a targeted player leaves the team. Likely a regression in 59fb1ff18fd32aade0e79bb6d8a7f92202c71335.

Fixes #2692.


cu-kai (approved):
Seems to fix the crash when tested on Bunker server.

PR #2694, opened 2023-05-15
Add wall climbing for dretch and mantis bots

sweet235:
Make dretch and mantis bots able to follow upward navcons by using the alien wall climb. This works as follows: if a dretch or mantis bot is over an upward navcon, it enables the wall climb and goes up. Additionally, it saves the current time in its mind.

3 seconds after encountering the navcon, it will disable the wall climb. This has been observed to be useful if a particularly difficult wall geometry makes a bot loop on a wall. It will fall down eventually.

This can be seen in action on my server on the map thunder, near the default alien base.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -830,8 +830,20 @@ Global Bot Navigation
 =========================
 */
 
+static constexpr int WALLCLIMB_MAX_TIME = 3000;
+
 static Cvar::Cvar<int> g_bot_upwardNavconMinHeight("g_bot_upwardNavconMinHeight", "minimal height difference for bots to use special upward movement.", Cvar::NONE, 100);
 static Cvar::Cvar<int> g_bot_upwardLeapAngleCorr("g_bot_upwardLeapAngleCorr", "is added to the angle for mantis and dragoon when leaping upward (in degrees).", Cvar::NONE, 20);
+
+static void BotClimbToGoal( gentity_t *self )
+{
+	glm::vec3 ownPos = VEC2GLM( self->s.origin );
+	glm::vec3 dir = self->botMind->nav().glm_dir();
+	BotAimAtLocation( self, ownPos + 100.f * dir );
+	self->botMind->cmdBuffer.upmove = -127;
+	self->botMind->cmdBuffer.forwardmove = 127;
+	self->botMind->cmdBuffer.rightmove = 0;
+}
 
 void BotMoveUpward( gentity_t *self, glm::vec3 nextCorner )
 {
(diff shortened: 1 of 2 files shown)


src/sgame/sg_bot_nav.cpp
```diff
@@ -830,17 +830,37 @@ Global Bot Navigation
 =========================
 */
 
+static constexpr int WALLCLIMB_MAX_TIME = 3000;
```
bmorel, line 833: I'd be curious to play with variable to see how useful this mechanism is. Could it be possible to have it as as cvar?

src/sgame/sg_bot_nav.cpp
```diff
@@ -830,17 +830,37 @@ Global Bot Navigation
...
+
 static Cvar::Cvar<int> g_bot_upwardNavconMinHeight("g_bot_upwardNavconMinHeight", "minimal height difference for bots to use special upward movement.", Cvar::NONE, 100);
 static Cvar::Cvar<int> g_bot_upwardLeapAngleCorr("g_bot_upwardLeapAngleCorr", "is added to the angle for mantis and dragoon when leaping upward (in degrees).", Cvar::NONE, 20);
 
+static void BotClimbToGoal( gentity_t *self )
```
bmorel, line 838: I would just check for `BG_ClassHasAbility( pcl, SCA_WALLCLIMBER )` and return. This way, the code would be more generic, and this could be moved out of the switch. Notably, it would enable adv grangering when we'll have building bots.

illwieckz:
The hard-coded `WALLCLIMB_MAX_TIME` seemed a bit hacky at first, then I remember recast's navcon can only work on close navmeshes anyway… so that's likely fine. Otherwise we may want to evaluate the travel time from the distance and add a margin. This is not blocking though, I'll simply add a `HACK:` flag in the code with a comment.

bmorel:
all that stuff is really hackish anyway, but as many features as possible (regarding navigation) are needed before we can possibly cleanup the code to make it more reliable, and i.e. have it's efficiency scale with bot's level (no, a SK1 bot should not be able to travel as fast as a SK9 one while wallwalking or pouncing). This is "easy" to do, just apply some modifier, but it's polishing: we need things to actually work before going to that step.

src/sgame/sg_bot_nav.cpp
sweet235: Not really. That would work for dretch and advanced granger. The mantis, however, will have to decide if a leap or a wall climb is the way to go. Right now, it goes for the wall climb if over a navcon, and for the leap if attacking. That is not optimal.

src/sgame/sg_bot_nav.cpp
sweet235: We can do that later. I am already guilty of adding so many cvars that it is becoming increasingly difficult to set them up.

PR #2695, opened 2023-05-15
Don't grab cursor during loading if windowed

slipher:
Enhanced version of #2688 that doesn't show the cursor in full-screen mode.

According to my tests, the game already issues an appropriate call to trap_SetMouseMode when the loading ends (Deltas if focused or SystemCursor if not), so additional calls at the end of CG_Init are not needed.


illwieckz (approved):
I have the same patch in one of my local branches trying various ways to do this. 😁️

LGTM, then.

PR #2698, opened 2023-05-16
Reset bot stuck timer when attacking or standing on medistat

sweet235:
We can reduce the stuck timer in the .bt files to as low as 1 second if we merge this.

### DIFF
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -556,8 +556,6 @@ static void HArmoury_Use( gentity_t *self, gentity_t*, gentity_t *activator )
 	G_TriggerMenu( activator->client->ps.clientNum, MN_H_ARMOURY );
 }
 
-void BotResetStuckTime( gentity_t * );
-
 static void HMedistat_Think( gentity_t *self )
 {
 	int       entityList[ MAX_GENTITIES ];
@@ -709,10 +707,6 @@ static void HMedistat_Think( gentity_t *self )
 			{
 				BG_AddUpgradeToInventory( UP_MEDKIT, client->ps.stats );
 			}
-		}
-		else if ( player->r.svFlags & SVF_BOT )
-		{
-			BotResetStuckTime( player );
 		}
 	}
 	// we lost our target
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -556,6 +556,8 @@ static void HArmoury_Use( gentity_t *self, gentity_t*, gentity_t *activator )
 	G_TriggerMenu( activator->client->ps.clientNum, MN_H_ARMOURY );
 }
 
+void BotResetStuckTime( gentity_t * );
+
 static void HMedistat_Think( gentity_t *self )
 {
 	int       entityList[ MAX_GENTITIES ];
@@ -707,6 +709,10 @@ static void HMedistat_Think( gentity_t *self )
 			{
 				BG_AddUpgradeToInventory( UP_MEDKIT, client->ps.stats );
 			}
+		}
+		else if ( player->r.svFlags & SVF_BOT )
+		{
+			BotResetStuckTime( player );
 		}
 	}
 	// we lost our target
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1159,6 +1159,16 @@ AINodeStatus_t BotActionFollow( gentity_t *self, AIGenericNode_t *node )
 	return BotMoveToGoal( self ) ? STATUS_RUNNING : STATUS_FAILURE;
 }
 
+// TODO: Currently this only works for human bots. Make it work for alien bots too.
+static bool BotCurrentlyHealingWithDedicatedBuildable( gentity_t *self )
+{
+	if ( G_Team( self ) == TEAM_HUMANS && ( self->client->ps.stats[ STAT_STATE ] & SS_HEALING_2X ) )
+	{
+		return true;
+	}
+	return false;
+}
+
 static AINodeStatus_t BotActionReachHealA( gentity_t *self );
 static AINodeStatus_t BotActionReachHealH( gentity_t *self );
 AINodeStatus_t BotActionHeal( gentity_t *self, AIGenericNode_t *node )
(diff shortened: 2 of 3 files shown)


src/sgame/sg_buildable.cpp
```diff
@@ -708,6 +710,10 @@ static void HMedistat_Think( gentity_t *self )
 				BG_AddUpgradeToInventory( UP_MEDKIT, client->ps.stats );
 			}
 		}
+		else if ( player->r.svFlags & SVF_BOT )
+		{
+			BotResetStuckTime( player );
```
necessarily-equal, line 715: Can we put that in BotActionHeal though? Like after checking that we are healing on medistat or booster speed (`SS_HEALING_…`)

necessarily-equal (commented):
Thank you for doing that, I wanted to do a PR like that for long.

slipher:
> We can reduce the stuck timer in the .bt files to as low as 1 second if we merge this.

What do you mean by that?

src/sgame/sg_buildable.cpp
necessarily-equal: Here's an (untested) suggestion, to make implementing this easier.
```diff
diff --git a/src/sgame/sg_bot_ai.cpp b/src/sgame/sg_bot_ai.cpp
index 4947d8f67..a9730b764 100644
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1159,6 +1159,31 @@ AINodeStatus_t BotActionFollow( gentity_t *self, AIGenericNode_t *node )
 	return BotMoveToGoal( self ) ? STATUS_RUNNING : STATUS_FAILURE;
 }
 
+// Note: this function assumes that the goal is valid, and is a buildable
+static bool BotCurrentlyHealingWithDedicatedBuildable( gentity_t *self )
+{
+	switch( self->client->ps.stats[ STAT_STATE ] )
+	{
+		case SS_HEALING_2X:
+			// medistation healing if human, good
+			return G_Team( self ) == TEAM_HUMANS;
+		case SS_HEALING_8X:
+			// booster healing if alien, also good
+			return G_Team( self ) == TEAM_ALIENS;
+		case SS_HEALING_4X:
+		{
+			// for human, this is medkit healing, return false.
+			// for aliens, this can be creep healing, in which case
+			// this is good only if we are not aiming towards a booster.
+			return G_Team( self ) == TEAM_ALIENS
+				&& self->botMind->goal.getTargetedEntity()->s.modelindex == BA_A_BOOSTER;
+		}
+		default:
+			return false;
+	}
+}
+
 static AINodeStatus_t BotActionReachHealA( gentity_t *self );
 static AINodeStatus_t BotActionReachHealH( gentity_t *self );
 AINodeStatus_t BotActionHeal( gentity_t *self, AIGenericNode_t *node )
@@ -1197,6 +1222,12 @@ AINodeStatus_t BotActionHeal( gentity_t *self, AIGenericNode_t *node )
 		return STATUS_FAILURE;
 	}
 
+	// If we are properly healing, waiting without moving is good
+	if ( BotCurrentlyHealingWithDedicatedBuildable( self ) )
+	{
+		BotResetStuckTime( self );
+	}
+
 	if ( G_Team( self ) == TEAM_HUMANS )
 	{
 		return BotActionReachHealH( self );
```

I think those `SS_HEALING_[248]X` flags are a huge mess to be honest

necessarily-equal: Testing show that this doesn't work because the `SS_HEALING_4X` and `SS_HEALING_8X` flags are never set for an alien bot. For some reason (bot code executing before buildable code maybe?).

You can ignore this thread honestly, we can fix that later. Or you can integrate this with a comment on the 4X and 8X cases, because it works reliably for humans.

sweet235:
Depending on map and base location, bots may end up on top of crates or pipes frequently. They might even spawn there. On these places, there may or may not be a navcon for the bot's class. So we rely on the unstick mechanism to make them run/jump down and hope this will make them land on a navmesh.

Currently, we wait for 10 seconds before we trigger this. This is far too long because it makes the bot an easy kill. Even if not killed, the bot does not do anything useful in this time.

There are a few situations in which we _do_ want bots to stand still, among these are:

- A bot waiting for a door to open
- A bot healing using a buildable (medistat, booster, alien stuff)
- A bot attacking an enemy building

We do reset the stuck timer near the booster in the .bt files. This PR resets the stuck timer when healing on the medistat and when attacking.

Using this, we can reduce the stuck timer to one or two seconds. If we ensure that bots will move forward in the first few unstick seconds, they will handle doors too.

src/sgame/sg_buildable.cpp
sweet235: This is a better method indeed.

slipher:
1 second is way too early to start unstick maneuvers. This would trigger all the time in normal play, even in the middle of fights (current PR notwithstanding). This is not what you want since if stuck detection activates, it takes precedence over all other behaviors so the bot forgets its goals and can't fight or navigate. The unstick behavior is an emergency backstop to prevent the bot being sidelined for minutes at a time.

Regardless, the PR itself does make sense.

bmorel:
> Using this, we can trigger the unstick mechanism after one or two seconds.

I'm not convinced that 1 or 2 seconds are long enough, but it's not important as the number of occurrences and places where this happen has drastically reduced with autojump and the other support of navcons.
It's mostly useful on crates, and for that we could use much better ways to detect the need to move on or even the way where to move.
10s is definitely too much, though. I think my own experiments led me to a value around 5s, which was still slow but hey, better than "infinity 10s"

The main problem with the unstick code is that it's very naive, and should be implemented a better way. Detecting we're on a crate is trivial enough, so moving down shuold be easy, if there's no navcon saying otherwise.
As for the rest, it's naivety makes it so that bots never notice they're stuck in yocto's hole or karith's rock border.

> If we ensure that bots will move forward in the first few unstick seconds, they will handle doors too.

Could work yes, but same opinion as above: it's easy to detect a door is currently opening, and when it's wide enough. I would also like to improve that movers handling, as it has many problems (having to wait for door to be fully opened to update blockers being one of them, some automatic doors not being detected as such is another, and usstremor horribly slow doors are a pain by themselves).

sweet235:
The discussion is spiralling out of control. This PR does not even touch the unstick trigger time. But it does fix a bug where human bots trigger it while healing on the medistat.

Gireen (approved):
lgtm

PR #2699, opened 2023-05-18
Add time expressions for map rotation control

slipher:
@bmorel If you have a specific idea how the `weekday` should work feel free to suggest it.

### DIFF
--- a/src/sgame/sg_maprotation.cpp
+++ b/src/sgame/sg_maprotation.cpp
@@ -311,6 +311,38 @@ static bool ParseValue( const char *token, mrValue_t &result )
 	{
 		result = rand() & 1;
 	}
+	else if ( !Q_stricmp( token, "time" ) ) // 0000 to 2359
+	{
+		// We use UTC time since localtime doesn't work in NaCl and maprotations would mostly
+		// be used on hosted servers which are normally configured to UTC anyway.
+		qtime_t t;
+		Com_GMTime( &t );
+		result = t.tm_hour * 100 + t.tm_min;
+	}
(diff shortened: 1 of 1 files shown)


illwieckz:
`0` for Sunday is fine.

illwieckz:
I mean, starting with Sunday is fine.

Some people starts the week with Sunday but using `1` for Sunday, example:

- https://support.microsoft.com/en-us/office/weekday-function-60e44483-2ed1-439f-8bd0-e404c190949a

bmorel:
I can live with any format, this is not important. What is important is the feature itself. And that code is reasonably clean, ofc.

I do though, consider that having _only_ the weekday starting with 0 feels a bit inconsistent with the rest, but as said, I don't care much about that.

From what I've read, LGTM. This will be helpful for that campaign server I want to do, or for varying rotations a bit. But my server setup still needs lot of work, with all the navcons and layouts we now have, setting up a correct server became a lot less trivial in a year (not complex, per se, but many small tasks to do, over 300+ maps *yawn* but at least I have the list of maps in the rotation: next step, implement the conditions, map per map, for those ~80 maps. And then install the navcons, the layouts, and see how everything works together. Ah, and fix bots. OFC.)

illwieckz:
> I do though, consider that having only the weekday starting with 0 feels a bit inconsistent with the rest, but as said, I don't care much about that.

Same. Starting with `1` for weekday like monthday and yearmonth looks better to me.

bmorel:
Thanks, it's far from enough to me.

LGTM.

illwieckz:
> OK I made it 1 (Sunday) through 7 (Saturday)

Excellent!

PR #2701, opened 2023-05-19
translation: ignore breton for now, enable french

illwieckz:
- disable shipping breton as it's not ready yet
- enable shipping french as it starts to be usable
- the superfluous en.pot had been removed


slipher:
LGTM

PR #2702, opened 2023-05-19
cg_rocket_data_source: sanitize server data

illwieckz:
Always display a server and map name even if the server is doing hacky stuff, do not display negative player numbers.

- Strip non-printable characters in all case (prevents servers to cheat the list sorting with them).
- Strip spacing characters and colors in a temporary variable and if the stripped string is empty, display a fallback (for example display the server address if the server name is empty).

Before:

[sanitize server list](https://dl.illwieckz.net/b/unvanquished/wip/sanitize-server-list/unvanquished_2023-05-19_211658_000.png)

After:

[sanitize server list](https://dl.illwieckz.net/b/unvanquished/wip/sanitize-server-list/unvanquished_2023-05-19_211712_000.png)


Gireen (approved):
optional, instead of the address fallback it could be a generic text name like "Servername"

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -34,6 +34,47 @@ Maryland 20850 USA.
...
+		str.end() );
+}
+
+static std::string SanitizeString( std::string original, std::string fallback )
+{
+	std::string printable = original;
+	StripNonPrintingCharacters( printable );
+
+	std::string stripped = printable;
+	Color::StripColors( stripped );
```
slipher, line 67: You have to use the return value of StripColors

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -34,6 +34,47 @@ Maryland 20850 USA.
...
+}
+
+static void StripSpacingCharacters( std::string& str )
+{
+	str.erase(
+		std::remove_if(
+			str.begin(),
+			str.end(),
+			[]( unsigned char c ){
+				return std::isspace( c );
```
slipher, line 56: Don't use std::isspace because the behavior is not specified in different locales. Use Str::cisspace

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -34,6 +34,47 @@ Maryland 20850 USA.
 
 #include "cg_local.h"
 
+static void StripNonPrintingCharacters( std::string& str )
+{
+	str.erase(
```
slipher, line 39: What are you trying to do here? std::isprint is another function that should not be used since it doesn't have clearly specified behavior. Is it intended to delete all non-ASCII unicode?

src/cgame/cg_rocket_datasource.cpp
illwieckz: Something is probably wrong in that `StripColors` code, I first did that:

```
std:string stripped = Color::StripColors( original );
```

But then all servers were color-stripped.

It looks like `Color::StripColors` modifies the string in place and the return is useless.

If that's not expected, we have a bug.

src/cgame/cg_rocket_datasource.cpp
illwieckz: I'm looking for a way to filter out non-printable characters, things like ASCII `bell`, Unicode `ZWSP`, etc.

src/cgame/cg_rocket_datasource.cpp

src/cgame/cg_rocket_datasource.cpp
illwieckz: Well, it looks like it was a bug. Too bad I didn't kept the code triggering it (earlier version of this PR), as it was the exact opposite (and with GCC) of:

- https://github.com/DaemonEngine/Daemon/issues/839

The gcc bug I got with an early version of the commit was: input and ouput are both stripped.
The clang bug being: neither input and output are stripped.

illwieckz: Ah no I get it, my earlier code was doing:

```
std::string stripped = Color::StripColors( node->name );
```

But `node->name` is a `char*` and then the function modifies the input in place…

We have a very bad api on that part.

This is the function:

```

char *StripColors( char *string )
{
	char *out = string;

	for ( const auto& token : Parser( string ) )
	{
		Str::StringView text = token.PlainText();
		memmove( out, text.begin(), text.size() );
		out += text.size();
	}

	*out = '\0';
	return string;
}
```

GCC modifies both strings, Clang modifies none.

src/cgame/cg_rocket_datasource.cpp
slipher: We don't have an API for Unicode printable characters. A filter `c >=0 && c < 32` would work against ASCIIbetical sorting abuse at least

illwieckz: Well, I can keep all characters for the printed string, as sorting abuse can be taken care in sorting code.

We still need to strip the non-printing characters to detect empty server names only made of not-printable characters (what the current hackish server does if I'm right).

illwieckz:
> optional, instead of the address fallback it could be a generic text name like "Servername"

@Gireen the advantage of `???` is that we probably don't have to care about translating this.

src/cgame/cg_rocket_datasource.cpp
illwieckz: So, even if we don't take care about sortable strings, we still want to strip those non-printable characters here.

Reading this page: https://en.cppreference.com/w/cpp/string/byte/isprint
maybe we would want to use `std::setlocale` with `(unvanquished language).UTF-8`?

src/cgame/cg_rocket_datasource.cpp
illwieckz: Fixed.

src/cgame/cg_rocket_datasource.cpp
slipher: > Reading this page: https://en.cppreference.com/w/cpp/string/byte/isprint
> maybe we would want to use `std::setlocale` with `(unvanquished language).UTF-8`?

No. That API has no ability to handle Unicode and would just say (the constituent bytes of) any Unichode character is unprintable.

PR #2708, opened 2023-05-20
silence clang warning for bitwise instead of logical operators

### LINKED ISSUE #2652: sg_cmds.cpp: warning: use of bitwise '|' with boolean operands [-Wbitwise-instead-of-logical]
This is a warning I get with ICX 2023.1.0 (which is based on very recent LLVM clang):

```
src/sgame/sg_cmds.cpp:3160:9: warning: use of bitwise '|' with boolean operands [-Wbitwise-instead-of-logical]
        return Cmd_Sell_upgradeItem( ent, UP_LIGHTARMOUR ) |
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/sgame/sg_cmds.cpp:3160:9: note: cast one or both operands to int to silence this warning
src/sgame/sg_cmds.cpp:3160:9: warning: use of bitwise '|' with boolean operands [-Wbitwise-instead-of-logical]
        return Cmd_Sell_upgradeItem( ent, UP_LIGHTARMOUR ) |
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/sgame/sg_cmds.cpp:3160:9: note: cast one or both operands to int to silence this warning
src/sgame/sg_cmds.cpp:3160:9: warning: use of bitwise '|' with boolean operands [-Wbitwise-instead-of-logical]
        return Cmd_Sell_upgradeItem( ent, UP_LIGHTARMOUR ) |
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                           ||
src/sgame/sg_cmds.cpp:3160:9: note: cast one or both operands to int to silence this warning
src/sgame/sg_cmds.cpp:3254:10: warning: use of bitwise '|' with boolean operands [-Wbitwise-instead-of-logical]
                return Cmd_Sell_weapons( ent ) | Cmd_Sell_upgrades( ent );
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                               ||
src/sgame/sg_cmds.cpp:3254:10: note: cast one or both operands to int to silence this warning
```

_Edit_: I also get this with AOCC 4.0.0 which is based on Clang 14.0.6.

illwieckz:
Fixes #2652:

- https://github.com/Unvanquished/Unvanquished/issues/2652

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -2497,10 +2497,19 @@ static bool Cmd_Sell_upgrades( gentity_t *ent )
 
 static bool Cmd_Sell_armour( gentity_t *ent )
 {
+	#if defined(__clang_major__) && __clang_major__ >= 14
+		_Pragma("clang diagnostic push")
+		_Pragma("clang diagnostic ignored \"-Wbitwise-instead-of-logical\"")
+	#endif
+
 	return Cmd_Sell_upgradeItem( ent, UP_LIGHTARMOUR ) |
 	       Cmd_Sell_upgradeItem( ent, UP_MEDIUMARMOUR ) |
 	       Cmd_Sell_upgradeItem( ent, UP_BATTLESUIT ) |
 	       Cmd_Sell_upgradeItem( ent, UP_RADAR );
+
+	#if defined(__clang_major__) && __clang_major__ >= 14
+		_Pragma("clang diagnostic pop")
+	#endif
 }
 
 static bool Cmd_Sell_internal( gentity_t *ent, const char *s )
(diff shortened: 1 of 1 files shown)


slipher:
You should disable the warning in CMake instead.

illwieckz:
But what if someone writes `|` instead of `||` as a mistake one day?

slipher:
Hmm you know what, maybe this code actually is kind of bad and should be changed. Not because of anything with the `|` operator specifically, but rather because it's questionable for any arithmetic or bitwise operator to be used with side-effect-producing expressions on both sides. This is because it's unspecified whether the left or right side evaluates first, so the algorithm is different depending on the compiler's whim. Selling items in a random order is not a bug, but it's preferable not to have unnecessary randomness in the program. We should write it like `bool a = sideEffectProducingExpr1(); bool b = sideEffectProducingExpr2(); return a || b;`

illwieckz:
I prefer that indeed.

PR #2718, opened 2023-05-25
Glm extract

bmorel:
some safe commits extracted from #2547 .
Goal is to reduce possible git conflicts considering how long this is taking to me to debug that annoying NaN problem. It's likely I could get other "safe" commits from there, but it would trigger git conflicts, so I'd rather avoid it.
I also have other commits starting to instrument the code to hunt the NaN, not sure this should be added there as well.
If there could be an easy way to detect NaN usage (like signals or exceptions, or at least aborts) it would really be great, but for now I could not find anything usable.

### DIFF
--- a/src/sgame/BaseClustering.cpp
+++ b/src/sgame/BaseClustering.cpp
@@ -127,7 +127,7 @@ namespace BaseClustering {
 
 			// Trace from mean buildable's origin towards cluster center, so that the beacon does
 			// not spawn inside a wall. Then use MoveTowardsRoom on the trace results.
-			trace_t tr = G_RayTrace( VEC2GLM( mean->s.origin ), center, 0, MASK_SOLID, 0);
+			trace_t tr = G_PointTrace( VEC2GLM( mean->s.origin ), center, 0, MASK_SOLID, 0);
 			Beacon::MoveTowardsRoom(tr.endpos);
 
 			// Prepare beacon flags.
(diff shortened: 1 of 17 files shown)


bmorel:
About the RayTrace interface, it's also part of other changes I have in mind, to simplify a lot the tracing interface in many situations (passing less parameters and making it more obvious the trace's goal).
This one simply replaces the many trace calls that don't use mins and maxs. I believe it may be possible to expose simpler interfaces to trace calls that would possibly allow to grab some very minor performance boosts in addition to the readability (because less checks and calculations done), but my other changes in that regard are not really mature and might be more controversial.

slipher (commented):
G_RayTrace sounds a little funny since it evokes rendering. Maybe G_PointTrace? The idea is you're moving a point through space (as opposed to a box like CM_BoxTrace).

For a new tracing API maybe it would be better to use `gentity_t*` rather than the entity number? The entity number can be error-prone - we had a bug where someone thought 0 is a good default for passEntityNum but this made the trace pass through the player in the first client slot.

bmorel:
> G_RayTrace sounds a little funny since it evokes rendering. Maybe G_PointTrace? The idea is you're moving a point through space (as opposed to a box like CM_BoxTrace).

Sure, why not.

> For a new tracing API maybe it would be better to use gentity_t* rather than the entity number? The entity number can be error-prone - we had a bug where someone thought 0 is a good default for passEntityNum but this made the trace pass through the player in the first client slot.

That's a good point.

I also expect to get specialised APIs for class_t/buildable_t and stuff like that which are used _a lot_ by the AI code at least, so let's try to make this 1st one as best as possible, so that other ones can follow more easily when the time will come.
Thanks for your input.

bmorel:
> For a new tracing API maybe it would be better to use gentity_t* rather than the entity number? 

Reading at the code anew, I see that the number used here varies. Sometimes it's extracted from the `s` member, and sometimes from the `r` member (entityState vs entityShared I think?).
I think it's safe and a good idea to use a single one of those, but which one would be better, I'm not certain. Probably the shared one?

PR #2720, opened 2023-05-26
make the skillset feature opt-out

bmorel:
This works by adding a cvar named: g_bot_defaultSkill

If this cvar is empty, which is the default, behavior won't change. Otherwise, it must be filled with "key:value" pairs, separated by a space (' ').
When a bot is added that have a skill >= key, then it gets the "value" skill automatically.
There are sanity checks that prevents non-existing skills to be tried to be added.

In a reasonable futur, I'd like to add a "/bot listSkills" command that will list all available skills. This could be then used to build a GUI that helps building the cvar.

### DIFF
--- a/src/sgame/sg_bot_local.h
+++ b/src/sgame/sg_bot_local.h
@@ -127,6 +127,41 @@ enum bot_skill
 
 using skillSet_t = std::bitset<BOT_NUM_SKILLS>;
 
+class botSkillTreeElement_t {
+public:
+	Str::StringRef name;
+	bot_skill skill;
+	int cost;
+	// function that determines if this skill is available
+	std::function<bool(const gentity_t *self, skillSet_t existing_skills)> predicate;
+	// skills that are unlocked once you unlock this one
+	std::vector<botSkillTreeElement_t> unlocked_skills;
+
+	botSkillTreeElement_t(
+			Str::StringRef name_,
+			bot_skill skill_,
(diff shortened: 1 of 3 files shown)


bmorel:
I would not call this "clean" but considering the situation I don't think I can make clean _and_ simple to review.
This keeps the feature opt-in, I'll let someone more used to fiddle with skills build a sane default (@sweet235 maybe?) to have it opt-out.

necessarily-equal:
I think this is mostly a duplicate of https://github.com/Unvanquished/Unvanquished/pull/2697. The differences were:
* this PR uses a single cvar, I've updated https://github.com/Unvanquished/Unvanquished/pull/2697 to do that also because I think that is a good idea. This is not a difference anymore.
* this PR disables the skilltree if the list is set, while on mine the skilltree is still used if there is budget left. This makes it easier to disable the skilltree randomness (one doesn't have to set the budget to 0), but on the other hand it is less flexible because it doesn't allow mixed list+randomness.

I think #2697 is the way to go here.

bmorel:
I disagree with you here, but we should wait for the input of the person which did the skillset balancez maintenance since now around 6 months. That's @sweet235 .
I do have my server too, but the skillset and navcons and number of maps and rotation changes made it too complex to maintain so I don't consider this a working server. Not Yet. This PR is going to help on that, though.

Issue #2706, opened 2023-05-20
2023-05: Project direction; Urgent, and Important things to do 


illwieckz:
Explanation about this document:

The idea for this kind of issue is to cover a given time. Let's say we will do one every semester. Because this one is the first one, this one will last until next release. Then every semester we will close the current one and do a new one. In the next document, tasks done will be removed, new tasks may be added, importantness or urgentness may be re-evaluated, etc.

Right now, many of those issues are issues everyone waits on each others, so nothing moves forward. It happens that sometime things among the more important ones are also the ones with the least personal interest. So I encourage people to look at those issues even if that's not the topic they are naturally inclined to, otherwise some of them will never be fixed.

Things can be classified as urgent and non-urgent, important and non-important. Urgent is not something higher than Important. Those are different classifications. There can be non-urgent important things, and non-important urgent things.

Things are usually organized this way:

&nbsp;|urgent|non-urgent
-|-|-
important||
non-important||

To simplify things, I only focus on the important line for now.

----

Now the document:

## Important and Urgent

- [x] **writing:** write a blog post calling for translation contributions

- [ ] **writing:** gather the changes and start writing the next release blog post

This is both urgent and important, and one has to do it. I remind this is not an effortless task and this is a release blocker. One cannot complain the release is not done if that is not already done.

- [ ] **release:** get a release built and published within 2 months (before August 2023).

Because this will be a 3+ months release, this is only achievable if someone starts writing the blog post one or two months before (that means now, May 2023).

- [ ] **game:** https://github.com/Unvanquished/Unvanquished/issues/2700

- [x] **game:** UI for selecting the language

- [x] **game:** bad bot behavior related to skillset
This one is also considered a release blocker.

- [x] **engine:** https://github.com/DaemonEngine/Daemon/issues/798
We better want this for the next release, if we cannot have it, we should not wait that much.

- [x] **engine:** https://github.com/DaemonEngine/Daemon/issues/839 (Clang regression)
We better want this for the next release, if we cannot have it, we should not wait that much.

- [ ] **game:** https://github.com/Unvanquished/Unvanquished/issues/2682 (Clang regression)
We better want this for the next release, if we cannot have it, we should not wait that much.

About Clang regressions, we need to know if those are compiler a bugs. At least one of those bugs don't modify the variable that should be modified, so there may be many hidden bugs everywhere else if that's a compiler bug.

[...]

PR #2721, opened 2023-05-30
Revert "Rotate bounding box correctly in G_CM_LinkEntity"

bmorel:
This reverts commit ec4df1d0139e70e99b5251245c160dc697b45f8d.

Said commit introduces the regression which is talked about in https://github.com/Unvanquished/Unvanquished/issues/2706#issuecomment-1565340318

I only did quick testing (that I described in aforementioned discussion), and would appreciate if @cu-kai or @sweet235 could test it on more situations.

Just as a reminder:
`mission-one`, `/setviewpos -3950 2193 3; noclip; devteam a`

### DIFF
--- a/src/sgame/sg_cm_world.cpp
+++ b/src/sgame/sg_cm_world.cpp
@@ -444,16 +444,16 @@ void G_CM_LinkEntity( gentity_t *gEnt )
 	// set the abs box
 	if ( gEnt->r.bmodel && ( angles[ 0 ] || angles[ 1 ] || angles[ 2 ] ) )
 	{
-		glm::mat3 matrix = RotationMatrix( VEC2GLM( angles ) );
+		// expand for rotation
+		float max;
 
-		glm::vec3 mins = matrix * VEC2GLM( gEnt->r.mins );
-		glm::vec3 maxs = matrix * VEC2GLM( gEnt->r.maxs );
+		max = RadiusFromBounds( gEnt->r.mins, gEnt->r.maxs );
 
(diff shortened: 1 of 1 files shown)


sweet235:
I confirm that this fixes the problem on mission-edge too.

sweet235:
On the other hand, we are having a problem with the tremship door if we revert this, as described in https://github.com/Unvanquished/Unvanquished/pull/2280.

bmorel:
Well, I'd rather have an AI regression than a physics/collision regression, but yes, ideally we should properly implement the obstacle handling for func_rotating_doors.
My guess is, the problem in our obstacle handling is that we use `entityShared_t gentity_t::r;` instead of `entityState_t gentity_t::s;` (yes, I know, those names are really nicely chosen)

sweet235:
> Well, I'd rather have an AI regression than a physics/collision regression

A reasonable opinion.

> ideally we should properly implement the obstacle handling for func_rotating_doors

Maybe this is just about finding the right place to do this rotation calculation.

slipher:
https://github.com/Unvanquished/Unvanquished/commit/ec4df1d0139e70e99b5251245c160dc697b45f8d is in fact geometrically wrong. Not sure why I LGTMed it the first time :) To find the optimal bounds you would need to compute the mins/maxs over all 8 corners of the rotated box, unlike that code which just checks 2 of them.

slipher:
> To find the optimal bounds you would need to compute the mins/maxs over all 8 corners of the rotated box, unlike that code which just checks 2 of them.

I propose to do this in #2724

PR #2722, opened 2023-05-31
Skilltree refactoring commits (4/?)

necessarily-equal:
It is needed for https://github.com/Unvanquished/Unvanquished/pull/2697 and https://github.com/Unvanquished/Unvanquished/pull/2635. Although the commits are new, the code change themselves date from mid-april.

Next PR in this series: #2697


slipher:
LGTM

bmorel:
so, we're going to merge more experiments with a release approaching soon and the AI problem considered blocking?
I think it is bad idea.

necessarily-equal:
Oh I also wanted to say that if we want to make the skill list per-level for the release, then this is what really matters for this release and this is what should be heavily tested.

bmorel:
I see my disagreement was ignored. Noted.

PR #2724, opened 2023-06-01
Fix bugs with rotated bbox and custom vote commands

slipher:


DolceTriade:
LGTM. Could you add more detail aroudn what's wrong in the bugged vrsion in the commit message?

bmorel:
on this branch, running `/setviewpos -3950 2193 3; noclip; devteam a`  no longer works.

slipher:
> Could you add more detail aroudn what's wrong in the bugged vrsion in the commit message?

Done.



> on this branch, running `/setviewpos -3950 2193 3; noclip; devteam a` no longer works. This is how I test the physics problem on mission-one, as said is other places.

I wasn't sure whether to read this as saying the bug is fixed or not fixed, but I tried it and the PR does fix the bug which can be seen in 0.54.

bmorel:
@slipher pasting this set of commands in the console normally sends the player to a specific position, which allows to check quickly if the collision problem still happens. The line does not works on your branch.

That was just a quick test, and it at least seems to bring this behavior change, that I would consider a regression. I have not tested further, being busy on other things.

slipher:
> @slipher pasting this set of commands in the console normally sends the player to a specific position, which allows to check quickly if the collision problem still happens. The line does not works on your branch.

Those commands appear to work fine for me. Please be more specific than "does not works"

bmorel:
I'm not teleported to the correct place, which is on one of the rotating things that had the bug.

slipher:
It worked for me. As I mentioned above, I successfully tested the rotating thing with a `map-mission-one-b7_0.0.2.dpk`

bmorel:
weird, I'm certain I got all submodules updated, and tried several times. I'll give it another shot in few hours and will try to dig more if I still have a problem.

sweet235 (approved):

bmorel:
So, checked anew. I could not see blocking problems with rotators or func_door_rotating, so the physics and the bots seem to work.

What seems new to me is that using `noclip` while being in a team collides with rotators, now. And maybe this is why the reproductor failed, maybe `/setviewpos` fails if it would send the client inside geometry, and in combination with `/noclip` it makes the command I used "not work".
In any case, this new behavior exist before those commits.

PR #2726, opened 2023-06-07
Invalidate GentityRef on player death

slipher:
The existing behavior is to be invalidated only when the client disconnects. The existing behavior turned out to be good enough to prevent some crashes (by using GentityRef instead of gentity_t *), but I always intended it to be invalidated on player death.


sweet235:
LGTM

PR #2727, opened 2023-06-08
human bots see an enemy: they shoot at it

sweet235:
I am convinced that currently, bots will only pick an enemy as target if the enemy's position can be projected downwards onto the navmesh. This does affect human bots in particular. Build some alien building over an unreachable navmesh part, or over no navmesh part: they will not attack it.

This PR relaxes this condition. Human bots are allowed to pick a target that is not over the navmesh. They will stand still and fire at it, as long as it is visible. Human bots can always do that reasonably, as they have the blaster for targets too far away for the primary weapon. If the target is not a building, and moves onto the navmesh, they proceed as usual.

Here is a video illustrating the problem: https://unvanquished.greboca.com/F-A-I-L.html . That place is not reachable for human bots. So the aliens were perfectly safe there. With this PR, a human bot will fire at them if it sees them.

Recently, I spent some time making alien bots go to previously unreachable places (with navcons provided). This PR makes human bots shoot at previously unseen places, for symmetry.

Edit:
That video is a bad example, as human bots might not see the aliens there because of the reactor. I hope you get the idea though.

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -806,6 +806,8 @@ static bool TargetInOffmeshAttackRange( gentity_t *self )
 		return distSquare < Square( ( 100 * 8192 ) / CHAINGUN_SPREAD );
 	case WP_MACHINEGUN:
 		return distSquare < Square( ( 50 * 8192 ) / RIFLE_SPREAD );
+	case WP_FLAMER:
+		return distSquare < Square( 250 ); // determined by trial and error
 	default:
 		return true;
 	}
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -788,6 +788,29 @@ AINodeStatus_t BotActionSay( gentity_t *self, AIGenericNode_t *node )
 	return STATUS_SUCCESS;
 }
 
+// Check if a human bot should switch to blaster to attack an offmesh target.
+// TODO: Add different ranges for buildings vs moving targets.
+//       If the target can move, we could allow a higher range for spready guns.
+// This function has a small overlap with BotTargetInAttackRange.
+static bool TargetInOffmeshAttackRange( gentity_t *self )
+{
+	float distSquare = DistanceToGoalSquared( self );
+	switch ( BG_PrimaryWeapon( self->client->ps.stats ) )
+	{
+	case WP_HBUILD:
+	case WP_PAIN_SAW:
+		return false;
+	case WP_SHOTGUN:
+		return distSquare < Square( ( 50 * 8192 ) / SHOTGUN_SPREAD );
+	case WP_CHAINGUN:
+		return distSquare < Square( ( 100 * 8192 ) / CHAINGUN_SPREAD );
+	case WP_MACHINEGUN:
+		return distSquare < Square( ( 50 * 8192 ) / RIFLE_SPREAD );
+	default:
+		return true;
+	}
+}
+
 // TODO: Move decision making out of these actions and into the rest of the behavior tree
 AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 {
@@ -813,7 +836,12 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 		return STATUS_SUCCESS;
 	}
 
-	if ( !mind->nav().havePath )
+	if ( !mind->nav().havePath && !mind->hasOffmeshGoal )
+	{
+		return STATUS_FAILURE;
+	}
+
+	if ( mind->hasOffmeshGoal && !BotTargetIsVisible( self, self->botMind->goal, MASK_OPAQUE ) )
 	{
 		return STATUS_FAILURE;
 	}
@@ -861,6 +889,47 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 	}
 
 	// We have a valid visible target
+
+	if ( mind->hasOffmeshGoal )
+	{
+		// The target is visible, but not on the navmesh
+		ASSERT( G_Team( self ) == TEAM_HUMANS );
+		if ( !BotTargetIsVisible( self, self->botMind->goal, MASK_SHOT ) )
+		{
+			return STATUS_SUCCESS;
+		}
+		bool inRange = TargetInOffmeshAttackRange( self );
+		bool couldChangeToPrimary = BG_GetPlayerWeapon( &self->client->ps ) == WP_BLASTER && !WeaponIsEmpty( BG_PrimaryWeapon( self->client->ps.stats ), &self->client->ps );
+		if ( !inRange && BG_GetPlayerWeapon( &self->client->ps ) != WP_BLASTER )
+		{
+			G_ForceWeaponChange( self, WP_BLASTER );
+		}
+		else if ( inRange && couldChangeToPrimary )
+		{
+			G_ForceWeaponChange( self, WP_NONE );
+		}
+
+		// check if the target has moved onto the navmesh
+		if ( mind->nav().directPathToGoal )
+		{
+			// target is on a navmesh now, switch to primary weapon if we have
+			// the blaster selected (the onmesh target code below assumes this)
+			if ( couldChangeToPrimary )
+			{
+				G_ForceWeaponChange( self, WP_NONE );
+			}
+			self->botMind->setHasOffmeshGoal( false );
+		}
+		else
+		{
+			BotAimAtEnemy( self );
+			BotStandStill( self );
+			BotFireWeaponAI( self );
+			return STATUS_RUNNING;
+		}
+	}
+
+	// The target is visible, and on the navmesh
 
 	bool inAttackRange = BotTargetInAttackRange( self, self->botMind->goal );
 	self->botMind->enemyLastSeen = level.time;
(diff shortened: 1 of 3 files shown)


sweet235:
From private communication, rephrased: the [modest] code duplication from `BotTargetInAttackRange` to `TargetInOffmeshAttackRange` is ugly.

That is correct. However, `BotTargetInAttackRange` decides if and where a bot should _move_. `TargetInOffmeshAttackRange` only decides if a human bot should switch to the blaster.

necessarily-equal:
I do not have time to review or test this, but I just wanted to say thank you for tackling this. :smiley:

src/sgame/sg_bot_ai.cpp
```diff
@@ -813,7 +836,12 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 		return STATUS_SUCCESS;
 	}
 
-	if ( !mind->nav().havePath )
+	if ( !mind->nav().havePath && !mind->hasOffmeshGoal )
+	{
+		return STATUS_FAILURE;
+	}
+
+	if ( mind->hasOffmeshGoal && !BotTargetIsVisible( self, self->botMind->goal, MASK_OPAQUE ) )
```
slipher, line 846: Why is MASK_OPAQUE used here and MASK_SHOT in the other place?

src/sgame/sg_bot_local.h
```diff
@@ -179,6 +179,9 @@ struct botMemory_t
 	int lastNavconTime;
 	int lastNavconDistance;
 
+	void setHasOffmeshGoal( bool val ) { hasOffmeshGoal = val; }
```
slipher, line 182: Do we really need this variable? Why can't we just have action fight give up whenever the target is out of sight AND there is no path?

src/sgame/sg_bot_ai.cpp
```diff
@@ -788,6 +788,29 @@ AINodeStatus_t BotActionSay( gentity_t *self, AIGenericNode_t *node )
 	return STATUS_SUCCESS;
 }
 
+// Check if a human bot should switch to blaster to attack an offmesh target.
+// TODO: Add different ranges for buildings vs moving targets.
+//       If the target can move, we could allow a higher range for spready guns.
+// This function has a small overlap with BotTargetInAttackRange.
+static bool TargetInOffmeshAttackRange( gentity_t *self )
```
slipher, line 795: Maybe you can extract like a GetEffectiveWeaponRange function out of BotTargetInAttackRange?

src/sgame/sg_bot_local.h
sweet235: Currently, they will move into the direction of an invisible target even without a path, until `g_bot_chasetime` seconds pass. In the case of mantis, marauder, dragoon this might trigger an upward attack, which can make the bot reach the target. In the other cases, the bot will usually just run against a wall.

This is the current behavior, and I don't want to be responsible for changing it. So I introduced this variable. It is only set for human bots. Maybe it would be better to split the attack action into one function for each team.

slipher: OK so the current algorithm is to give up the target if it has not been visible for g_bot_chasetime, great. So why can't we just keep the current algorithm for everything and just add the ability to set things without a path as the target?

bmorel: How would that work with aliens which only have melee?
I suspect that if one did the same with aliens as you seem to suggest, we'd soon have dragoon that constantly stuck themselves in places they don't know how to escape from, waiting for a prey to unstuck them.
One would also need a trace for aliens, as their "missiles" (their body) is considerably bigger than any human weapon, so they might be able to see something they don't know how to reach, while being truly unable to attack it still.

The current algorithm is also well known to make bots jump in death pits because they ignore navmesh while fighting their target. If I understand correctly what you suggest, your idea would increase that problem, unlike this PR which makes human bots stand by when they're attacking offmesh targets.

src/sgame/sg_bot_ai.cpp
bmorel: if we really wanted to make things clean on the weapon range codes (this is not the only place), they we'd need to move those values from hard-coded into the weaponAttributes_t, so that we'd be able to get rid of that kind of mess everywhere (same goes for obituaries actually).
I remember doing efforts in that direction a while ago, but it was not well received...

src/sgame/sg_bot_local.h
sweet235: > So why can't we just keep the current algorithm for everything and just add the ability to set things without a path as the target?

That would make them move forward when attacking an enemy without a path. But this will often make the bot move to a place where it cannot see the enemy anymore.

src/sgame/sg_bot_ai.cpp
```diff
@@ -862,6 +892,47 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
...
+		ASSERT( G_Team( self ) == TEAM_HUMANS );
+		if ( !BotTargetIsVisible( self, self->botMind->goal, MASK_SHOT ) )
+		{
+			return STATUS_SUCCESS;
+		}
+		bool inRange = TargetInOffmeshAttackRange( self );
+		bool couldChangeToPrimary = BG_GetPlayerWeapon( &self->client->ps ) == WP_BLASTER && !WeaponIsEmpty( BG_PrimaryWeapon( self->client->ps.stats ), &self->client->ps );
+		if ( !inRange && BG_GetPlayerWeapon( &self->client->ps ) != WP_BLASTER )
+		{
+			G_ForceWeaponChange( self, WP_BLASTER );
```
slipher, line 907: Is there some code to ensure the bot switches back to the primary weapon after finishing with the off-mesh target?

src/sgame/sg_bot_ai.cpp
sweet235: I'm not entirely sure. I went for MASK_OPAQUE first, because this is used nearby. But it turns out that this will make bots fire at their own buildings if these are in the line of sight. I forgot to change this occurance.

src/sgame/sg_bot_ai.cpp
sweet235: There is now.

sweet235:
A few remarks:

### Primary weapon vs blaster

Our current code (without this PR) ensures that if a human bot has the blaster active, the primary weapon is empty or a ckit. This PR allows for activating the blaster when the primary weapon is not empty but has too short a range, thus invalidating this invariant. The resulting confusion (albeit managable) was enough to make me introduce a serious bug where bots would not switch back to the primary weapon in some cases.

That bug is fixed now. A bot might walk around holding a blaster, but switch back to the primary weapon when seeing the next target. I find this to be funny, as a bot with a blaster is not always cheap food anymore. But it is a little ugly.

### Seperate code path for offmesh goals

This PR makes a bot remember if the goal it picked has not been reachable via navmeshes. Quite correctly, it has been remarked that this is a hack. Currently, if a bot's goal leaves the navmesh-reachable space, the bot will try to chase it for a few seconds. Especially for alien bots, this can succeed frequently.

The case of a human bot shooting an offmesh goal is different. Moving into any direction can make the goal invisible, and the bot has no idea if and when this will happen. So I made it stand still as long as the goal is still not navmesh-reachable. If the goal is a player and not visible anymore, the bot will not attempt to chase them.

### Code duplication

There is still moderate code duplication between `BotTargetInAttackRange` and the local `TargetInOffmeshAttackRange`. I'm still contemplating if having different ranges for different attack modes is a good or a bad thing. For example: when shooting a moving target, a weapon's spread hurts a little less than when shooting a building.

bmorel:
> Code duplication

Paying a close attention to the values, this is *not* a duplicate. Values are different.
The original close code snippet uses 50, 100 and 60 multiplyers of 8192 for shotgun, rifle, chaingun.
The not-so-duplicated code uses 50, 50, and 100 for shotgun, rifle, chaingun.

For flamer, the original code takes a complete algorithm while new proposal provides a hand-crafted value. I have to say, I'm unsure why flamer would be so complex to estimate a rough range and prefer the hard-coded one even if it might be less accurate (remains to be seen, bots never were strong with flamer).

It have to be said that I have no idea why the values 8192, 50, 60, and 100 are used or where they come from.

On the topic of having separate score functions, I do not know. I know that considering my game style and aiming, I'm better with a smg or shotgun in hands than a mass driver, but I don't really think we're ready for that kind of decision from bots, so different functions seems a bit useless to me.
Besides, the problem of spray is not really firing at something immobile, but at target that are big enough considering the distance to have good chances to deal them serious damages.
This is taken in consideration in none of the duplicated codes.

bmorel:
Also note we have no clue where the original values come from. I don't know if the "manual guess" comment for flamer apply to all, but if yes, at least we have some info on how to get something better if needed to, while not risking to break code we are not entirely certain why it's written the way it is.

Still LGTM.

necessarily-equal:
IMHO, it's worth merging with or without that ~deduplication. I'd rather have the fix that not.

bmorel:
I'd like to note that this is running on my server too, now.

src/sgame/sg_bot_ai.cpp
sweet235: There is not really code duplication, but rather a conceptual overlap. I do not think this should be blocking, as it is commented and not hidden in any way. I agree that we should have a closer look at BotTargetInAttackRange, but let's do this in a seperate PR.

sweet235:
I'll merge this as it is if there is no objection.

src/sgame/sg_bot_ai.cpp
slipher: That's OK but please add a comment about why the ranges might need to be different.

PR #2731, opened 2023-06-19
Quote() constness and cleanup

bmorel:
Quote() function iterates over 4 buffers for no reason. Get rid of that.
Also use a std::string to store the cached data, which prevents to meet too long strings.
Last, make it's returned value const, as it should.

### DIFF
--- a/src/sgame/sg_utils.cpp
+++ b/src/sgame/sg_utils.cpp
@@ -943,15 +943,17 @@ std::string G_EscapeServerCommandArg( Str::StringRef str )
 // Escape a command for used in server commands (sent from client to server)
 // Difference from Cmd::Escape and normal command parsing is that newlines are allowed
 // (for commands that have multi-line output)
-char *Quote( Str::StringRef str )
+// WARNING: this function can only be called BUFFERSIZE times by the same function,
+// or bad things will happen!
+const char *Quote( Str::StringRef str )
 {
-	static char buffer[ 4 ][ MAX_STRING_CHARS ];
+	int constexpr BUFFERSIZE = 256; // must be a power of 2
+	static std::string buffer[BUFFERSIZE];
 	static int index = -1;
 
-	index = ( index + 1 ) & 3;
-	Q_strncpyz( buffer[ index ], G_EscapeServerCommandArg( str ).c_str(), sizeof( buffer[ index ] ) );
-
-	return buffer[ index ];
+	index = ( index + 1 ) & ( BUFFERSIZE - 1 );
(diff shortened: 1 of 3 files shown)


slipher:
Multiple buffers are needed in places such as this.
```
			ADMP( va( "%s %s %s %d", QQ( N_("^3builder:^* $1$ built by $2$^*, buildlog #$3$") ),
				  Quote( buildingName ), Quote( builder ), MAX_CLIENTS + level.buildId - i - 1 ) );
```

slipher:
> As in, the order in which parameters expression are resolved is up to the compiler to decide.

That's correct. But the code above still works correctly regardless of which one is evaluated first.

bmorel:
This is quite dangerous though, as in, it means you can't call Quote() more than 4 times this way, and there's nothing explicitly saying that limitation or other safeguard.

I wonder if we could improve on this?

src/sgame/sg_utils.cpp
```diff
@@ -943,15 +943,17 @@ std::string G_EscapeServerCommandArg( Str::StringRef str )
 // Escape a command for used in server commands (sent from client to server)
 // Difference from Cmd::Escape and normal command parsing is that newlines are allowed
 // (for commands that have multi-line output)
-char *Quote( Str::StringRef str )
+// WARNING: this function can only be called BUFFERSIZE times by the same function,
+// or bad things will happen!
+const char *Quote( Str::StringRef str )
 {
-	static char buffer[ 4 ][ MAX_STRING_CHARS ];
+	int constexpr BUFFERSIZE = 256; // must be a power of 2
```
slipher, line 950: Surely 8 or 16 is enough? Also BUFFERSIZE is not a good name as this is a number of buffers, not the size of a buffer.

bmorel: I wrote this after talking with @cu-kai on IRC, and it seems he needs to run this function in a loop with MAX_CLIENTS iterations.

> Also BUFFERSIZE

Have to admit I didn't took much time at finding a variable name. Initially I wanted to get rid of that limit entirely, perhaps using a vector, but my short tinkering didn't let me find a satisfying solution (looked uglier than current).

Tbh I wonder how much this stuff makes sense.

slipher: You should never store the pointer returned by this in a variable. The max limit should just accommodate when you use several of them in a single line like the print(%s %s %s) things where each argument is quoted.

bmorel:
superseeded by #2735

PR #2734, opened 2023-06-19
AI: aim at center of enemy if not a human

### LINKED ISSUE #2733: Bots miss half their shots against tyrants with weapon that have spread
As seen here bots aim at the model origin, which for rants is near the floor, resulting in them missing half their shots with weapons that have spread (rifle, shotgun, chaingun):
unvanquished_2023-06-19_211837_000

bmorel:
Previously, code was only aiming at origin, which make weapons with spread much less efficient.
It might be desirable to make bots aim at the high of an enemy's BBox (the head) in more cases, though, to reduce shootings at friends when targeting bigger enemies (tyrants, overmind, reactor, armoury).

Fix #2733

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1467,15 +1467,16 @@ glm::vec3 BotGetIdealAimLocation( gentity_t *self, const botTarget_t &target, in
 	//this retrieves the target's species, to aim at weak point:
 	// * for humans, it's the head (but code only applies an offset here, with the hope it's the head)
 	// * for aliens, there is no weak point, and human bots try to take missile's speed into consideration (e.g. for luci)
-	if ( !isTargetBuildable && targetTeam == TEAM_HUMANS )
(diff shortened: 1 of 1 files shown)


src/sgame/sg_bot_util.cpp
```diff
@@ -1467,15 +1467,17 @@ glm::vec3 BotGetIdealAimLocation( gentity_t *self, const botTarget_t &target, in
...
-		{
-			aimLocation[2] += targetEnt->r.maxs[2] * 0.85;
-		}
-
+		aimLocation[2] += targetEnt->r.maxs[2] * 0.85;
+	}
+	else
+	{
+		// aim at middle
+		float height = targetEnt->r.maxs[2] - targetEnt->r.mins[2];
```
slipher, line 1479: mins and maxs are origin-relative. Do
`aimLocation[2] += 0.5f * (mins[2] + maxs[2])`

bmorel: I have not verified, just integrated your change.

slipher:
LGTM

bmorel:
Gave this a last round of testing, works as expected.

PR #2737, opened 2023-06-21
Make bot skills non-random by default

slipher:
Stacks on #2736; only the last commit is new.

Based on https://wiki.unvanquished.net/wiki/Bot_design#Distributing_Skills_over_Bot_Levels but with some changes to avoid setting skills while lacking a dependent skill.


necessarily-equal (approved):
LGTM

PR #2738, opened 2023-06-21
do not regenerate if there is OM

bmorel:
Aliens regenrating when OM and, usually, whole base, is down is mostly used to troll the human team by making games needlessly long. This patch attempts to fix that.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/server_setup.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/server_setup.rml
@@ -63,6 +63,19 @@
 				<input cvar="g_humanAllowBuilding" type="checkbox" />
 				<h3><translate>Allow human building</translate></h3>
 			</row>
+			<row>
+				<h3><translate>Alien regeneration conditions</translate></h3>
+				<input type="range" min="0" max="2" step="1" cvar="g_alien_reqOMToRegen" />
+				<p>
+					&nbsp;&nbsp;<translate>Current:</translate><inlinecvar cvar="g_alien_reqOMToRegen" type="number" format="%.0f"/><br/>
+					<translate>
+						&nbsp;&nbsp;&nbsp;0: always<br/>
+						&nbsp;&nbsp;&nbsp;1: requires either an egg or an overmind (active)<br/>
+						&nbsp;&nbsp;&nbsp;2: requires either a granger, an egg or an overmind (active)<br/>
+					</translate>
+					<ilink onclick='Events.pushevent("exec reset g_alien_reqOMToRegen", event)'> (reset) </ilink>
+				</p>
+			</row>
 
 			<h2><translate>Physics</translate></h2>
 			<row>
--- a/src/sgame/sg_active.cpp
+++ b/src/sgame/sg_active.cpp
@@ -31,6 +31,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 bool ClientInactivityTimer( gentity_t *ent, bool active );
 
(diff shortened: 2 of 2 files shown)


cu-kai:
Perhaps health regeneration could be disabled only if there's no hope of rebuilding the OM, such as a situation where there is no granger, no overmind and no spawns?

DolceTriade:
I would be ok with no eggs + no OM == no health regen. health regen can start when OM is up again.

bmorel:
I have to admit, I was not expecting that much in favor, considering how strict it is.
I think adding an active egg to the list of requirements makes sense, too.

I'll improve by making it an option, which would default on current behavior, and a GUI option for solo players to configure their games locally.
I Guess the values this option could takes would be:

* traditional (as we are currently)
* no builder + no egg + no OM
* no egg + no OM
* no OM (but I now agree this one would be very strict)

I think that should be ok?

slipher:
I think this would have very little use in a solo game and should be omitted from the GUI to avoid clutter.

Gireen:
I think this change robs Unvanquished of a fun element of its gameplay, to solve a problem where there is none.
The hunt for the last remaining alien is exiting and not perceived as getting trolled. 
It creates situations that are some of the most atmospheric and memorable of the game.

Disabling regeneration makes it near impossible for a lone alien to still win the game with skilled maneuvers. 
At the same time it does not prevents but even encourages an alien to just running away and hide.

VReaperV:
> I think this change robs Unvanquished of a fun element of its gameplay, to solve a problem where there is none.
> The hunt for the last remaining alien is exiting and not perceived as getting trolled.
> It creates situations that are some of the most atmospheric and memorable of the game.

I've never once seen a situation where it is fun. It's always just a boring waste of time either looking for a dretch hiding in some vent or a mara/goon running around.

> Disabling regeneration makes it near impossible for a lone alien to still win the game with skilled maneuvers.

It's near impossible regardless though.

> At the same time it does not prevents but even encourages an alien to just running away and hide.

The point is that you can still hit the alien, it's just not gonna regenerate all the lost health right after.

bmorel:
> I worry about adding too much configurability before there is demand for it, but I agree in principle about one of the top three options.

I'll do that.

> I think this would have very little use in a solo game and should be omitted from the GUI to avoid clutter.

I am talking about the server config GUI, which already allows fine tuning of few things. I think it would not hurt to have this option in there.

About "robs the game"... well, this mechanics robs the time of many players, who then quits. I've seen many quitting in the middle of such boring hunt.

slipher:
> > I think this would have very little use in a solo game and should be omitted from the GUI to avoid clutter.
> 
> I am talking about the server config GUI, which already allows fine tuning of few things. I think it would not hurt to have this option in there.

My point is that the server config GUI generally cannot be used to configure online servers, so having the option there is not useful. Having options that are not useful makes the menu less useful since it gets harder to find the important ones.

bmorel:
Updated the PR so that it gives the 3 options, defaulting to current behavior:

* always regenerate
* regen if there's any of egg, OM
* regen if there's any of egg, OM, builder

Now, to be honest, my tests shows a bug in the GUI: the `reset` button does not works, I have no idea why.
Here is a screenshot of the GUI though (with red rectangle to the new area):



I did some quick testing locally, but I will deploy this with the strictest configuration on my own server on next update (and use a name that does not implies testing is in progress as I currently do)

bmorel:
About the fact that GUI can't be used by remote admins, it's true, but it's also something I *want* to be fixed someday.
I think it's also important to keep such important options easy to tweak for solo/LAN players, and the GUI is definitely useful here, even if imperfect.
OTOH since I currently have a small problem with the GUI, I could also do the lazy thing and remove that part, but I think it'd be a waste (would be better to understand why the stupid reset button does not works, right?).

PR #2740, opened 2023-06-24
Don't print '5 minutes remaining!' as a plural translation.

cu-kai:
This fixes the time warning only showing '5' on the CenterPrint message.


bmorel:
Tested. This works.

bmorel (approved):

slipher:
Apparently the number is supposed to go before the message also like `cp_tr_p 5 "blah $1$" 5`. But anyway you still can't really do it right because there is no translation marker for a deferred plural translation. We would need to add something like `NP_("singular string", "plural string")` that always returns the first one.

LGTM

bmorel:
which is why the come have different handle for when the remaining time is 1 minute. This also fixes a bug we noticed recently, that is, that this code prints only "5".

PR #2744, opened 2023-06-27
Cleanup

bmorel:
This cherry-picks some cleanup commits of #2539 which requires conflict solving. Those commits don't bring any conflict nor behavioral change.


slipher:
LGTM

PR #2745, opened 2023-06-27
bots: human use medkit when go for heal

bmorel:
This commit makes humans use their medkit pre-emptively when needing to heal. This will reduce their time camping on medistation at the very least.


slipher:
LGTM

Issue #2196, opened 2022-09-11
Administration enhancements


cu-kai:
The administration system in this game in the current form "gets the job done", but I think we could go further.

I'd like at least some sort of feature parity with Der Bunker (Tremulous): https://github.com/enneract/AA-QVM (staging branch) and as such many of the below features are already implemented in that QVM. Of course the code and featuress will need modernising to Unvanquished standards. I think I will be able to implement at least some of these via PRs.

Note that this list may change over time.

Each additional command should be accompanied by a suitable admin flag.

I won't consider this issue "closed" until the following things have been implemented:

Improvements to existing systems
===
- [ ] `/showbans`: Ban list format should be updated to be closer to what we use on bunker, which is IMO more readable:
![image](https://user-images.githubusercontent.com/13281185/189525479-27b854bb-b305-484f-b522-c8c8cb395aff.png)
- [ ] `/revert`: allow reverting a specific action. Current behaviour is: all actions since that are also reverted. Syntax should be: `/revert #5`, where `#5` is the specific buildlog entry you wish to revert. Retain original behaviour by not using the `#` operator.
- [ ] `/revert`: allow reverting the last X actions quickly. Syntax should be similar to: `/revert x5 h` to revert the last 5 human buildable actions. Retain original behaviour by not specifying a team or using the `x` operator.

New features
===
- [ ] NEW: `.NOVOTE` & `.NOCHAT` flags: self explanatory. Prevent a level or player from calling votes, or sending chat messages. Note the `.` before the flag, denoting that it is a "negative" flag. This would allow an admin with `/flag` access to assign this flag to somebody without the need to have that flag themselves. This would require changes to the `/flag` and `/unflag` commands as we don't currently support "negative" flags.
- [ ] NEW: `STEALTH` / `SUPERSTEALTH` mode: When admins perform actions, we should differentiate between how the admin is identified. 
By default, admins should be identified by their netname (connected name). 
If they have the `STEATH` flag, they are identified by their registered name only (current Unv behaviour).
If they have the `SUPERSTEALTH` flag, they are not identified (chat message would say "someone <performed action>").
- [ ] NEW: `/adminlog` command: Log admin actions which have occurred this game. Senior server admins can then recall a list of actions performed by admins. Here is an example of the current implementation on Tremulous: 
![image](https://user-images.githubusercontent.com/13281185/189524515-dd4af41d-25cf-4062-80b5-27a3407ce5a9.png)
[...]

PR #2746, opened 2023-06-29
sg_admin.cpp: improve the ban list

cu-kai:
This PR changes the ban list to the same style as used on Bunker (Tremulous) and as previously discussed in #2196.

I have attached a screenshot of what the new banlist will look like.
Screenshot_20230629_125311

Furthermore, I also reverted a change I made in #2735 which changed the behaviour of `G_ADMIN_NAME` and more appropriately used `G_user_name` instead.

If nobody objects, I would like to store the level of the admin who created the ban to include this on the list. I haven't made that change yet as it will break compatibility with any existing admin.dat which contains bans.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -4176,6 +4176,7 @@ static int ban_out( void *ban, char *str )
 	char          date[ 11 ];
 	g_admin_ban_t *b = ( g_admin_ban_t * ) ban;
 	char          *made = b->made;
+	bool          active;
 
 	if ( !str )
 	{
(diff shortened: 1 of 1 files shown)


src/sgame/sg_admin.cpp
```diff
@@ -4194,35 +4194,45 @@ static int ban_out( void *ban, char *str )
 
 	date[ i ] = 0;
 
-	if ( !b->expires || b->expires - t > 0 )
+	if ( !b->expires )
 	{
 		G_admin_duration( b->expires ? b->expires - t : -1,
```
slipher, line 4199: Unnecessary test of `b->expires` (there's another one a few lines down)

src/sgame/sg_admin.cpp
```diff
@@ -4194,35 +4194,45 @@ static int ban_out( void *ban, char *str )
...
-	             MAX_DURATION_LENGTH - 1,
-	             duration,
-	             Color::ToString( ( strchr( b->ip.str, '/' ) ) ? Color::Red : Color::White ).c_str(),
+		Q_strncpyz( duration, "(expired)", sizeof( duration ) );
+		d_color = Color::Green;
+	}
+
+	Com_sprintf( str, MAX_STRING_CHARS, 
+	             "^3------------------------------\n"
+	             "    ^7Name:     %s^7\n"
```
slipher, line 4219: There's no use in having these `^7`s before a newline

src/sgame/sg_admin.cpp
```diff
@@ -4194,35 +4194,45 @@ static int ban_out( void *ban, char *str )
...
 	             b->ip.str,
-	             date,
 	             b->banner,
+	             b->reason,
+	             date, 
 	             Color::ToString( G_ADMIN_BAN_IS_WARNING( b ) ? Color::Yellow : Color::Red ).c_str(),
-	             G_ADMIN_BAN_IS_WARNING( b ) ? "WARNING" : "BAN",
-	             b->reason );
+	             G_ADMIN_BAN_IS_WARNING( b ) ? "WARNING " : "BAN ",
+	             ( atoi( time ) > 0 ? "^7- expiring in " : "^7- " ),
```
slipher, line 4233: Seems simpler to construct one string above that would combine `time` `d_color` and `duration` instead of this ugly atoi hack.

cu-kai: I've removed the atoi hack while keeping the original string.

sweet235 (approved):

PR #2749, opened 2023-07-02
Implement default navgen config (applies to all maps)

### LINKED ISSUE #2748: allow changing default configuration of navmesh generator
I believe it is currently possible to configure navmesh generator elements for a single map, but I do not think it is possible to overload the default configuration? It would be be nice to be able to do that.

slipher:
A config can be placed at `<homepath>/default.navcfg`

TODO: testing

Fixes #2748.

### DIFF
--- a/src/shared/bot_nav_shared.cpp
+++ b/src/shared/bot_nav_shared.cpp
@@ -157,6 +157,8 @@ static void ParseOption( Str::StringRef name, Str::StringRef value, Str::StringR
 
 // A configuration file can be placed in the VFS or in <homepath>/game/ that will
 // customize the values in NavgenConfig (the former daemonmap flags).
+// To set the default for all maps, place it at /default.navcfg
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):
Will changing this file invalidate the navmeshes or do we need to manually regen all navmeshes?

sweet235:
> Will changing this file invalidate the navmeshes or do we need to manually regen all navmeshes?

These values end up in the header of a navmesh file. Near the end of the file, we do this:
```
	if ( 0 != memcmp( &header.config, &config, sizeof(NavgenConfig) ) )
	{
		return "Navgen config changed";
	}
```

This compares the file's config with the config requested by the .navcfg files. If these are not equal, new navmeshes are generated automatically.

LGTM

slipher:
If someone has done a basic test that this works it can be merged.

Issue #2729, opened 2023-06-09
Wall climbing alien bots can be blocked by a building on top


sweet235:
Dretch and mantis will activate the wall climb when following a navcon that leads upwards. Usually, these navcons will be placed such that these bots climb a wall quite vertically. They will arrive at a rather small predefined navcon end on top of the wall.

Put some building at the top edge: they will fail to go over the edge. Then they will fall down, and most probably try to follow the navcon again; and do this in a loop.

There is not much to do about the dretch, but the mantis could and should attack if it touches a human building.

I am solely responsible for this.

PR #2758, opened 2023-07-12
dretch and mantis bots: attack enemies while wall climbing

sweet235:
Fix https://github.com/Unvanquished/Unvanquished/issues/2729 regarding enemy players and buildables. Wall climbing dretch and mantis perform an upward trace while wall climbing. If that trace hits an enemy, they will attack and increase the time limit for the wall climb, until about 10 seconds have passed.

If a wall climbing bot is blocked by one of its own team's buildables, there is still not much we can do. But the team should notice eventually and remove the buildable.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -855,11 +855,12 @@ static void BotMaybeAttackWhileClimbing( gentity_t *self )
 	BG_BoundingBox( static_cast<class_t>( ownClass ), &playerMins, &playerMaxs, nullptr, nullptr, nullptr );
 
 	glm::vec3 origin = VEC2GLM( self->s.origin );
-	glm::vec3 dir = glm::vec3( 0.0f, 0.0f, 1.0f );
+	glm::vec3 forward, right, up;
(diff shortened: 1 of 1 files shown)


src/sgame/sg_bot_nav.cpp
```diff
@@ -845,6 +845,48 @@ static int WallclimbStopTime( gentity_t *self )
...
+static void BotMaybeAttackWhileClimbing( gentity_t *self )
+{
+	ASSERT( self->botMind->lastNavconDistance > 0 );
+
+	int ownClass = self->client->ps.stats[ STAT_CLASS ];
+	glm::vec3 playerMins, playerMaxs;
+	BG_BoundingBox( static_cast<class_t>( ownClass ), &playerMins, &playerMaxs, nullptr, nullptr, nullptr );
+
+	glm::vec3 origin = VEC2GLM( self->s.origin );
+	glm::vec3 dir = glm::vec3( 0.0f, 0.0f, 1.0f );
```
slipher, line 858: Walls are not necessarily exactly vertical. I think it would make more sense to use the direction the bot is facing

sweet235: Yes

slipher:
LGTM

PR #2759, opened 2023-07-15
bots: invert check if a healing source is horizontally placed

sweet235:
Current code tries to prevent alien bots from picking a building to heal at, if this building is on walls or ceilings (except booster). This is reasonable, as this would make the bot stuck doing strange jumps in most cases. However, the current check is _inverted_, meaning that bots will _only_ pick such bad healing targets.

Please do not ask me why current code uses the dot product with a vector `(0,0,1)` to extract the vertical coordinate of a vector. I have no idea at all.

I have tried this with the default eggs on plat23. More tests could not hurt.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -246,7 +246,7 @@ botEntityAndDistance_t BotGetHealTarget( const gentity_t *self )
 		// We skip buildings on roof or walls as bots can't wallwalk.
 		// Unless it's a booster, as boosters are often put on walls
 		// and have a large area of effect.
-		if ( candidate.ent && ( glm::dot( VEC2GLM(candidate.ent->s.origin2), up ) <= MIN_WALK_NORMAL || buildable == BA_A_BOOSTER ) )
+		if ( candidate.ent && ( glm::dot( VEC2GLM(candidate.ent->s.origin2), up ) >= MIN_WALK_NORMAL || buildable == BA_A_BOOSTER ) )
 		{
 			return candidate;
 		}
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -246,6 +246,9 @@ botEntityAndDistance_t BotGetHealTarget( const gentity_t *self )
 		// We skip buildings on roof or walls as bots can't wallwalk.
 		// Unless it's a booster, as boosters are often put on walls
 		// and have a large area of effect.
+		// TODO:
+		// - The dot product is used to simply extract z component, maybe remove it
+		// - Maybe check if the building is over the navmesh
 		if ( candidate.ent && ( glm::dot( VEC2GLM(candidate.ent->s.origin2), up ) >= MIN_WALK_NORMAL || buildable == BA_A_BOOSTER ) )
 		{
 			return candidate;


src/sgame/sg_bot_util.cpp
```diff
@@ -246,7 +246,7 @@ botEntityAndDistance_t BotGetHealTarget( const gentity_t *self )
 		// We skip buildings on roof or walls as bots can't wallwalk.
 		// Unless it's a booster, as boosters are often put on walls
 		// and have a large area of effect.
-		if ( candidate.ent && ( glm::dot( VEC2GLM(candidate.ent->s.origin2), up ) <= MIN_WALK_NORMAL || buildable == BA_A_BOOSTER ) )
+		if ( candidate.ent && ( glm::dot( VEC2GLM(candidate.ent->s.origin2), up ) >= MIN_WALK_NORMAL || buildable == BA_A_BOOSTER ) )
```
necessarily-equal, line 249:
```suggestion
		if ( candidate.ent && ( candidate.ent->s.origin2[2] >= MIN_WALK_NORMAL || buildable == BA_A_BOOSTER ) )
```

necessarily-equal: And with that change you can remove the `static constexpr glm::vec3 up( 0.0f, 0.0f, 1.0f );` above.

necessarily-equal (approved):
Ah, good catch. That would be why the bots go to the OM very often, and sometime act weird on healing.

Indeed, as you asked for confirmation on IRC, dot product with (0,0,1) extracts the z coordinate. IMO it's easier to just do `x[2]` so maybe we could change that also. I did put a suggestion thread for that.

LGTM with and without those two suggested changes. Thanks

src/sgame/sg_bot_util.cpp
sweet235: > That would be why the bots go to the OM very often

Changing this might actually be a problem. Assume no booster is present. If there are eggs or leeches far from base, most bots might go there to heal. Result: they will not defend their base. Until we implement better bot strategy, I would argue that bots should go to the overmind frequently. That is what I do myself.

> and sometime act weird on healing

That is the bug I am trying to fix. Currently, once the overmind is down, the bots stop to defend the rest of the base. Instead, they do strange jumps below the remaining eggs until fully healed.

I could simply add a comment to explain what the result of the dot product is. We might add a check for the vertical distance to the navmesh soon. Then we can refactor this. (Checking the vertical distance would also be useful for painsaw bots.)

necessarily-equal: > > That would be why the bots go to the OM very often
> 
> Changing this might actually be a problem. Assume no booster is present. If there are eggs or leeches far from base, most bots might go there to heal. Result: they will not defend their base. Until we implement better bot strategy, I would argue that bots should go to the overmind frequently. That is what I do myself.

Actually overmind has higher priority compared to eggs and drills in all cases, and I don't think it can be built slanted, so it's probably always selected if available. This also means that in theory the bug you are afraid of shouldn't trigger. (Or overmind is dead).

And about the small change to make it more readable, it's ok if you want to refactor it another day.

(Btw, another thing I'm noticing now is that the code dosen't check if the buildable is reachable from navmesh. But that's not easy to fix, so should probably be left for another day)

PR #2760, opened 2023-07-22
improve bot lcannon handling: do not always charge to the target's full health

sweet235:
Recently, human bots using lcannon are extremely easy to kill with big aliens. The reason is that they are too easy to predict. This is surprising, as they are quite deadly against small aliens.

How to easily kill even a skill 9 lcannon bot: use a dragoon. Notice that the bot will _always_ charge its lcannon up to the dragoon's 200 HP. This will trigger the beep. As soon as you hear the beep, pounce upwards to fly in a nice parabola (one quite safe direction is tangential to some small circle around the bot on the floor). The bot will likely not hit you. After that, you have enough time to get close the bot and hit it once. You are in no danger at all until you hear the beep again. Repeat until the bot is dead. You can learn to do this without being hit in a few minutes.

Compare this to trying to kill a prifle or chaingun bot. Good luck with that.

This PR adds two small improvements to lcannon handling:

- If you have been hit by a human in the last 4 seconds, we can assume the bot heard you cry and can thus estimate your health. Make the bot only charge up to your actual health in this case. This might make it shoot before the beep. 
- If you go very near the bot, you are likely attacking it. Allow the bot to release the shot if you are closer than 100 units (and the charge is above 40%).

### DIFF
--- a/src/sgame/sg_bot_local.h
+++ b/src/sgame/sg_bot_local.h
@@ -121,6 +121,7 @@ enum bot_skill
 	BOT_A_AIM_HEAD,
 	BOT_A_AIM_BARBS, // precisely target barbs
 	BOT_H_PREDICTIVE_AIM, // predict where to aim depending on weapon and enemy speed
+	BOT_H_LCANNON_TRICKS,
 
 	BOT_NUM_SKILLS
 };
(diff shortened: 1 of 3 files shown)


DolceTriade (approved):
Do you think this should be a higher skill level bot "skill"?

sweet235:
It could be, but which part? We have 3 lcannon tricks for bots:

- Knowing the health for 4 seconds
- Firing below the desired charge when in danger
- Using the secondary fire against small/weak aliens

DolceTriade:
I'm less familiar with the bot code, but I thought it could just be another one of the tricks for bots.

sweet235:
No problem, I think I understand the bot skills in principle. I added a bot skill `lcannon-tricks`, this makes the bot "hear" the enemy's health. By default, bots with skill level 5 and above have it.

The other two are just basic lcannon handling, I'd say.

slipher:
LGTM

PR #2761, opened 2023-07-22
command revert: remove beacons, and maybe reduce team's spawn counter

sweet235:
The admin command `revert` is arguably the most important command, especially when you have new players around during a serious game. It has two bugs:

- When reverting the construction of a building, its beacon will persist.
- When reverting the construction of a spawn, the team's spawn counter is not changed. Thus the game cannot end anymore. (This is a bad one)

Fix both. A beacon for the enemy team might remain until the enemy team looks again.

### DIFF
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -2682,6 +2682,17 @@ void G_BuildLogRevert( int id )
 										 ent->num(), BG_Buildable( ent->s.modelindex )->name );
 						}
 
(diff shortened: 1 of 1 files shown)


slipher:
I think the underlying problem here is that an entity with a health component is being removed without being killed first so that HandleDie functions can perform cleanup. Other nonviolent means such as deconstruction always "kill" the entity first, in the CBSE sense. So I propose another approach in #2763.

PR #2763, opened 2023-07-25
Fix admin build revert by "killing" reverted buildables

slipher:
See #2761 for a description of bugs this would fix.

An entity with a HealthComponent should never be freed without being killed first, so that we can use HandleDie to maintain invariants and ensure cleanup.

Using G_Deconstruct we should get these things handled automatically:
- Decrease momentum for the team owning the buildable 
- Refund build points
- Decrement spawn counter (in SpawnerComponent::HandleDie)
- Remove non-enemy beacons (in BuildableComponent::HandleDie)

TODO:
- Testing
- Check momentum stuff more


slipher:
This is ready to go. I have tested that everything works with momentum and build points. (I found another pre-existing bug which gives free build points in some cases.)

src/sgame/components/BuildableComponent.cpp
```diff
@@ -84,7 +84,7 @@ void BuildableComponent::HandleDie(gentity_t* killer, meansOfDeath_t meansOfDeat
 	}
 
 	// If not deconstructed, add all build points to queue.
```
DolceTriade, line 86: update comment?

DolceTriade (approved):

PR #2765, opened 2023-08-11
Don't use in-place StripColors

slipher:
The in-place version will be deleted due to triggering compiler bugs in clang. Also it did not even make sense to use it any of the existing cases; in all of them the version targeting a separate buffer is strictly superior.

See https://github.com/DaemonEngine/Daemon/issues/839


illwieckz:
Great! I would like to know why clang is going mad at it, but we better want to use better code anyway.

illwieckz:
I guess this obsoletes:

- https://github.com/DaemonEngine/Daemon/pull/878

DolceTriade (approved):

PR #2767, opened 2023-08-14
Revert "Increase painsaw "range" for bot calculations"

sweet235:
This reverts commit 3145e8e7860a6d54fc5c00c92672c3919ef50c32. Currently, painsaw bots will stop 60 units short of an alien building to attack. Removing the number 60 in the code fixes this.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1306,8 +1306,7 @@ bool BotTargetInAttackRange( const gentity_t *self, botTarget_t target )
 			secondaryRange = 0;
 			break;
 		case WP_PAIN_SAW:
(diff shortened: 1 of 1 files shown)


slipher:
This was done to support #2539. #2539 without the range change made painsaw bots play like a triggerbot. However #2539 was never merged. At some point I LGTMed it but then @bmorel added another commit (which did not LGTM), and later closed it without merging.

If we never merge #2539 I guess it's OK to revert.

sweet235:
I re-opened https://github.com/Unvanquished/Unvanquished/pull/2539 as https://github.com/Unvanquished/Unvanquished/pull/2769. We must either merge it, or do this revert.

PR #2768, opened 2023-08-15
bots avoid ground flames

sweet235:
There are frequent complaints about alien bots dying in ground flames. Currently, they simply ignore them. That makes them appear quite suicidal. This can be exploited by human players to gain lots of funds.

This PR eases this problem by making the bots jump when they are damaged by ground flames. This already helps a lot. The jump is not done if the bot is close to its next path corner. This is because, by jumping, the bot might miss the corner and turn around to go back into the flames.

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -400,6 +400,8 @@ Bot Thinks
 =======================
 */
 
+static Cvar::Cvar<bool> g_bot_jumpOverFire("g_bot_jumpOverFire", "if bots should jump over ground flames", Cvar::NONE, true);
+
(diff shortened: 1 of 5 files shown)


src/sgame/sg_bot.cpp
```diff
@@ -400,6 +400,8 @@ Bot Thinks
 =======================
 */
 
+static Cvar::Cvar<bool> g_bot_jumpOverFire("g_bot_jumpOverFire", "if bots should jump over ground flames", Cvar::NONE, true);
```
slipher, line 403: s/if/whether/

illwieckz:
@sweet235 do you know what's the status of this?

PR #2769, opened 2023-08-15
Ai fight range

sweet235:
This is a copy of https://github.com/Unvanquished/Unvanquished/pull/2539


slipher:
Whichever commit contains `blurrDist` should be removed. That's the part added later that looks wrong.

There is a merge conflict to fix

sweet235:
I could try to update this. But we might as well choose to merge https://github.com/Unvanquished/Unvanquished/pull/2767. What do we want to do?

slipher:
Opened #2771 as a quick fix for the painsaw vs. buildable issue.

I have a vague memory that you also worked on some sort of blaster switching logic -- a topic treated also by this PR. Did that get merged? If so maybe it conflicts too much and we'll have to pick out only some of the commits here.

PR #2771, opened 2023-08-17
Fix bot painsaw range issues

slipher:
Fix the bug reported by @sweet235 in #2767. But also implement one of the ideas of #2539 which is to stop bots from *always* firing the painsaw, even when there are no enemies around, which can lead to friendly fire.

Includes one refactoring commit from #2539.


sweet235:
I do not feel like we should think too hard about friendly fire here. Friendly fire is a (rather big) problem we have with all weapons. A summary of my findings about friendly fire is in https://github.com/Unvanquished/Unvanquished/issues/2775.

slipher:
Well it seems good regardless of friendly fire. Not always firing the painsaw just improves the verisimilitude.

Plus the painsaw doesn't share the same code path as other weapons. Even if you get the traces to work correctly, it wouldn't help avoid FF from the painsaw.

PR #2773, opened 2023-08-23
Make human bots go to the nearest armoury when out of ammo

sweet235:
Currently, we never make human bots decide to pick up new ammo. When out of ammo, they continue to fight with the blaster. Even after their current enemy target is killed or lost, they will pick new enemy targets to attack with the blaster. Usually, this results in their sudden demise.

Why do not we see blaster bots all the time? Usually, there is an armoury next to a medistation. When low on health, they go there and trigger the automatic refill. The other reason is the purchase of better equipment (might be just a grenade/firebomb) at the nearest armoury.

This PR  makes the bot fight the current enemy target with the blaster, once out of ammo. But it does not allow human bots to choose new enemy targets after that, if the nearest armoury is reachable. Instead, the existing buy action is triggered, even when the bot has everything it wants to buy. This results in ammo refill.

Two drawbacks of this approach:
- This only considers the nearest armoury. However, the existing buy action has the same restriction
- This does not allow for energy weapon refill at the reactor or drills

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1510,7 +1510,8 @@ AINodeStatus_t BotActionBuy( gentity_t *self, AIGenericNode_t *node )
 		}
 
 		//we have every upgrade we want to buy
-		if ( numContain == numUpgrades )
+		if ( numContain == numUpgrades
+			 && ( !WeaponIsEmpty( BG_PrimaryWeapon( self->client->ps.stats ), &self->client->ps ) || !g_bot_resupply.Get() ) )
 		{
 			return STATUS_FAILURE;
 		}
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -1192,12 +1192,13 @@ bool BotChangeGoal( gentity_t *self, botTarget_t target )
 
 	if ( g_bot_resupply.Get()
 		 && G_Team( self ) == TEAM_HUMANS
+		 && target.targetsValidEntity()
 		 && G_Team( target.getTargetedEntity() ) == TEAM_ALIENS
 		 && WeaponIsEmpty( BG_PrimaryWeapon( self->client->ps.stats ), &self->client->ps ) )
 	{
 		botTarget_t closestArmoury;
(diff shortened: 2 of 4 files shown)


slipher:
This would be better as part of the behavior tree. There are `percentAmmoClip` and `percentClips` functions there. Also there is `resupplyScore`, which was contributed by @bmorel but not used in master's behavior tree.

sweet235:
> This would be better as part of the behavior tree

I tried. For hours in fact. But I concluded that this is not feasible. The reason is that bots are "alerted to enemies" automatically. Even if the .bt file does not run the action `fight`, the enemy will still be the bot's goal. So the bot will run towards it (but not attack it). I tried some ugly hacks using `decorator return ( STATUS_FAILURE )`, but could not manage to do it.

So I decided to change the part of the C++ code that selects an enemy as a bot's goal. I believe this is the only way to avoid having enemy goals when out of ammo. This is easy and straightforward to do. In fact, this is PR consists mainly of this one change. The rest is a one line hack to make the buy action run in case ammo is needed.

I considered adding an action `resupply` instead. But then I asked myself why on earth anyone would want their bots to buy weapons, but not get ammo when needed. On the other hand, a resupply action could handle reloading energy weapons at other buildings.
 
> There are percentAmmoClip and percentClips functions

But we removed the more useful `percentAmmo`.

> there is resupplyScore [...] not used in master's behavior tree

Not used for good reason. I bet it has the same problems as `healScore`, in particalur to depend on the distance. You can lock our current alien bots in their base by hitting them with an mdriver shot as soon as they come out. Proximity of the booster gives a high heal score, so the bots turn around immediately. I would propose to remove `healScore` from our behavior trees.

slipher:
One way you could do it in the BT is to make the bots bail out of the `alertedToEnemy` block if they have no ammo, by changing this:

```
	sequence
	{
		condition alertedToEnemy
		selector
		{
			... //huge block here
		}
	}
```
to this:
```
	sequence
	{
		condition alertedToEnemy
		condition percentAmmoClip > 0 || percentClips > 0
		{
			selector
			{
				... //huge block here
			}
		}
	}
```

Recall that the form of condition followed by a braced block continues checking the condition every frame and exits the block if it becomes false, whereas 
```
sequence {
	condition foo
	action bar
}
```
checks `condition foo` once, and does not recheck it as long as `action bar` is running.

PR #2774, opened 2023-08-23
Fix debug draw drawing quads 

slipher:
I found out that DU_DRAW_QUADS doesn't work. I don't know if quadrilaterals are ever used by any of our bot debug drawing stuff.


sweet235:
LGTM

PR #2776, opened 2023-08-26
fix bug in bot behavior function percentClips

sweet235:
For weapons without multiple clips (lgun, chaingun, flamer, lcannon), the behavior tree function `percentClips` returns 1.0. This is inconsistent with the rest of the codebase, which has 0 (for example: https://github.com/Unvanquished/Unvanquished/blob/master/pkg/unvanquished_src.dpkdir/configs/weapon/lgun.attr.cfg#L13). Fix this by returning 0.0 instead of 1.0.

Notice that this number does never count the current clip, but rather gives the number of times a weapon can be reloaded.

### DIFF
--- a/src/sgame/sg_bot_parse.cpp
+++ b/src/sgame/sg_bot_parse.cpp
@@ -153,7 +153,7 @@ static AIValue_t haveUpgrade( gentity_t *self, const AIValue_t *params )
 // Only used to reduce duplicated code.
 static AIValue_t Ratio( int current, int max, bool infiniteAmmo )
 {
-	if ( max == 0 || infiniteAmmo )
+	if ( infiniteAmmo )
 	{
 		return AIBoxFloat( 1.f );
 	}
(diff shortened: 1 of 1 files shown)


src/sgame/sg_bot_parse.cpp
```diff
@@ -164,6 +164,7 @@ static AIValue_t Ratio( int current, int max, bool infiniteAmmo )
 static AIValue_t percentAmmoClip( gentity_t *self, const AIValue_t* )
 {
 	ASSERT( self && self->client );
+	ASSERT( wpa->maxAmmo > 0 );
```
slipher, line 167: Assertion seems wrong. I think maxAmmo is 0 for infinite ammo weapons

cu-kai (approved):

src/sgame/sg_bot_parse.cpp
bmorel: wrong. Adv goon have maxAmmo = 3 for barbs, which are also infinite.

PR #2780, opened 2023-08-29
fill beyond level.team[team].botFillTeamSize to balance

sweet235:
The current situation: let's assume we requested to fill both teams with bots to 3 players, and nobody joined the game yet. We have 3 bots on each team. Now assume one user joins one team, and five users join the other team. Then we have one user and two bots on one team, and five users on the other team.

Change this behavior such that the smaller team is filled to 5 players in this case. The excess bots are removed as soon as the other team shrinks.

This is only done if the requested fill sizes of both teams are equal. If they are not, the game config rejected a balanced fill explicitely.

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -687,6 +687,9 @@ static void G_BotCheckDefaultFill()
 	}
 }
 
+static Cvar::Cvar<bool> g_bot_fillEvenHumans("g_bot_fillEvenHumans", "whether the human team is filled beyond the chosen number of bots if the alien team has more players than that", Cvar::NONE, false);
+static Cvar::Cvar<bool> g_bot_fillEvenAliens("g_bot_fillEvenAliens", "whether the alien team is filled beyond the chosen number of bots if the human team has more players than that", Cvar::NONE, false);
+
 void G_BotFill(bool immediately)
 {
 	static int nextCheck = 0;
@@ -715,8 +718,15 @@ void G_BotFill(bool immediately)
 	{
 		std::vector<int> current; // list of filler bots
 		int target; // if <0, too many bots, if >0, not enough
+		int actualFill;
 	} fillers[ NUM_TEAMS ] = {};
 	int missingFillers = 0;
+
+	// by default, fill according to level.team[ team ].botFillTeamSize
+	for ( team_t team : { TEAM_ALIENS, TEAM_HUMANS } )
+	{
+		fillers[ team ].actualFill = level.team[ team ].botFillTeamSize;
+	}
 
 	for (int client = 0; client < MAX_CLIENTS; client++)
 	{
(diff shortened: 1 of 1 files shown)


slipher:
I don't find this very compelling, as a 5v1 is probably very unbalanced, regardless of whether bots are added to compensate. We have `g_teamForceBalance` to prevent the players from joining in an unbalanced way.

Also this would make a config like `g_teamBalanceCountBots on`, `g_teamForceBalance 1`, `bot fill 3` fail to work as expected. If you have enforced team size balancing but bots are counted as players and new bots always joint to balance, then the team size balancing restriction would have no actual effect unless the server runs out of slots. So if we do merge this it should be an opt-in cvar.

sweet235:
> So if we do merge this it should be an opt-in cvar

Very well, did that.

> We have g_teamForceBalance to prevent the players from joining in an unbalanced way

Do we, really?

sweet235:
Let me explain why I plan to do that. Our player community can be divided into some expert players, and some casual players. The range in players' skill can be quite high. Frequently, I have to play against a team of 2 to 4 players in an attempt to balance a game. And there are stronger players than me.

In these situations, new players can connect and join the already bigger team. This creates the scenario described above. Almost always this is followed by a vote or admin intervention to add a bot. Naturally, people ask why we do not automate this.

slipher:
> > We have g_teamForceBalance to prevent the players from joining in an unbalanced way
> 
> Do we, really? Counting players does not give us a good measure for balance. But this is not what this PR is about.

Is there some reason you don't want to use it? If making the team sizes equal is not useful for balance, I don't see the point of this PR.

cu-kai:
> > > We have g_teamForceBalance to prevent the players from joining in an unbalanced way
> > 
> > 
> > Do we, really? Counting players does not give us a good measure for balance. But this is not what this PR is about.
> 
> Is there some reason you don't want to use it? If making the team sizes equal is not useful for balance, I don't see the point of this PR.

Because if the newer, lower-skilled players can't join their preferred team (admittedly usually Humans), chances are they won't play at all and might leave the server. There is no denying that the Human team is easier to play from the perspective of a new player, which in itself is also a problem outside the scope of the PR.

I think this is a good partial solution to this problem; on the map testing server and on Bunker we do not force team balance. In a more competitive environment this would be desirable, and the option should remain available. I can't think of a Tremulous server that _didn't_ force team balance.

sweet235:
Indeed, I know no server that has `g_teamForceBalance` enabled. It seems the player community has moved on to take this as given, and we fail to catch up in the codebase.

Another reason to disable forced balance is to allow anyone to help a single new player, who will frequently end up in the human team on an otherwise empty server. There they will build 10 turrets, which the alien bots will eat. Then they get overrun by big aliens and quit.

> If making the team sizes equal is not useful for balance, I don't see the point of this PR

There are two measures here. One is to require the same number of players on each team (plus or minus one). Another is to require both teams to be filled with bots to an equal number. Given the social impossibility of the former, the latter is what this is about.

slipher:
> Another reason to disable forced balance is to allow anyone to help a single new player, who will frequently end up in the human team on an otherwise empty server.

There is the `g_teamForceBalance 2` option which allows N vs. 0 as a special case. (This was contributed by illwieckz so I guess his server uses it?)

Another good option is `g_teamForceBalance 1` with `g_teamBalanceCountBots 1`. Let's say you have `bot fill 5`, then you can have up to 5v1 (or maybe 6v1?) because the bots are included in the count.

DolceTriade:
Is the motivation here to always have "even" (as in number) teams regardless of the fill settings?

I'm not sure that the name `g_bot_fillBeyond` is a great name. I would suggest something like `g_bot_botFillEvenTeam`. I assume this cvar operates independently of `bot fill`? So regardless of whether bot fill is enabled, if this cvar is enabled if one team has 5 players humans and another team has 4 humans, the 4 player team will get a bot?

If so, this sounds like a reasonable feature.

slipher:
> I think another mode to `g_forceTeamBalance` could work too (ie, mode 3 == fill team with less players with bots)

I was also thinking about this, as adding bots seems like a way to balance team size that conflicts with the method of preventing players from joining the larger team. But then you could combine the two to make bots fill spots left by players leaving a team, but not allow players to join in an unbalanced way.

Another possibility is to make it part of the `bot fill` command. I suggest the syntax `bot fill 4+` meaning that the team size is filled to at least 4, but maybe more to keep the team sizes equal. In this way the effect of the bot fill commands will still disappear after a map change, but without having to have an unprincipled exception that `g_bot_fillBeyond` is ignored if the requested fill size is 0.

bmorel:
This change allows admins to focus on game instead of constantly adjusting bot fill number when teams are of uneven numbers of players.
Today's games shown that this is a useful change, as they happened on cu-kai's server which does not have this feature, and teams were regularly imbalanced because one team would have one more player than the other on frequent basis as players left and joined.

illwieckz:
I believe `g_teamForceBalance` is a good option for a game that has thousands of players in game at any time in the day, making possible to assume there is a kind of average skill, either because people chose servers according to their skills, either because the average people is not new. Our gaming community is not in such situation and I truly believe this option is better for small communities, then for us in our curren situation.

illwieckz (approved):
Assuming it is tested and it works, LGTM.

sweet235:
Fixed a bug where the actual number of filler bots could be below `botFillTeamSize`. Rebased on master.

Current state:
- Boolean cvars `g_bot_fillEvenHumans` and `g_bot_fillEvenAliens`
- If activated by cvar, and if the other team has more players than `botFillTeamSize`, then add fillers to make teams even
- Do not do this if `botFillTeamSize` is zero

DolceTriade:
LGTM.

src/sgame/sg_bot.cpp
```diff
@@ -727,10 +737,35 @@ void G_BotFill(bool immediately)
 		}
 	}
 
+	int biggerTeamSize = -1;
+	for ( team_t team : { TEAM_ALIENS, TEAM_HUMANS } )
```
slipher, line 741: I don't understand what the calculation in this loop is supposed to be doing. If you want to get the size of a team excluding filler bots, you can subtract the length of the filler bots list from the team size. I think the body of the loop could just be
```
if ( <relevant cvar>.Get() )
   fillers[ team ].actualFill = std::max( fillers[ team ].actualFill, level.team[ team ].numClients - fillers[ team ].current.size();

PR #2781, opened 2023-08-30
human bots refill ammo at armouries

sweet235:
This supersedes https://github.com/Unvanquished/Unvanquished/pull/2773 and is described there.

PR #2782, opened 2023-08-31
human weapons: refill the current clip too

sweet235:
At the armoury (or maybe reactor or drill), do not only give a player's clip weapon the maximum number of clips, but fill the current clip too. Current behavior is to only give the maximum number of clips.

I suppose the current behavior was implemented after noticing that one can circumvent the reload time when being close to an armoury.  You have to stop shooting for a short while to trigger the refill, but the reload time is noticably longer.

Turns out that you cannot host even the most slightly modded server without every player you know complaining about this. At each (complete or non-disciplined) refill, a player has to camp near the armoury for the reload time. This is conceived as very annoying. So I have been doing this on my server since I forgot when.

In the end, any serious player will learn how to circumvent the reload time anyway. One way is to simply buy the current weapon again. If you have a bind for that, I guess are you are already doing this. You can even use `p_weapon` and `vstr` to make a refill bind just like Tremulous had it.

Recently (in https://github.com/Unvanquished/Unvanquished/pull/2781), I ran into the problem of making human bots camp for the reload time when refilling ammo. I considered to make them buy the same weapon again instead. But then I decided to open this PR.

### DIFF
--- a/src/sgame/sg_weapon.cpp
+++ b/src/sgame/sg_weapon.cpp
@@ -137,6 +137,11 @@ static bool CanUseAmmoRefill( gentity_t *self )
 		return false;
 	}
 
+	if ( ps->weapon == WP_BLASTER )
+	{
+		return false; // don't allow refill if they have the blaster equipped
+	}
+
 	if ( wa->maxClips == 0 )
 	{
 		// clipless weapons can be refilled whenever they lack ammo
@@ -187,7 +192,7 @@ bool G_RefillAmmo( gentity_t *self, bool triggerEvent )
 		}
 	}
 
-	G_ForceWeaponChange( self, (weapon_t) self->client->ps.stats[ STAT_WEAPON ] );
+	G_ForceWeaponChange( self, (weapon_t) self->client->ps.weapon );
 
 	return true;
 }
--- a/src/sgame/sg_weapon.cpp
+++ b/src/sgame/sg_weapon.cpp
@@ -178,7 +178,6 @@ bool G_RefillAmmo( gentity_t *self, bool triggerEvent )
 
 		if ( triggerEvent )
 		{
-			G_AddEvent( self, EV_CLIPS_REFILL, 0 );
 			G_AddEvent( self, EV_AMMO_REFILL, 0 );
 		}
 	}
--- a/src/sgame/sg_weapon.cpp
+++ b/src/sgame/sg_weapon.cpp
@@ -187,7 +187,7 @@ bool G_RefillAmmo( gentity_t *self, bool triggerEvent )
 		}
 	}
 
-	G_ForceWeaponChange( self, (weapon_t) self->client->ps.weapon );
+	G_ForceWeaponChange( self, (weapon_t) self->client->ps.stats[ STAT_WEAPON ] );
 
 	return true;
 }
--- a/src/sgame/sg_weapon.cpp
+++ b/src/sgame/sg_weapon.cpp
@@ -142,9 +142,9 @@ static bool CanUseAmmoRefill( gentity_t *self )
 		// clipless weapons can be refilled whenever they lack ammo
 		return ( ps->ammo != wa->maxAmmo );
 	}
-	else if ( ps->clips != wa->maxClips )
+	else if ( ps->clips != wa->maxClips || ps->ammo != wa->maxAmmo )
 	{
-		// clip weapons have to miss a clip to be refillable
+		// clip weapons can be refilled whenever they lack ammo or clips
 		return true;
 	}
 	else
@@ -169,10 +169,12 @@ bool G_RefillAmmo( gentity_t *self, bool triggerEvent )
 	if ( BG_Weapon( self->client->ps.stats[ STAT_WEAPON ] )->maxClips > 0 )
 	{
 		GiveMaxClips( self );
+		GiveFullClip( self );
 
 		if ( triggerEvent )
 		{
 			G_AddEvent( self, EV_CLIPS_REFILL, 0 );
+			G_AddEvent( self, EV_AMMO_REFILL, 0 );
 		}
 	}
 	else
--- a/src/sgame/sg_weapon.cpp
+++ b/src/sgame/sg_weapon.cpp
@@ -130,32 +130,15 @@ static bool CanUseAmmoRefill( gentity_t *self )
 	}
 
 	ps = &self->client->ps;
-	wa = BG_Weapon( ps->stats[ STAT_WEAPON ] );
+	// This is the *currently equipped* weapon, so you don't get refill while wielding blaster
+	wa = BG_Weapon( ps->weapon );
 
 	if ( wa->infiniteAmmo )
 	{
 		return false;
 	}
 
-	if ( ps->weapon == WP_BLASTER )
-	{
-		return false; // don't allow refill if they have the blaster equipped
-	}
-
-	if ( wa->maxClips == 0 )
-	{
-		// clipless weapons can be refilled whenever they lack ammo
-		return ( ps->ammo != wa->maxAmmo );
-	}
-	else if ( ps->clips != wa->maxClips || ps->ammo != wa->maxAmmo )
-	{
-		// clip weapons can be refilled whenever they lack ammo or clips
-		return true;
-	}
-	else
-	{
-		return false;
-	}
+	return ps->clips != wa->maxClips || ps->ammo != wa->maxAmmo;
 }
 
 /**
--- a/src/sgame/sg_weapon.cpp
+++ b/src/sgame/sg_weapon.cpp
@@ -187,6 +187,8 @@ bool G_RefillAmmo( gentity_t *self, bool triggerEvent )
 		}
 	}
 
+	G_ForceWeaponChange( self, (weapon_t) self->client->ps.weapon );
+
 	return true;
 }
 
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -1083,9 +1083,8 @@ void CG_EntityEvent( centity_t *cent, vec3_t position )
 			break;
 
 		case EV_AMMO_REFILL:
-		case EV_CLIPS_REFILL:
 		case EV_FUEL_REFILL:
-			// TODO: Add different sounds for EV_AMMO_REFILL, EV_CLIPS_REFILL, EV_FUEL_REFILL
+			// TODO: Add different sounds for EV_AMMO_REFILL, EV_FUEL_REFILL
 			trap_S_StartSound( nullptr, es->number, soundChannel_t::CHAN_AUTO, cgs.media.itemFillSound );
 			break;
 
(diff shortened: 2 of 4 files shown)


slipher:
The long-term DPS of the rifle is about 34, in the field. This bumps it up to 62 for someone camping by the Armory. Yes, you can get almost as good of a rate by using buy binds, but I see infinite clip size for camping humans as more of an exploit that should be fixed than a feature that should be promoted.

Anyway with this code, to fire at the optimal rate, you still need the stupid trick of stopping when you have 1 bullet left in the clip instead of exhausting it. But the "reloading" delay is a bit less compared to buy binds, like 0.2 s vs. 0.4.

cu-kai (approved):
We've had this in place on the map testing server for months now, after many complaints from players about upstream behaviour. I do not believe it negatively impacts the gameplay enough to be an issue, and as @sweet235 said, good players were abusing the "buy weapon" trick beforehand anyway.

Note that I do _not_ host @sweet235 patch on my server and to this day get complaints about it when there are games on there.

The code is sound and the feature has been tested thoroughly.

sweet235:
> I see infinite clip size for camping humans as more of an exploit that should be fixed than a feature that should be promoted

Well, this calls for a meta argument. Because we are having a conflict between this abstract line of reasoning and user experience. I would argue that user experience is much more important than a minor gameplay problem. But it seems to me that in the past, dev team has favoured abstract reasoning over users' opinions.

src/sgame/sg_weapon.cpp
```diff
@@ -142,9 +142,9 @@ static bool CanUseAmmoRefill( gentity_t *self )
 		// clipless weapons can be refilled whenever they lack ammo
 		return ( ps->ammo != wa->maxAmmo );
 	}
-	else if ( ps->clips != wa->maxClips )
+	else if ( ps->clips != wa->maxClips || ps->ammo != wa->maxAmmo )
```
slipher, line 145: There is no longer a need to special-case clipless weapons. This if/else if/else chain can be replaced with `return ps->clips != wa->maxClips || ps->ammo != wa->maxAmmo`

src/sgame/sg_weapon.cpp
```diff
@@ -169,10 +169,12 @@ bool G_RefillAmmo( gentity_t *self, bool triggerEvent )
 	if ( BG_Weapon( self->client->ps.stats[ STAT_WEAPON ] )->maxClips > 0 )
 	{
 		GiveMaxClips( self );
+		GiveFullClip( self );
 
 		if ( triggerEvent )
 		{
 			G_AddEvent( self, EV_CLIPS_REFILL, 0 );
+			G_AddEvent( self, EV_AMMO_REFILL, 0 );
```
slipher, line 177: This makes two instances of the same sound play at once. Just delete EV_CLIPS_REFILL from the enum.

slipher (commented):
Player feedback usually tends to suggest buffing whichever side they are currently playing. So if something feels annoying we should pay attention, but we should be skeptical of the proposed solutions.

Anyway some suggestions to clean up the code...

bmorel:
What this change does is this: it makes bind users less advantaged compared. Which means it makes the game more fair for newcommers.

If that "advantage" is bad, then the real fix would be to make sell/buy actions take time, an easy way to do this would be to make buying a weapon with empty clips. But my bet is that this would be rather unpopular.

sweet235:
If any of you is confused about why this is so controversial, I can assure you: so am I. Luckily, there is a way to give you an easy method to refill all ammo. Just add this to your file `autogen.cfg`:

```
set sweet_head "if "
set sweet_refill_rifle "set sweet_tail \" = 9 \\\"/sell weapons; buy rifle\\\"\"; concat sweet_cmd sweet_head p_weapon sweet_tail; vstr sweet_cmd"
set sweet_refill_shotgun "set sweet_tail \" = 11 \\\"/sell weapons; buy shotgun\\\"\"; concat sweet_cmd sweet_head p_weapon sweet_tail; vstr sweet_cmd"
set sweet_refill_lgun "set sweet_tail \" = 12 \\\"/sell weapons; buy lgun\\\"\"; concat sweet_cmd sweet_head p_weapon sweet_tail; vstr sweet_cmd"
set sweet_refill_mdriver "set sweet_tail \" = 13 \\\"/sell weapons; buy mdriver\\\"\"; concat sweet_cmd sweet_head p_weapon sweet_tail; vstr sweet_cmd"
set sweet_refill_chaingun "set sweet_tail \" = 14 \\\"/sell weapons; buy chaingun\\\"\"; concat sweet_cmd sweet_head p_weapon sweet_tail; vstr sweet_cmd"
set sweet_refill_flamer "set sweet_tail \" = 15 \\\"/sell weapons; buy flamer\\\"\"; concat sweet_cmd sweet_head p_weapon sweet_tail; vstr sweet_cmd"
set sweet_refill_prifle "set sweet_tail \" = 16 \\\"/sell weapons; buy prifle\\\"\"; concat sweet_cmd sweet_head p_weapon sweet_tail; vstr sweet_cmd"
set sweet_refill_lcannon "set sweet_tail \" = 17 \\\"/sell weapons; buy lcannon\\\"\"; concat sweet_cmd sweet_head p_weapon sweet_tail; vstr sweet_cmd"
set sweet_refill "vstr sweet_refill_rifle; vstr sweet_refill_shotgun; vstr sweet_refill_lgun; vstr sweet_refill_mdriver; vstr sweet_refill_chaingun; vstr sweet_refill_flamer; vstr sweet_refill_prifle; vstr sweet_refill_lcannon"
bind N "vstr sweet_refill"
```

Start the game: now you can press N to refill all ammo whenever you are near an armoury. Change the `N` in the last line to any key you would like to use. You can find a list of all key names [here](https://github.com/DaemonEngine/Daemon/blob/master/src/common/KeyIdentification.cpp#L45).

slipher:
I have no longer been opposing the basic idea. It is the stage of fixing up the code

slipher:
> Luckily, there is a way to give you an easy method to refill all ammo. Just add this to your file `autogen.cfg`:

Hey, that's a great example of [p_ vars usable for "cheating"](https://github.com/Unvanquished/Unvanquished/pull/2677) :D  Although it could still be done without them, as long as you don't mind a center print warning, with `sell rifle; buy rifle; sell shotgun; buy shotgun;` etc.

sweet235:
That is not cheating. Nor will you prevent any actual cheating by randomly removing cvars once you find one particular "cheat". The console is turing-complete (the proof is in https://github.com/Unvanquished/Unvanquished/pull/2677). I will not give an example of an actual cheat. I will, however, remark that all (or almost all) player actions are available as console commands.

illwieckz:
Can we get the reload animation when refilling the current clip?

illwieckz:
> Can we get the reload animation when refilling the current clip?

Well, I guess cheaters will then use sell/buy commands to avoid this.

What if we always play reload animation after buying something?
For weapon without reload animation we may just wait.

I find it weird when I go to an armoury and my current clip is refilled without me doing a clip reload, forcing the reload animation may not only reduce the risk of cheating, but would also make the game more realistic.

sweet235:
> Well, I guess cheaters will then use sell/buy commands to avoid this.

I would not put it like this, because I do not want to accuse the majority of our regular users of cheating. This is simply something you do if you have binds to buy weapons. Considering the context sensitivity of our armoury menu, I cannot blame anyone for using raw binds.

illwieckz:
I mean someone with intention of getting an unfair advantage would still be able to do that. I'm not saying anyone using bind is a cheater, it's our fault if we give two ways to do one thing and one gives advantage. Anyway selling/buying command are not reloading command, so using selling/buying to reload is not the purpose of those commands.

I never meant “cheaters” designate actual people, it's about naming a potential threat we have to prepare response for.

sweet235:
I would argue that the very nature of an open source game makes client-side cheat prevention meaningless.

illwieckz:
> I would argue that the very nature of an open source game makes client-side cheat prevention meaningless.

Open source just makes some things easier, but does not make anticheat prevention meaningless. Every time some game project argues they have to be closed source to prevent  cheating, cheaters are laughing. Even closed-source games require literal rootkits in kernel space to try to catch (and not prevent) cheating.

VReaperV:
> I mean someone with intention of getting an unfair advantage would still be able to do that. I'm not saying anyone using bind is a cheater, it's our fault if we give two ways to do one thing and one gives advantage. Anyway selling/buying command are not reloading command, so using selling/buying to reload is not the purpose of those commands.
> 
> I never meant “cheaters” designate actual people, it's about naming a potential threat we have to prepare response for.

It's something that any player can do using in-game functionality though though. And myself and, as I've seen, other people that change binds/cvars/configs have told people in the game how to use those or do specific things with them, so it's not some secret knowledge either.

slipher:
> Can we get the reload animation when refilling the current clip?

The reload animation lasts a couple of seconds while the delay for the weapon to be ready after this PR's reload is less than 0.5 s. Using the reload animation would give a false impression of the weapon not being ready

cu-kai:
With permission from @sweet235, I have updated this PR. You will now be forced to change to your current weapon upon receiving ammo at the armoury, creating a slight delay as well as playing the appropriate animation.

illwieckz:
> Using the reload animation would give a false impression of the weapon not being ready

I don't think the weapon should be ready when the current clip is refilled, in all cases.

This is exactly what I asked for.

slipher:
> > Using the reload animation would give a false impression of the weapon not being ready
> 
> I don't think the weapon should be ready when the current clip is refilled, in all cases.
> 
> This is exactly what I asked for.

If you make the weapon unready for the duration of the reload animation, that would be quite a nerf for humans -- whenever you get close to the armoury you become unable to fire for 2 seconds... the original conception was to help humans by giving them easier quick reloads. Anyway the weapon switch idea seems like a happy medium as long as the bugs are fixed.



> With permission from @sweet235, I have updated this PR. You will now be forced to change to your current weapon upon receiving ammo at the armoury, creating a slight delay as well as playing the appropriate animation.

The reload triggers even if the blaster is currently equipped, which seems kinda wonky.

Also in testing this again I found a bug in the original commit. If you sell your weapon at the armory (leaving yourself with only the blaster), a refill event is triggered.

cu-kai:
Re-equipping your current weapon is what Tremulous did too. Perhaps it could instead force you to swap to the weapon in your slot?

slipher:
The behavior of the weapon switch commit is t

> Re-equipping your current weapon is what Tremulous did too. Perhaps it could instead force you to swap to the weapon in your slot?
Not sure I understand which way you're suggesting, but the current behavior is to switch from blaster to blaster.

cu-kai:
> > Re-equipping your current weapon is what Tremulous did too. Perhaps it could instead force you to swap to the weapon in your slot?
> 
> Not sure I understand which way you're suggesting, but the current behavior is to switch from blaster to blaster.

I'm suggesting enforcing switching to the weapon in the slot, i.e. whatever your `self->client->ps.stats[ STAT_WEAPON ]` is. I've now made that change.

src/sgame/sg_weapon.cpp
slipher: Fixed

slipher:
LGTM

Kai's latest update also fixes the existing bug of getting a refill event when selling your weapon (without buying another).

PR #2784, opened 2023-09-02
Weapon-related refactoring

slipher:
- Partially get rid of the inapt abstraction of buildable attacks as "weapons" (WP_HIVE, WP_LOCKBLOB, WP_ROCKETPOD, WP_MGTURRET). Weapons used by human players have several things in common:
    - Can be equipped by players
    - Firing events are generated by pmove; the actual firing is implemented elsewhere in response to these events
    - Configuration values from the files `configs/weapons/<name>.attr.cfg` are used
  
  The buildable weapons use none of this. The only part of of the "weapon" infrastructure they use is that for some of them a weapon firing event is used to trigger sounds or other client-side effects, which are configured in `models/weapons/<name>/weapon.cfg`. This is the case for turret and rocketpod firing. (The trapper also has a firing sound but it is implemented differently.) So I was able to completely remove WP_HIVE and WP_LOCKBLOB but not the other two.
  
- Clean up some ugly global variable usage in sg_weapons.cpp. In doing so I discovered and fixed a bug where the mara zap effect flickers in and out of existence.

Also there is a commit adding a `lifetime` attribute to missile configs. It's not very relevant but comes from a branch working on CBSE for missiles, with which I don't want to create merge conflicts.


src/sgame/sg_weapon.cpp
```diff
@@ -636,7 +625,7 @@ HIVE
 Target tracking for the hive missile.
 ================
 */
-static void HiveMissileThink( gentity_t *self )
+void HiveMissileThink( gentity_t *self )
```
DolceTriade, line 639: nit: Since this is a global function, should we do the G_ prefix? Should this function just be moved to the Component? (same comment for the rocket pod)

DolceTriade (approved):
nice clean ups

src/sgame/sg_weapon.cpp
slipher: I am working on CBSE for missiles, so when the hive missile has its own component I can move it there

DolceTriade: Sgtm

PR #2785, opened 2023-09-02
Fix alien HUD being drawn for spectators

slipher:
This fixes the alien HUD being inappropriately drawn at various times, e.g. for spectators after a map_restart.

Also remove the onSpawn event type for BSP entities. This seems to be a remnant of Q3; you can't spawn from a map entity in Unvanquished.

### DIFF
--- a/src/sgame/sg_entities.cpp
+++ b/src/sgame/sg_entities.cpp
@@ -527,7 +527,6 @@ static const entityCallEventDescription_t gentityEventDescriptions[] =
 		{ "onFree",      ON_FREE      },
 		{ "onReach",     ON_REACH     },
 		{ "onReset",     ON_RESET     },
-		{ "onSpawn",     ON_SPAWN     },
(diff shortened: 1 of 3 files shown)


src/sgame/sg_entities.cpp
```diff
@@ -527,7 +527,6 @@ static const entityCallEventDescription_t gentityEventDescriptions[] =
 		{ "onFree",      ON_FREE      },
 		{ "onReach",     ON_REACH     },
 		{ "onReset",     ON_RESET     },
-		{ "onSpawn",     ON_SPAWN     },
```
DolceTriade, line 530: Do you think this is functionality we might want to keep for mods/Lua?

DolceTriade (approved):

src/sgame/sg_entities.cpp
slipher: The code relating to this is all specific to BSP entities so I doubt it could be good with a Lua integration.

PR #2786, opened 2023-09-02
Small cleanup

slipher:


src/cgame/cg_predict.cpp
```diff
@@ -174,12 +174,12 @@ static void CG_ClipMoveToEntities( const vec3_t start, const vec3_t mins,
 		{
 		case traceType_t::TT_CAPSULE:
 		case traceType_t::TT_AABB:
-			CM_TransformedBoxTrace( &trace, start, end, mins, maxs, cmodel, mask, skipmask, origin, angles, collisionType );
+			CM_TransformedBoxTrace( &trace, start, end, mins, maxs, cmodel, MASK_ALL, 0, origin, angles, collisionType );
```
DolceTriade, line 177: `mask` + `skipmask` are now unused? We should remove the parameter or respect it.

DolceTriade (approved):

src/cgame/cg_predict.cpp
slipher: CM_TransformedBoxTrace can be used to trace against either the world or a specific entity's bounding box. When tracing against the world the masks are useful, but a specific entity only has one content type so the masks are pointless.

DolceTriade: I think we should document this...

slipher: I realized the two different cases, tracing against a bbox or tracing against a map model, are actually combined in the same call sites. So I should only switch it to MASK_ALL for the bbox case. Anyway I will drop the commit from this PR for now and revisit later.

PR #2789, opened 2023-09-06
Select class to spawn as in behavior tree

slipher:
With an optional `selectClass { ... }` prelude in the behavior tree, a bot can choose to spawn with a different class/weapon. The logic there parses and executes the same way as the normal BT. There's an example in the code comments.

Also I added the `numOurBuildings` function for BTs just so I could write a simplistic example. I think @sweet235 would be able to use it though, if porting the random builder logic from C++ to BT.

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -608,6 +608,14 @@ void G_BotIntermissionThink( gclient_t *client )
 	client->readyToExit = true;
 }
 
+void G_BotSelectSpawnClass( gentity_t *self )
+{
+	if ( self->botMind->behaviorTree && self->botMind->behaviorTree->classSelectionTree )
+	{
+		BotEvaluateNode( self, self->botMind->behaviorTree->classSelectionTree );
+	}
+}
+
 // Initialization happens whenever someone first tries to add a bot.
 // Assuming the meshes already exist, this incurs some delay (a few tenths of a second), but on
 // servers bots are normally added at the beginning of the round so it shouldn't be noticeable.
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -502,6 +502,38 @@ static bool EvalConditionExpression( gentity_t *self, AIExpType_t *exp )
 	return false;
 }
 
+AINodeStatus_t BotSpawnNode( gentity_t *self, AIGenericNode_t *node )
+{
+	if ( !( self->client->ps.pm_flags & PMF_QUEUED ) )
+	{
+		return STATUS_FAILURE;
+	}
+
+	const AISpawnNode_t *spawn = reinterpret_cast<AISpawnNode_t *>( node );
+
+	switch ( G_Team( self ) )
+	{
+	case TEAM_ALIENS:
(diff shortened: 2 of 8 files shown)


sweet235:
Another reason this will be useful is granger bots extinguishing burning buildings. I found out that we already have a working implementation of advanced grangers spitting at human targets quite accurately. This can easily be adapted to put out fires.

slipher:
Any thoughts about how it should look in the BT file? I'm not sure a tree is the best way, for example maybe it should be just a boolean expression saying whether to spawn as builder.

sweet235:
Maybe a boolean expression is simpler, at least as long as we do not change the rules of the game. Notice that we will want bot to spawn as advanced grangers, when possible. So the we would need to either make `spawnAs` fail for `PCL_ALIEN_BUILDER0`, or include some function to check if advanced grangers are unlocked.

That would look like this:

```
selectClass {
    selector {
	condition team == TEAM_ALIENS && numOurBuildings(E_A_OVERMIND) == 0 {
            selector {
                spawnAs PCL_ALIEN_BUILDER0_UPG
		spawnAs PCL_ALIEN_BUILDER0
            }
	}
	...
    }
}
```

Or like this:

```
selectClass {
    selector {
	condition team == TEAM_ALIENS && numOurBuildings(E_A_OVERMIND) == 0 {
            selector {
                condition isUnlocked(PCL_ALIEN_BUILDER0_UPG) {
                    spawnAs PCL_ALIEN_BUILDER0_UPG
                }
		spawnAs PCL_ALIEN_BUILDER0
            }
	}
        ...
    }
}
```

A boolean flag would make this a little shorter.

sweet235:
> A boolean flag would make this a little shorter.

But we would have to introduce some action to spawn as the better builder. Maybe the proposed method is simpler.

sweet235:
LGTM

slipher:
> But we would have to introduce some action to spawn as the better builder.

Well, we consider offering a choice between the builder types to be unimportant enough that there is no menu for it in the cgame and it can only be done through commands. With a boolean expression, automatically choosing the highest unlocked builder class seems good enough.

But IMO there is a stronger reason to choose the behavior tree: the decision criteria may be complex. Instead of some 10-line monstrosity of a single boolean expression, it would be more comprehensible to use the behavior tree.

sweet235:
Let's do it like this. The worst that can happen is that we change this later.

bmorel:
FWIW (likely not much, and I no longer consider BTs such a good option anyway, for unv's complexity) I think bot's "brain code" should be in G_ClientThink, so that bots would actually be able to decide how to spawn, or perhaps even if they should go to spectator or change team depending on the needs. Bots spectating would make sense, as it would avoid them to disconnect when they're replaced, loosing all their credits in the process, and being broke when they connect because a player left.
They should also likely not be just savagely kicked, but either wait until they die or are out of fight, which would fix the quite annoying problem of: you're shooting at a bot, hoping to acquire some credits, and it gets kicked. Or the bot is leading an attack and you were assisting it (or assisted by it), and gets kicked, breaking the flow of the game.

But I doubt this would be a simple code change in any case.

PR #2793, opened 2023-09-24
Documentation for trap_Trace

slipher:
Document `trap_Trace`, the sgame's extension of tracing functions. (The main tracing documentation for the engine is added in https://github.com/DaemonEngine/Daemon/pull/931). Unfortunately the extra field `entityNum` filled in by the gamelogic has bad and inconsistent semantics.

I found a lot of places which have bugs due to using tracing APIs incorrectly. This is mostly because the engine's trace API is very hard to understand without documentation.  I made comments on some of them, as well as on places which may work correctly but rely on the questionable entityNum semantics.

An area that needs more investigation is whether the attack traces work correctly when playing with high latency. When unlagged is activated, the other clients' positions are temporarily changed to their past positions, so it is possible that they overlap the attacking player. There is some code that tries to avoid prevent this but it never corrects the overlap for more than one other client, so clearly it won't always work.

I added a `trace` command which prints the results of a trap_Trace call for debugging.

Also there are some minor bugfixes and cleanups related to tracing. I didn't touch any of the major bugs.

### DIFF
--- a/src/cgame/cg_predict.cpp
+++ b/src/cgame/cg_predict.cpp
@@ -211,8 +211,12 @@ void CG_Trace( trace_t *result, const vec3_t start, const vec3_t mins, const vec
 
 	CM_BoxTrace( &t, start, end, mins, maxs, 0, mask, skipmask, traceType_t::TT_AABB );
 	t.entityNum = t.fraction != 1.0 ? ENTITYNUM_WORLD : ENTITYNUM_NONE;
-	// check all other solid models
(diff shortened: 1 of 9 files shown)


illwieckz:
For memories:

```
2023-10-08 00:05:28 +0200 <perturbed`> any objections to https://github.com/DaemonEngine/Daemon/pull/931  https://github.com/Unvanquished/Unvanquished/pull/2793 ?
2023-10-08 00:20:21 +0200 <illwieckz> none
```

PR #2794, opened 2023-09-27
Fix bugs with missile tracing

slipher:
Add a new trace API G_Trace2 which always collides when there is something overlapping the start point, unlike standard tracing. Use this to fix a bug that some missiles don't explode when overlapping a player.

### DIFF
--- a/src/sgame/sg_missile.cpp
+++ b/src/sgame/sg_missile.cpp
@@ -25,6 +25,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
 #include "sg_local.h"
+#include "sg_cm_world.h"
 #include "Entities.h"
 #include "CBSE.h"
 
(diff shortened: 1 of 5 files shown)


src/sgame/sg_missile.cpp
```diff
@@ -332,20 +333,20 @@ static int ImpactHive( gentity_t *ent, trace_t*, gentity_t *hitEnt )
...
+static void MissileImpact( gentity_t *ent, const trace2_t *trace )
 {
 	int       dirAsByte, impactFlags;
 	const missileAttributes_t *ma = BG_Missile( ent->s.modelindex );
 	gentity_t *hitEnt   = &g_entities[ trace->entityNum ];
 	gentity_t *attacker = &g_entities[ ent->r.ownerNum ];
 
 	// Returns whether damage and hit effects should be done and played.
-	std::function<int(gentity_t*, trace_t*, gentity_t*)> impactFunc;
+	int (*impactFunc)( gentity_t *, const trace2_t *, gentity_t * );
```
DolceTriade, line 349: why go back to C funcs?

slipher: My instinct was to use the most specific type possible so that it is more efficient. Doesn't really matter though

Hopefully I will get the impact stuff as a CBSE message

DolceTriade (approved):

PR #2806, opened 2023-10-06
Fix shader-replacing entities

### LINKED ISSUE #2796: shader replacement entity: regression in habitat
In tremulous, it is possible for a mapper to change a texture used in a brush which does not belongs to an entity.
This feature is used at least on [habitat](https://users.unvanquished.net/~sweet/pkg/map-habitat_1.1.dpk) from yalt and participates to the feedback players receive from the map, which received significant effort to give players a good immersion.
The source of that map can be found [here](https://github.com/Masmblr/ioQuake3-Tremulous-Maps).

It is possible to enable (but not disable) some elevators in this room:

Before doing so, they will appear like this:

In tremulous, but not in unvanquished, at least those "lights" textures will change from red to green:

I think @cu-kai or @sweet235 are able to confirm.

slipher:
Thanks to @bmorel for researching and identifying the fix

Fixes #2796. Tested by me

### DIFF
--- a/src/sgame/sg_entities.cpp
+++ b/src/sgame/sg_entities.cpp
@@ -740,6 +740,13 @@ void G_EventFireEntity( gentity_t *self, gentity_t *activator, gentityCallEvent_
 	gentityCall_t call;
 	call.activator = activator;
 
+	// Shader replacement. Example usage: map "habitat" elevator activation, red to green light replacement
+	if ( self->shaderKey && self->shaderReplacement )
+	{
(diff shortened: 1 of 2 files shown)


slipher:
@bmorel says:
> That's not enough. You also need to disable the deletion of entity when they don't have a target (they have, but it's a shader, and this is not considered by unv's code).

Well the red lights became green by doing this. Is there some other effect that's missing?

slipher:
OK I added the other missing bit that makes the green lights pulsating, and makes the control panels light up after pressing a button.

DolceTriade:
lgtm

PR #2809, opened 2023-10-09
Don't zero hive missile velocity before death

slipher:
> The motivation for this was to simplify the code in preparation for
porting to CBSE. However it is also an aesthetic improvement. With the
old behavior, you typically get a cluster of bees spinning in a tight
circle. The spinning, alpha-fading circle of bees looks stupid. With the
new behavior they tend to disperse in various directions which seems
better.

Also remove nearby dead code.


DolceTriade:
Is this purely aesthetic or does this alter gameplay as well?

slipher:
I suppose there's a tiny impact on gameplay since the window in which the missile moves and seeks enemies is 100ms longer. The lifetime can be shortened in the config file if this is a concern

DolceTriade (approved):

DolceTriade:
Don't think it's that big of a deal.

PR #2810, opened 2023-10-09
Make overcharging lucifer cannon actually explode at the muzzle point

slipher:
The old behavior was that the projectile actually travels a short distance before exploding. But it was actually possible for it to hit something before the programmed self-destruction, and score a direct hit for 265 damage. This seems like an illogical mechanic and a bug, so ensure that the missile explodes at its start point without being allowed to hit anything.

### DIFF
--- a/src/sgame/sg_missile.cpp
+++ b/src/sgame/sg_missile.cpp
@@ -460,12 +460,11 @@ static void MissileImpact( gentity_t *ent, trace_t *trace )
 void G_ExplodeMissile( gentity_t *ent )
 {
 	vec3_t dir;
-	vec3_t origin;
 	const missileAttributes_t *ma = BG_Missile( ent->s.modelindex );
 
-	BG_EvaluateTrajectory( &ent->s.pos, level.time, origin );
+	glm::vec3 origin = VEC2GLM( ent->r.currentOrigin );
 	SnapVector( origin );
-	G_SetOrigin( ent, VEC2GLM( origin ) );
+	G_SetOrigin( ent, origin );
 
 	// we don't have a valid direction, so just point straight up
 	dir[ 0 ] = dir[ 1 ] = 0;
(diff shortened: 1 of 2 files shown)


cu-kai:
I have used this in the past to kill a target right in front of me. You have to be quite lucky to be in a position where this is actually effective, but nonetheless is there a good reason to change this behaviour other than "it seems like a bug to me"? This change would be gameplay altering and I'm not sure I support it.

DolceTriade:
I think that an explosion should also hurt surroundings

slipher:
The current behavior doesn't square with the mechanics of the weapon. The lucifer cannon projectile is supposed to explode only when it hits something. Here we have a goofy 3rd mode where it acts like a grenade launcher with a 50 ms fuse.

> This change would be gameplay altering

What is this supposed to imply? It's literally true, but the change makes no meaningful impact on balance.



> I think that an explosion should also hurt surroundings

The fallout works same as before (except not slightly in front of the player). The difference is that you can't score a direct hit.

sweet235:
> But it was actually possible for it to hit something before the programmed self-destruction, and score a direct hit for 265 damage

That has been the agreed upon behavior for one or two decades. Why is this a problem?

slipher:
> > But it was actually possible for it to hit something before the programmed self-destruction, and score a direct hit for 265 damage
> 
> That has been the agreed upon behavior for one or two decades. Why is this a problem?

In what sense is it agreed upon? It was probably unintentional by the code's author, and very few players will have noticed it since an unlikely combination of conditions is needed (overcharging, an entity being very close and being able to gauge the amount of damage done to the entity).

cu-kai:
> > > But it was actually possible for it to hit something before the programmed self-destruction, and score a direct hit for 265 damage
> > 
> > 
> > That has been the agreed upon behavior for one or two decades. Why is this a problem?
> 
> In what sense is it agreed upon? It was probably unintentional by the code's author, and very few players will have noticed it since an unlikely combination of conditions is needed (overcharging, an entity being very close and being able to gauge the amount of damage done to the entity).

Never underestimate the player's ability to notice small quirks like this. I do not believe this should be changed at this stage.

DolceTriade:
Seems that this is not a desirable change?

slipher:
Superseded by #2825.

PR #2811, opened 2023-10-13
Migrate missiles to CBSE (part 1)

slipher:
This is stacked on top of #2810 and #2794. Only the last 3 commits are new.

This first part goes as far as creating a CBSE entity for each missile entity. Later parts will improve the encapsulation by moving more code and data into the components.


sweet235:
I propose to break out of the endless refactoring loop. The best we get out of this is no new bugs and no new features. But usually, new bugs will slip in.

CBSE should never have been merged, as it was never complete. The same goes for GLM. Both are not likely to be completed any time soon. Maybe it would be better to remove them both.

slipher:
> CBSE should never have been merged, as it was never complete. The same goes for GLM.

Those are both large undertakings which couldn't realistically be completed all at once. In each case there was a strong developer consensus to start the project. (Well, I wasn't really there for early CBSE discussions but a lot of devs contributed to it so I can deduce that.)

sweet235:
> Those are both large undertakings which couldn't realistically be completed all at once.

Looking at the current code base, one may doubt if they ever will be completed. It seems to be a common theme that people start some migration, but then leave the project. We end up with a messy mix of old and new methods.

src/sgame/sg_missile.cpp
```diff
@@ -438,13 +445,17 @@ static void MissileImpact( gentity_t *ent, const trace2_t *trace )
 
 void G_ExplodeMissile( gentity_t *ent )
 {
+	if ( ent->s.modelindex == MIS_FIREBOMB )
+	{
+		G_FirebombMissileIgnite( ent );
+	}
+
```
DolceTriade, line 452: This is kinda hacky to special case MIS_FIREBOMB here? Is this just to avoid the boilerplate of defining a new component for this?

src/sgame/sg_missile.cpp
```diff
@@ -517,7 +528,9 @@ static trace2_t MissileTrace( gentity_t *ent )
 	return result;
 }
 
-void G_RunMissile( gentity_t *ent )
+// Move missile along its trajectory and detect collisions.
+// Returns true if the missile has ceased to exist
+bool G_MoveMissile( gentity_t *ent )
```
DolceTriade, line 533: I would have expected this function and all others to live in the missile component?

src/sgame/components/RocketpodComponent.cpp
```diff
@@ -265,9 +265,13 @@ void RocketpodComponent::Shoot(const glm::vec3& direction) {
...
-	gentity_t* missile = G_SpawnMissile(MIS_ROCKET, entity.oldEnt, entity.oldEnt->s.pos.trBase, &direction[0],
-	                                    target->oldEnt, RocketThink, level.time + ROCKET_TURN_PERIOD);
-	missile->timestamp = level.time + BG_Missile(MIS_ROCKET)->lifetime;
+	gentity_t* m = G_NewEntity(HAS_CBSE);
+	RocketMissileEntity::Params params;
+	params.oldEnt = m;
+	params.Missile_attributes = BG_Missile(MIS_ROCKET);
+	m->entity = new RocketMissileEntity{ params };
+	G_SetUpMissile(m, MIS_ROCKET, entity.oldEnt, entity.oldEnt->s.pos.trBase, &direction[0]);
+	m->target = target->oldEnt;
```
DolceTriade, line 274: Should this have a helper function to avoid this boiler plate?

DolceTriade (commented):
This should have no gameplay changes right?

bmorel:
> CBSE should never have been merged, as it was never complete. The same goes for GLM. Both are not likely to be completed any time soon. Maybe it would be better to remove them both.

I'll allow myself to answer this partially, as I'm the person behind GLM move. There are few things to say here, which are quite different from the CBSE problem:

1. GLM introduction allowed to get rid of another NiH 3D vector library (said lib which was apparently introduced with CBSE);
2. GLM aims at reducing the size of the codebase maintained by upstream, while CBSE is a NiH implementation of ECS, mixing several languages (because why not). ECS are complex things to implement _correctly_, but they have real strengths, described succinctly by [godot engine devs](https://godotengine.org/article/why-isnt-godot-ecs-based-game-engine/). In my opinion, none of the positive aspects described there apply to CBSE, and many are not useful for a game like unvanquished anyway;
3. you say that it (glm migration) is unlikely to be completed. Well, you are both right and wrong. On my side of the code, I have migrated the whole server-side gamelogic and most of the shared gamelogic as well. I have slowed down my effort on this since I stepped out of the project and no longer host a server, but really, at least for shared/ I can get this finished in few days if I want, I even have a branch with the very delicate bg_pmove.cpp migration. I didn't touched cgame for a good reason: this is the only part of the code I don't know enough to be able to cherry-pick things from upstream in few minutes.

I believe I do not have my word to say on that PR though, nor on any unvanquished's code change, as my codebase already diverged way too much (around 200 commits by myself, plus some cherry picks from here, patched to apply, obviously). I just wanted to defend the glm migration.

Note that reverting the glm migration can likely be done in a very reasonable amount of time. The ABI of glm and q3 vec lib are close enough, you would mostly have to replace everything that starts with "glm::" by their counterpart. In this direction, it's going to be easy, trivial, even. You will then be able to enjoy the unreadable code that results from that.

src/sgame/sg_missile.cpp
slipher: The current working version of my [missile CBSE branch](https://github.com/slipher/Unvanquished/tree/mis-cbse) has that code moved into the missile component. Just trying to break it up into smaller pieces

src/sgame/sg_missile.cpp
slipher: Besides having a firebomb component (which seems fine as it does have its own specialized logic), I guess I'd need to add another CBSE message to allow the behavior to be specialized here. I'm a little loath to add extension points that don't have at least 2 users.

src/sgame/components/RocketpodComponent.cpp
slipher: Maybe, although it would have to be a template which is annoying.

What if we had the code generator make a builder-pattern struct for the Params thing? Like `RocketMissileEntity::Params().oldEnt(m).Missile_attributes(BG_Missile(MIS_ROCKET))` etc.

slipher:
> This should have no gameplay changes right?

Right, once you factor out #2810.

src/sgame/components/RocketpodComponent.cpp
DolceTriade: I think I template is way cleaner and readable.

(probably doesn't compile...)
```cpp
template<classname T>
gentity_t* NewMIssile(mis missle_t) {
        gentity_t* m = G_NewEntity(HAS_CBSE);
	T::Params params;
	params.oldEnt = m;
	params.Missile_attributes = BG_Missile(mis);
	m->entity = new T{ params };
	G_SetUpMissile(m, MIS_ROCKET, entity.oldEnt, entity.oldEnt->s.pos.trBase, &direction[0]);
	m->target = target->oldEnt;
        return m;
}
```

DolceTriade:
I think we should factor out #2810 from this PR. I tested this and stuff seems ok to me.

slipher:
Superseded by #2825.

src/sgame/components/RocketpodComponent.cpp
slipher: It works well as long as all missile types have the same params. But in one of my later commits I have the hive and rocket targets being passed in params, so then you have to pass around half-finished param structs and it's not that nice.

PR #2813, opened 2023-10-14
Make CBSE not a submodule

### LINKED ISSUE #2812: Make CBSE not a submodule
The concept of CBSE is well-known in game dev under the name "ECS" (entity-component system). Our implementation of it is quite bare-bones; meanwhile there are many alternative implementations. So it is quite unlikely that other developers will want to use it. It would be more expedient to have the CBSE generator in the Unvanquished repository so we can easily make changes.

I would move it from `src/utils/` to the top-level `tools/` I guess.

slipher:
Also move the rest of src/utils/ to tools/.

Fixes #2812.


DolceTriade (approved):

illwieckz:
Do we want to import history (I can do it)?

illwieckz:
This `temp/cbse-import` branch imports CBSE history, so if you rebase your PR on it, `git blame` and other tools processing history (like looking for file rename, etc.) will work:

- https://github.com/Unvanquished/Unvanquished/tree/temp/cbse-import

illwieckz:
It works for me.

I recommend to add `/src/utils` to gitignore to avoid listing `src/utils` as not committed when old submodules files are remaining (which can be pretty annoying when bisecting).

PR #2814, opened 2023-10-14
add quake and emoticons attributes to inlinecvar element

illwieckz:
Made to please @cu-kai for:

- https://github.com/DaemonEngine/Daemon/pull/937

This is totally untested.

- rocket: add `quake` and `emoticons` attributes to `inlinecvar` element
- dpkdir/ui: enable `quake coloring` for `ui_errorMessage` printing


cu-kai (approved):
Can confirm this allows the error messages to be formatted correctly.

cu-kai:
As we are messing with error.rml here, would it be possible to look into centering the window as part of this PR?

DolceTriade (approved):

DolceTriade:
I think centering the window and making it static should be a separate PR

PR #2819, opened 2023-10-16
fix: banned player names are now shown during attempted connection

### LINKED ISSUE #2816: Banned users' names aren't shown during attempted connection
When a banned player connects, admins are notified of this by an admin chat message, however the banned player's netname is never shown.

The output is:
`AdminMsg: -1 "console": Banned player  tried to connect from 127.0.0.1 (ban #2) `

For some reason, although `ent` is valid (and other fields are valid within), the netname is an empty string so `G_user_name` also fails to fall back to the ban name. The solution could be to exclusively use the name they had when they were banned however I think it preferable to display netname if possible.

cu-kai:
fixes #2816

When attempting to connect, show the accused's name used while connecting. If it differs from the name in the ban entry, show them both.


slipher:
Use G_ClientCleanName so that it's consistent with what the name would end up as in-game and so they can't spam the log with control characters

cu-kai:
> Use G_ClientCleanName so that it's consistent with what the name would end up as in-game and so they can't spam the log with control characters

Good point. I've now fixed this.

src/sgame/sg_admin.cpp
```diff
@@ -1702,9 +1705,17 @@ static void G_admin_ban_message(
 
 	if ( areason && ent )
 	{
+		trap_GetUserinfo( ent->num(), userinfo, sizeof( userinfo ) );
+
+		const char *s = Info_ValueForKey( userinfo, "name" );
+
+		G_ClientCleanName( s, name, sizeof( name ), ent->client );
```
slipher, line 1712: I think you should put nullptr for client here because if not it may do some goofy thing with incrementing a counter of the number of unnamed players...

This sort of thing is why we need refactoring :)

slipher:
LGTM

PR #2825, opened 2023-10-23
Migrate missiles to CBSE (part 1+2)

slipher:
This contains a commit to prevent lcannon overcharge direct hits, which faced opposition in #2810. However the last commit restores the original behavior.


slipher:
It's easy to restore the original overcharge behavior after having the missileAttributes_t customization commit (" MissileComponent: store attributes per missile"), but would be a lot of work without that. That's why I'm redoing the PR with several additional commits as compared to #2811.

DolceTriade:
Overall seems ok. Will review more closely after testing

DolceTriade (approved):

DolceTriade:
Tested it in a bunch of bot games and I didn't notice any regressions.

PR #2826, opened 2023-10-24
Modification of the menus

Grise3:
List of the changes/creations :

1. Main menu

- "Colors and symbols" moved to the player options menu
- list of the Emoticons created in the player options menu
- "Movement & Cosmetic" options  moved to a new window.

2.  Ingame menu :

- "Democracy" part moved to a new window
- "Bot" part moved to a new window

4. Global :

- Some spacing problems are fixed
- The text color is now white. It is more readable.
- Design updated ( buttons, ilink)
- Now, there is a space between the emoticons and the text, in the "gameplay guide"


DolceTriade:
Very cool!

Do you think you could add screenshots? To take screenshot in the menu, you can press `~` to open the menu and then:

`delay 1000 screenshotJPEG` and close the console and wait a second to let the screenshot happen.

DolceTriade:
All in all looks good I think. Do you think you could center the main menu? That' the only bug I see.








PS: If you wanna get fancy with it, you could add RCSS transitions for animations when opening and closing menus: https://mikke89.github.io/RmlUiDoc/pages/rcss/animations_transitions_transforms.html#interpolation

illwieckz:
While chatting with Grise3 I said  that centering the menu looked like a workaround for not having in the center what the player is expecting to find at startup: the server list (or the server creation form).

We don't have to put the menu in the center (and we better not do that) if the center of the screen is always occupied by what the user is doing.

illwieckz:
Also, opened windows will cover the main menu, so a centered main menu is only good at startup, but then becomes annoying as soon as a window is opened.

I then recommend to keep the main menu on the right for now even if that's annoying with the current layout of other things, but centering the main menu to fix one issue (first thing the player see is not centered) is bringing many more issues and if we try to fix all those issues to merge this PR this PR will never be merged. This PR brings enough changes, dealing with menu position should better be done later to reduce the burden.

pkg/unvanquished_src.dpkdir/ui/options_mouvements_cosmetic.rml
```diff
@@ -0,0 +1,51 @@
+<rml>
+	<head>
+		<link type="text/rcss" href="/ui/shared/basics.rcss" />
+		<link type="text/template" href="/ui/shared/window.rml" />
+	</head>
+	<body id="options_mouvements_cosmetic"  template="window" style="width: 45em; margin: 1em;">
```
illwieckz, line 6: Here “mouvement” is french instead of “movement” in english (same for the file name). :wink:

pkg/unvanquished_src.dpkdir/ui/options_ui.rml
```diff
@@ -9,16 +9,18 @@
...
+				<option value="en_EN">English</option>
 				<option value="fr_FR">Français</option>
 				<option value="ru_RU">Русский</option>
-				<option value="en_EN">English</option>
-				<option value="de_DE">Deutsch</option>
+				<option value="de_DE">Deutsch</option> <!--only translated at 39% -->
 				<option value="it_IT">Italiano</option>
-				<option value="es_ES">Español</option>
+				<option value="es_ES">Español</option> <!--translated at 19%-->
+				<option value="fi_FI">Español</option>
```
illwieckz, line 21: Finnish instead of Español :wink:

Grise3:
Ho and, I've added [unv] in the emoticon list :)
For don't "center" the main menu, 
replace
 <centerbar> </centerbar> by <sidebar> </sidebar>
and remove the comment of <hackyspacer/>

for show the other colors 
Screenshot_20231025_140429

Grise3:
> All in all looks good I think. Do you think you could center the main menu?

Well maybe I can. I need to find a way :)
I can try to do a 'real' centerbar that work for all the window. 


> Very cool!

Thx :) 

> Do you think you could add screenshots?

Do did it :)
There is a  screenshot in my precedent post.
when the cursor is on the button, the text became yellow and there is a lighter grey and a lighter cyan

@illwieckz I put the main menu in the right, then?

illwieckz:
Sorry for the late reply, I think that most of the changes are very good, however, I found a bug!

On the main menu, both on welcome screen and in game, pressing the `tab` key moves the whole menu to the left. I guess it's a relic of the old scrollbar mechanism:

unvanquished_2023-11-05_045732_000

unvanquished_2023-11-05_045652_000

unvanquished_2023-11-19_184921_000

unvanquished_2023-11-19_184928_000

illwieckz:
Then, while I agree the centered main menu looks very good in screenshot, it's an actual regression in usability (even if current usability is not that good).

When a player starts the game for the first time, the main menu is covered by the welcome form and then not easily accessible.

unvanquished_2023-11-01_013534_000

Once we open the server list, the main menu is behind and then actions from it are less accessible than before:

unvanquished_2023-11-19_222129_000

Right now the centered main menu adds more problems than it solves, which is sad because I like the way it looks!

To make a centered main menu works properly, we need a huge revamp of how we deal with windows, meaning we're not ready yet for a centered main menu.

And if we do a huge revamp, we may not center it, we may do better than that: we don't have to center the main menu if we open at startup what the player is looking for: the server list. But this is another topic.

Right now I suggest to keep the menu on the right in order to keep it accessible while other windows are opened.

Grise3:
The problem was that the size of the button was 100% I put it at 87% and it is good 
in options_player.rml, it was "</p>Your name visible to others.</p>"  I changed it to "<translate>Your name visible to others.</translate>" Now, the sentence will be translated.

Grise3:
All the button are now to 87%.

@illwieckz : I put the menu in the right, in the same place than the "old" menu.


Screenshot_20231213_122619

There are some button size in the "old" menu (and I don't corrected them) that are not "good"

Screenshot_20231213_122745

I will correct them for another PR.

Grise3:
Screenshot_20231216_100017

illwieckz:
I'm going to merge this to a feature branch before doing some rebase and adding some commits and then merging it to `master`. This way this PR is flagged as merged.

illwieckz:
I had to fix a regression where the “Server setup” menu entry was removed by mistake, now added back.

I added other commits improving the positionning of menus and windows.

illwieckz:
I pushed some other fix commits.

illwieckz:
I pushed the updated `.pot` file so Weblate can fetch it.

PR #2827, opened 2023-10-24
Fix double-send of info queries on server refresh

slipher:
When refreshing the serverlist, the "globalservers" or "localservers" console command is used to make the engine query the master server again and rebuild the list of potential servers. The command is buffered and not executed until after the cgame frame finishes. Meanwhile the cgame immediately sends trap_LAN_ResetPings and trap_LAN_UpdateVisiblePings which causes pings to be sent for all the servers in the *old* list. As soon as the new list arrives from the master, everything on the new list is pinged as well.

### DIFF
--- a/src/cgame/cg_rocket_datasource.cpp
+++ b/src/cgame/cg_rocket_datasource.cpp
@@ -248,6 +248,24 @@ void CG_Rocket_BuildServerList()
 	int netSrc = rocketInfo.currentNetSrc;
 	const char *srcName = CG_NetSourceToString( netSrc );
 
+	// Hack to avoid calling UpdateVisiblePings in the same frame that /globalservers
+	// is sent, which would ping the old server list
+	if ( rocketInfo.realtime == rocketInfo.serversLastRefresh )
+	{
+		return;
+	}
+
+	int numServers = trap_LAN_GetServerCount( netSrc );
+
+	// Still waiting for a response...
+	// HACK: assume not done if there are 0 servers, although it could be true
+	if ( numServers <= 0 )
+	{
+		return;
+	}
+
+	trap_LAN_MarkServerVisible( netSrc, -1, true );
+
 	// Only refresh once every 200 ms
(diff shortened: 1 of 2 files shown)


src/cgame/cg_rocket_datasource.cpp
```diff
@@ -251,7 +251,11 @@ void CG_Rocket_BuildServerList()
...
 	if ( rocketInfo.realtime < 200 + rocketInfo.serversLastRefresh )
 	{
-		trap_LAN_UpdateVisiblePings( netSrc ); // keep the ping queue moving
+		if ( rocketInfo.data.serverListPopulated )
+		{
+			trap_LAN_UpdateVisiblePings( netSrc ); // keep the ping queue moving
+		}
+
 		return;
 	}
```
DolceTriade, line 260: Why not just move this code after line 274 so we don't need to add another bool to track state?

slipher: I realized that my code had a regression in that there can be a delay up to 200 ms before UpdateVisiblePings is called for the first time, and rewrote it. I ended up getting rid of the bool and instead have a magic behavior when `rocketInfo.serversLastRefresh == rocketInfo.realtime`

DolceTriade (commented):
Is there a proper fix in daemon instead of waiting a frame for globalservers goes out?

Should we always update ping on global servers? Can we do that instead?

slipher:
> Is there a proper fix in daemon instead of waiting a frame for globalservers goes out?

I guess we could add a trap call for refreshing the serverlist instead of using a command.

> 
> Should we always update ping on `/globalservers`? Can we do that instead?

You mean it should automatically send pings instead of requiring trap_LAN_UpdateVisiblePings to be called every frame? Yeah, maybe. These APIs could use a redesign in general but I'm still not sure what would be a good design, so for now I'm trying to improve the situation without any compatibility breaks.

DolceTriade:
fair enough

DolceTriade (approved):

PR #2830, opened 2023-10-26
Fix warning that breaks CI

slipher:


DolceTriade:
If it breaks CI, how was it submitted?

DolceTriade (approved):

slipher:
The CI doesn't block anything... code authors have to check the status themselves to avoid submitting broken code.

DolceTriade:
huh

PR #2831, opened 2023-10-28
Move map entity-related gentity_t fields into sub-struct

slipher:
Code taken from @bmorel's fork, with light touchups.

Not sure how the pull request etiquette should work for this. Can I merge it without additional approvals since @bmorel is the author and I can be considered the approver?


illwieckz:
> Not sure how the pull request etiquette should work for this. Can I merge it without additional approvals since @bmorel is the author and I can be considered the approver?

Yes.

DolceTriade (approved):

PR #2188, opened 2022-09-09
Bot evolve fix and improvement

necessarily-equal:
This fixes the alien bot regression in master and also fixes https://github.com/Unvanquished/Unvanquished/issues/2181.

PR #2833, opened 2023-10-29
allow bots to evolve when the cost is zero

sweet235:
Bots cannot evolve from grangers to dretch or from granger to advanced granger. Since we merged a bot build action, this is a bug now. Fix it.

To see the current problem, set an alien bot to use `build.bt`, and then set it to use `default.bt`. If it has less than 2 morph points, it will be a battle granger.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -613,7 +613,7 @@ AINodeStatus_t BotActionEvolve( gentity_t *self, AIGenericNode_t* )
 	{
 		if ( cl.item == currentClass )
 		{
-			break;  // no smaller classes left to try
+			break;  // no bigger classes left to try
 		}
 
 		evolveInfo_t info = BG_ClassEvolveInfoFromTo( currentClass, cl.item );
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -613,8 +613,11 @@ AINodeStatus_t BotActionEvolve( gentity_t *self, AIGenericNode_t* )
 	{
 		evolveInfo_t info = BG_ClassEvolveInfoFromTo( currentClass, cl.item );
 
-		if ( info.evolveCost > 0 // no devolving or evolving to the same
-				&& BotEvolveToClass( self, cl.item ) )
+		bool isBuilder = currentClass == PCL_ALIEN_BUILDER0 || currentClass == PCL_ALIEN_BUILDER0_UPG;
+		if ( ( info.evolveCost > 0     // no devolving or evolving to the same
+			   || ( isBuilder && cl.item == PCL_ALIEN_LEVEL0 )
+			   || ( currentClass == PCL_ALIEN_BUILDER0 && cl.item == PCL_ALIEN_BUILDER0_UPG ) )
+			 && BotEvolveToClass( self, cl.item ) )
 		{
 			return STATUS_SUCCESS;
 		}
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -611,13 +611,14 @@ AINodeStatus_t BotActionEvolve( gentity_t *self, AIGenericNode_t* )
 
 	for ( auto const& cl : classes )
 	{
+		if ( cl.item == currentClass )
+		{
+			break;  // no smaller classes left to try
+		}
+
 		evolveInfo_t info = BG_ClassEvolveInfoFromTo( currentClass, cl.item );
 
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):
Basically allow bots to go from builder -> dretch and granger -> adv granger?

sweet235:
I think the old code never tried to handle the case of the granger classes. It simply assumed that evolving costs morph points. I added these cases, where no morph point is spent:

- Granger evolves to dretch
- Advanced granger evolves to dretch
- Granger evolves to advanced granger

src/sgame/sg_bot_util.cpp
```diff
@@ -613,8 +613,11 @@ AINodeStatus_t BotActionEvolve( gentity_t *self, AIGenericNode_t* )
 	{
 		evolveInfo_t info = BG_ClassEvolveInfoFromTo( currentClass, cl.item );
 
-		if ( info.evolveCost > 0 // no devolving or evolving to the same
-				&& BotEvolveToClass( self, cl.item ) )
+		bool isBuilder = currentClass == PCL_ALIEN_BUILDER0 || currentClass == PCL_ALIEN_BUILDER0_UPG;
+		if ( ( info.evolveCost > 0     // no devolving or evolving to the same
+			   || ( isBuilder && cl.item == PCL_ALIEN_LEVEL0 )
+			   || ( currentClass == PCL_ALIEN_BUILDER0 && cl.item == PCL_ALIEN_BUILDER0_UPG ) )
+			 && BotEvolveToClass( self, cl.item ) )
```
necessarily-equal, line 620: Thanks for fixing this! I believe this is equivalent to this:

```suggestion
		if ( ( !info.isDevolving && BotEvolveToClass( self, cl.item ) )
```

Interestingly I had the same problem when indicating the evolutions in the UI.

cu-kai (approved):

src/sgame/sg_bot_util.cpp
slipher: LGTM to this version

sweet235: I find this to be a little silly. First, we had evolve. Then we made an exception, and now we have devolve. Once we make the next exception, we will already be out of words.

slipher: The distinction between evolving and devolving is that you have to be close to the overmind to devolve. 

Maybe we should allow any evolution or devolution though? Looking at `build.bt`, it seems to expect being able to devolve: 
```
                                                action moveTo( E_A_OVERMIND )
                                                selector
                                                {
                                                        action evolveTo( PCL_ALIEN_BUILDER0_UPG )
                                                        action evolveTo( PCL_ALIEN_BUILDER0 )
                                                }
```

sweet235: `evolveTo` can indeed devolve

sweet235: To clarify: this PR changes the BT action `evolve`, which is usually done by each alien bot at each frame. It lets the bot choose the class to evolve to. This is different from the action `evolveTo( CLASS )`, tells the bot to evolve to `CLASS`.

My dislike of the term "devolving" stems from the fact that "evolve" and "devolve" sound like opposites, when in fact devolve is simply an evolve with an additional condition to be satisfied. Consequently, most of our evolve code handles devolves gracefully. I would rather NUKE all mentions of "devolve", and simply embrace the fact that there are (dynamic) conditions to be satisfied for an evolve to succeed.

sweet235: Here is a (probably incomplete) list of reasons why your attempt to evolve/devolve can fail:

- You are too poor
- You are trying to get bigger, but the map geometry is blocking you
- You are trying to get bigger, but a player is blocking you
- You are trying to get bigger, but a building is blocking you
- Your wall climb is active
- You are currently attacking
- You are trying to devolve too far away from the overmind

In the end, we should not make bots try to handle all these cases. It is best to make them run around and try until they succeed.

sweet235:
All right, I use `info.isDevolving` now. What's in a name? That which we call a rose by any other name would smell as sweet.

necessarily-equal:
I introduced `.isDevolving` at the time for situations where while you *could* devolve, you wouldn't want to do that ever, to gray them out in the UI. Maybe you can find a better name someday.

Then, about this:

> In the end, we should not make bots try to handle all these cases. It is best to make them run around and try until they succeed.

I totally agree. That's why I introduced https://github.com/Unvanquished/Unvanquished/pull/2175 and https://github.com/Unvanquished/Unvanquished/pull/2188. But I also tried to keep `botCanEvolveTo` functional, so you should be able to check if you can evolve with:

```c++
if ( BotCanEvolveToClass( self, c ) && G_AlienEvolve( self, c, false, /* dryRun = */ true ) )
```

But let's not get trapped into too much discussion. This PR is LGTM (without testing from me though).

PR #2834, opened 2023-10-29
make bots build in default.bt

sweet235:
This is my ~~first~~ attempt at making bots build bases ~~autonomously~~ (edit: they share information in the "blackboard" now). If there are build points available, or if there is no reactor/overmind or no miner, one bot in the team will build. The decision about which buildable to build is not part of this PR.

~~This is work in progress. A few problems to address before we can merge this:~~

- ~~Bots will build even if there are users on their team, they probably should not~~
- ~~In case bots cannot buy a ckit or devolve to granger, they might suicide within less than 20 seconds~~
- ~~Aliens bots might not build a booster for a long time, because they already used up their BP~~

### DIFF
--- a/pkg/unvanquished_src.dpkdir/bots/default.bt
+++ b/pkg/unvanquished_src.dpkdir/bots/default.bt
@@ -6,7 +6,7 @@ selectClass
 		{
 			selector
 			{
-				condition numUsersInTeam == 0 && usableBuildPoints >= chosenBuildableCost && !teamateIsClass( PCL_ALIEN_BUILDER0 ) && !teamateIsClass( PCL_ALIEN_BUILDER0_UPG )
+				condition levelTime >= 60000 && numUsersInTeam == 0 && usableBuildPoints >= chosenBuildableCost && !teamateIsClass( PCL_ALIEN_BUILDER0 ) && !teamateIsClass( PCL_ALIEN_BUILDER0_UPG )
 				{
 					spawnAs PCL_ALIEN_BUILDER0
 				}
@@ -17,7 +17,7 @@ selectClass
 		{
 			selector
 			{
-				condition numUsersInTeam == 0 && usableBuildPoints >= chosenBuildableCost && !teamateHasWeapon( WP_HBUILD )
+				condition levelTime >= 60000 && numUsersInTeam == 0 && usableBuildPoints >= chosenBuildableCost && !teamateHasWeapon( WP_HBUILD )
 				{
 					spawnAs WP_HBUILD
 				}
--- a/pkg/unvanquished_src.dpkdir/bots/default.bt
+++ b/pkg/unvanquished_src.dpkdir/bots/default.bt
@@ -6,7 +6,7 @@ selectClass
 		{
 			selector
 			{
-				condition buildPoints >= chosenBuildableCost && !teamateIsClass( PCL_ALIEN_BUILDER0 ) && !teamateIsClass( PCL_ALIEN_BUILDER0_UPG )
+				condition buildPointsNonneg >= chosenBuildableCost && !teamateIsClass( PCL_ALIEN_BUILDER0 ) && !teamateIsClass( PCL_ALIEN_BUILDER0_UPG )
 				{
 					spawnAs PCL_ALIEN_BUILDER0
 				}
@@ -17,7 +17,7 @@ selectClass
 		{
 			selector
 			{
-				condition buildPoints >= chosenBuildableCost && !teamateHasWeapon( WP_HBUILD )
+				condition buildPointsNonneg >= chosenBuildableCost && !teamateHasWeapon( WP_HBUILD )
 				{
 					spawnAs WP_HBUILD
 				}
--- a/pkg/unvanquished_src.dpkdir/bots/default.bt
+++ b/pkg/unvanquished_src.dpkdir/bots/default.bt
@@ -6,7 +6,7 @@ selectClass
 		{
 			selector
 			{
-				condition levelTime >= 60000 && ( numUsersInTeam == 0 || numOurBuildings( E_A_SPAWN ) < 2 ) && usableBuildPoints >= chosenBuildableCost && !teamateIsClass( PCL_ALIEN_BUILDER0 ) && !teamateIsClass( PCL_ALIEN_BUILDER0_UPG )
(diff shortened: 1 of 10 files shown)


DolceTriade:
This is what's live on your server?

DolceTriade:
Do you think this makes sense as a tactic?

sweet235:
> This is what's live on your server?

No. This is a cleaned up version. On my server, I change the bot behaviors automagically to make a bot build. But we worked on that, and now we can integrate this into the default behavior.

> Do you think this makes sense as a tactic?

Why not. Just add `build` to `g_tacticBehaviors`, and you can make any number of bots build. Notice, however, that this PR is only about `default.bt`, `build.bt` is an already existing example file on how to use the build action.

sweet235:
This is ready for review.

pkg/unvanquished_src.dpkdir/bots/default_humans.bt
```diff
@@ -2,6 +2,50 @@ selector
 {
 	behavior use_medkit
 
+	sequence
+	{
+		condition levelTime >= 60000
```
slipher, line 7: This stuff should be a block-style condition like `condition <expr> { ... }` rather than a sequence of condition then action, so that once there is another builder on the team the bot gives up on trying to become builder (see https://wiki.unvanquished.net/wiki/Formats/Behavior_tree#The_condition_keyword). Same goes for other team's bt

src/sgame/sg_bot_parse.cpp
```diff
@@ -332,6 +337,11 @@ static AIValue_t numOurBuildings( gentity_t* self, const AIValue_t *params )
 	return AIBoxInt( count );
 }
 
+static AIValue_t numOurSpawns( gentity_t* self, const AIValue_t* )
+{
+	return AIBoxInt( level.team[ G_Team( self ) ].numSpawns );
```
slipher, line 342: `numSpawns` updates on spawn (of the spawning structure) and death so the semantics are in fact identical to `numOurBuildings` FWIW.

pkg/unvanquished_src.dpkdir/bots/default_aliens.bt
```diff
@@ -1,5 +1,48 @@
...
+		selector
+		{
+			sequence
+			{
+				condition class != PCL_ALIEN_BUILDER0 && class != PCL_ALIEN_BUILDER0_UPG
+				selector
+				{
+					condition timeSinceMyTimerUpdate >= 20000
+					{
+						action suicide
```
slipher, line 18: If suicide is going to be used to reach the builder state, you really need some coordination so I think it would be better to use the change BT thing like you said you had on the server. Otherwise if there is no OM or no armory, the bots are all going to suicide simultaneously.

Alternatively the suicide bit could be removed. 20 seconds is a kind of long time to go without anyone on the team being killed anyway

pkg/unvanquished_src.dpkdir/bots/default_aliens.bt
```diff
@@ -1,5 +1,48 @@
 selector
 {
+	sequence
+	{
+		condition levelTime >= 60000
+		condition numUsersInTeam == 0 || numOurSpawns < 2
```
slipher, line 6: What's the story behind the num spawns < 2 exception?

sweet235: If all users are in the spawn queue and you have 0 or 1 spanws left, the bots will build new spawns. Even if there is are users on their team. Incidently, this forces a building user to leave 2 spawns at all times, which is universally accepted to be correct.

pkg/unvanquished_src.dpkdir/bots/default_aliens.bt
sweet235: What I did on my server turned out to be confusing to users and admins. I have come to believe that doing these things in default.bt is preferrable to changing BTs by backgroung magic.

I think a better way would be to make a general purpose mechanism for bots to share such information. Some suggested we call this a "blackboard".

sweet235: I gave each bot a "transient" field on the "blackboard", that is: an integer in its mind that is reset to zero at each frame. I also added a function to make bots the non-zero fields in their team. This is enough to ensure that only one living bot tries to become the builder.

sweet235:
When testing this, granger bots will not evolve to dretches, as described in https://github.com/Unvanquished/Unvanquished/pull/2833. For a quick test, simply set `g_freeFundPeriod 1` to make them evolve to something bigger than a dretch.

pkg/unvanquished_src.dpkdir/bots/default_aliens.bt
sweet235: Notice that, if no BT change is ever done automatically, we can allow any BT changes using the tactic command. No confusion will arise.

pkg/unvanquished_src.dpkdir/bots/default.bt
```diff
@@ -4,11 +4,25 @@ selectClass
...
+				}
+				spawnAs PCL_ALIEN_LEVEL0
+			}
 		}
 		condition team == TEAM_HUMANS
 		{
-			spawnAs WP_MACHINEGUN
+			selector
+			{
+				condition levelTime >= 60000 && ( numUsersInTeam == 0 || numOurBuildings( E_H_SPAWN ) < 2 ) && usableBuildPoints >= chosenBuildableCost && !teamateHasWeapon( WP_HBUILD )
```
slipher, line 20: So that the num spawns cvars work correctly I guess `numOurBuildings( E_H_SPAWN ) < 2` should be replaced with `(numOurBuildings( E_H_SPAWN ) < 2 && numOurBuildings( E_H_SPAWN ) < cvar( "g_bot_buildNumTelenodes" )` ?

P.S. you can split the condition on multiple lines. No magic whitespace in BT

slipher (commented):
OK looking pretty good, the blackboard method is nice.

One thing I would like to see is a cvar to disable the building. (can use the `cvar` function in a condition)

src/sgame/sg_bot_parse.cpp
```diff
@@ -394,6 +404,62 @@ static AIValue_t stuckTime( gentity_t *self, const AIValue_t* )
...
+}
+
+static AIValue_t myTimer( gentity_t *self, const AIValue_t* )
+{
+	return AIBoxInt( level.time - self->botMind->myTimer );
+}
+
+static AIValue_t levelTime( gentity_t *self, const AIValue_t* )
+{
+	return AIBoxInt( level.time );
```
slipher, line 441: Use `level.matchTime` so that it has the correct value after a `map_restart`

slipher:
LGTM

PR #2835, opened 2023-10-29
show g_maxMiners value in teamstatus

cu-kai:
If g_maxMiners is set, this should also be shown in the teamstatus print:

image


slipher:
LGTM

PR #2836, opened 2023-10-29
G_admin_action only when an in-game admin uses /bot

cu-kai:
This makes the output of the /bot command behave consistently. Now:

 - The console will never be identified when using /bot (from experience, this gets messy the more mapconfigs you have)
 - Any real player will be identified when using /bot ingame for any purpose aside from changing bot names (with all outputs adhering the rules of [`G_admin_action`](https://github.com/Unvanquished/Unvanquished/blob/bd1e8d58b4def06c661d8f97bbee15c4e787138b/src/sgame/sg_admin.cpp#L565-L566))

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -6099,11 +6099,27 @@ static bool BotAddCmd( gentity_t* ent, const Cmd::Args& args )
 	{
 		ADMP( QQ( N_( "Can't add a bot" ) ) );
 	}
+	else
+	{
+		if ( ent )
+		{
+			G_admin_action( QQ( N_("^3bot:^* $1$^* added a bot to the $2$ team with behavior ^3$3$$4$") ),
+			                "%s %s %s %s", 
+			                ent, 
+			                Quote( va( "%s%s^*", 
+			                           ( team == TEAM_ALIENS ? "^i" : "^d" ), 
+			                           BG_TeamName( team ) ) ), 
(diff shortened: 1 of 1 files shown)


src/sgame/sg_admin.cpp
```diff
@@ -6153,30 +6169,26 @@ static bool BotFillCmd( gentity_t *ent, const Cmd::Args& args )
 		return false;
 	}
 
-	for (team_t team : teams)
+	for (team_t t : teams)
```
cu-kai, line 6172: This change appeases the AppVeyor process. While this code isn't buggy this is a re-definition of `team` as I have already defined it at a higher scope.

src/sgame/sg_admin.cpp
```diff
@@ -6395,6 +6423,19 @@ bool G_admin_bot( gentity_t *ent )
 			// Will ignore non-bots, unfortunately you can't change human skill with a command
 			G_BotSetSkill( i, skill );
 		}
+		if ( ent )
+		{
+			std::string nameStr = "^3both teams^*";
+
+			if ( team != TEAM_NONE ){
+				nameStr = va( "the %s%s^* team", ( team == TEAM_ALIENS ? "^i" : "^d" ), BG_TeamName( team ) );
+			}
```
DolceTriade, line 6432: nit: if formatting is funky.

DolceTriade (approved):
except for if formatting

PR #2837, opened 2023-10-30
human bots walk when low on stamina

sweet235:
Because otherwise, they often get stuck in their own base, running against a building until they regain enough stamina to jump. This is especially bad for battlesuits. Do not make them walk while in combat, though.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -2545,12 +2545,10 @@ void BotCalculateStuckTime( gentity_t *self )
 	self->botMind->lastThink = level.time;
 }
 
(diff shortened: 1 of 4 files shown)


DolceTriade:
seems reasonable lgtm

src/sgame/sg_bot_util.cpp
```diff
@@ -2545,6 +2545,19 @@ void BotCalculateStuckTime( gentity_t *self )
 	self->botMind->lastThink = level.time;
 }
 
+static Cvar::Cvar<int> g_bot_walkAfterCombat( "g_bot_walkAfterCombat", "how many milliseconds must pass after combat until human bots are allowed to walk when low on stamina", Cvar::NONE, 3000 );
+
```
DolceTriade, line 2549: Is this cvar actually helpful?

sweet235: No idea.

sweet235: I NUKED it.

PR #2838, opened 2023-10-30
listplayers: split the printed count into players and bots

cu-kai:
This now tells people exactly how many real players and how many bots are connected.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -4047,8 +4047,10 @@ bool G_admin_listplayers( gentity_t *ent )
 	bool        canset = G_admin_permission( ent, "setlevel" );
 	bool	canseeWarn = G_admin_permission( ent, "warn" ) || G_admin_permission( ent, "ban" );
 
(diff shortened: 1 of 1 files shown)


DolceTriade:
output is <real players>, <bots> ( <total players> )?

cu-kai:
> output is `<real players>, <bots> ( <total players> )`?

Exactly

DolceTriade (approved):

slipher:
`(5 total)` rather than just `(5)` please

cu-kai:
> `(5 total)` rather than just `(5)` please

"total" cannot be translated. I know we have other deficiencies in our translations but I think it's obvious enough what the (total) is. It's only shown when there are bots present.

slipher:
> "total" cannot be translated.

It could if you want. 2 different format strings depending on whether there are bots.

cu-kai:
That does feel like unnecessary code duplication for something as simple as this. Do you consider this blocking?

illwieckz:
(Approved by me)

PR #2842, opened 2023-11-01
fix bots sometimes building two armouries, medistations, etc instead of only one

sweet235:
Currently, building bots rely on `self->botMind->closestBuildings` to find out if there is an existing armoury. This excludes a currently building armoury, which takes longer to complete than the build timer to expire. The result is that bots regularly build two drills, two armouries, two medistations, etc.

Additionally, current bot build code cannot specify to build, say, 3 armouries; whatever this would be good for.

A straightforward way to fix this would be to count each buildable type for each bot. But this would be rather expensive. The current implementation of bot building scales well only when avoiding loops over entities.

This PR counts the buildables in the existing main loop over all entities the server does at each frame. These numbers can be used in the next frame as a sufficiently accurate estimate for the building bots. This is as fast as it gets.

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1294,18 +1294,12 @@ static bool build( gentity_t *self, buildable_t toBuild )
 		return false;
 	}
 
-	if ( level.team[ G_Team( self ) ].botBuildingInThisFrame )
-	{
-		return false;
-	}
-
 	self->client->ps.stats[ STAT_BUILDABLE ] &= ~SB_BUILDABLE_MASK;
 	self->client->ps.stats[ STAT_BUILDABLE ] |= toBuild;
 	if ( self->client->ps.stats[ STAT_MISC ] == 0 )
 	{
 		BotFireWeapon( WPM_PRIMARY, &self->botMind->cmdBuffer );
 	}
-	level.team[ G_Team( self ) ].botBuildingInThisFrame = true;
 	return true;
 }
 
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -1289,13 +1289,7 @@ static bool build( gentity_t *self, buildable_t toBuild )
 		G_ForceWeaponChange( self, WP_HBUILD );
 	}
 
-	if ( SB_BUILDABLE_TO_IBE( self->client->ps.stats[ STAT_BUILDABLE ] ) != IBE_NONE )
-	{
-		return false;
-	}
-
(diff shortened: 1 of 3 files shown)


src/sgame/sg_main.cpp
```diff
@@ -2211,6 +2211,8 @@ void G_RunFrame( int levelTime )
 
 	G_CheckPmoveParamChanges();
 
+	int numBuildables[ BA_NUM_BUILDABLES ] = { 0 };
```
slipher, line 2214: Remove the 0, just `{}` (as is there should be a compile warning about too few elements to initialize)

slipher (commented):
I recall complaining about the use of `closestBuildings` while reviewing the first PR :)

Looks fine, assuming there is no more than one bot builder per team.

sweet235:
Now this allows for any number of builders in a team. At each frame, only bot per team may build a buildable. The action `buildNowChosenBuildable` fails now if the building does not succeed.

src/sgame/sg_bot_ai.cpp
```diff
@@ -1289,7 +1289,13 @@ static bool build( gentity_t *self, buildable_t toBuild )
 		G_ForceWeaponChange( self, WP_HBUILD );
 	}
 
-	self->client->ps.stats[ STAT_BUILDABLE ] = toBuild;
+	if ( SB_BUILDABLE_TO_IBE( self->client->ps.stats[ STAT_BUILDABLE ] ) != IBE_NONE )
+	{
+		return false;
+	}
+
+	self->client->ps.stats[ STAT_BUILDABLE ] &= ~SB_BUILDABLE_MASK;
```
slipher, line 1302: The buildable UI error state is only updated every 100 ms. Also the use of `|=` doesn't make sense here.

slipher:
> Looks fine, assuming there is no more than one bot builder per team.

To be clear, I'm not saying this is something that needs to be solved. It LGTM if you remove the last commit attempting to support more than one builder.

src/sgame/sg_bot_ai.cpp
sweet235: I try to set only the bits indicated in `SB_BUILDABLE_MASK` to `toBuild`. Is this wrong?

slipher: Never mind I misread it.

slipher:
Oops, I had never read the `botBuildingInThisFrame` code until now. The stuff I intended to say should be reverted is the STAT_BUILDABLE stuff, which will not work because it is not updated every frame. But the `botBuildingInThisFrame` doesn't seem great either because if you add more than one builder, all of them except one would do nothing most of the time.

Why not merge the approved changes and open a new PR for STAT_BUILDABLE and/or botBuildingInThisFrame stuff?

PR #2843, opened 2023-11-03
sgame command cleanup and documentation

slipher:


DolceTriade:
@sweet235 Do you generally do navcons locally or on a running server?

cu-kai:
> @sweet235 Do you generally do navcons locally or on a running server?

The only way to use navedit currently is in a local devmap session. It cannot be used on a live server.

sweet235:
Reading the commits, one can get the impression that it used to be possible to use these commands on a server. But I have no idea how. I have only ever used this locally too.

DolceTriade (approved):

PR #2844, opened 2023-11-03
buildlog: it is not possible for a building to be "unpowered"

cu-kai:
There are only 6 fates, this commit removes the irrelevant one from the buildlog code and fixes a bug where buildings that are "removed" are incorrectly shown as "unpowered" in the buildlog.


slipher:
LGTM

PR #2849, opened 2023-11-11
make bots always use adv granger when they should

sweet235:
Handle the two missing cases:

- spawn as adv granger when unlocked
- evolve from granger to adv granger when unlocked


illwieckz (approved):
LGTM, assuming you tested that works.

sweet235:
We did test this.

cu-kai (approved):
Tested on Bunker. LGTM.

PR #2854, opened 2023-11-17
Implement buildable death sounds

cu-kai:
This commit implements new feedback sounds for when buildables are initially killed. There are no new assets added, this code uses legacy assets previously unused by any Unvanquished code.


illwieckz:
I find funny that all Tremulous and Unvanquished releases shipped those sounds files since 2006 and maybe even before without making use of them. 😄️

I believe the human sound file sounds a bit like a cheap medieval RTS, but that's something we can fix later, and the file is not provided by the PR so I don't mind.

illwieckz (approved):
LGTM.

illwieckz:
That was tested on Der Bunker and the code looks good to me.

PR #2856, opened 2023-11-25
README enhancements

### LINKED ISSUE #2715: Questionable modding instructions in README
1. It is suggested to put all new/changed files for the mod in a new dpkdir which will constitute the package to distribute to players. This only works if you make sure to check out the latest release tag as the base for your mod branch, but no mention is made of this. In practice I think most people want to develop against the latest version of the code. In that case you need to either package the whole unvanquished_src.dpkdir or use Urcheon to get only the changed files.
2. It is suggested to put packages in a `pkg/` co-located with `daemon`. (This suggestion is made in other sections besides modding as well.)  IMO this is a poor workflow because it is frequently necessary to delete the build folder, be it because you switched compilers, a new external_deps package was released, or everything is just mysteriously broken. So I believe non-temporary assets should never be placed in the build folder. I always put released DPKs in the `pkg/` in the default homepath which is much nicer because then they are automatically found from *any* build directory without any extra command line flags.

slipher:
Fixes #2715.


README.md
```diff
@@ -130,24 +137,27 @@ cd Unvanquished/
...
 ```
 
 ### Linux cross-compile to Windows
 
 💡️ For a 32-bit build use the `cross-toolchain-mingw32.cmake` toolchain file instead.
 
 ```sh
-cmake -H. -Bbuild -DCMAKE_TOOLCHAIN_FILE=cmake/cross-toolchain-mingw64.cmake
-cmake --build build -- -j4
+mkdir build-win64; cd build-win64
```
illwieckz, line 158: Isn't it better to do?

```sh
mkdir -p build/win64; cd build/win64
cmake ../.. -DCMAKE_TOOLCHAIN_FILE=cmake/cross-toolchain-mingw64.cmake
…
```

Maybe personal taste, but I'm still surprised when I find people not knowing about it. I'm doing this since probably 10 years, so [I don't know what was annoying about ignoring only the `build/` folder](https://github.com/Unvanquished/Unvanquished/pull/2326).

illwieckz: Actually the path to the toolchain is wrong anyway, it is either:

```sh
mkdir -p build/win64; cd build/win64
cmake ../.. -DCMAKE_TOOLCHAIN_FILE=../../daemon/cmake/cross-toolchain-mingw64.cmake
```

or:

```sh
mkdir -p build-win64; cd build-win64
cmake .. -DCMAKE_TOOLCHAIN_FILE=../daemon/cmake/cross-toolchain-mingw64.cmake
```

or:

```sh
cmake -H. -Bbuild/win64 -DCMAKE_TOOLCHAIN_FILE=daemon/cmake/cross-toolchain-mingw64.cmake
```

or:

```sh
cmake -H. -Bbuild-win64 -DCMAKE_TOOLCHAIN_FILE=daemon/cmake/cross-toolchain-mingw64.cmake
```

README.md
```diff
@@ -130,24 +137,27 @@ cd Unvanquished/
...
 ```
 
 ### MSYS2
 
 ```sh
-cmake -H. -Bbuild -G "MSYS Makefiles"
-cmake --build build -- -j4
+mkdir build; cd build
+cmake .. -G "MSYS Makefiles"
+make -j4
```
illwieckz, line 150: I guess this is the most “old cmake-compatible” way, otherwise less old cmake can do:

```sh
cmake -S. -Bbuild -G"MSYS Makefiles"
cmake --build build --parallel 4
```

Which is totally generator agnostic.

If one wants to run build commands in folder he can do:

```sh
cmake -S. -Bbuild -G"MSYS Makefiles"
cmake --build build --parallel 4
```

or:

```sh
mkdir -p build ; cd build
cmake -S.. -B. -G"MSYS Makefiles"
cmake --build . --parallel 4
```

or:

```sh
mkdir -p build ; cd build
cmake .. -G"MSYS Makefiles"
cmake --build . --parallel 4
```

README.md
```diff
@@ -107,6 +108,12 @@ You can do `./download-paks --help` for more options.
 ./download-paks build/pkg
 ```
 
+The above snippet stores the downloaded paks in the
+[libpath](https://wiki.unvanquished.net/wiki/Libpath) of the Daemon engine that you are building,
```
illwieckz, line 112: The “in the libpath” wording may suggest the dpk archives are not stored in `libpath/pkg` but in `libpath`.

illwieckz:
- `README: nuke modding instructions`:   
LGTM.
- `README: encourage downloading paks to the homepath`:  
It could benefit from a less-ambiguous phrasing.
- `README: cd into build for cmake commands`:  
There is at least one error we may fix (path to toolchain is wrong and was already wrong).

For commands we recommends we may have a more deep talk about what we want to recommend.

README.md
slipher: OK I changed the wording.

README.md
slipher: Fixed the toolchain path.

Yeah that's how I have it set up on my machine - all the build directories are subdirectories of `build`. Just keeping it simple for the README. Also if in other parts of the README we suggest that `build` itself is used, I don't want to make suggestions that would have one build dir inside another if used together.

slipher:
Merged the LGTMed commit.

illwieckz (approved):
LGTM

Issue #1112, opened 2019-04-20
single server browser list for LAN and Internet servers


illwieckz:
Just put LAN servers above Internet servers if they exist.

Like [Quetoo](http://quetoo.org/):

[![quetoo server browser](https://dl.illwieckz.net/b/quetoo/screenshots/20171227-121016-000.quetoo-server-browser.png)](https://dl.illwieckz.net/b/quetoo/screenshots/20171227-121016-000.quetoo-server-browser.png)

Simplifies UI.

This requires changes in [unvanquished_src.dpkdir](http://github.com/UnvanquishedAssets/unvanquished_src.dpkdir) too.

Dimitriio:
Hello @illwieckz,
i am working on it. I have done for now the UI. As you can see, i removed the tabs and have added the "Source" column.
But i have an issue with the data. For now it is split in 2 arrays. And it seems quite difficult merging them without changing how it's done as all events are called through the generic `dataSourceCmd_t`.
![Screenshot from 2020-07-18 22-25-45](https://user-images.githubusercontent.com/9897777/87861419-56d64c80-c946-11ea-950c-0b5a767fb874.png)

illwieckz:
Hi, that's nice to see! :+1: 

Do we need to merge them? We may just list LAN servers atop Internet server list, but in this case the issue is: how to not have two heading cells (“Name, Source, Map…”), and we may also have a scrolling issue: if we do two lists we may want to scroll a kind of frame containing the two lists and not to scroll the lists themselves…

Dimitriio:
Well, the source of the datagrid is only one array, so if we want one datagrid to be shown (with the sorting), we need to merge them, atleast in the exposed data that the grid will use. 

For now the source can be `server_browser.internet` or `server_browser.local` so to show both, it needs some adjustements. The model is inside the variable `rocketInfo.data.servers[ netSrc ]` where netSrc correspond to `AS_LOCAL 0`, `AS_GLOBAL 1`, `AS_FAVORITES 2`

And what could be bothering too is both filtering & sorting if we consider using 2 datagrid.

We could keep `rocketInfo.data.servers[ netSrc ]` and just directly add both LAN & internet to the data but we will loose the capabilities to do it separatly for only internet or local in the future. And it won't have the same comportment as exec, build, clean, ect that will still need the `netSrc` to be working (except if we remove the generecity of all methods, but i dunno the actual impacts).

slipher:
[...]

PR #2857, opened 2023-11-29
spawn server browser at startup

illwieckz:
As mentionned in #2826:

- https://github.com/Unvanquished/Unvanquished/pull/2826

I believe the centering of the main menu is a workaround for the issue of not providing in the center of the screen what the user is actually looking for.

The user is looking for a game to play, the user is not looking for an UI to use.

There is no main menu to use to join a game if the server browser is automatically opened at game startup, meaning the main menu placement becomes non-critical.

About joining games:

- This change reduces the amount of clicks to join an Internet server.
- This change reduces the amount of clicks to join a LAN server.
- This change keeps the same amount of clicks to start a server.

About other actions:

- This change keeps the same amount of clicks to do other actions like tweaking options.

I see no drawback on the topic of user interaction.  Everything is either easier to do, or like before.

This effort is a continuation on what was started in UnvanquishedAssets/unvanquished_src.dpkdir#14:

- https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/14

We already show to the user the very first thing he looks for while being a new user (nickname and language), now we would show to the user the very first thing he looks for while not being a new user (server list).


illwieckz:
Further improvements can be done over this:

- listing LAN servers atop of net servers in a single list: https://github.com/Unvanquished/Unvanquished/issues/1112
- add a checkbox to not look for Internet servers

This may eventually be done too:

- make the server creation form a tab of that window
- remember what was the previous tab opened (server browser or server creation)

But this is not for this PR, I just mean this PR is part of a whole UI redesign thought for years to makes things easier for users, we're just slow to implement it.

slipher:
It's kind of obnoxious for the game to immediately start sending a bunch of network traffic on startup, even though it is unknown whether the user intends to go online.

Grise3:
>  That's why I would like to see an option to not query Internet.

And right now, if you have a bad network, the time it takes for the information to be received, everything will be frozen, you won't be able to do anything.
Imagine how great that is
you run the game : you have to wait for can use the menus :D

for the window, something like this can be done : 
whynot

PR #2859, opened 2023-11-29
ui: fix server label in server list

illwieckz:
Fix the display of the server label in server list.

I don't know what the original code was for, but deleting it fixes all the bugs:

- it doesn't seem to eat or hide the first character
- emoticons work

If I use this `featured.txt` file with unvanquished master server:

```
[official]
	206.189.56.213:27960
	82.165.236.249:27960
```

I get this in game:

unvanquished_2023-11-29_193755_000


slipher:
LGTM

PR #2863, opened 2023-12-03
Fix attack warning bugs

### LINKED ISSUE #2472: `EV_WARN_ATTACK` message may contain an incorrect location
I have observed on both Bunker and Sweet's server on the map `operation-dretch` - the location marker for a base which is under attack is sometimes incorrect, displaying a location nowhere near where the base is being attacked. This does not always happen and the locations it occurs in are inconsistent. 

This may be reproducible on operation-dretch by opening the 2nd forward base and letting the alien bots destroy the armoury within. The game reports that the final Overmind room is under attack, and although that room is nearby (but separated by a thick wall), there are closer location markers than the Overmind room.

The other location I saw this occur was near the end of the final halls on that map, an armoury was destroyed and the server reported "Under attack - 1st large room" which is at the opposite end of the map.

slipher:
The commit `Simplify EV_WARN_ATTACK location lookup` might not fix #2472 (I haven't really tried to reproduce it), but at least it makes things simpler so there is less chance to go wrong.

### DIFF
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -1286,22 +1286,13 @@ void CG_EntityEvent( centity_t *cent, vec3_t position )
 			// if eventParm is non-zero, there is a nearby main structure (OM/RC). Humans are only notified
 			// of base attacks if RC is in range. Aliens are notified of any base attacks, but base attacks
 			// without the OM in range do not give you the location of the attack.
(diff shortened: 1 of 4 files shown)


cu-kai:
It doesn't fix #2472. I have tested this patch against my branch and on the map `grangermaze`.

cu-kai (approved):
Now confirmed to fix the issue with `EV_WARN_ATTACK`. Thanks!

PR #2876, opened 2023-12-15
@bmorel: MAP: TOGGLE spawnflag for doors: ignore wait

cu-kai:
This is a patch by @bmorel which I have been asked to submit on his behalf for political reasons.

This patch will allow `func_door` and `func_door_rotating` entities to be toggled open/closed instead of waiting for x seconds to close and removes the necessity for mappers to use an ugly and time consuming `func_train` hack to create toggle-able doors. This patch is required for the new map `grangermaze` to work correctly and both @sweet235 and myself run this on our servers with no side effects.

Bitflag value 256 will apply here and on successful approval and merging of this PR I will also submit the relevant (already-tested) patch to https://github.com/Unvanquished/unvanquished-mapeditor-support.

This PR can be tested with the latest version of the "test-doors" map in https://github.com/UnvanquishedAssets/UnvanquishedTestAssets

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -45,7 +45,9 @@ static Cvar::Range<Cvar::Cvar<int>> g_bot_ignoreSmallObstacles(
 		"0 is disabled. Should only be enabled on maps with known problem."
 		"Maximum is 18, which is bots' stepsize.", Cvar::NONE, 0, 0, 18 );
 
-constexpr int ACTIVATE = 0x80; // spawnflags with this bit can be triggered by `+activate`
+constexpr int ACTIVATE = 1 << 7; // spawnflags with this bit can be triggered by `+activate`
+constexpr int TOGGLE   = 1 << 8; // those doors and buttons will ignore the "wait" value and will not automatically return in position
+
 /*
 ===============================================================================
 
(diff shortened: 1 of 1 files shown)


src/sgame/sg_spawn_mover.cpp
```diff
@@ -1859,7 +1880,7 @@ void SP_func_door_rotating( gentity_t *self )
 			self->die = Think_MoverDeath;
 		}
 	}
-	else
+	else if ( not ( TOGGLE & self->mapEntity.spawnflags ) )
```
slipher, line 1883: Why do other doors need to create a "door_sensor" entity but toggling doors don't?

cu-kai: Because you don't want a door that requires being toggled to automatically open when a player approaches it (as it would never close again, or would close if somebody else approached the door while in the open state).

Doors with a name/targetname will also not create the `door_sensor`.

slipher: I guess if the door is toggled by another entity then it has to have a name? So the "else if" would never be reached anyway.

cu-kai: Yes, it would need a name in that context. That being said, hypothetically, the entity could be fired by other means (entityFire command, even one summoned by lua).

slipher:
LGTM

PR #2881, opened 2023-12-23
reduce g_bot_alienAimDelay to 150 (as it used to be)

sweet235:
Use the same aim delay for alien bots as for human bots again. Lower numbers mean faster aim. This delay has been experimentally worsened during the random skilltree experiments, but never changed back once the random skills were disabled by default.


illwieckz (approved):
If that's what used to be, LGTM.

sweet235:
Also, me and @cu-kai use this on our servers.

PR #2883, opened 2023-12-27
cgame: preselect a map in server creation window, remember it

illwieckz:
Preselect a map in server creation window, this means a newcomer has less actions to do to create a server when trying the game for the first time, and then remember the last started map.

unvanquished_2023-12-27_024133_000


src/cgame/cg_rocket_datasource.cpp
```diff
@@ -1119,16 +1121,33 @@ static void CG_Rocket_BuildMapList( const char* )
...
+		{
+			rocketInfo.data.mapIndex = i;
+		}
 
 		Rocket_DSAddRow( "mapList", "default", buf );
 	}
 
+	if ( rocketInfo.data.mapIndex == -1 )
+	{
+		if ( rocketInfo.data.mapList.size() != 0 )
```
DolceTriade, line 1146: can combine these two conditions. can also write as:

`rocketInfo.data.mapIndex == -1 && !rocketInfo.data.mapList.empty()`

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -1137,7 +1156,20 @@ static void CG_Rocket_CleanUpMapList( const char* )
 
 static void CG_Rocket_SetMapListIndex( const char*, int index )
 {
-	rocketInfo.data.mapIndex = index;
+	if ( index >= 0 && index < int( rocketInfo.data.mapList.size() ) )
```
DolceTriade, line 1159: if we're checking, let's also warn when the else case happens

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -1119,16 +1121,33 @@ static void CG_Rocket_BuildMapList( const char* )
 	Rocket_DSClearTable( "mapList", "default" );
 	CG_LoadMapList();
 
+	rocketInfo.data.mapIndex = -1;
+
 	for ( size_t i = 0; i < rocketInfo.data.mapList.size(); ++i )
 	{
 		char buf[ MAX_INFO_STRING ] = { 0 };
+
+		std::string mapLoadName = rocketInfo.data.mapList[ i ].mapLoadName;
```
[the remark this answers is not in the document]

src/cgame/cg_rocket_datasource.cpp

src/cgame/cg_rocket_datasource.cpp
illwieckz: Well, this part was cargo-cult copy-paste, but I added the warning.

illwieckz: Well, that's not a good idea, the game prints this warning at startup:

```
Warn: Map list index -1 is out of range [0, 0[ 
Warn: Map list index -1 is out of range [0, 0[ 
Warn: Map list index -1 is out of range [0, 0[ 
```

DolceTriade: so before we were setting mapIndex to -1 and everything was working ok? Does this mean with our current code, we cannot "unselect" a map? -1 means no selection

illwieckz: I assume we cannot unselect a map but I don't see that as an issue, emptying `g_defaultMap` will select the first one, one cannot create a server without a map anyway. It also makes the UI consistent by always displaying the levelshot, meaning the position of the UI is always consistent: I can center it vertically with a levelshot displayed and expect it to always look good.

illwieckz:
I added a commit that forces the height of the levelshot. I forced `16:9` ratio. This is not too bad, even on old legacy maps that had `4:3` levelshots.

This prevents the server creation window size to change dynamically while scrolling maps, hence preventing the “Start” button to move up and down while scrolling map, and it would prevent a nasty map with a nasty `1:1000` levelshot to move the “Start” button off screen.

slipher:
I don't have time to review or test, but I would ask to ensure that this does *not* cause a map pak to be loaded before opening the map menu.

DolceTriade (approved):

DolceTriade:
> I don't have time to review or test, but I would ask to ensure that this does not cause a map pak to be loaded before opening the map menu.

This shouldn't do that

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -34,6 +34,8 @@ Maryland 20850 USA.
 
 #include "cg_local.h"
 
+static Cvar::Cvar<std::string> g_defaultMap("g_defaultMap", "default map to play", Cvar::ARCHIVE, "plat23");
```
slipher, line 37: Should be `cg_`

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -1137,7 +1153,20 @@ static void CG_Rocket_CleanUpMapList( const char* )
...
+		rocketInfo.data.mapIndex = index;
+		g_defaultMap.Set( rocketInfo.data.mapList[ index ].mapLoadName );
+	}
+}
+
+static int CG_Rocket_GetMapListIndex( const char* )
+{
+	if ( rocketInfo.data.mapList.size() != 0 )
+	{
+		CG_Rocket_BuildMapList( nullptr );
```
slipher, line 1167: Build the list only if it is already nonempty? Seems weird. Probably this shouldn't be here at all?

illwieckz: Ah no it should be `== 0`, well, I copypasted `CG_Rocket_GetResolutionListIndex` the cargo cult way, so I don't know if that code is needed, but when duplicating, that should be `==`. I'll fix that.

slipher: Since the code worked even with `!=0` this shouldn't be needed at all.

PR #2887, opened 2024-01-02
New crosshairs

Grise3:
New crosshairs,
they are more visible
The crosshairs have a black outline, which makes them more visible on a white surface.
note : they're bigger too, but we can change in the settings.
I've chaned the values of the crosshairs sizes in the settings, it is more easy for the users :D
Now it is :
Tiny  0.4
Small = 0.5
Regular = 0.6
Huge = 0.8


Grise3:
the crosshairs :

list

the indicators : 

indicator

illwieckz:
Great! You can also delete the `pkg/unvanquished_src.dpkdir/gfx/feedback/crosshairs.xcf` to avoid confusion as it is not anymore the source of the new images.

For the crosshair scale we may add some factor in code to keep the original `cg_crosshairSize` value.

Grise3:
`For the crosshair scale we may add some factor in code to keep the original cg_crosshairSize value.`
We can, but if we have to do it in C++ I can't help.

DolceTriade:
can we get in game screenshots?

Grise3:
Do you want screenshots of all the weapons + indicators?

DolceTriade:
sure, that would be ideal.

Grise3:
Well, currently I can't, I don't have my computer. But, I have this with me :
Yvw
 [](url)
Screenshot_20240102_211427

Screenshot_20240102_211454

Grise3:
@DolceTriade here is the rest :

Screenshot_20240103_185209

Screenshot_20240103_185258

Screenshot_20240103_185332

Screenshot_20240103_185420

Screenshot_20240103_185530

Screenshot_20240103_185636

Screenshot_20240103_185726


Is that what you wanted?

DolceTriade:
Some of these are pretty hard to see imo. I'll test it out later to confirm myself

Grise3:
Which ones are hard to see?

DolceTriade:
Just based on those screenshots...all of them :(

Grise3:
Sure, we are on the fire, there are the damage numbers.... This hide it :)

Maybe you have to try them for see :D

Grise3:
When I see  #2921, I think that this PR  is not that useful.
and after what said I don't remember who, the crosshairs have to be in SVG.

Then, these crosshairs are not useful :)
Problem is : For svg files, as said @cu-kai (I think It was him) for svg crosshairs we need something that know how to do that ^^

Anyways, I close this PR, and I focus on the UI

PR #2888, opened 2024-01-02
tools: add git-version to print a package version from git references

illwieckz:
This is a simple wrapper around `daemon/tools/git-version/git-version`, but reporting Unvanquished version.

This way, when doing that:

```
daemon/tools/git-version/git-version
tools/git-version/git-version
```

The first command prints the engine version string, and the second one the Unvanquished version string.

The Unvanquished versions string is expected to be the same of the unvanquished dpk version string, and then usable to be used as unizip version string and symbols archive version string.

See also:

- https://github.com/DaemonEngine/Daemon/pull/1009

### DIFF
--- a/tools/git-version/git-version
+++ b/tools/git-version/git-version
@@ -1,0 +1,35 @@
+#! /usr/bin/env python3
+
+# Daemon BSD Source Code
+# Copyright (c) 2024, Daemon Developers
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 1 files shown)


illwieckz:
This is a simple wrapper to make things easy for users using the default repository layout.

If we want to compute version strings in Unvanquished CMake, we better want to do instead:

```sh
"${DAEMON_DIR}/tools/git-version/git-version"
"${DAEMON_DIR}/tools/git-version/git-version" "${CMAKE_CURRENT_SOURCE_DIR}"
```

To get both Daemon and Unvanquished version string while relying on `${DAEMON_DIR}`.

illwieckz:
Once merged I'll make `release-scripts/build-release` to use it instead of the `printVersion` bash implementation this script uses.

I plan to make the docker-build script producing the unizip with version strings using such script.

illwieckz:
Obsolete.

PR #2899, opened 2024-01-12
Fix bind menu regressions

slipher:


illwieckz (approved):
I will have to trust you on this one, LGTM.

PR #2900, opened 2024-01-15
Allow using binds in some circle menus

### LINKED ISSUE #1195: Opening build menu disable wallwalk when using non-toggled wallwalk
I'm surprised to not find an already created issue for that since I'm sure I talk about it since a long time…

But when you use non-toggled wallwalk and open the build menu, the build menu intercept your keys so you fall.

slipher:
Stacked on #2899. It's not really entwined with that, but first I had to fix the bind menu, to ensure that I am not breaking it.

Engine companion: https://github.com/DaemonEngine/Daemon/pull/1030

> It is now possible to use binds while the evolve menu, build menu, or
beacon menu is open. So, for example, you can walk forward while
choosing an alien class, or use wallwalk in non-toggling mode while
choosing a structure as Advanced Granger.

> A menu may be configured to allow binds at the same time by adding the
"passthrough" keyword in rocket.txt. This means that the menu will only
capture keystrokes that do something in the menu (e.g. numbers in circle
menus), while unconsumed keystrokes will be processed as binds. Note
that mouse clicks and mouse movements are still always captured by any
open menu.

> I excluded the armory menu because it might be frustrating for purchases
to fail due to having moved too far from the armory.

Fixes #1195.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/alien_build.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/alien_build.rml
@@ -121,7 +121,7 @@
 	end
 </script>
 </head>
-<body id="alien_build" class="circlemenu alien" onkeydown="CirclemenuHandleKey(event, document, AlienBuildableClick)" onshow="BuildAlienBuildMenu(document)">
+<body id="alien_build" class="circlemenu alien" onkeydown="CirclemenuHandleKey(event, document, AlienBuildableClick)" onshow="CloseOtherCirclemenus(event) BuildAlienBuildMenu(document)">
 <tabset class="circlemenu">
 <tab><translate>Build</translate></tab>
 <panel>
--- a/pkg/unvanquished_src.dpkdir/ui/alien_evo.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/alien_evo.rml
@@ -136,7 +136,7 @@
 	end
 </script>
 </head>
-<body id="alien_evo" class="circlemenu alien" onkeydown="CirclemenuHandleKey(event, document, EvolutionClick)" onshow="BuildEvolveMenu(document)">
+<body id="alien_evo" class="circlemenu alien" onkeydown="CirclemenuHandleKey(event, document, EvolutionClick)" onshow="CloseOtherCirclemenus(event) BuildEvolveMenu(document)">
 <tabset class="circlemenu">
 <tab><translate>Evolve</translate></tab>
 <panel>
--- a/pkg/unvanquished_src.dpkdir/ui/human_build.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/human_build.rml
@@ -122,7 +122,7 @@
 	end
 </script>
 </head>
-<body id="human_build" class="circlemenu human" onkeydown="CirclemenuHandleKey(event, document, HumanBuildableClick)" onshow="BuildHumanBuildMenu(document)">
+<body id="human_build" class="circlemenu human" onkeydown="CirclemenuHandleKey(event, document, HumanBuildableClick)" onshow="CloseOtherCirclemenus(event) BuildHumanBuildMenu(document)">
 <tabset class="circlemenu">
 <tab><translate>Build</translate></tab>
 <panel>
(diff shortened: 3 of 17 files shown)


slipher:
It would be nice to get some testing on this, since input is something that behaves a bit different between operating systems. I only tested on Windows so far. Please inform me if your machine becomes wedged, unable to receive inputs...

DolceTriade:
@sweet235 Do you think we could run this on your server to get feedback?

slipher:
> @sweet235 Do you think we could run this on your server to get feedback?

No because it requires engine changes.

slipher:
I found a bug which is that if you open more than one circle menu at once (e.g. alien build + evolve), after closing the top one, the next one does not automatically grab focus (so it doesn't respond to keyboard input). Probably I should just close any other circle menus whenever opening one of them.

DolceTriade:
tested and seemed ok to me

slipher:
> found a bug which is that if you open more than one circle menu at once (e.g. alien build + evolve), after closing the top one, the next one does not automatically grab focus (so it doesn't respond to keyboard input). Probably I should just close any other circle menus whenever opening one of them.

Fixed this. Now when you open a circle menu other circle menus will close.

Also I added a new feature: if you disable `cg_circleMenusCaptureMouse` you can keep using the mouse to move around and use mouse button binds while in a circle menu. (So you have to select something with number keys.)

src/cgame/cg_rocket_events.cpp
```diff
@@ -70,6 +71,24 @@ static void CG_Rocket_EventHide()
...
+	{
+		if ( menu.passthrough && menu.id && skip != menu.id )
+		{
+			if ( Rml::ElementDocument *document = menuContext->GetDocument( menu.id ) )
+			{
+				document->Hide();
+			}
+		}
+	}
+}
```
DolceTriade, line 91: Can we just implement this in pure lua to avoid the round trip through our event system?

DolceTriade: Basically I think this could just be a "HideOthers()" function that would close all non

slipher: I think it's a good idea to use the pushevent thing in case `Hide` or other document functions are not reentrant. Wouldn't want to have the hide function called while the show function is executing or anything like that.

DolceTriade: Personally, it seems like a HideOthers is way more applicable. In fact, I'm not even sure the passthrough part of this is relevant. You probably want to close *all* other menus except the circle menu that just opened. 

Lastly, if you choose not to make this change, at the very least you shouldn't hard code the id. You should be able to get it by calling `document.id` or the equivalent lua.

DolceTriade (approved):
This is fine for now, but for future cleanup, we should avoid using our hacky events to control elements.

They were added because we didn't have Lua. We should just call `Blur()`

I don't think we have to worry about being re-entrant because this is how RmlUI was intended to be used...

slipher:
Thanks for the review; this seems a lot better than the initial CloseOtherPassthrough at least.



> This is fine for now, but for future cleanup, we should avoid using our hacky events to control elements.
> 
> They were added because we didn't have Lua. We should just call `Blur()`
> 
> I don't think we have to worry about being re-entrant because this is how RmlUI was intended to be used...

Maybe, but Rmlui has a history of crashing so I definitely wouldn't trust it to react well in case our code did something stupid like close or open a document that was already in the process of being closed.

DolceTriade:
> Maybe, but Rmlui has a history of crashing so I definitely wouldn't trust it to react well in case our code did something stupid like close or open a document that was already in the process of being closed.

I guess. Theoretically, using our events and using the lua directly should do the exact same thing. I don't think our events run on the next frame or anything. They run immediately, so they will ultimately call `document->Blur()` just like `document:Blur()` in the Lua code.

slipher:
Using pushevent ensures that the stack will be unwound between document show/hide calls. With calling them directly, you can potentially call Show or Hide on a document which is already executing one of these functions further up in the stack.

PR #2902, opened 2024-01-17
voting improvements - add colours and formatting

cu-kai:
This PR changes the vote style to be less ugly. Now it will no longer to sometimes wrap text in an ugly way if the string is too long. It will also now use Quake colour codes and display name colours correctly.

A CenterPrint message is also sent to the relevant clients when the vote is started.

image

The countdown timer will become red after 10 seconds in the hope that it draws more attention to itself.

image

In future perhaps we could replace the check / cross emoticons to be nicer than these ones but for now I think this is acceptable.


illwieckz:
You can merge that PR and submit the translatable stuff later.

PR #2901, opened 2024-01-17
generate navmeshes for advanced granger, make it climb walls (using navcons)

sweet235:
This works if accompanied by a PR for `res-players` (https://github.com/UnvanquishedAssets/res-players_src.dpkdir/pull/16), where we currently make the advanced granger bots use the same navmeshes as the normal granger. If we give the advanced granger its own navmesh, it can have its own navcons too.

PR #2905, opened 2024-01-23
Put navgen use-other-class setting in code + optionally generate less

slipher:
Some ideas discussed in #2901 comments. The settings for which player classes should use another class's navmesh instead of generating its own are moved into the code so that res-players does not have to be updated. Introduce a cvar to generate less classes so that generation takes less time.

Thoughts on whether the cvar should be on or off by default?

If people like this I will make a PR to res-players to remove the setting, and also add a UI checkbox for the cvar.

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1065,7 +1065,13 @@ static void GenerateNavmeshes()
 	std::string mapName = Cvar::GetValue( "mapname" );
 	std::vector<class_t> missing;
 	NavgenConfig config = ReadNavgenConfig( mapName );
-	for ( class_t species : RequiredNavmeshes() )
+	bool reduceTypes;
+	if ( !Cvar::ParseCvarValue( Cvar::GetValue( "g_bot_navmeshReduceTypes" ), reduceTypes ) )
+	{
+		ASSERT_UNREACHABLE(); // sgame should have been loaded and this cvar initialized
+	}
+
+	for ( class_t species : RequiredNavmeshes( reduceTypes ) )
 	{
 		fileHandle_t f;
 		std::string filename = NavmeshFilename( mapName, species );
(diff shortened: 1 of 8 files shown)


sweet235:
Well, isn't this premature optimization? Why bother adding these obstacles to the problems of people working on this in the future?

illwieckz:
The maintenance cost of tricks to save on navgen generation time by mutualizing them is too high compared to the cost of waiting for a navmesh to be generated.

The cost of waiting for navmesh to be generated is just a one time waiting for some time and forget it. There is zero mental load.

The cost of tricks brings maintenance cost: dealing with the tricks in the future, making sure we don't introduce regressions when implementing other things, we will have to mind them, etc. This just adds mental load for our future self.

I also agree this is too early, we're still exploring “how to make basic things”.

illwieckz:
Actually the navmesh reduction could be made fully automatic by using the player properties as key in a cache, basically instead of caching a light human and medium human navmesh, we would cache a “navmesh for a player with this size, this capability, etc.”

slipher:
OK, I made the cvar off by default so you don't have to worry about it. Only people obsessed with load times will turn it on.

I also updated the cvar's behavior to substitute the human battlesuit mesh for non-battlesuit types since the only difference is that battlesuit is taller.

Out of curiosity to see what proportion of time is saved, I recorded the time for each class's navgen on Thunder:

```
Granger: 2003
Advanced Granger: 2008
Dretch: 3333
Mantis: 3333
Marauder: 1574
Advanced Marauder: 1396
Dragoon: 1260
Advanced Dragoon: 1061
Tyrant: 852
Base Human: 3305
Battlesuit: 3265
```

For this map it would be 44% faster with the cvar. The savings are proportionally greater than the number of skipped classes, because the shared navmeshes are concentrated in classes with small size, which take longer to generate.

sweet235:
I think I have another use case for this mechanism: when you have a map where the navmesh generation fails for some, but not all, classes. And no tweaking you can think of can unravel this mystery. This happens to me occasionally.

Quite often, you can resort to using the navmesh of a bigger class instead. We could integrate the proposed mechanism into the .navcfg files later.

sweet235:
LGTM

slipher:
https://github.com/UnvanquishedAssets/res-players_src.dpkdir/pull/17 will clean up the player configs

Issue #2357, opened 2023-01-07
Some big maps get only partial navmeshes


sweet235:
I tried this with recent `daemonmap`, but I think this is deprecated, so I report it here.

Here is an example of a map where a part did not receive a navmesh:
![unvanquished_2023-01-07_095906_000](https://user-images.githubusercontent.com/110394607/211166638-74475e93-d5d5-4e86-94c2-a1d64c567b85.jpg)

The map is this one:
https://users.unvanquished.net/~sweet/pkg/map-dryout_2%2Bb5a.dpk

Other maps where this happens are:
https://users.unvanquished.net/~sweet/pkg/map-foggy_0%2Bb1.dpk
https://users.unvanquished.net/~sweet/pkg/map-cube_0%2Bb2.dpk

necessarily-equal:
Some things to consider:
1. Sometimes the navmesh can be slightly under the ground, which seems to be the case at some places in your screenshot, like here:
![image](https://user-images.githubusercontent.com/59283660/211192044-27056e93-584b-4883-9500-cc948c104afd.png)
2. As you said daemonmap is deprecated. And navemesh generation actually changed a bit for "patch" https://github.com/Unvanquished/Unvanquished/pull/1646#discussion_r780622717. "Patch" are wavy shapes instead of convex polygons ("brush"). This means that for the wavy parts like the outside you have shown the navmesh generation is different, relying only on more crude simplified shapes that are hopefully precise enough (recast does integer rounding anyhow, so .

necessarily-equal:
To test the new navmesh generator, you can open any map, do `/navgen all` and `/map_restart` in the console. This should be enough.

sweet235:
Using 0.54 ingame navgen, I get:
```
==== Bot Navigation Initialization ==== 
 loading navigation mesh file 'maps/dryout-level0.navMesh'... 
Warn: Can't load navmesh for level0: Cached navmesh generation failure (Map is too tall to generate a navigation mesh) 
Warn: Cannot load navigation mesh file for level0 
Failed to load navmeshes 
Warn: Navigation mesh files unavailable for this map 
```

I'm not sure what the error message wants to say. Are my old navmeshes too big?

slipher:
There was some code to mitigate the "map too tall" situation which I deleted right before the release because it didn't seem to work. Maybe it actually did work in some cases though. There is https://github.com/DaemonEngine/daemonmap/pull/13/files attempting to make a better version of it

[...]

PR #2909, opened 2024-01-24
add navcfg option cellSizeFactor

sweet235:
There is a hardcoded magic number controlling a crucial parameter of the navmesh generation: its "cell size". The cell size determines how big a navmesh's tiles will be.

As it turns out, https://github.com/Unvanquished/Unvanquished/issues/2357 can (at least almost) always be resolved by tweaking this number. This PR adds a field `cellSizeFactor` to the .navcfg files to allow for that. A mapper or server operator may choose any number they see fit.

The magic number is in `src/shared/navgen/nav.cpp`, it is the divisor 4 in:

```
const float cellSize = radius / 4.0f;
```

The `radius` is the radius of the class. 

This PR allows to specify any number (in a sensible range) instead of this number 4. We all hate divisions, so let's make this a multiplication by 0.25 by default.

For example, look at the map `void2` (you can find it here: https://github.com/Masmblr/map-Void2_src). With the default value of 0.25, you see this when doing `/navedit enable level0`:

unvanquished_2024-01-24_164056_000

Now, using this PR, create a file `maps/void2.navcfg`:

```
cellSizeFactor 0.5
```

Load the map again, and you will see this:

unvanquished_2024-01-24_164344_000

Notice how the size of the tiles has doubled (in each dimension). The choice of this number has influence on the path search at runtime, I assume fewer tiles mean more robotic behavior, but faster search. Notice also that anything making the bots use the whole map is better than restricting them to one half.

### DIFF
--- a/src/shared/bot_nav_shared.cpp
+++ b/src/shared/bot_nav_shared.cpp
@@ -146,6 +146,18 @@ static void ParseOption( Str::StringRef name, Str::StringRef value, Str::StringR
 			return;
 		}
 	}
+	else if ( Str::IsIEqual( name, "cellSizeFactor" ) )
+	{
+		if ( floatValue < 0.1f || 10.f < floatValue )
+		{
+			Log::Warn( "%s: incorrect value for cellSizeFactor '%f': must be between 0.1 and 10", file, floatValue );
+		}
+		else if ( floatValue > 0.0f )
+		{
+			config.cellSizeFactor = floatValue;
+			return;
+		}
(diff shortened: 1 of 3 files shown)


illwieckz:
I actually wonder if we can automatically guess the smaller cellSizeFactor for a given map based on its size.

But I assume it's not only a matter of map width, because of verticality there can be many layers requiring many tiles in a small cube.

slipher:
LGTM

PR #2910, opened 2024-01-26
Fix portal rotation

VReaperV:
Complementary pr to: https://github.com/DaemonEngine/Daemon/pull/1037

New behaviour uses spawnflags bit 3 (8). Since maps without it only used roll to compensate for incorrect code, roll will now be set to 0 without that flag (unless no spawnflags are set i. e. bobbing rotate, where roll will be used as normally).

Portal rotation will now be set by sgame only by using self->s.angles2. For portal cameras that use "target" keyword roll of 0 is assumed, unless specified otherwise with the "roll" keyword. Removed clientNum usage, angles will now always have the correct angle.

Removed the usage of DirToByte() and ByteToDir() functions for portals because they were resulting in incorrect angles.

### DIFF
--- a/src/cgame/cg_ents.cpp
+++ b/src/cgame/cg_ents.cpp
@@ -627,18 +627,11 @@ static void CG_Portal( centity_t *cent )
 	refEntity_t ent{};
 	VectorCopy( cent->lerpOrigin, ent.origin );
 	VectorCopy( s1->origin2, ent.oldorigin );
-	ByteToDir( s1->eventParm, ent.axis[ 0 ] );
-	PerpendicularVector( ent.axis[ 1 ], ent.axis[ 0 ] );
 
-	// negating this tends to get the directions like they want
-	// we really should have a camera roll value
-	VectorSubtract( vec3_origin, ent.axis[ 1 ], ent.axis[ 1 ] );
-
-	CrossProduct( ent.axis[ 0 ], ent.axis[ 1 ], ent.axis[ 2 ] );
+	AnglesToAxis( s1->angles2, ent.axis );
 	ent.reType = refEntityType_t::RT_PORTALSURFACE;
 	ent.oldframe = s1->misc;
 	ent.frame = s1->frame; // rotation speed
-	ent.skinNum = s1->clientNum / 256.0 * 360; // roll offset
 
 	// add to refresh list
 	trap_R_AddRefEntityToScene( &ent );
--- a/src/sgame/sg_spawn_gfx.cpp
+++ b/src/sgame/sg_spawn_gfx.cpp
@@ -179,7 +179,6 @@ env_portal_*
 */
 static void gfx_portal_locateCamera( gentity_t *self )
 {
-	vec3_t    dir;
 	gentity_t *target;
 	gentity_t *owner;
 
@@ -224,19 +223,29 @@ static void gfx_portal_locateCamera( gentity_t *self )
 
 	if ( target )
 	{
-		VectorSubtract( target->s.origin, owner->s.origin, dir );
-		VectorNormalize( dir );
+		axis_t cameraAxis;
+		VectorSubtract( target->s.origin, owner->s.origin, cameraAxis[0] );
+		CrossProduct( cameraAxis[0], axisDefault[2], cameraAxis[1] );
+		VectorInverse( cameraAxis[1] );
+		CrossProduct( cameraAxis[0], cameraAxis[1], cameraAxis[2] );
+		AxisToAngles( cameraAxis, self->s.angles2 );
+
+		// Legacy: Old behaviour used roll to workaround incorrect code
+		// Legacy: But use roll if no rotation and speed flags are not set
+		if ( owner->mapEntity.spawnflags & 8 || ( !( owner->mapEntity.spawnflags & 8 ) && self->s.misc && !self->s.frame ) ) {
+			self->s.angles2[2] = owner->s.angles[2];
+		}
 	}
 	else
 	{
-		glm::vec3 angles = VEC2GLM( owner->s.angles );
-		glm::vec3 dir_ = VEC2GLM( dir );
-		G_SetMovedir( angles, dir_ );
-		VectorCopy( &angles[0], owner->s.angles );
-		VectorCopy( &dir_[0], dir );
-	}
+		VectorCopy( owner->s.angles, self->s.angles2 );
 
-	self->s.eventParm = DirToByte( dir );
+		// Legacy: Old behaviour used roll to workaround incorrect code
+		// Legacy: But use roll if no rotation and speed flags are not set
+		if ( !( owner->mapEntity.spawnflags & 8 ) && !( self->s.misc && !self->s.frame ) ) {
+			self->s.angles2[2] = 0.0;
+		}
+	}
 }
 
 void SP_gfx_portal_surface( gentity_t *self )
@@ -261,15 +270,11 @@ void SP_gfx_portal_surface( gentity_t *self )
 
 void SP_gfx_portal_camera( gentity_t *self )
 {
-	float roll;
-
 	VectorClear( self->r.mins );
 	VectorClear( self->r.maxs );
 	trap_LinkEntity( self );
 
-	G_SpawnFloat( "roll", "0", &roll );
-
-	self->s.clientNum = roll / 360.0f * 256;
+	G_SpawnFloat( "roll", va( "%f", self->s.angles[2] ), &self->s.angles[2] );
 }
 
 /*


slipher:
There is seemingly an attempt to parse "roll" in SP_gfx_portal_camera:

```
	G_SpawnFloat( "roll", "0", &roll );

	self->s.clientNum = roll / 360.0f * 256;
```

Does this work at all?

VReaperV:
Also, self->s.clientNum which is used there later gets passed down to the engine where it is used for some of the portal "bobbing" code.

illwieckz:
Something looks wrong on pulse:

unvanquished_2024-01-26_194217_000

VReaperV:
> So the expected behavior of roll is to (a) cause a "bobbing" effect, plus (b) overwrite the 3rd member of "angles"? Is this all for backwards compatibility with Tremulous?

Only (b) is needed. I think "roll" being used for bobbing is an error since [docs](https://github.com/Unvanquished/unvanquished-mapeditor-support/blob/bec3db5fcf397fc15dc8eb5b664b941721729082/build/netradiant/unvanquished.game/pkg/entities.ent#L1329) say that spawnflags should be used for that. It's probably safe to remove it, especially since the engine code tries to use the value passed from clientNum to make the camera bob when spawnflags dictate it shouldn't.

VReaperV:
> Something looks wrong on pulse:

That is another old map that was erroneously setting roll which was then getting ignored. I've uploaded the updated version of the map [here](https://users.unvanquished.net/~reaper/maps/portal-rotation-fix/map-pulse_2%2Bbeta3-1.1.1.dpk). Can you check if that works for you?

slipher:
Are we sure this documentation is really right? Maybe someone just made that stuff up about roll.

VReaperV:
I think the doc is right. I've looked at the .bsp for the pulse map in tremulous (`map-pulse-beta3-1.1.0.pk3`) and it also uses both "spawnflags" "4" and "roll". Tremulous just ignores the roll angle and both cameras have no bobbing effect.

illwieckz:
> That is another old map that was erroneously setting roll which was then getting ignored. I've uploaded the updated version of the map [here](https://users.unvanquished.net/~reaper/maps/portal-rotation-fix/map-pulse_2%2Bbeta3-1.1.1.dpk). Can you check if that works for you?

That fixes the issue, indeed!

VReaperV:
It is possible that clientNum was *intended* to pass the roll value to the engine, but it evidently never worked - release and current master clients display cameras without roll even if it's specified. Either way, it's better to have a straightforward implementation instead of some weird roundabout ways that are not even working anyway.

slipher:
Since Pulse has `spawnflags & 4` this disables the bobbing motion (according to code reading; I didn't test). `spawnflags & 4` -> `<portal surface entity>.s.misc == 0` -> `refEntity_t::oldFrame == 0` -> rotation/bobbing disabled. So you would need a different test case to check whether this stuff works.

illwieckz:
Something looks weird to me, while mirrors render as I expect them to:

unvanquished_2024-01-26_221513_000

unvanquished_2024-01-26_221616_000

Portals seems to have all their cameras on the floor (which is not true, unless I did something wrong):

unvanquished_2024-01-26_221530_000

unvanquished_2024-01-26_221458_000

illwieckz:
All those portal/camera/position entities are on the same plane, and not on the floor:

20240126-222119-000 test-portal-recursion

VReaperV:
> Since Pulse has `spawnflags & 4` this disables the bobbing motion (according to code reading; I didn't test). `spawnflags & 4` -> `<portal surface entity>.s.misc == 0` -> `refEntity_t::oldFrame == 0` -> rotation/bobbing disabled. So you would need a different test case to check whether this stuff works.

There's another codepath in R_GetPortalOrientations() when refEntity_t::oldframe == 0: it takes refEntity_t::skinNum as the rotation. It's supposedly set to the roll value in gfx_portal_locateCamera() (self->s.clientNum = roll / 360.0f * 256, then in CG_Portal() ent.skinNum = s1->clientNum / 256.0 * 360), but that seems to be doing nothing.

VReaperV:
> Portals seems to have all their cameras on the floor (which is not true, unless I did something wrong):

The portal cameras are 32 qu above the floor, while the centers of the portal surfaces are 64 qu above ground. If you do /setviewpos to the portal camera origins it will in fact be the same view.

illwieckz:
> > Portals seems to have all their cameras on the floor (which is not true, unless I did something wrong):
> 
> The portal cameras are 32 qu above the floor, while the centers of the portal surfaces are 64 qu above ground. If you do /setviewpos to the portal camera origins it will in fact be the same view.

Ah! I now get it, thanks!

src/sgame/sg_spawn.cpp
```diff
@@ -732,6 +734,9 @@ static void G_ParseField( const char *key, const char *rawString, gentity_t *ent
 
 			( ( float * ) entityDataField ) [ 0 ] = tmpFloatData[ 0 ];
 			( ( float * ) entityDataField ) [ 1 ] = tmpFloatData[ 1 ];
+			if ( !Q_stricmp( key, "angles" ) && entity->preserveRoll ) { // HACK: "roll" keyword must overwrite "angles"
+				break;                                                   // Proper fix requires rewriting entity parsing
+			}
```
illwieckz, line 739: I prefer if you format the comment this way:

```c++
			// HACK: "roll" keyword must overwrite "angles"
			// Proper fix requires rewriting entity parsing
			if ( !Q_stricmp( key, "angles" ) && entity->preserveRoll ) {
				break;                                                   
			}
```

illwieckz:
You probably know better than me what @slipher said here:

> Since Pulse has `spawnflags & 4` this disables the bobbing motion (according to code reading; I didn't test). `spawnflags & 4` -> `<portal surface entity>.s.misc == 0` -> `refEntity_t::oldFrame == 0` -> rotation/bobbing disabled. So you would need a different test case to check whether this stuff works.

So if there is something else to do, LGTM after that something else is done, otherwise you already get my LGTM.

slipher:
I still don't get why we would want to have the `roll` key overwriting part of `angles`. The documentation may suggest that, but if the documentation goes against the code, existing maps, and common sense, then it is the documentation that should be fixed.

That said, if we do implement this semantic, it would be cleaner to add it to the existing `roll` parser in SP_gfx_portal_camera, rather than trying to do it with the table-based parser.

VReaperV:
> I still don't get why we would want to have the `roll` key overwriting part of `angles`. The documentation may suggest that, but if the documentation goes against the code, existing maps, and common sense, then it is the documentation that should be fixed.

Well, existing maps do suggest the same usage as documentation. E. g. map `gauntlet` has portal cameras with both "roll" and "angles" with 2 values. The only way that would make sense is if it's supposed to overwrite (since "angles" always reads 3 values).

> That said, if we do implement this semantic, it would be cleaner to add it to the existing roll parser in SP_gfx_portal_camera, rather than trying to do it with the table-based parser.

I don't think the existing one is working, but I'll check again to make sure.

VReaperV:
I tried it without the roll parser changes and clientNum is 0 at gfx_portal_locateCamera() regardless of what roll is.

slipher:
> Roll does get parsed properly there apparently. It looks worse to me to parse it there, but I can change it to use clientNum instead.

I guess it doesn't really matter whether the number is stored in `clientNum` or `angles` (assuming the 3rd member of angles will be ignored). The important thing is to avoid adding a bunch of ugly code to the table-based parser.

The table-based parser is a questionable paradigm anyway since there is no telling which attributes belong to which types of entities. It enforces the god struct antipattern.

VReaperV:
One of the test maps uses the 3rd value of angles for cameras, and grangermaze attempts to use it as well. I can just make the parsing of roll default to the second value of angles. This would avoid the ugly hack, and would work with old (and updated) maps, new maps and test maps.

src/sgame/sg_spawn.cpp
VReaperV: I've removed that hack altogether now in favour of a different approach.

VReaperV:
Tested with all maps with portals we have, this works correctly for old maps. Test maps will need an adjustment to use spawnflags bit 3.

VReaperV:
> So if there is something else to do, LGTM after that something else is done, otherwise you already get my LGTM.

This is all working properly now, since roll was only used in old maps to compensate for non-working code it is set to 0 for old maps now (new maps can set spawnflag bit 3 to use the new behaviour).

If that looks good to you, I'll merge this in a bit.

slipher:
So you don't think it would be a good idea to make a different name for a portal entity that avoids the buggy legacy behavior? If the old one needs a "don't be buggy" flag that seems like a perfect time to actually deprecate the old class (not necessarily implying we will ever remove it) and create a new one. The `gfx_portal_camera` / `gfx_portal_surface` aliases have never been used by any map so they could be utilized for this.

Requiring a flag for "no camera wiggling" seems obnoxious as well but I have no idea what that's about.

VReaperV:
> So you don't think it would be a good idea to make a different name for a portal entity that avoids the buggy legacy behavior?

Yes, I think making a duplicate entity for only this purpose is a bad idea. It would also mean duplicating gfx_portal_locateCamera() and SP_gfx_portal_surface() with very few changes or something like that.

> not necessarily implying we will ever remove it

You can't completely remove it anyway, it needs to at least be an alias to support old maps.

> Requiring a flag for "no camera wiggling" seems obnoxious as well but I have no idea what that's about.

It's a relic of quake 3.

DolceTriade:
I think we're going to have to accept some annoying flags and hacks to maintain q3 compat. I think this change looks good

VReaperV:
Additionally, we *can* remove the gfx_portal_\* aliases completely but not the misc_portal_\* ones.

illwieckz:
I will not be against having a “correct from the start” alternate entity, but to me it is out of scope of this PR.

illwieckz (approved):
LGTM.

PR #2915, opened 2024-01-29
building human bot buys armour

sweet235:
Because why not

### DIFF
--- a/pkg/unvanquished_src.dpkdir/bots/default_humans.bt
+++ b/pkg/unvanquished_src.dpkdir/bots/default_humans.bt
@@ -31,12 +31,11 @@ selector
 					{
 						action suicide
 					}
-					sequence
+					selector
 					{
-						decorator return( STATUS_SUCCESS )
-						{
-							action equip
-						}
+						action buy( WP_HBUILD, UP_BATTLESUIT )
+						action buy( WP_HBUILD, UP_MEDIUMARMOUR )
+						action buy( WP_HBUILD, UP_LIGHTARMOUR )
 						action buy( WP_HBUILD )
 					}
 					condition alertedToEnemy


DolceTriade:
didn't test but it seems reasonable

sweet235:
The `action equip` tries to do that, but it seems this stopped to work at some time. But we can specify which armour to buy explicitely, as I found out. So this PR does that.

slipher:
LGTM but there are 3 other cases of `buy( WP_BUILD )` that should probably get the same treatment. In defend.bt, build.bt, and (a second one in) default_humans.bt.

sweet235:
And a 4th case: a bot spawns with the ckit.

sweet235:
Here is your essay:

These BT instructions used to make a bot buy armour and ckit:
```
decorator return( STATUS_SUCCESS )
{
    action equip
}
action buy( WP_HBUILD )
```

Bisecting suggests that this does not work anymore since https://github.com/Unvanquished/Unvanquished/commit/100f802fd12df1e3db5a477812a6f966a9701b18

The sequence `decorator ... { action equip }; action buy( WP_HBUILD )` occurs two times in our code. Had its behavior not been changed, we would be fine. I created the second occurance after observing that the first occurance worked at that time.

It is true that this PR does not yet fix both occurances. But the reason is, as I am led to believe, that https://github.com/Unvanquished/Unvanquished/commit/100f802fd12df1e3db5a477812a6f966a9701b18 changed this without testing it. Now the best we can do is to fix this step by step.

I shall not attempt to merge this. Someone seems to own all my PRs, and be trusted to change or close them as they please. I will leave the further decisions to the powers that be.

slipher:
Probably we should have two different code paths for "just buy this equipment and don't change anything else" and "outfit oneself with exactly this set of equipment" instead of trying to do all of them with the same action.

PR #2916, opened 2024-02-05
Encapsulate Medistation state in CBSE component

slipher:
I still need to test this before merging. I was on an airplane or whatever and didn't test it at the time I wrote it.


sweet235:
I think that we should revisit our long-term goals. We are maintaining a big codebase, and times change.

CBSE was introduced in the early, wild days of Unvanquished development. The first commit was 45e600fcd594f8aae59499f231a6839ff294f53a, a decade ago. In these times, a lot was being done. People were improving the engine greatly, and more modern models, maps, textures were crafted to match these improvements. A stated goal from these times was to gradually move from C to C++, as this has obvious benefits.

One benefit of C++ is its improved type system. Modern C++ still has some quirks that go back to C, but when used cautiously, C++ can catch more errors at compile time.

CBSE is a concept described in https://en.wikipedia.org/wiki/Component-based_software_engineering. Our current implementation uses a rather obscure piece of software named "Python Jinja". As far as I understand, this creates the monstrosity you can find in `src/sgame/backend/CBSEBackend.cpp` (this directory is created at build time).

This is sledge hammer software engineering: when facing problems in a programming language, simply generate some source files in a python script. This approach loses all type safety. If we make a change in how we generate the files in `src/sgame/backend`, we will catch any type errors only much later, when the C++ compiler tries to process these files. And the error messages will likely be not helpful at all. This is just as evil as plain old C macros.

Clearly, we should not be doing this for another decade. But what do we want to do instead? At a first glance, it seems that something as simple as Java's interfaces could solve this problem in a properly type-safe manner. Maybe we should investigate methods to use templates instead of a custom code generator. Template hell is a dangerous thing, but is it really as dangerous as a naive custom code generator?

We are working in a different decade now: our game is quite stable, and we are competing with non open-source games even on linux. Our code base is what it is, refactorings can introduce bugs, but not improve the user experience. But hunting down these eventual bugs slows us down.

Therefore, i propose we abandon the paradigm of moving everything to C++, especially to our current CBSE.

sweet235:
oops

sweet235:
Well, someone has to say it: you are blocking progress.

slipher:
The way I see it there are 3 basic duties that CBSE performs for us:
- Encapsulate the state and logic into (relatively) independent units
- Define interfaces that can be implemented by more than one component (several components of an entity can have `HandleDie`, unlike the `die` function pointer which can only have one implementation)
- Conveniently iterate over entities of a given type

It's possible to provide these facilities without using a code generator. Examples of alternate implementations of Entity-Component Systems (the more common name for CBSE in the game industry) include https://github.com/skypjack/entt and https://github.com/alecthomas/entityx. These use elaborate C++ template metaprogramming rather than a code generator. It's really a matter of taste which one you prefer. Some people find the generated code refreshingly simple compared to the C++ template horrors. Others consider code generation a nuclear weapon of programming and run away screaming.

It kind of sucks to maintain our own ECS implementation but it's not that time-consuming. Nor are there any features in the alternate implementations I particularly want, so I personally wouldn't find it worthwhile to migrate to one of these other ones. And perhaps Team Anti-C++-Templates will even turn out to be bigger than Team Anti-Code-Generation and people will complain about the new thing even more.

> How can we make our CBSE implementation type-safe?

What do you mean by this? It's not really different from changing a C++ header. If you make a mistake, an error will be shown when you attempt to build, same as with writing C++ code.

slipher:
I have now tested the code

DolceTriade:
Change itself looks fine. I'll test more later. Are there any implications for bots that need adjusting with the move to CBSE?

src/sgame/components/MedipadComponent.cpp
```diff
@@ -1,13 +1,179 @@
...
-	if (entity.oldEnt->target && entity.oldEnt->target->client) {
-		entity.oldEnt->target->client->ps.stats[STAT_STATE] &= ~SS_HEALING_2X;
+	if (target_) {
+		target_->client->ps.stats[STAT_STATE] &= ~SS_HEALING_2X;
+	}
+}
+
+gentity_t* MedipadComponent::GetTarget() const
+{
+	return target_.get();
```
Viech, line 45: Should this return `nullptr` if `target_` is `nullptr`?

slipher: Yes it should, and does.

slipher:
> Are there any implications for bots that need adjusting with the move to CBSE?

Sure there were some places in the bot code touching the medistation state and they have been migrated.

DolceTriade (approved):
My testing was ok

PR #2917, opened 2024-02-06
Prevent (d)evolving to current class

cu-kai:
This fixes an exploit where one can use `/class level3upg` to immediately recharge barbs near the overmind.

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -2045,6 +2045,14 @@ bool G_AlienEvolve( gentity_t *ent, class_t newClass, bool report, bool dryRun )
 		return false;
 	}
 
+	if ( newClass == currentClass )
+	{
+		// prevent recharging adv dragoon barbs instantly by the overmind, for example.
+		// also, fail silently - we don't need to tell the
+		// player they cannot evolve to their current class.
+		return false;
+	}
+
 	// all the checks passed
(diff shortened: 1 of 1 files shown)


src/sgame/sg_cmds.cpp
```diff
@@ -2045,6 +2045,11 @@ bool G_AlienEvolve( gentity_t *ent, class_t newClass, bool report, bool dryRun )
 		return false;
 	}
 
+	if ( newClass == currentClass )
```
DolceTriade, line 2048: do you need to respect `report` and report back soemthign to the user?

cu-kai: I discussed this on IRC with @illwieckz:
```
<+cu-kai> anyway: that commit silently prevents it
<+cu-kai> (do we really need to feed back for this?)
<+illwieckz> no
<+illwieckz> evolving to itself seems to be a bug to me
<+cu-kai> i agree
<+cu-kai> so simply fail silently, instead of "You cannot evolve to your current class"?
<+cu-kai> if so: i can make a PR for this now if you like
<+illwieckz> yes please
```

DolceTriade: Then please add a comment about this in the code.

DolceTriade (approved):

DolceTriade (approved):

PR #2921, opened 2024-02-09
Use proper scaling for crosshair outlines

VReaperV:
Adds a crossHairSizeNoBorder to the weapon info. This is the size of the visible part of the crosshair, either width or height (whichever is larger). This is required because crossHairSize is the size of the image, not the actual visible part of the crosshair. All crosshair images we have are 256x256, which corresponds to size 64.

Both sizes are in virtual coordinates: the game maps the screen to 640x480 coordinates for UI, regardless of screen resolution.
The crosshairSizeNoBorder can be found as `(pixelCoord2 - pixelCoord1 + 1) / imageSize * crossHairSize`
pixelCoord2 and pixelCoord1 - coordinates in pixels of the closest and furthest visible pixels (either x or y, wherever the larger part ofthe crosshair is).
imageSize - size of the crosshair image dimension in pixels

If no crosshairSizeNoBorder is specified, it defaults to crossHairSize.
You can find the updated weapon config files at https://github.com/UnvanquishedAssets/res-weapons_src.dpkdir/pull/16.

### DIFF
--- a/src/cgame/cg_weapons.cpp
+++ b/src/cgame/cg_weapons.cpp
@@ -879,6 +879,24 @@ static bool CG_ParseWeaponFile( const char *filename, int weapon, weaponInfo_t *
 			wi->crossHairSize = size;
 
 			continue;
+		} else if ( !Q_stricmp( token, "crosshairSizeNoBorder" ) ) {
+			float size;
+
+			token = COM_Parse( &text_p );
+
+			if ( !token ) {
+				break;
+			}
+
+			size = atof( token );
+
+			if ( size < 0.0f ) {
+				size = 0.0f;
+			}
+
+			wi->crossHairSizeNoBorder = size;
+
+			continue;
 		}
 		else if ( !Q_stricmp( token, "disableIn3rdPerson" ) )
 		{
@@ -939,6 +957,11 @@ static bool CG_ParseWeaponFile( const char *filename, int weapon, weaponInfo_t *
 
 		Log::Warn( "unknown token '%s'", token );
 		return false;
+	}
+
+	if ( wi->crossHairSizeNoBorder == 0.0f && wi->crossHair ) {
+		Log::Warn( "weapon %s crossHairSizeNoBorder not set or is 0, setting to crossHairSize", filename );
+		wi->crossHairSizeNoBorder = wi->crossHairSize;
 	}
 
 	return true;
--- a/pkg/unvanquished_src.dpkdir/ui/options_ui.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_ui.rml
@@ -183,6 +183,10 @@
 							<input type="range" min="-5" max="5" step="0.05" cvar="cg_crosshairOutlineScale"/>
 							<translate>Scale:</translate>&nbsp;<inlinecvar cvar="cg_crosshairOutlineScale" type="number" format="%0.2f"/>
 						</row>
+						<row>
+							<input type="range" min="-5" max="5" step="0.05" cvar="cg_crosshairOutlineOffset"/>
+							<translate>Offset:</translate>&nbsp;<inlinecvar cvar="cg_crosshairOutlineOffset" type="number" format="%0.2f"/>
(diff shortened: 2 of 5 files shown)


Grise3:
That mean that the size of the crosshair will not be the size of the picture, but something set? And if nothing is set it use the picture size?

slipher:
Please post screenshots showing the difference

VReaperV:
It prevents the outline from getting "detached" when changing crosshair size or switching to different weapons with different-sized crosshairs.

Old code using the same scale:
unvanquished_2024-02-11_005745_000
unvanquished_2024-02-11_005750_000
Or when switching weapons with different crosshair sizes:
unvanquished_2024-02-11_005831_000
unvanquished_2024-02-11_005827_000

New code:
unvanquished_2024-02-11_010156_000
unvanquished_2024-02-11_010205_000

I've also added an offset value so the outline looks the same thickness when using different weapons:
unvanquished_2024-02-11_011209_000
unvanquished_2024-02-11_011214_000

VReaperV:
> That mean that the size of the crosshair will not be the size of the picture, but something set? And if nothing is set it use the picture size?

Only the size of the outline. Yes, if it is not set it will just use the image size basically.

src/cgame/cg_weapons.cpp
```diff
@@ -941,6 +959,11 @@ static bool CG_ParseWeaponFile( const char *filename, int weapon, weaponInfo_t *
 		return false;
 	}
 
+	if ( wi->crossHairSizeNoBorder == 0.0f && wi->crossHair ) {
```
slipher, line 962: Can we get rid of the warning please? It's going to spam everyone until the next release is out.

sweet235:
LGTM

src/cgame/cg_weapons.cpp
sweet235: So we shall release soon

VReaperV: This shouldn't really affect anyone though? To get the warning, one would need to run master cgame without master res-weapons. If a server operator uses this change, they're expected to use the updated res-weapons as well. (I assume that no one other than developers really runs master gamelogic on a local server with release assets)

I can comment out the warning until the next release though.

slipher: > To get the warning, one would need to run master cgame without master res-weapons.

Exactly. Running with every asset package from source would be a pain in the ass so we want it to be usable with just the master Unvanquished repo.

DolceTriade: I don't think it's that big of a deal to require the latest resources. Annoying for sure, but I think we're probably going to be ready for a release soon enough.

slipher: It is a big deal to set up your environment with the latest asset packages because many of them require compilation with the asset toolchain and can't be used directly from source. Assuming that a release will happen very soon is a bad idea.

DolceTriade: I guess "require" was a strong word. A warning log is hardly a big deal.

slipher: That would be 15 or so warning logs, one for each weapon.

This warning doesn't really seem useful for preventing mistakes anyway... The number would need to be updated whenever the image is changed, but there's no way to detect that.

Grise3: >That would be 15 or so warning logs, one for each weapon.

currently we have a lot of warnings on the games
and there is this too : "[further messages like this will be suppressed] 
"
Then in fact, the result is something like this : 

Screenshot_20240217_082851

It is not a real problem. we have warnings too, and we just ignore them.

Screenshot_20240217_083131

like this warnings messages. I think that one year ago we had them too. 
Then warnings is not a problem.

slipher: All right, full spam ahead if that's what you guys want :stuck_out_tongue:

VReaperV:
Conflict resolve gone slightly wrong.

PR #2922, opened 2024-02-11
FIX: tutorial memory corruption

DolceTriade:
Code looks ok. Will merge later if no one has any comments.

### DIFF
--- a/src/cgame/cg_key_name.cpp
+++ b/src/cgame/cg_key_name.cpp
@@ -59,7 +59,7 @@ static int bindTeam;
 void CG_SetBindTeam(team_t team) {
     switch (team) {
     case TEAM_NONE:
-        bindTeam = 3; // BIND_TEAM_SPECTATORS
+        bindTeam = NUM_TEAMS; // BIND_TEAM_SPECTATORS
         break;
     case TEAM_HUMANS:
     case TEAM_ALIENS:
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -2252,7 +2252,7 @@ void          CG_DestroyTestTS_f();
 //
 // cg_tutorial.c
 //
-const char *CG_TutorialText();
+const std::string& CG_TutorialText();
 
 //
 // cg_beacon.c
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -3209,7 +3209,7 @@ static void CG_Rocket_DrawTutorial()
 		return;
 	}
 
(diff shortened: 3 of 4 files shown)


slipher:
If we're going to return std::string we should just return it by value and get rid of the static buffers.

DolceTriade:
Isn't that a bunch of extra copies and allocations?

slipher:
> Isn't that a bunch of extra copies and allocations?

This commit by its nature adds a bunch of extra copies and allocations. Writing in a static char buffer avoids allocations, while writing to std::string is going to reallocate it a dozen times or so while building the string.

But the change I'm suggesting shouldn't add any additonal extra allocations as the compiler can automatically "move" the string when the return value is a local variable ("NRVO").

DolceTriade:
Just for clarification, `std::string::clear()` shouldn't deallocate any buffers per call right? So after the first frame where we do a bunch of allocations, subsequent calls won't do additional allocations unless we exceed the buffer?

src/cgame/cg_key_name.cpp
```diff
@@ -59,7 +59,7 @@ static int bindTeam;
 void CG_SetBindTeam(team_t team) {
     switch (team) {
     case TEAM_NONE:
-        bindTeam = 3; // BIND_TEAM_SPECTATORS
+        bindTeam = NUM_TEAMS; // BIND_TEAM_SPECTATORS
```
slipher, line 62: Remove unrelated change

slipher:
> Unlike for [std::vector::clear](https://en.cppreference.com/w/cpp/container/vector/clear), the C++ standard does not explicitly require that [capacity](https://en.cppreference.com/w/cpp/string/basic_string/capacity) is unchanged by this function, but existing implementations do not change capacity. This means that they do not release the allocated memory (see also [shrink_to_fit](https://en.cppreference.com/w/cpp/string/basic_string/shrink_to_fit)).

Interesting. So the number of allocations may not be so much.

Still, the value of the change in the current form seems questionable since it doesn't actually make things any more memory-safe. There are still pointers/references to a buffer that may be overwritten with different content. If we make it return by value then the change will actually have a good motivation. We can use `reserve` if multiple allocations per call is a concern.

VReaperV:
Tutorial text without this change on recent master produces random garbage in tutorial text:
image
Haven't tried it with this change yet.

src/cgame/cg_key_name.cpp
DolceTriade: It was part of the original change by freem. I think he wanted to use a constant rather than a naked number.

VReaperV:
This patch fixes the garbage in tutorial text on my end.

DolceTriade:
I'll merge this later if no one objects. Thanks for verifying

PR #2924, opened 2024-02-16
crosshair: Add a preview crosshair to the crosshair menu

DolceTriade:
This builds on reaper's work to allow previewing the crosshair. We do this by modifying the existing crosshair code to allow overriding the weapon to show. If we override the weapon, always show the crosshair. Further, remove an used color parameter.

Once my Lua code is merged, we won't have to hardcode weapons, but this works well for now imo.

https://github.com/Unvanquished/Unvanquished/assets/1097524/69308137-fbb9-4477-b46d-d15a40f6f842

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/options_ui.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_ui.rml
@@ -12,7 +12,7 @@
 					<!-- Keep in sync with language menu in options_welcome.rml -->
 					<h3><translate>Language</translate> </h3>
 					<p> <translate>Some languages are not fully translated.</translate> </p>
-		
+
 					<!-- NOTE: it's a bad idea to put languages translated at less than 50% in the list, right? -->
 					<select cvar="language">
 						<option value="fr">Français</option>
@@ -79,6 +79,29 @@
 			</panel>
 			<tab><translate>Crosshair</translate></tab>
 			<panel class="crosshair">
+				<row>
+					<h3><translate>Weapon</translate></h3>
+					<select name="crosshair_weapon" value="rifle" onchange='document:GetElementById("crosshair"):SetAttribute("weapon", event.current_element.attributes["value"])'>
+						<option value="level0"><translate>Aliens</translate></option>
+						<option value="blaster"><translate>Blaster</translate></option>
+						<option value="rifle"><translate>Rifle</translate></option>
+						<option value="psaw"><translate>Pain Saw</translate></option>
+						<option value="shotgun"><translate>Shotgun</translate></option>
+						<option value="lgun"><translate>Lasgun</translate></option>
+						<option value="mdriver"><translate>Mass Driver</translate></option>
+						<option value="chaingun"><translate>Chaingun</translate></option>
+						<option value="flamer"><translate>Flamer</translate></option>
+						<option value="prifle"><translate>Pulse Rifle</translate></option>
+						<option value="lcannon"><translate>Lucifer Cannon</translate></option>
+					</select>
+
+					<p/>
+				</row>
+				<row>
+					<div style="height: 4em; padding-left: 6em; padding-top: 4em;">
+						<crosshair id="crosshair"/>
+					</div>
+				</row>
 				<row>
 					<h3><translate>Crosshair visiblity</translate></h3>
 					<select name="drawcrosshair" cvar="cg_drawCrosshair">
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -698,16 +698,30 @@ private:
 class CrosshairHudElement : public HudElement {
 public:
 	CrosshairHudElement( const Rml::String& tag ) :
-			HudElement( tag, ELEMENT_BOTH, true ),
-			color_( Color::White ) {
+			HudElement( tag, ELEMENT_ALL, true ),
+			// Use as a sentinel value to denote that no weapon is
+			// explicitly set.
+			weapon_( WP_NUM_WEAPONS ) {
 	}
 
-	void OnPropertyChange( const Rml::PropertyIdSet& changed_properties ) override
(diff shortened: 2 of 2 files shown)


illwieckz:
That's excellent!

On my end the video was not playing in the browser, I had to download it, here is a webm conversion:

[output.webm](https://github.com/Unvanquished/Unvanquished/assets/2311649/977140cd-b6cb-4a4e-8fac-d57ded288505)

Grise3:
"" Video can't be played  because  the file is corrupt. ""

@DolceTriade that is really great to add a preview crosshair to the crosshair menu

@illwieckz what you say is a good idea :)

Note that I'm note sure that a color can be good for all the maps. (color of the crosshairs)

DolceTriade:
I guess we could add a bunch of backgrounds in a different slider but I think the autocolor might handle that ok?

Grise3:
> I guess we could add a bunch of backgrounds in a different slider but I think the autocolor might handle that ok?

The autocolor?

Grise3:
After reflection I don't think that the UI that you did really good, @dolcetriade.

I think that the better would make a window for the crosshairs options, and in the window do like the "keybindings" menu, With kinds of “tabs”.
for example : tabs sepzreat like this :  
*crosshairs size and visibility
* colors of the crosqhairs
* colors of the outline

(for example) The goal would be to have a small window, for ppls that have a little screen ;) and for don't have too much things in the window. 

And if the crosshairs are disabled, can we change the crosshairs colors?

VReaperV:
> I think that the better would make a window for the crosshairs options, and in the window do like the "keybindings" menu, With kinds of “tabs”. for example : tabs sepzreat like this :
> 
> * crosshairs size and visibility
> * colors of the crosshairs
> * colors of the outline

Having to switch between different tabs to change one thing (crosshair) wouldn't be good.

Grise3:
The tabs was to make the window smaller... I think that currently, on some screens, the window that you did is too big and can be outside the window :)
But true, the tabs is not a really good idea.

>I think crosshair being visible or not is independent of crosshair colors. This is just meant to be a single pane of glass for you to configure your crosshair for the weapons you choose.

Well, I was talking about the view of the crosshairs.

Maybe that would be great if the view is like we see ingame? I mean the size, the colors, the visibility....

>There are definitely further improvements to be made, but this is a step forward from what was there before.

Sure! It's a big step forward from that we had before. I really think that you said is great :)

I'm just afraid that the window will protrude on either side of the screen on small screens.

And for the other, like the options it was just something like a proposition.

I admit that what I wrote wasn't  clear.

DolceTriade:
No worries. The concern about big menus on small screens is valid. We can make the crosshair preview smaller if we need to.

illwieckz:
I would like to reduce the amount of windows, because windows only works with mouses. But I agree the current proposed UI can be perfect. I don't mind: I prefer to have this in next release than nothing. We can improve later.

Since we're discussing UI, one quick thing I thought was to do to columns, one for croshair itself, one for outline. They both have the same options (size, color…) so this may even look good.

Anyway, the currently proposed design is good enough to be merged.

VReaperV:
Yeah, we could use some more horizonatally filled windows in general.

DolceTriade:
Ultimately, we need a UI revamp. If no one objects, I'll merge this shortly

illwieckz (approved):
LGTM.

Grise3:
> No worries. The concern about big menus on small screens is valid. We can make the crosshair preview smaller if we need to.

Maybe it can be smaller (but with that, that mean that we can't see it that much) but after, another solution (not really a good solution) is to do a scrollable window, Ith a small size. It is not a good idea but nvn.

@DolceTriade : I was looking the video on a phone, and after looking on a computer, I see that the window is not that big (it is big, but when I look on my phone it's really bigger, idk why)

The crosshairs are becoming good guys! 

> I would like to reduce the amount of windows, because windows only work with mouses. But I agree the current proposed UI can be perfected. I don't mind: I prefer to have this in next release than nothing. We can improve later, even with a future PR before release.
> 
> Since we're discussing UI, one quick thing I thought was to do two columns, one for crosshair itself, one for outline. They both have the same options (size, color…) so this may even look good.
> 
> Anyway, the currently proposed design is good enough to be merged.

It would be cool to reduct, yeah. If it is reduced or we found another solution, we could use controllers! 

> Ultimately, we need a UI revamp. If no one objects, I'll merge this shortly

I can do this :)
well, I can try to do this :)
I allready started with my new menus, I can continue!

slipher:
Please avoid merging pull requests less than 24 hours after opening them.

Grise3:
> Please avoid merging pull requests less than 24 hours after opening them.

Why? All was okay for say that the modifications where great. With that, why wait with a delay?  We where 4 for say that it was good.

DolceTriade:
Is the rule to allow everyone to get a chance to review it. I mistimed it. Will remember next time

Grise3:
> Is the rule to allow everyone to get a chance to review it. I mistimed it. Will remember next time

Ha, there is a rule for that? I didn't knew 

Sorry @slipher

Grise3:
Ha...
@DolceTriade 

Just for say : 
when I'm in the main menu, i have this : 

>Warn: [string "document:GetElementById("crosshair"):SetAttri..."]:1: attempt to index a nil value (local 'document')
>stack traceback:
>        [string "document:GetElementById("crosshair"):SetAttri..."]:1: in function <[string >"document:GetElementById("crosshair"):SetAttri..."]:1> 


no idea if it is caused by these modifications

and : 
Screenshot_20240217_095950

when I "start" the window, it don't show the crosshair. I have to chage the crosshair of the weapon like "show" the blaster crosshair.

Eny idea of why?

Grise3:
https://github.com/Unvanquished/Unvanquished/assets/148749571/6ea99bf8-8f64-49a7-a4b4-b4c946f4e410

DolceTriade:
I'm out of town this weekend. I'll fix it when I get back. It sounds related to this

PR #2925, opened 2024-02-17
HUD: use exponential moving average FPS counter

slipher:
This makes the FPS counter be constant instead of flickering when the frame rate is steady.


illwieckz:
I like this! It makes reading the fps easier.

DolceTriade (approved):

PR #2928, opened 2024-02-18
add code for see if the server is hosted by user, for show/hide menus

Grise3:
with these modifications, these menus will be not accessible for the users that don't host : 
* Bots 
* Server setup

Note that the bot menu work with /bot.

for can enable it for all the users, we have to make votes for remove/add only *ONE* bot per team, like in the current bot menu.

Screenshot_20240218_125942


slipher:
I guess these two things should just be part of the "Host menu" section which is hidden already.

Grise3:
@slipher yeah true. I did it, it is updated.

PR #2930, opened 2024-02-22
lua: Add framework for shared Lua and add some Lua integration for bg_misc stuff

DolceTriade:
With this change, we add lua to shared so it can be used for the sgame soon. We also add Lua APIs for accessing Unvanquished weapon/class/buildable/upgrade info (with the caveat that BG_InitAllConfigs() must be called and paks must be loaded).

### DIFF
--- a/src/shared/LuaLib.inl
+++ b/src/shared/LuaLib.inl
@@ -1,0 +1,351 @@
+/*
+ * This source file is part of libRocket, the HTML/CSS Interface Middleware
+ *
+ * For the latest information, see http://www.librocket.com
+ *
+ * Copyright (c) 2008-2010 CodePoint Ltd, Shift Technology Ltd
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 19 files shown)


src/shared/LuaLib.inl
```diff
@@ -0,0 +1,351 @@
...
+        {
+            //In case this is an address that has been pushed
+            //to lua before, we need to set it to nil
+            lua_pushnil(L); // ->[4] = nil
+            lua_setfield(L,-2,name); //represents t[k] = v, [-2 = 3] = t -> v = [4], k = <ClassName>; pop [4]
+        }
+
+        if(IsReferenceCounted<T>())
+        {
+			// TODO: Remove
```
slipher: So what is the story with object life cycle? Do we need that "DO NOT TRASH" shit? I don't remember how it works but if so we might have a problem since this library appears to be based on a really old version that doesn't have my [crash fix](https://github.com/mikke89/RmlUi/commit/23f2e148ec9730e353b520c665c1d1aa089b8cc6) (and even still uses the `Rocket` namespace).

Also when importing code like this it would be nice to have the unmodified version of the imported code (not included in the build) as the first commit, with a note as to which commit in the external project it came from, and only hack it to suit our needs in the second commit.

src/shared/bg_lua.cpp
```diff
@@ -2,7 +2,7 @@
 ===========================================================================
 
 Daemon GPL Source Code
-Copyright (C) 2012 Unvanquished Developers
+Copyright (C) 2012 Unv Developers
```
slipher, line 5: lol

src/shared/lua/Weapons.h
```diff
@@ -0,0 +1,66 @@
+/*
+ = =================*=========================================================
+
+ Daemon GPL Source Code
```
slipher: Daemon?

slipher (commented):
So what's the point of having these APIs in the cgame? Is it supposed to replace some of the data sources?

The way you've done it is all right but it occurs to me that weapons/classes/upgrades/buildables don't really need to have a type with getter functions. It's just a constant key/value mapping so you could simply put the weapon attributes in a Lua table.

PR #2932, opened 2024-02-24
tyrant bots cancels charge when in attack range

sweet235:
(insert an essay about why players cancel the trample when in swipe range)

You might say that a player will not move for 2 or 5 frames when releasing the forward key for a short time. I say it is better for the bot to cancel it in one frame than not to cancel it at all.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -2100,6 +2100,10 @@ void BotFireWeaponAI( gentity_t *self )
 			}
 			else
 			{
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

slipher:
Is this supposed to make it play better or is the concern that the current behavior is somehow unfairly strong?

sweet235:
Current behavior makes tyrants jokingly easy to kill. Once their trample is active, they will hit you at most once until the trample time is over. This hit does less damage than a single swipe.

PR #2933, opened 2024-02-24
alien bots: do not chase players when attacked by turrets

sweet235:
If attacked by a turret: choose a new goal. Most often, this will be one of the turrets. Attacking them is a good idea.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -2493,6 +2493,17 @@ static gentity_t *BotPopEnemy( enemyQueue_t *queue )
 
 void BotPain( gentity_t *self, gentity_t *attacker, int )
 {
+	// alien bots should not chase human players when attacked by turrets
+	if ( G_Team( self ) == TEAM_ALIENS
+		 && G_Team( attacker ) != TEAM_NONE
+		 && attacker->s.modelindex == BA_H_MGTURRET
+		 && self->botMind->goal.targetsValidEntity()
+		 && self->botMind->goal.getTargetedEntity()->s.eType == entityType_t::ET_PLAYER )
+	{
+		BotChangeGoalEntity( self, self );  // reset goal
+		return;
+	}
+
 	if ( G_Team( attacker ) != TEAM_NONE
 		&& !G_OnSameTeam( self, attacker )
 		&& attacker->s.eType == entityType_t::ET_PLAYER


src/sgame/sg_bot_util.cpp
```diff
@@ -2493,6 +2493,17 @@ static gentity_t *BotPopEnemy( enemyQueue_t *queue )
 
 void BotPain( gentity_t *self, gentity_t *attacker, int )
 {
+	// alien bots should not chase human players when attacked by turrets
+	if ( G_Team( self ) == TEAM_ALIENS
+		 && G_Team( attacker ) != TEAM_NONE
+		 && attacker->s.modelindex == BA_H_MGTURRET
+		 && self->botMind->goal.targetsValidEntity()
+		 && self->botMind->goal.getTargetedEntity()->s.eType == entityType_t::ET_PLAYER )
```
DolceTriade, line 2501: Do you think there should be some time of heuristic for tradeoff between pursuing into a base (ie, human attacker equipment > bot value) or is this sufficient.

Can this be abused by players by simply building a turret to have bots discourage people from attacking?

Lastly, should this consider rocket pods as well, or is the range too great for this to be effective?

src/sgame/sg_bot_util.cpp
```diff
@@ -2493,6 +2493,17 @@ static gentity_t *BotPopEnemy( enemyQueue_t *queue )
 
 void BotPain( gentity_t *self, gentity_t *attacker, int )
 {
+	// alien bots should not chase human players when attacked by turrets
+	if ( G_Team( self ) == TEAM_ALIENS
+		 && G_Team( attacker ) != TEAM_NONE
+		 && attacker->s.modelindex == BA_H_MGTURRET
```
slipher, line 2499: This should check that the attacker is a buildable before reading modelindex

src/sgame/sg_bot_util.cpp
sweet235: A better heuristic can never hurt. The current heuristic is in `BotFindBestEnemy`.

I assume it can be abused. But the current situation can be abused even more. Build a wall of turrets, make an alien bot chase you, run to the wall and jump over it. The bots will not even try to attack the turrets. A tyrant will get stuck there, because it cannot jump over.

Rocket pod range is too high for a simple solution.

sweet235:
The thing is: with our current gameplay, destroying a turret is almost always a more valuable goal than killing a player.

necessarily-equal:
Beside Slipher's comment, this seems like a nice hack. Maybe we could avoid it by going across all visible enemies every (few) frames and retargeting, but in the meantime this seems nice.

PR #2934, opened 2024-02-24
add more buildable types as rush targets

sweet235:


DolceTriade:
So what this does is makes bots prioritize some defensive structures instead of just heading for RC/OM + spawns?

slipher:
> So what this does is makes bots prioritize some defensive structures instead of just heading for RC/OM + spawns?

I wouldn't called it "prioritizing" since they only go for those if they don't know the location of any main structure or spawns.

LGTM

sweet235:
> So what this does is makes bots prioritize some defensive structures instead of just heading for RC/OM + spawns?

The actual priority is given in `BotGetEnemyPriority( gentity_t *self, gentity_t *ent )`, see https://github.com/Unvanquished/Unvanquished/pull/2935 for a recent tweak there. This PR only changes what `action rush` in the .bt files does. Currently, `action rush` will fail if the bots do not know of enemy spawns or the main buildable. Making them rush for other buildings too will often make them discover new buildings.

DolceTriade (approved):

DolceTriade:
Makes sense. Thanks!

Issue #2886, opened 2023-12-30
Brazilian portuguese (pt-br)


marcoaureliocardoso:
Can you enable brazilian portuguese (pt-br) on Weblate?

illwieckz:
Hi, if I'm right you can add a new language yourself by clicking the big “+” button on the [Weblate project page](https://hosted.weblate.org/projects/unvanquished/unvanquished/), here is a direct link to the page to add a language: [add new](https://hosted.weblate.org/new-lang/unvanquished/unvanquished/).

![Capture d’écran du 2023-12-30 09-06-50 1](https://github.com/Unvanquished/Unvanquished/assets/2311649/8f1f00ea-cb36-4af6-ad21-894ee05ca87f)

Note that for now the game only supports one variant of a language and will make no difference between Portuguese and Brazilian portuguese. So I recommend to contribute to the Portuguese language to begin with.

marcoaureliocardoso:
Thanks for the quick response.

That translation present on Weblate has already been started with Portuguese from Portugal and, considering the differences with the Brazilian language, I believe that players from Brazil would be much happier with the game in their language.

Regarding your suggestion, I could even collaborate with the existing translation on Weblate, but I don't completely master the nuances of Portuguese as spoken in Portugal.

marcoaureliocardoso:
I completed the translation to pt_BR.
https://hosted.weblate.org/projects/unvanquished/unvanquished/pt_BR/

I hope it is useful for the project. 😉

slipher:
We can't automatically detect country variants for now, but I think they could be added for selection in the menu.
Like `pt` -> Portuguese (Portugal), `pt_BR` -> Portuguese (Brazil). (Assuming that tinygettext maps `pt` to the Portugal variant; it could be the other way around).

Given that pt_PT is mostly incomplete, we could just disable it for now and make pt_BR the only one.

PR #2941, opened 2024-03-02
[tinygettext] Fix language matching, compare to lowercase-d specs

### LINKED ISSUE #2938: Needs manual intervention for Chinese language support
I’d like to contribute some variants of Chinese translation, but I’ve found some problems.

After analyzing I came up with fixes. Pick one of these:

## The easy way

We *need* specific Chinese variants, namely `zh_CN`, `zh_HK` and `zh_TW`, but Weblate platform doesn’t have these:
(As for why *need*… see explanations later)
Screenshot_20240302_124913
Screenshot_20240302_125006

In the mean while, weblate says:
> This list includes only basic languages. Please contact project maintainers to add others.

So if maintainers are able to add those variants manually (and I can access them) the problem is solved.

## The hard way

As shown in screenshots above, Weblate do has Chinese variants called `zh_Hans`, `zh_Hant_HK`, and `zh_Hant`.
They loosely corresponds to `zh_CN`, `zh_HK` and `zh_TW`.

But the *code* doesn’t allow us to use `zh_Han*` correctly.

To fix, we need to patch (or vendor) the tinygettext library:
- Add `zh_Han*` variants to [`language.cpp`](https://github.com/Unvanquished/Unvanquished/blob/ae89e2591c1bf52ab8311e1f9dd43efd6cf9b65c/libs/tinygettext/src/language.cpp#L281):
```cpp
  { "zh", "Hans",     nullptr, "Chinese (simplified)"        },
  { "zh", "Hant_HK",  nullptr, "Chinese (Hong Kong)"         },
  { "zh", "Hant",     nullptr, "Chinese (traditional)"       },
```
- Work out a way to let the helper function [`convertFilename2Language`](https://github.com/Unvanquished/Unvanquished/blob/ae89e2591c1bf52ab8311e1f9dd43efd6cf9b65c/libs/tinygettext/src/dictionary_manager.cpp#L269) to don’t capitalize so-called “country specification” in these cases (`zh_HANS` is wrong, and makes all [matches](https://github.com/Unvanquished/Unvanquished/blob/ae89e2591c1bf52ab8311e1f9dd43efd6cf9b65c/libs/tinygettext/src/dictionary_manager.cpp#L136) score the same, all variants will use a same `.po` file)
  - In my mind, matching after making both filename and contrycode lowercase should be ok

## The unofficial way

I could also keep Chinese translations by myself in a git repository as a community effort.
(see [one of my similar work](https://github.com/NaitLee/xon-zh-grand-l10n), quality guaranteed 😉)
For this case you can have a dedicated dpk for distribution, if that’s acceptable.

NaitLee:
[tinygettext] Fix language matching, compare to lowercase-d specs, instead of forcibly making "country" spec uppercase (which could go wrong when that field isn’t really a country)

Fixes #2938 the hard way.

Note that this fix assumes that results from `Language::from_env` are only used for comparison to preset languages, rather than being used directly.

### DIFF
--- a/libs/tinygettext/src/dictionary_manager.cpp
+++ b/libs/tinygettext/src/dictionary_manager.cpp
@@ -125,7 +125,7 @@ DictionaryManager::get_dictionary(const Language& language)
         if (has_suffix(*filename, ".po"))
         { // ignore anything that isn't a .po file
 
-          Language po_language = Language::from_env(convertFilename2Language(*filename));
+          Language po_language = Language::from_env(*filename);
 
           if (!po_language)
           {
@@ -258,41 +258,6 @@ DictionaryManager::set_filesystem(std::unique_ptr<FileSystem> filesystem_)
 {
   filesystem = std::move(filesystem_);
 }
-// ----------------------------------------------------------------------------
-/** This function converts a .po filename (e.g. zh_TW.po) into a language
- *  specification (zh_TW). On case insensitive file systems (think windows)
- *  the filename and therefore the country specification is lower case
- *  (zh_tw). It Converts the lower case characters of the country back to
- *  upper case, otherwise tinygettext does not identify the country
- *  correctly.
- */
-std::string DictionaryManager::convertFilename2Language(const std::string &s_in) const
-{
-    std::string s;
-    if(s_in.substr(s_in.size()-3, 3)==".po")
-        s = s_in.substr(0, s_in.size()-3);
-    else
-        s = s_in;
-
-    bool underscore_found = false;
-    for(unsigned int i=0; i<s.size(); i++)
-    {
-        if(underscore_found)
-        {
-            // If we get a non-alphanumerical character/
-            // we are done (en_GB.UTF-8) - only convert
-            // the 'gb' part ... if we ever get this kind
-            // of filename.
-            if(!::isalpha(s[i]))
-                break;
-            s[i] = static_cast<char>(::toupper(s[i]));
-        }
-        else
-            underscore_found = s[i]=='_';
-    }
-    return s;
-}   // convertFilename2Language
-
 
 } // namespace tinygettext
 
--- a/libs/tinygettext/src/language.cpp
+++ b/libs/tinygettext/src/language.cpp
@@ -281,6 +281,9 @@ static const LanguageSpec languages[] = {
   { "zh", "CN", nullptr, "Chinese (simplified)"        },
   { "zh", "HK", nullptr, "Chinese (Hong Kong)"         },
   { "zh", "TW", nullptr, "Chinese (traditional)"       },
+  { "zh", "Hans",     nullptr, "Chinese (simplified)"        },
+  { "zh", "Hant_HK",  nullptr, "Chinese (Hong Kong)"         },
+  { "zh", "Hant",     nullptr, "Chinese (traditional)"       },
   { "zu", nullptr,    nullptr, "Zulu"                        },
   { nullptr, nullptr,    nullptr, nullptr                          }
 };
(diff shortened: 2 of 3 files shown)


NaitLee:
[From #2886](https://github.com/Unvanquished/Unvanquished/issues/2886#issuecomment-1872522333):
> We can't automatically detect country variants for now,

If I didn’t get it wrong, this is fixed too.

DolceTriade (commented):
Does this break behavior for existing translations which don't have the country set?

NaitLee:
> Does this break behavior for existing translations which don't have the country set?

No, the matching is simply made loose, current Unvanquished translations are working well.

I’d suggest to also give it some tests when upstreaming, for some other projects are using tinygettext too, and some among them have better translation support.

libs/tinygettext/src/dictionary_manager.cpp
```diff
@@ -258,41 +258,6 @@ DictionaryManager::set_filesystem(std::unique_ptr<FileSystem> filesystem_)
 {
   filesystem = std::move(filesystem_);
 }
-// ----------------------------------------------------------------------------
-/** This function converts a .po filename (e.g. zh_TW.po) into a language
- *  specification (zh_TW). On case insensitive file systems (think windows)
- *  the filename and therefore the country specification is lower case
- *  (zh_tw). It Converts the lower case characters of the country back to
- *  upper case, otherwise tinygettext does not identify the country
- *  correctly.
```
DolceTriade, line 267: this logic seems required no?

On windows the filenames could be `zh_hant_hk.po`

And thus won't be identified by tinygettext?

NaitLee: You can see `Language::match` now compares language specs in lowercase.

For the purpose of `DictionaryManager::get_dictionary`, tinygettext compares pre-defined languages (they have correct cases) to languages from filenames, and then *uses a pre-defined language that is closest to the best file*. So we don’t really need to “correct” that so-called “country” spec, but rather match it loosely, ignoring case.

I’d test on Windows when I’m available. This fix works well on Linux (I have `zh_Han*.po` language files locally, that I’m working on)

DolceTriade: Ah yes, I see.

DolceTriade (approved):

sweet235:
This has been approved, and may be merged.

@NaitLee leave a comment if you want to merge this, I will do it.

libs/tinygettext/src/language.cpp
```diff
@@ -281,6 +281,9 @@ static const LanguageSpec languages[] = {
   { "zh", "CN", nullptr, "Chinese (simplified)"        },
   { "zh", "HK", nullptr, "Chinese (Hong Kong)"         },
   { "zh", "TW", nullptr, "Chinese (traditional)"       },
+  { "zh", "Hans",     nullptr, "Chinese (simplified)"        },
+  { "zh", "Hant_HK",  nullptr, "Chinese (Hong Kong)"         },
+  { "zh", "Hant",     nullptr, "Chinese (traditional)"       },
```
illwieckz, line 286: Isn't it confusing if the “long name” (in parenthesis) is the same as other values? Can those lang code be fully exchanged? Is zh_Hant always usable instead of zh_TW as a code for the same translation (and the reverse being true), basically zh_TW being an alias to zh_Hant?

NaitLee: For `zh_Hans` the `Han` represents 「漢」“Chinese” and `s` for “simplified”. I think it is the `zh_CN` to be changed to `Chinese (China)` instead, because the `CN` part really means a country.

Which one to use depends on our choices of infrastructure (as the situation stated in #2938), and we only choose one between `zh_Hans` and `zh_CN`, as their difference can be disregarded after then.

I’d differentiate them as you said.

illwieckz: Thanks.

PR #2944, opened 2024-03-03
navgen: Use parallel threads to generate navmeshes

DolceTriade:
This greatly speeds up navmesh generation since now we can generate all the classes at once.

For comparison: generating plat23 takes ~30s serially but takes ~3s in parallel. Anthil takes 7 min serially but takes 30s in parallel.

We use C++'s regular thread interface which is supported by NaCL as well. One interesting thing about using multiple threads in the gamelogic is that we cannot make trap calls on threads since our trap call interface is single threaded (ie, we send a message and immediately wait for a reply on the same socket. If we send two messages, we will mess up replies for one or the other). Therefore, to work around this problem, there is a new `AsyncRunOnMainThread` interface which will asynchronously run trap calls on the main thread when you call `RunTasks()`. This works well enough.

To control parallelism, we added two cvars (for cgame and sgame): `cg_navgen_maxThreads` and `g_navgen_maxThreads`. Setting these to zero or below will result in all navmeshes being generated at once. By default, we generate them all at once as this is the fastest. It might be required to limit parallelism for some more unique maps like epic5 where it might eat up all your RAM and cause daemon to OOM.

Lastly, you can create navmeshes without starting the game using daemonded:

```
./daemonded -pakpath ../pkg -pakpath ~/.local/share/unvanquished/base/pkg -set vm.sgame.type 3 -set vm.cgame.type 3 +devmap <MAP_NAME> +navgen all +quit
```

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -157,6 +157,9 @@ Cvar::Cvar<float> cg_motionblurMinSpeed("cg_motionblurMinSpeed", "minimum speed 
 Cvar::Cvar<bool> cg_spawnEffects("cg_spawnEffects", "desaturate world view when dead or spawning", Cvar::NONE, true);
 
 static Cvar::Cvar<bool> cg_navgenOnLoad("cg_navgenOnLoad", "generate navmeshes when starting a local game", Cvar::NONE, true);
+static Cvar::Cvar<int> cg_navgen_maxThreads(
+	"cg_navgen_maxThreads", "Maximum number of threads to use when generating navmeshes. A value of zero or below means unlimited.",
+	Cvar::NONE, 0);
 
 // search 'fovCvar' to find usage of these (names come from config files)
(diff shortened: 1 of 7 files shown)


slipher:
I think we could get rid of the "async" stuff except logs if we just write the files on the main thread.

Will take a closer look when I get back from vacation

slipher:
I see 2 nontrivial shortcomings:
- sgame background navgen is not cancelable. If you change map while it is running, the server will freeze until all threads finish generating their current class.
- The possibility to run on a single thread is removed. I think we should keep that around in case of further porting efforts to platforms without thread support.

I might be asking a lot so I will take a stab at fixing it and see how feasible it is.

illwieckz:
@DolceTriade I assume this one is obsoleted by:

- https://github.com/Unvanquished/Unvanquished/pull/2955

and can be closed?

DolceTriade:
yes.

Issue #2926, opened 2024-02-17
My map voxelstorm11 has incomplete navmeshes for small classes


sweet235:
For example, do

```
/navedit enable level0
/setviewpos 1730 8874 3853
```

Strangely, _reducing_ the `cellSizeFactor` in `voxelstorm11.navcfg` can give complete navmeshes. But the tiles are so small that the search is far too slow at bot runtime.

The map is here: https://users.unvanquished.net/~sweet/pkg/map-voxelstorm11_1.dpk

Notice that the map is big, but not very huge.

sweet235:
Here is a screenshot of the level4 navmesh at this place:

![unvanquished_2024-02-18_103848_000](https://github.com/Unvanquished/Unvanquished/assets/110394607/fd3ccda0-29cc-4c52-9468-1fe5fe7e0b84)

And here is the level0 navmesh:

![unvanquished_2024-02-18_103915_000](https://github.com/Unvanquished/Unvanquished/assets/110394607/ef5d33a7-3612-4c91-abb0-0b8ae3f0a879)

Notice how the level0 navmesh ends at the upper right corner in this image.

DolceTriade:
Is this an artifact of the stair size? is this something that happens in all maps with stairs?

sweet235:
> Is this an artifact of the stair size? is this something that happens in all maps with stairs?

I have no idea about the exact cause. But I am convinced it is not the stair size.

Something similar happens at some maps with considerable vertical extent, like map-cube.

In situations like this, we can simply use the level4 navmesh for all classes. This is not optimal, but not too bad either: one simply copies the file `voxelstorm11-level4.navMesh` to `voxelstorm11-level0.navMesh`. But at some point in the future, the server will overwrite these files.

Notice how we moved the responsibility for this from the mapper to the server.

PR #2947, opened 2024-03-06
increase recast option EXPECTED_LAYERS_PER_TILE

sweet235:
Fix https://github.com/Unvanquished/Unvanquished/issues/2926.

It seems we set an upper bound of tiles for recast. It depends on `EXPECTED_LAYERS_PER_TILE`, in `nav.cpp`, we use this number to set something called `tcparams.maxTiles`.

I had to increase the number to as much as 12, in order to create complete navmeshes for map-cube.

### DIFF
--- a/src/shared/navgen/navgen.h
+++ b/src/shared/navgen/navgen.h
@@ -34,7 +34,7 @@
 #include "shared/bg_public.h"
 
 static const int MAX_LAYERS = 32;
-static const int EXPECTED_LAYERS_PER_TILE = 4;
+static const int EXPECTED_LAYERS_PER_TILE = 12;
 
 struct ladder_t
 {


DolceTriade:
What impact does this have on other maps? Any difference in time or memory?

sweet235:
> What impact does this have on other maps? Any difference in time or memory?

I think not, but who can tell. But who can tell for sure. I'll run regen all my navmeshes using this, and report back.

slipher:
Apparently increasing the limit on tiles *decreases* the limit on some other thing? From NavmeshGenerator::WriteFile()

```
// there are 22 bits to store a tile and its polys
int tileBits = rcMin( ( int ) dtIlog2( dtNextPow2( d_->tcparams.maxTiles ) ), 14 );
int polyBits = 22 - tileBits;
```

Maybe this is a good thing to investigate for why part of the navmesh disappears in large maps

sweet235:
By trying this, I did not spot any problem yet.

sweet235:
This is not the solution. Some big maps still do not get complete navmeshes without this.

sweet235:
I have been doing this for about a year now, and did not encounter any problems. Some other maps I made require this too.

PR #2948, opened 2024-03-06
add navgen option walkableRadiusFactor

sweet235:
There is a margin between the edge of a navmesh, and the actual geometry causing this edge. Look at the level4 navmesh on map antares for example:

unvanquished_2024-03-06_103603_000

Notice how these margins cause a gap in the navmesh, where the white circle is. The tyrant bots will not leave their base on this path, despite the fact that a tyrant fits in there. 

This is very unfortunate, as there is a similar gap at the other exit as well. So tyrant bots cannot leave their base.

The margin is controlled by a value `walkableRadius` we tell recast to use. Add a navgen option `walkableRadiusFactor`, to tweak this value. For example, create a file `game/maps/antares.navcfg`:

```
walkableRadiusFactor 0.75
```

Restart, and look at the same navmesh:

unvanquished_2024-03-06_103634_000

There is one other important reason to keep the margins as small as we can: bots will only pick enemy buildings as targets if their origin is over the navmesh (with minor exceptions).

This is a very dangerous value to tweak. Make the margin too small, and bots might get stuck at little edges on the walls.

### DIFF
--- a/src/shared/bot_nav_shared.cpp
+++ b/src/shared/bot_nav_shared.cpp
@@ -167,6 +167,18 @@ static void ParseOption( Str::StringRef name, Str::StringRef value, Str::StringR
 			return;
 		}
 	}
+	else if ( Str::IsIEqual( name, "walkableRadiusFactor" ) )
+	{
+		if ( floatValue < 0.1f || 10.f < floatValue )
+		{
+			Log::Warn( "%s: incorrect value for walkableRadiusFactor '%f': must be between 0.1 and 10", file, floatValue );
+		}
+		else
+		{
+			config.walkableRadiusFactor = floatValue;
+			return;
+		}
+	}
 	else
 	{
 		Log::Warn( "%s: unknown navgen setting '%s'", file, name );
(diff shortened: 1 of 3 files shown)


DolceTriade:
Should this be a per-class value?

sweet235:
> Should this be a per-class value?

Probably. Some navgen options should be. But we did not implement per-class configs yet.

sweet235:
> This is a very dangerous value to tweak.

Indeed. Setting it to 0.1, and generating navmeshes for map kaiji-starsidehotel crashes the server.

DolceTriade:
Why did it crash? Can 0.1 be a valid value for some maps but cause others to crash?

sweet235:
> Why did it crash?

I suspect these narrow bridges:

unvanquished_2024-04-09_181016_000

I could not make any bot cross them yet.

> Can 0.1 be a valid value for some maps but cause others to crash?

I'm not sure yet.

sweet235:
Note to self: it has been communicated to me in private that a minimum of 2 could be enough for both values.

sweet235:
Rebased to fix conflict with master.

Reduced the allowed range of values to the interval from 0.6 to 2.0. This prevents problems on all known maps.

PR #2946, opened 2024-03-05
cgame: use easier-to-translate strings with less argument-ordering problems

illwieckz:
cgame: use easier-to-translate strings with less argument-ordering problems

Obsolete:

- https://github.com/Unvanquished/Unvanquished/pull/2885

PR #2949, opened 2024-03-06
cgame: better obituaries strings, <3 cu-kai

illwieckz:
Better obituaries strings, see:

- https://github.com/Unvanquished/Unvanquished/pull/2946

### DIFF
--- a/src/cgame/cg_event.cpp
+++ b/src/cgame/cg_event.cpp
@@ -472,44 +472,44 @@ static void CG_Obituary( entityState_t *ent )
 				break;
 
 			case MOD_LEVEL1_CLAW:
-				message = G_( "%s%s ^*was sliced by %s%s%s^*'s claws" );
-				messageAssisted = G_( "%s%s ^*was sliced by %s%s%s^*'s claws; %s%s%s^* assisted" );
+				message = G_( "%s%s ^*was sliced by %s%s%s^*'s mantis" );
+				messageAssisted = G_( "%s%s ^*was sliced by %s%s%s^*'s mantis; %s%s%s^* assisted" );
 				break;
 
 			case MOD_LEVEL2_CLAW:
-				message = G_( "%s%s ^*was shredded by %s%s%s^*'s claws" );
-				messageAssisted = G_( "%s%s ^*was shredded by %s%s%s^*'s claws; %s%s%s^* assisted" );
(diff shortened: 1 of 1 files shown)


DolceTriade (commented):
Would it be better to have emojis here rather than names?

Otherwise LGTM

cu-kai:
If you use capital letters for the class names, I will be happy :smile:

illwieckz:
> If you use capital letters for the class names, I will be happy 😄

To me it looks like capital letters was an unintended side effect of building the string with variables (and variable not having any context about the sentence they are used in), existing strings were already like this:

- `%s%s ^*was incinerated by %s%s%s^*'s firebomb"`
- `"%s%s ^*was gently nibbled by %s%s%s^*'s granger"`

illwieckz:
> Would you prefer it if we didn't use the capital letters?

Yes.

cu-kai (approved):

illwieckz:
As an exception I will not wait tomorrow to merge this because it fixes wrong strings I merged early in the day and those wrong strings were automatically pushed to weblate and I don't want to waste translator's time who may translate the wrong strings for nothing.

NaitLee:
“player’s tyrant” sounds like “the tyrant that `player` is controlling”, a bit weird, but ok.
Indeed I think all of these are fine. (I don’t mind to re-translate)

```
"%s%s ^*was mauled by %s%s%s^*'s claws"
"%s%s ^*was mauled by %s%s%s^*'s tyrant"
"%s%s ^*was mauled by %s%s%s^*'s Tyrant"
```

And yes, I have capitalized aliens’ names in my translations because I thought they are just names and that’s consistent in a Chinese paragraph. I should (or I must) eventually give them meaningful Chinese names for sure, so that majority of players could understand them.

illwieckz:
> “player’s tyrant” sounds like “the tyrant that `player` is controlling”, a bit weird, but ok.

Yeah, right now it's the expected meaning.

illwieckz:
If it's more consistent with chinese language to capitalize this in chinese it's fine. Sometime capitalization is language specific. For example in English we say « in English » (capitalized), but in French we say « en français » (not capitalized). I guess you'll know better what to do for Chinese and also in the end it's better if all sentences are consistent together for a single language (even if two options are possible for the same language).

PR #2950, opened 2024-03-07
make human bots climb ladders without jumping

sweet235:
Human bots can climb ladders. They will use forward traces to detect them. When provided with an appropriate navcon, they will "press and hold the jump key" to climb the ladder.

This fails when their stamina is too low. A battlesuit bot stuck at the bottom of a ladder is a common sight for that reason.

Change how they climb ladders, make them look up instead of jumping. This is what players do.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -523,12 +523,19 @@ static const gentity_t* BotGetPathBlocker( gentity_t *self, const glm::vec3 &dir
 	return nullptr;
 }
 
+typedef enum
+{
+	OBSTACLE_NONE,
+	OBSTACLE_JUMPABLE,
+	OBSTACLE_LADDER
+} obstacle_t;
+
 // checks if jumping would get rid of blocker
 // return true if yes
 //
 // This currently doesn't handle jumping over map geometry, but that could be
 // nice some day.
-static bool BotShouldJump( gentity_t *self, const gentity_t *blocker, const glm::vec3 &dir )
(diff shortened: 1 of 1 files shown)


sweet235:
Here is a test scenario for both the current and the proposed climb method. Create a file `maps/parpax-human_naked.navcon`:

```
navcon 3
2664 98 -2254 2694 834 -2249 30 1 63 0
2705 834 -2258 2669 834 -2416 30 1 63 0
2705 834 -2258 2671 834 -2063 30 1 63 0
```

Load map parpax, and do this:

```
/bot add * h 5
/g_freeFundPeriod 1
```

Soon, the bot will be rich and use the ladder to reach the alien base.

src/sgame/sg_bot_nav.cpp
```diff
@@ -740,11 +752,17 @@ static bool BotAvoidObstacles( gentity_t *self, glm::vec3 &dir, bool ignoreGeome
...
-	if ( BotShouldJump( self, blocker, dir ) )
+	obstacle_t obst = BotDetectObstacle( self, blocker, dir );
+	if ( obst == OBSTACLE_JUMPABLE )
 	{
 		BotJump( self );
 		return false;
 	}
+	else if ( obst == OBSTACLE_LADDER )
+	{
+		dir.z += 200;
```
DolceTriade, line 763: do we ever have to worry about this overflowing or does this get clamped somewhere else?

sweet235: That is actually `+= 200.f`. As far as I know, the maximum z coordinate is around 65536. The maximum float value is around 3e38.

DolceTriade (approved):

PR #2952, opened 2024-03-11
Print ladders with a command instead of as a side effect of navgen

slipher:
Here's an illustration of a ladder in Yocto which got a better result (before vs. after):

unvanquished_2024-03-10_235627_000

unvanquished_2024-03-11_024414_000


DolceTriade:
Will test this on other maps with ladders.

DolceTriade:
What's the expected usage of the ladder command?

slipher:
It gives a starting point for making navcons for ladders

DolceTriade:
I guess I don't understand the screenshots. Before, were ladders just printed but no other action was taken? Now we can list ladders on demand?

What do the screenshots have to do with the changes

slipher:
> I guess I don't understand the screenshots. Before, were ladders just printed but no other action was taken? Now we can list ladders on demand?

Yes, before they were spammed during navmesh generation; now they are listed on demand.

> 
> What do the screenshots have to do with the changes

I wrote some debug draw code (not part of the PR) to show the ladder start and end points.

PR #2954, opened 2024-03-15
add Chinese language to the UI menu for selection

NaitLee:
After #2941, both client and server can handle locale code with “country specification” correctly. They will, say, distinguish `zh_Hans` vs `zh_Hant` or `en_US` vs `en_GB`, not mixing them up.

This adds 3 variants of Chinese language (`zh_Hans`, `zh_Hant_HK` and `zh_Hant`) to the UI menu for selection, as they are completed (>50% progress), and will just work.
(“completed” as “100% progress”, proof reading & quality fixes going on 😄)

Note that language detection isn’t fixed at the moment, so a clean startup on a machine with `LANG=zh_*` would always select a specific variant, whichever the code encounters first.


DolceTriade (approved):

illwieckz:
Thanks. 👍️

PR #2955, opened 2024-03-15
Use multiple threads to generate navmeshes

slipher:
Some improvements compared to #2944, in particular:
- keeps code for generating on main thread only for sgame background navgen
- can cancel navgen if the sgame shuts down so as not to hang the server
- fixes regression that map-level errors are no longer cached

Also I added menu options.

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1125,16 +1125,16 @@ static void GenerateNavmeshes()
 		cg.navmeshLoadingFraction = classesCompleted / classesTotal;
 		trap_UpdateScreen();
 
-		navgen.StartGeneration( species );
+		std::unique_ptr<NavgenTask> task = navgen.StartGeneration( species );
 		do
 		{
-			float fraction = ( classesCompleted + navgen.SpeciesFractionCompleted() ) / classesTotal;
+			float fraction = ( classesCompleted + task->FractionCompleted() ) / classesTotal;
 			if ( fraction - cg.navmeshLoadingFraction > 0.01 )
 			{
 				cg.navmeshLoadingFraction = fraction;
 				trap_UpdateScreen();
 			}
-		} while ( !navgen.Step() );
+		} while ( !navgen.Step( *task ) );
 		++classesCompleted;
 	}
 	cg.loadingNavmesh = false;
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1117,7 +1117,7 @@ static void GenerateNavmeshes()
 	trap_UpdateScreen();
 
 	NavmeshGenerator navgen;
-	navgen.EnqueueTasks( mapName, missing );
+	navgen.LoadMapAndEnqueueTasks( mapName, missing );
 	navgen.StartBackgroundThreads( cg_navgenMaxThreads.Get() );
 	navgen.WaitInMainThread( []( float progress ) {
 		if ( progress >= cg.navmeshLoadingFraction + 0.01f )
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -158,6 +158,9 @@ Cvar::Cvar<float> cg_motionblurMinSpeed("cg_motionblurMinSpeed", "minimum speed 
 Cvar::Cvar<bool> cg_spawnEffects("cg_spawnEffects", "desaturate world view when dead or spawning", Cvar::NONE, true);
 
 static Cvar::Cvar<bool> cg_navgenOnLoad("cg_navgenOnLoad", "generate navmeshes when starting a local game", Cvar::NONE, true);
+static Cvar::Cvar<int> cg_navgenMaxThreads(
+	"cg_navgenMaxThreads", "Maximum number of threads to use when generating navmeshes",
+	Cvar::NONE, 2);
 
 // search 'fovCvar' to find usage of these (names come from config files)
 // 0 means use global FOV setting
@@ -1075,7 +1078,7 @@ bool CG_ClientIsReady( int clientNum )
 static void GenerateNavmeshes()
 {
 	std::string mapName = Cvar::GetValue( "mapname" );
-	std::vector<class_t> missing;
+	std::bitset<PCL_NUM_CLASSES> missing;
 	NavgenConfig config = ReadNavgenConfig( mapName );
 	bool reduceTypes;
 	if ( !Cvar::ParseCvarValue( Cvar::GetValue( "g_bot_navmeshReduceTypes" ), reduceTypes ) )
@@ -1089,7 +1092,7 @@ static void GenerateNavmeshes()
 		std::string filename = NavmeshFilename( mapName, species );
 		if ( BG_FOpenGameOrPakPath( filename, f ) < 0 )
 		{
-			missing.push_back( species );
+			missing.set(species);
 			continue;
 		}
 		NavMeshSetHeader header;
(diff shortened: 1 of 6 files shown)


src/shared/navgen/nav.cpp
```diff
@@ -40,13 +40,35 @@
...
+}
+
+void UnvContext::RunOnMainThread( std::function<void()> f )
+{
+	mainThreadTasks_.push_back( std::move( f ) );
+}
+
+void UnvContext::DoMainThreadTasks()
+{
+	for ( auto &f : mainThreadTasks_ )
```
DolceTriade, line 69: Does this need to be cleared at the end of the function? If not, please comment.

src/shared/navgen/nav.cpp
```diff
@@ -61,43 +83,52 @@ static NavgenStatus::Code CodeForFailedDtStatus(dtStatus status)
 
 static int tileSize = 64;
 
-void NavmeshGenerator::WriteFile() {
-	if ( d_->status.code == NavgenStatus::TRANSIENT_FAILURE )
+void NavmeshGenerator::WriteFile( const NavgenTask &t ) {
```
DolceTriade, line 86: Can we leave comments about which functions are main thread only here too?

src/shared/navgen/nav.cpp
```diff
@@ -1009,16 +1037,49 @@ static NavgenStatus rasterizeTileLayers( Geometry& geo, rcContext &context, int
...
+	std::string names;
+
+	for ( int i = PCL_NUM_CLASSES; --i != PCL_NONE; )
+	{
+		if ( !classes[ i ] )
+		{
+			continue;
+		}
+
+		Init( mapName );
```
DolceTriade, line 1058: Just init once outside the main loop to make the intent clear?

DolceTriade:
What are the numbers with and without patch for:
 - Anthill
 - cube
 - plat23

I'll help test. Let's see if we can get this into 0.55

slipher:
All right, I tested like this `time( ./daemonded.exe -set vm.sgame.type 1 -pakpath /c/unv/Unvanquished/pkg (-set g_bot_navgen_maxThreads 7) +devmap plat23 +navgen all +quit )` for blocking navgen and this `time( ./daemonded.exe -set vm.sgame.type 1 -pakpath /c/unv/Unvanquished/pkg -set g_bot_navgen_maxThreads 7 -set logs.level.sgame.navgen debug +devmap plat23 +bot fill 5 )` (plus manually clicking X at the end, so less accurate) for background navgen. Tested with NaCl on MinGW. I have a 4 cores with hyperthreading processor.

Branch | Navgen Type | plat23 | cube | anthill
--- | --- | --- | --- | ---
master | blocking navgen | 0m4.614s | 1m18.648s | 3m47.651s
master | background navgen | 0m6.513s | 1m20.323s | N/A
#2955 | blocking navgen (7 threads) | 0m2.399s | 0m26.809s | 1m14.501s
#2955 | background navgen (7 threads) | 0m2.657s | 0m29.411s | 1m18.005s


Also experimented a bit to see if the hyperthreading works well or not, I think not really. For example in a 4-threaded Anthill run I got 51542ms and 14083ms for Tyrant while with 7 threads I got 38567ms and  9689ms for Tyrant.

src/cgame/cg_main.cpp
```diff
@@ -158,6 +158,9 @@ Cvar::Cvar<float> cg_motionblurMinSpeed("cg_motionblurMinSpeed", "minimum speed
 Cvar::Cvar<bool> cg_spawnEffects("cg_spawnEffects", "desaturate world view when dead or spawning", Cvar::NONE, true);
 
 static Cvar::Cvar<bool> cg_navgenOnLoad("cg_navgenOnLoad", "generate navmeshes when starting a local game", Cvar::NONE, true);
+static Cvar::Cvar<int> cg_navgenMaxThreads(
```
DolceTriade, line 161: consider using the max threads possible by default. no reason not to.

DolceTriade (commented):
Also I need to get around to testing this on a shit VPS which only has a single core. tbh, I don't think we should expect people to run these under nacl, but we'll solicit feedback from others too about this situation regardless.

DolceTriade:
> Also experimented a bit to see if the hyperthreading works well or not, I think not really. For example in a 4-threaded Anthill run I got 51542ms and 14083ms for Tyrant while with 7 threads I got 38567ms and 9689ms for Tyrant.

when I get back home, I'll try running these numbers on my system since I have enough cores to run all the classes in parallel. For me, anthill took ~30s. I don't expect any regressions, but let's be double sure about what we're doing here.

src/cgame/cg_main.cpp
DolceTriade: https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency

src/shared/navgen/nav.cpp
slipher: The idea is to only call it once. I think there is a comment elsewhere about that but I went ahead and added a `clear()` just to avoid suspicious appearances.

src/shared/navgen/nav.cpp
slipher: I added comments to functions which can be called from a non-main thread since there are much fewer of those and they are the ones with restrictions.

src/shared/navgen/nav.cpp
slipher: Done.

src/cgame/cg_main.cpp
slipher: Wow it even works in Nacl. I put the default on hardware_concurrency() - 1 since on Windows at least that's the most you can use without making other apps sluggish.

For the server I am leaving it at 1 background thread by default to respect other services which may be running.

slipher: We should test that hardware_concurrency works on a Linux and a Mac system though

src/shared/navgen/nav.cpp
```diff
@@ -1048,29 +1049,36 @@ void NavmeshGenerator::EnqueueTasks( Str::StringRef mapName, std::bitset<PCL_NUM
...
-			continue;
-		}
+			if ( !classes[ i ] )
+			{
+				continue;
+			}
 
-		Init( mapName );
-		taskQueue_.push_back( StartGeneration( Util::enum_cast<class_t>( i ) ) );
+			LoadMap( mapName );
```
DolceTriade, line 1063: loadmap twice?

DolceTriade:
So this looks good to me. I tested on my shit VPS and it seems to work well. We really shouldn't expect people to generate huge maps like anthill on their single core vps.

I'll solicit some more feedback but otherwise and test perhaps on some more maps, but then we should merge this prior to 0.55.

src/cgame/cg_main.cpp
slipher: Tested that this API works inside and outside Nacl on Linux and Mac.

DolceTriade (approved):

DolceTriade:
LGTM. Let's wait for sweet to give this a try and then we can merge.

sweet235:
My server does not have the means to test this. But this is about local games mostly, and you already tested this.

DolceTriade:
Cool, looks ready to merge.

sweet235:
I am still not sure if adding this complexity is worth it. Even if the code is free of bugs, it will make the server a much more dangerous thing to maintain for any sysadmin. A wrong command might spawn enough threads to grind the whole machine to a halt.

DolceTriade:
> I am still not sure if adding this complexity is worth it. Even if the code is free of bugs, it will make the server a much more dangerous thing to maintain for any sysadmin. A wrong command might spawn enough threads to grind the whole machine to a halt.

I think that this is a really cool feature that significantly reduces the time it takes to generate navmeshes. I think that there are reasonable safeguards to prevent server stoppages.

sweet235:
> I think that there are reasonable safeguards to prevent server stoppages.

I'm good, my server does nothing else. But once we merge this, people will probably have to put our server in a managed container on machines that do other work too.

PR #2957, opened 2024-03-21
add bot action extinguishFire, add extinguish_fire.bt, use them

sweet235:
Make one alien bot devolve to advanced granger, and extinguish burnings alien buildables (if possible). The bot will only put out the fire on burning buildables, not fire on the map geometry. This is similar to the human bots being able to repair damaged buildables, but maybe more useful in practice.

Extinguishing fire is only reasonable near the overmind, because a bot has to reach the overmind, then devolve, and then reach the burning buildables. If this procedure takes too long, the fire vanishes by itself or the buildable is destroyed. Spawning as advanced granger is not useful, as the spawn point might be far away from the burning buildables.

I hope this will contribute to the discussions about wheter the firebomb is too strong. As our current bots only build on the floor, their bases are vulnerable to fire.


src/sgame/sg_bot_ai.cpp
```diff
@@ -1544,6 +1545,53 @@ static AINodeStatus_t BotActionReachHealA( gentity_t *self )
 	return STATUS_RUNNING;
 }
 
+AINodeStatus_t BotActionExtinguishFire( gentity_t *self, AIGenericNode_t *node )
+{
+	botMemory_t const* mind = self->botMind;
+
+	if ( node != self->botMind->currentNode )
+	{
+		if ( !isBuilder( self ) )
```
slipher, line 1554: The granger check should be moved to the beginning of the function, outside the if statement, since it is possible to evolve without changing the current node.

Also it should check for advanced Granger only, not any builder.

slipher:
LGTM

PR #2959, opened 2024-03-27
improve marauder bot wall climb

sweet235:
Make it look up and move straight ahead while jumping. Set a timeout after which the bot goes to the navcon start and tries again.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -944,6 +944,9 @@ void BotMoveUpward( gentity_t *self, glm::vec3 nextCorner )
 	case PCL_ALIEN_LEVEL2:
 	case PCL_ALIEN_LEVEL2_UPG:
 		{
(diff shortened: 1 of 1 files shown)


DolceTriade:
> Set a timeout after which the bot goes to the navcon start and tries again.

Where is this change?

sweet235:
> Where is this change?

This is the same timeout we use for the smaller classes. Experiments suggest that this is appropriate for the marauder too. https://github.com/Unvanquished/Unvanquished/pull/2901 adds a comment to `WallclimbStopTime` explaining that we currently use the same time limit for all classes.

This PR is kind of a bug fix. The old code could make the marauder stuck in a bad location where it ended up by bad luck. It would suicide or loop until a human approaches. Few people noticed that because there are few marauder navcons around. The bot can still loop at a bad navcon, but at least it will repeatedly try to follow it from the start.

DolceTriade (approved):

PR #2961, opened 2024-03-30
dragoon bots: do not use pounce to flee when at a jumpable obstacle

sweet235:
That obstacle is typically some alien building on the floor, blocking the path to the booster/overmind. Do not charge the pounce in this case, as this makes it impossible to jump over the obstacle.

The jumpable obstacle is detected by a short trace. A few frames after triggering the jump, it will not be detected anymore. Then it is safe to charge the pounce again.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -720,7 +720,7 @@ static bool BotFindSteerTarget( gentity_t *self, glm::vec3 &dir )
 // This function tries to detect obstacles and to find a way
 // around them. It always sets dir as the output value.
 // Returns true on error
-static bool BotAvoidObstacles( gentity_t *self, glm::vec3 &dir, bool ignoreGeometry )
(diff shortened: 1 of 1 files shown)


src/sgame/sg_bot_nav.cpp
```diff
@@ -1146,14 +1147,16 @@ bool BotMoveToGoal( gentity_t *self )
...
+				 && obstacle != OBSTACLE_JUMPABLE )
 			{
 				wpm = WPM_SECONDARY;
 				magnitude = LEVEL3_POUNCE_JUMP_MAG;
 			}
 		break;
 		case PCL_ALIEN_LEVEL3_UPG:
-			if ( ps.weaponCharge < LEVEL3_POUNCE_TIME_UPG )
+			if ( ps.weaponCharge < LEVEL3_POUNCE_TIME_UPG
+				 && obstacle != OBSTACLE_JUMPABLE )
```
DolceTriade, line 1159: Please comment these lines

DolceTriade (approved):

PR #2962, opened 2024-03-31
add attack.bt

sweet235:
This is the fifth bot behavior players can use on my server. It makes a bot attack the enemy base.

For reference, I use this on my server:
```
set g_tacticBehaviors "default, defend, attack, follow, stay-here"
```

This is similar to `reckless.bt`, but more elaborate. I do not dare to change `reckless.bt`. It is the simplest interesting example behavior, and was the first behavior I studied while learning the behavior tree language.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/bots/attack.bt
+++ b/pkg/unvanquished_src.dpkdir/bots/attack.bt
@@ -1,0 +1,110 @@
+selectClass
+{
+	selector
+	{
+		condition team == TEAM_ALIENS
+		{
+			spawnAs PCL_ALIEN_LEVEL0
+		}
+		condition team == TEAM_HUMANS
+		{
+			spawnAs WP_MACHINEGUN
+		}
+	}
+}
+
+selector
+{
+	// reset the suicide timer
+	decorator return( STATUS_FAILURE )
+	{
+		action resetMyTimer
+	}
+
+	behavior unstick
+
+	condition team == TEAM_HUMANS
+	{
+		selector
+		{
+			behavior use_medkit
+
+			sequence
+			{
+				condition alertedToEnemy
+				condition percentAmmoClip > 0 || percentClips > 0
+				{
+					selector
+					{
+						sequence
+						{
+							//if goal is a close enough enemy building or if we're gonna die
+							condition ( percentHealth( E_SELF ) < 0.3 || ( distanceTo( E_GOAL ) < 400 && goalType == ET_BUILDABLE ) )
+							condition ( distanceTo( E_FRIENDLYBUILDING ) > 400 )
+							selector
+							{
+								// no need to check the upgrade is in inventory, since action would fail
+								action activateUpgrade( UP_FIREBOMB )
+								action activateUpgrade( UP_GRENADE )
+							}
+						}
+						action fight
+					}
+				}
+			}
+
+			sequence
+			{
+				// the bot is required to attack, only visit an armoury
+				// if one is near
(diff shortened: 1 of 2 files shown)


DolceTriade (approved):

pkg/unvanquished_src.dpkdir/bots/attack.bt
```diff
@@ -0,0 +1,110 @@
...
+		}
+	}
+}
+
+selector
+{
+	// reset the suicide timer
+	decorator return( STATUS_FAILURE )
+	{
+		action resetMyTimer
```
slipher, line 21: The suicide timer is `stuckTime`, not `myTimer`. "My timer" is not used anywhere in this BT so this seems pointless.

sweet235: This is the general purpose timer. It is not set or read in C++. But `default.bt` uses it to wait for 20 seconds before suiciding to spawn as builder. If you change a bot from `attack.bt` to `default.bt`, it might suicide immediately if this timer has not been reset in a while.

We should do this in all behavior intented to be available to regular users.

slipher: Oh so old state is not cleared when switching behaviors? We should fix that.

sweet235: I did not think about this problem when I added that timer. All I cared about was a timer independent of the paths taken in the behavior tree (`decorator timer` is closely tied to these paths and success/fail).

We could reset it on behavior changes. Indeed, we should reset bot targets on behavior change too. But what if someone wants to use that timer in an "inverted" way? Like setting it when some positive condition is met, lets say when `alertedToEnemy`. And later query that time to find out how long a bot has not sensed an enemy. Automatic reset would break this instead.

slipher: > I did not think about this problem when I added that timer. All I cared about was a timer independent of the paths taken in the behavior tree (`decorator timer` is closely tied to these paths and success/fail).
> 
> We could reset it on behavior changes. Indeed, we should reset bot targets on behavior change too. But what if someone wants to use that timer in an "inverted" way? Like setting it when some positive condition is met, lets say when `alertedToEnemy`. And later query that time to find out how long a bot has not sensed an enemy. Automatic reset would break this instead.

I guess the appropriate way to reset `myTimer` is to set it to the time that the behavior tree was set, rather than 0. Then it would behave the most similarly to as if the game had just started.

slipher: I opened #2971 to fix this.

sweet235: > I guess the appropriate way to reset `myTimer` is to set it to the time that the behavior tree was set, rather than 0. Then it would behave the most similarly to as if the game had just started.

That seems to be the most sensible choice.

slipher:
LGTM

Gireen:
ready to be merged. @sweet235 
if no one beats me to it i will merge it in a few days 😸

sweet235:
This PR is stalled by https://github.com/Unvanquished/Unvanquished/pull/2971, with which it conflicts conceptually. It would be better to resolve these conflicts before merging.

slipher:
The BT will work fine with or without #2971. Merging #2971 only means that we could do a small cleanup, removing a handful of lines from the BT.

PR #2964, opened 2024-03-31
add bt function timeSinceLastCombat

sweet235:
There is an old, reliable integer in each client state on the server: the last combat time. It is the last time when a player has been hurt by the opponent team. The original purpose for this number was probably to restrict players from joining the specator team to avoid being killed.

There have been discussions about this number. Should it account for other damage times as well? I think the conclusion was to keep its ancient behavior.

Add a bot behavior function to query this number. This will be useful for future PRs. A bot frequently needs to find out if it is currently in combat.

Edit:
We have a rule: do not merge anything that is not used. This rule does not apply here. This an extension to the bot configuration language.

### DIFF
--- a/src/sgame/sg_bot_parse.cpp
+++ b/src/sgame/sg_bot_parse.cpp
@@ -404,6 +404,11 @@ static AIValue_t stuckTime( gentity_t *self, const AIValue_t* )
 	return AIBoxInt( level.time - self->botMind->stuckTime );
 }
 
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

DolceTriade:
> We have a rule: do not merge anything that is not used. This rule does not apply here. This an extension to the bot configuration language.

tbh, I think that if you can show you will use it in subsequent PRs and you are merging things in chunks for reviewablity, I don't think this rule applies.

PR #2975, opened 2024-04-17
cgame: load IQM human model

illwieckz:
Load IQM human model.

Requires:

- https://github.com/UnvanquishedAssets/res-players_src.dpkdir/pull/19

### DIFF
--- a/src/cgame/cg_animdelta.cpp
+++ b/src/cgame/cg_animdelta.cpp
@@ -71,26 +71,13 @@ bool AnimDelta::ParseConfiguration(clientInfo_t* ci, const char* token2, const c
 
 bool AnimDelta::LoadData(clientInfo_t* ci)
 {
-	char newModelName[ MAX_QPATH ];
-	// special handling for human_(naked|light|medium)
-	if ( !Q_stricmp( ci->modelName, "human_naked"   ) ||
-		!Q_stricmp( ci->modelName, "human_light"   ) ||
-		!Q_stricmp( ci->modelName, "human_medium" ) )
-	{
-		Q_strncpyz( newModelName, "human_nobsuit_common", sizeof( newModelName ) );
-	}
-	else
-	{
-		Q_strncpyz( newModelName, ci->modelName, sizeof( newModelName ) );
-	}
-
 	refSkeleton_t base;
 	refSkeleton_t delta;
 	for ( int i = WP_NONE + 1; i < WP_NUM_WEAPONS; ++i )
 	{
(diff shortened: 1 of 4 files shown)


src/cgame/cg_animdelta.cpp
```diff
@@ -71,17 +71,7 @@ bool AnimDelta::ParseConfiguration(clientInfo_t* ci, const char* token2, const c
...
-	if ( !Q_stricmp( ci->modelName, "human_naked"   ) ||
-		!Q_stricmp( ci->modelName, "human_light"   ) ||
-		!Q_stricmp( ci->modelName, "human_medium" ) )
-	{
-		Q_strncpyz( newModelName, "human_nobsuit_common", sizeof( newModelName ) );
-	}
-	else
-	{
-		Q_strncpyz( newModelName, ci->modelName, sizeof( newModelName ) );
-	}
```
DolceTriade, line 84: why is this required?

slipher:
This should go into a for-x.y.z branch so that developers can keep working on master without having to acquire new assets.

Is there supposed to be any user-visible change from this?

illwieckz:
> This should go into a for-x.y.z branch

Yes, I'll change the target later. I wanted to push the branch early so @DolceTriade can put his hands on it and have everything wired up to test any new animation he may implement for the psaw.

> Is there supposed to be any user-visible change from this?

No. Any visible change should be considered a bug before being investigated.

The purpose of those branch is only to convert the file formats and file layout to the asset pipeline we currently have the best support for.

One can also see the benefit of using recommended conventions so we can rely on default loaders, meaning we can drop the dedicated loading implementation.

Migrating the file for the sake of convention is something I wanted to do for years but the purity win was not worth the effort. Now that @DolceTriade is working on models, making the files compatible with a better workflow and working production pipeline is worth the effort.

src/cgame/cg_animdelta.cpp
illwieckz: This code was required when our human files were not following conventions. Now that the files are following convention, this custom loader can be dropped to rely on the generic one instead. In fact we should drop the custom loader to not look for the old files.

This is useless custom code being deleted. This was basically a workaround for the original files not following conventions.

DolceTriade: we can drop the copy too then

illwieckz: Done.

DolceTriade:
Let's merge this now.

illwieckz:
I just rebased to run the CI on it, this was 325 commits behind…

PR #2977, opened 2024-04-28
cgame: Fix ckit taunt not working

DolceTriade:
We were comparing the wrong enum type. We were comparing a WP_ enum to an animation enum. This causes ckit taunt to not work. Add some comments in bg_public.h indicating that the animation enum order is important and load bearing.


slipher:
LGTM

PR #2981, opened 2024-05-04
cg_marks: implement `trap_CM_BatchMarkFragments` to process all impact marks in one call

illwieckz:
It requires:

- https://github.com/DaemonEngine/Daemon/pull/1114

It expands:

- https://github.com/Unvanquished/Unvanquished/pull/2847

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -317,8 +317,9 @@ static void CG_Creep( centity_t *cent )
 
 	if ( size > 0.0f && tr.fraction < 1.0f )
 	{
-		CG_ImpactMark( cgs.media.creepShader, tr.endpos, cent->currentState.origin2,
-		               0.0f, 1.0f, 1.0f, 1.0f, 1.0f, false, size, true );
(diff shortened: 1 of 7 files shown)


illwieckz:
The purpose is to only do a single `trap` call for all marks on the screen at once. This is expected to bring significant performance boost while many players are fighting around, and especially in alien base as alien buildable creep registers an impact mark every frame. Maybe dead bodies do the same too.

illwieckz:
Unfortunately it doesn't work, marks are nto displayed, so I did a mistake somewhere, and I'm looking for help to debug it.

illwieckz:
My `re.MarkFragments` in engine always return `0`…

illwieckz:
I don't get it. I entirely rewrote this branch from scratch, step by step, checking that everything was working on every step.
But as soon as I call `trap_CM_MarkFragmentsArray`, it stops to work.

Even if I copy-paste the original code into `trap_CM_MarkFragmentsArray`, even if I make `trap_CM_MarkFragmentsArray` calling `trap_CM_MarkFragments`.

illwieckz:
The PR is ready, commits will be squashed, no need to review them one by one.

slipher:
LGTM

PR #2984, opened 2024-05-06
HACK: rocket: use direct Cvar access for HUD cvars

illwieckz:
This is a HACK.

Those cvar checks in the HUD are doing one trap call per frame:

```
pkg/unvanquished_src.dpkdir/ui/hud_basics.rml
184:		<if cvar="cg_minimapActive" condition="==" value="1">
191:			<if cvar="cg_drawTimer" condition="==" value="1">
194:			<if cvar="cg_drawClock" condition="!=" value="0">
197:			<if cvar="cg_drawFPS" condition="==" value="1">
203:			<if cvar="cg_lagometer" condition="==" value="1">
224:	<if cvar="cg_drawSpeed" condition="!=" value="0">
```

Those cvars are defined in game client side, so we don't need a trap call at all.

This implementation is an ugly HACK.

What I would like to see instead, is that when a Cvar is defined from the client source code, it is added to a local Cvar Map, either automatically for all of them, either by an explicit call to some function similar to `Cvar::Latch`, something like:

```c
Cvar::Local(g_drawSpeed);
```

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -2298,6 +2298,7 @@ void CG_SetKeyCatcher( int catcher );
 //
 
 void CG_Rocket_Init( glconfig_t gl );
+std::string CG_Rocket_GetCvarValue( std::string );
 void CG_Rocket_LoadHuds();
 void CG_Rocket_Frame( cgClientState_t state );
 const char *CG_Rocket_GetTag();
--- a/src/cgame/cg_rocket.cpp
+++ b/src/cgame/cg_rocket.cpp
@@ -240,6 +240,194 @@ void CG_Rocket_Init( glconfig_t gl )
 	CG_SetKeyCatcher( rocketInfo.keyCatcher | KEYCATCH_UI );
 }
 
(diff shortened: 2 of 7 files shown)


illwieckz:
Wait… the wrong changes were cherry-picked, so weird.

illwieckz:
> Wait… the wrong changes were cherry-picked, so weird.

Fixed.

illwieckz:
When I merge this above all my “batch trap calls” calls, I get no fps difference as spectator with `cg_draw2D` being enabled or disabled. This not true yet with human and alien hud because of other things that are not investigated yet, but this baseline profits to anyone.

illwieckz:
We can use the same HACK for `RocketCvarInlineElement::OnUpdate`.

VReaperV:
I get no significant difference between `/cg_draw2D off` and `cg_draw2D on` on my end with this branch too (it's within 1ms of difference).

illwieckz:
Obsoleted by:

- https://github.com/DaemonEngine/Daemon/pull/1133

PR #2985, opened 2024-05-06
rocket: do not write cg_activeMinimap on every frame and fix toggling minimap display

illwieckz:
This code was inefficient in multiple ways:

- It wrote every frame (and doing a synchronous trap call) a special cvar to tell RmlUi the minimap was both available and enabled.
- The code to toggle the “available and enabled“ state was only run when the state was “available and enabled”, meaning disabling minimap and re-enabling it would not bring it back.
- The Rml was poorly written and displayed the things differently given the `<if>` markup was `true` or `false` at map load, meaning disabling minimap and re-enabling it would not display the HUD the same way as before disabling it.

Alongside the toggling fixes, it trades:

- 1 cvar read
- 1 cvar write

against:

- 2 cvar reads

Right now the code still does two synchronous trap calls per frame just to tell RmlUi to render the minimap, which is super inefficient, as inefficient as before. But the good news is that there are ways to make RmlUi reading cgame Cvars without trap calls, see #2984 for a HACKY proof of concept:

- https://github.com/Unvanquished/Unvanquished/pull/2984

Combining the two PRs means Unvanquished can do no trap calls at all to render the HUD, instead of doing 6 synchronous IPC trap calls per frame just to know if HUD elements should be displayed or not…

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/hud_basics.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/hud_basics.rml
@@ -84,19 +84,19 @@
 			display: block;
 			width: 12em;
 			height: 12em;
-
-			float: right;
-
 			background-color: black;
 		}
 
-		statsbox {
+		div#minimapbox {
+			float: right;
+		}
+
+		div#statsbox {
 			display: block;
 			width: 25em;
 			height: 12em;
-			padding: 0.5em;
-
 			float: right;
+			padding: 0.5em;
 		}
 
 		stat {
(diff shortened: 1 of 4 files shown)


pkg/unvanquished_src.dpkdir/ui/hud_basics.rml
```diff
@@ -181,9 +181,13 @@
 </head>
 <body id="hud_basics">
 	<region class="topright">
-		<if cvar="cg_minimapActive" condition="==" value="1">
-			<minimap />
-		</if>
+		<minimapbox>
```
slipher, line 184: Use `div` with a class name instead of making up a random element name?

slipher (approved):

pkg/unvanquished_src.dpkdir/ui/hud_basics.rml
illwieckz: Done.

Issue #1883, opened 2022-07-13
vendoring font cleanup


ghost:
I think @illwieckz will be quite interested with this.

The game currently loads 8 `.ttf` files at start:

* `fonts/Roboto-Bold.ttf`
* `fonts/Roboto-BoldItalic.ttf`
* `fonts/Roboto-Italic.ttf`
* `fonts/Roboto-Regular.ttf`
* `fonts/fontawesome-webfont.ttf`
* `fonts/DejaVuSansCondensed.ttf`
* `fonts/MaterialIcons-Regular.ttf`

I have serious doubts that all those fonts are actually used, so we can probably lazy-load some of the stuff there, for a start.

2nd, there's actually 13 files in that `fonts/` folder, here they are with their sizes in bytes:

* 643852	`DejaVuSansCondensed.ttf`
* 141564	`fontawesome-webfont.ttf`
* 107780	`MaterialIcons-Regular.ttf`
* 134556	`Roboto-BoldItalic.ttf`
* 127744	`Roboto-Bold.ttf`
* 135504	`RobotoCondensed-BoldItalic.ttf`
* 127340	`RobotoCondensed-Bold.ttf`
* 133908	`RobotoCondensed-Italic.ttf`
* 125332	`RobotoCondensed-Regular.ttf`
* 132440	`Roboto-Italic.ttf`
* 126072	`Roboto-Regular.ttf`
* 269	`roboto.ttf`
* 16336380	`unifont.ttf`

The RobotoCondensed files are never referenced anywhere AFAICT.

The `roboto.ttf` files contains this text:

> I AM A DUMMY FILE
>
> I'm working around a font/filesystem bug where an old Roboto Thin  (found in the alpha 0.25 package) overrides the normal Roboto font.  I do this by camping on it's filename (roboto.ttf) so it cannot load.
>
> Veyrdite, 2014-07-31
>
> (This is a bad hack)

Lastly, the bigger file (`unifont.ttf`) is 16 megs big, only loaded by engine, possibly for console. [This](http://unifoundry.com/unifont/index.html) is unifont, a huge font with as much as possible glyphs. While this might be ok (I'm not that sure tbh), I see some issues here:

* it means daemon can not draw text in console if the `.dpk` have a problem. In my opinion, this file should be installed by the Updated along daemon itself, so that daemon would be more self-contained. This would also prevent this file to be decompressed at each map loading, and it represents ~50% of the decompressed archive;
[...]

PR #2987, opened 2024-05-07
fonts: Delete placeholder roboto.ttf

DolceTriade:
We tend to regenerate unvanquished_*.dpk and ship less partial diffs, so this is safe to remove now.

Addresses part of #1883


illwieckz (approved):
LGTM.

PR #2989, opened 2024-05-08
cgame: introduce cg_lowQualityModels to load low quality models

illwieckz:
Add `cg_lowModels` to load low quality models.

The purpose of this cvar is to be used in `low` and `lowest` presets.

If `cg_lowModels` is enabled, when looking for `buildables/eggpod/eggpod.iqm`, load `buildables/eggpod/eggpod_low.iqm` if it exists.

It is only implemented for buildables for now, I may implement it for player models later.

The purpose is to load a low quality variant of a model for performance purpose.

This may be a low poly model, but I doubt we have the resource to produce some.

What we can do though, is to edit existing IQM models with IQE Browser and delete some bones.

Some low-end old graphics card can only hardware accelerate the model animations if the models have no more than 41 bones.

Some of our models have many more than that, sometime more than an hundreds of that, forcing the game to fallback on a slow CPU implementation.

But many of our models have some bones we can remove, like legs of buildables. Those legs are animated for moving when being hurt or things like that, but removing the leg animation bones and a few others like the head bones do not break the whole animation most of the time.

It looks like `Cvar::Latch` is not implemented yet on CGame side, this would be appreciated as the `cg_lowModels` cvar should be latched (we need to reload the models to see a difference).

This commit should works on `master` but I make this PR target `for-0.55.0/sync` because since it requires models that are not released, no one can use the cvar yet anyway, and targetting the `for-0.55.0/sync` makes things easier to me when dealing with the various repositories.

_Edit:_ It now also changes the md5anim prefix path to be like IQM, so we better target the future branch.

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -527,7 +527,7 @@ static bool CG_ParseBuildableSoundFile( const char *filename, buildable_t builda
 	return true;
 }
 
-static bool CG_RegisterBuildableAnimation( buildableInfo_t *ci, const char *modelName, int anim, const char *animName,
+static bool CG_RegisterBuildableAnimation( buildableInfo_t *ci, Str::StringRef fileBase, int anim, const char *animName,
 		bool loop, bool reversed, bool clearOrigin, bool iqm )
 {
 	char filename[ MAX_QPATH ];
@@ -535,15 +535,14 @@ static bool CG_RegisterBuildableAnimation( buildableInfo_t *ci, const char *mode
 
 	if ( iqm )
 	{
-		Com_sprintf( filename, sizeof( filename ), "models/buildables/%s/%s.iqm:%s", modelName, modelName, animName );
-		ci->animations[ anim ].handle = trap_R_RegisterAnimation( filename );
+		Com_sprintf( filename, sizeof( filename ), "%s.iqm:%s", fileBase.c_str(), animName );
 	}
-
 	else
 	{
-		Com_sprintf( filename, sizeof( filename ), "models/buildables/%s/%s.md5anim", modelName, animName );
-		ci->animations[ anim ].handle = trap_R_RegisterAnimation( filename );
+		Com_sprintf( filename, sizeof( filename ), "%s_%s.md5anim", fileBase.c_str(), animName );
 	}
+
+	ci->animations[ anim ].handle = trap_R_RegisterAnimation( filename );
 
 	if ( !ci->animations[ anim ].handle )
 	{
@@ -609,6 +608,16 @@ void CG_InitBuildables()
 		defaultHumanSounds[ j ] = trap_S_RegisterSound( filename, false );
 	}
 
+	std::vector<std::string> suffixes;
+	suffixes.reserve(2);
+	
+	if ( cg_lowQualityModels.Get() )
+	{
+		suffixes.push_back( "_low" );
+	}
+
+	suffixes.push_back( "" );
+
 	for ( buildable = (buildable_t)( BA_NONE + 1 ); buildable < BA_NUM_BUILDABLES;
 	      buildable = (buildable_t)( buildable + 1 ) )
 	{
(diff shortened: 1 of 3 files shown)


illwieckz:
Example, default models:

unvanquished_2024-05-08_113419_000

Low models I modified:

unvanquished_2024-05-08_113451_000

VReaperV:
> This may be a low poly model, but I doubt we have the resource to produce some.

We can probably generate them from the original model.

illwieckz:
The effort of doing low poly models is probably not worth it anyway, the gain would not be worth the effort.
The performance killer here is the amount of bones, which is easy to do for most buildables.

VReaperV:
I agree.

src/cgame/cg_buildable.cpp
```diff
@@ -526,21 +526,21 @@ static bool CG_ParseBuildableSoundFile( const char *filename, buildable_t builda
 	return true;
 }
 
-static bool CG_RegisterBuildableAnimation( buildableInfo_t *ci, const char *modelName, int anim, const char *animName,
+static bool CG_RegisterBuildableAnimation( buildableInfo_t *ci, const char *modelName, const char* foundSuffix, int anim, const char *animName,
```
slipher, line 529: How about a single argument with the contents of `fileBase` from the other function, instead of modelName+foundSuffix?

src/cgame/cg_main.cpp
```diff
@@ -182,6 +182,8 @@ Cvar::Cvar<std::string> cg_sayCommand("cg_sayCommand", "instead of talking, chat
 // TODO: only works for player models. Buildings and weapons are also relevant
 Cvar::Cvar<bool> cg_lazyLoadModels("cg_lazyLoadModels", "load models only when needed", Cvar::CHEAT, false);
 
+Cvar::Cvar<bool> cg_lowModels("cg_lowModels", "load low quality models when available", Cvar::NONE, false);
```
slipher, line 185: `cg_lowQualityModels`?

illwieckz: Done.

src/cgame/cg_buildable.cpp
illwieckz: Done.

I also changed the md5 animation file name for buildables in the process, so `fileBase` is now the same for both iqm and md5 models. Because of the file name difference there was a mistake in my initial code.

Before it was `models/spiker/attack.md5anim`, now  it's `models/spiker/spiker_attack.md5anim`.

For information this is what already does weapon models (player models still do the same way as buildable before), so this is a step forward path name unification and with a more robust naming scheme as you can copy/move the md5 model and the anim files without taking risk of overwriting files from another model.

We don't have md5 buildable models (all have been converted to IQM in the past) so this path name change doesn't require to modify assets.

illwieckz:
> How are you generating these models?

I did it the ugly way using [IqeBrowser](http://www.reinhard-robert.de/IqeBrowser), by deleting joints.

It only works if the surfaces the joint was modifying are always as a correct place once the animation for them are deleted. For example it works with buildable feets, once the joints are deleted, they just don't react when I fire at the buildable.

This doesn't work with things like human fingers: after deleting the finger joints moving the hand would keep the fingers at the base position, introducing huge distortions.

This quick&dirty method is good enough for most alien buildables, which was enough to prove the mechanism of loading low quality models works. It also proves we can increase performances in the process.

My modified egg has a minor glitch that can be seen very close, but using lowes preset produces more ugly things thant that minor glitch anyway for a very short frametime at death time, and may not be seen at low resolution people may use if they need lowest presets anyway. Also one may notice that a buildable body may react without the head reacting or things like that, but this is not something very obvious and the lack of esthetism is very minor, and as said, far less ugly than other options enabled with `lowest` presets. I consider that as long as it is not obviously wrong, it's OK for the lowest preset.

One may produce better “low quality models” using a full featured tool like Blender, for example making sure that deleting an eye or fingers joints keep the eye and the fingers moving with the head and hand. But what I did is already a good example of thing that works.

That's because I don't know how to do better and I don't have a quick&dirty solution for player models I currently only implemented this for buildables.

illwieckz:
Here is an example of a very old and very low-end (even for the time) ATi R300 X550 alongside an old AMD64 with DDR2 memory, the scene is plat23, `setviewpos 1900 1920 0 0 0`, which has all the default alien base in sight.

Before, 17fps:

unvanquished_2024-05-29_234942_000

After, 22fps:

unvanquished_2024-05-29_235045_000

illwieckz:
When enabling `r_smp`, this same config jumps from 19fps:

unvanquished_2024-05-30_003247_000

to 23 fps:

unvanquished_2024-05-30_003419_000

The reason why the difference is smaller with `r_smp` is that by providing low quality models it makes possible to move some model processing from CPU to GPU on such card, and then the CPU boost brought by `r_smp` is smaller since the CPU computes less things to begin with, but we can still see the performance boost in all cases:

low model|smp|fps
-|-|-
❌️ off|❌️ off|17
❌️ off|✅️ on|19
✅️ on|❌️ off|22
✅️ on|✅️ on|23 🎉️

DolceTriade:
I personally dislike the manual creation of assets. The added complexity doesn't seem worth the gain?

illwieckz:
> I personally dislike the manual creation of assets. The added complexity doesn't seem worth the gain?

I share the same opinion and we should save on manual actions. But here the feature is totally optional: if the feature is enabled and a low quality model exists, that model is loaded. It means no one is asked to produce the low quality models and the game runs without.

Once the feature is available, one is free to do nothing, or to spend 2 minutes by loading a buildable model in IqeBrowser, deleting the foot joints and saving the file under a new name, or to spend a whole week doing a low-poly alternative in Blender…

Given there exists options with such low cost operation (at least for buildables). I find it good to provide the feature. I have a local branch with almost all alien buildables providing a model with less bones in a way they become processable on low-end GPUs without breaking the render. The human buildables are already low on bones.

I want to be able to spend time on such model optimization if it works, and experience shows it works. Such optimization being a 2-minute task or a much longer operation, it's up to to any contributor to decide if it's worth the effort. For the models I modified, I consider it was worth the effort.

src/cgame/cg_buildable.cpp
```diff
@@ -527,24 +527,23 @@ static bool CG_ParseBuildableSoundFile( const char *filename, buildable_t builda
 	return true;
 }
 
-static bool CG_RegisterBuildableAnimation( buildableInfo_t *ci, const char *modelName, int anim, const char *animName,
+static bool CG_RegisterBuildableAnimation( buildableInfo_t *ci, std::string fileBase, int anim, const char *animName,
```
DolceTriade, line 530: nit: Str::StringRef

DolceTriade (approved):

slipher:
LGTM

illwieckz:
Something that is good too alongside possible better framerate is how much the CPU usage drops on such machine once low quality models are used.

Before, 60% CPU on both cores:

20240531-200909-000 unvanquished-htop

After, 30% CPU on both cores:

20240531-200955-000 unvanquished-htop

PR #2993, opened 2024-05-11
rocket: avoid doing 2 or 3 Cvar::Get() trap calls per OnUpdate() call per cvar element

illwieckz:
Avoid doing 2 or 3 `Cvar::Get()` trap calls per `OnUpdate()` call per `cvar` element per frame.

🤦‍♀️️🤦‍♀️️🤦‍♀️️

### DIFF
--- a/src/cgame/rocket/rocketConditionalElement.h
+++ b/src/cgame/rocket/rocketConditionalElement.h
@@ -90,7 +90,22 @@ public:
 
 	virtual void OnUpdate()
 	{
-		if ( dirty_value || ( !cvar.empty() && cvar_value != Cvar::GetValue( cvar.c_str() ).c_str() ) )
+		Rml::String value;
+		bool modified = false;
+
+		if ( dirty_value )
+		{
+			value = Cvar::GetValue( cvar );
+			modified = true;
+		}
+		else if ( !cvar.empty() )
+		{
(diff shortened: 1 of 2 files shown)


src/cgame/rocket/rocketConditionalElement.h
```diff
@@ -90,7 +90,26 @@ class RocketConditionalElement : public Rml::Element
 
 	virtual void OnUpdate()
 	{
-		if ( dirty_value || ( !cvar.empty() && cvar_value != Cvar::GetValue( cvar.c_str() ).c_str() ) )
+		Rml::String value;
+		bool modified = false;
+
+		if ( dirty_value )
+		{
+			value = Cvar::GetValue( cvar ).c_str();
```
DolceTriade, line 98: c_str() unnecessary now. Is a remnant of libRocket

DolceTriade (approved):

src/cgame/rocket/rocketConditionalElement.h
illwieckz: Removed.

DolceTriade (approved):

slipher:
There were only extra calls it the cvar value or attributes changed, right? It wouldn't make much a performance difference in the average case.

PR #2997, opened 2024-05-12
cgame: declare ui_winner locally, it's read by the HUD on every frame

illwieckz:
Declare `ui_winner` locally, it's read by the HUD on every frame.

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -174,6 +174,7 @@ static Cvar::Cvar<float> cg_fov_level4("cg_fov_level4", "field of view (degrees)
 static Cvar::Cvar<float> cg_fov_human("cg_fov_human", "field of view (degrees) for humans", Cvar::NONE, 0);
 
 Cvar::Cvar<bool> ui_chatPromptColors("ui_chatPromptColors", "chat prompts (e.g. 'Say:') are color-coded", Cvar::NONE, true);
+Cvar::Cvar<std::string> ui_winner("ui_winner", "FOR INTERNAL USE", Cvar::NONE, "");
 Cvar::Cvar<std::string> cg_sayCommand("cg_sayCommand", "instead of talking, chat field does this command?", Cvar::NONE, "");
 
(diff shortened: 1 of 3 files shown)


src/cgame/cg_main.cpp
```diff
@@ -174,6 +174,7 @@ static Cvar::Cvar<float> cg_fov_level4("cg_fov_level4", "field of view (degrees)
 static Cvar::Cvar<float> cg_fov_human("cg_fov_human", "field of view (degrees) for humans", Cvar::NONE, 0);
 
 Cvar::Cvar<bool> ui_chatPromptColors("ui_chatPromptColors", "chat prompts (e.g. 'Say:') are color-coded", Cvar::NONE, true);
+Cvar::Cvar<std::string> ui_winner("ui_winner", "FOR INTERNAL USE", Cvar::NONE, "");
```
DolceTriade, line 177: Cvar::ROM and give proper description.

illwieckz: We already use `FOR INTERNAL USE` with `cg_minimapActive`, the fact those are cvars is basically a HACK and we better implement a specific features for variables internal to the cgame and the ui.

illwieckz: If this cvar is `Cvar::ROM` then `ui_winner` can never be updated and then when the game ends the cvar remains empty and the scoreboard doesn't display the winning team name.

DolceTriade: Are you sure? Cvar::ROM just means that users cannot modify it. The value can be changed with `Cvar::SetValue()`

slipher: ROM, like LATCH, is not yet implemented for the gamelogic.

slipher: illwieckz is right that this should not be a cvar anyway. We could make something like a data source but for singletons

slipher:
LGTM

DolceTriade (approved):

PR #2998, opened 2024-05-12
tools: update translation-related scripts

illwieckz:
- disable the generation of `commands.pot`,
  we dont translate this for now and having to delete the file by hand on every rule of the script is annoying.
- add `update_po.sh` script,
  this automatically updates the existing `.po` files using the `.pot` files.

It looks like I never found the time to push `update_po.sh` before.

### DIFF
--- a/tools/translation/generate_pot.sh
+++ b/tools/translation/generate_pot.sh
@@ -2,7 +2,7 @@
 
 # ===========================================================================
 #
-# Copyright (c) 2017-2020 Unvanquished Developers
+# Copyright (c) 2017-2024 Unvanquished Developers
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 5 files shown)


DolceTriade (approved):

tools/translation/update_po.sh
```diff
@@ -0,0 +1,51 @@
+#! /usr/bin/env bash
+
```
slipher, line 2: What is this program for?

tools/translation/generate_pot.sh
```diff
@@ -24,49 +24,64 @@
...
 
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 repo_dir="$(realpath "${script_dir}/../..")"
 dpk_dir="${repo_dir}/pkg/unvanquished_src.dpkdir"
 pot_dir="${dpk_dir}/translation"
-main_pot_file="${pot_dir}/game.pot"
-commands_pot_file="${pot_dir}/commands.pot"
 
-temp_pot_file="$(mktemp)"
+. "${script_dir}/translation.conf"
```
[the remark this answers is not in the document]
DolceTriade: sets basic config vars

tools/translation/update_po.sh
DolceTriade: `/usr/bin/env` is like `which`.  `/usr/bin/env` is a "well known path" for env and will work on distros which do not have `bash` at `/usr/bin` like NixOS (ie, doing this is more portable than `/usr/bin/bash`

tools/translation/generate_pot.sh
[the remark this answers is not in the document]
slipher: Is it the same as `source`? If so please use that instead of cryptically golfing

tools/translation/update_po.sh
slipher: lol I meant what is `update_po.sh` for

tools/translation/generate_pot.sh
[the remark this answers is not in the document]
DolceTriade: Yes, it's the same as source. Technically speaking, source is a bash-ism and `. foo` is more portable, but we do enforce bash anyways so its probably fine.

tools/translation/update_po.sh
illwieckz: Update the strings in `.po` files from the `.pot` file.

slipher: Weblate does that automatically in our current workflow right? I'm not saying we shouldn't have the script but I would like to see a comment about that if is the case.

tools/translation/generate_pot.sh
```diff
@@ -24,49 +24,64 @@
...
-	-a ! -name sg_admin.cpp \
-	-a ! -name sg_cmds.cpp \
-	-a ! -name sg_maprotation.cpp \
-| sort \
-| xgettext --from-code=UTF-8 \
-	-j -o "${temp_pot_file}" \
-	-k_ -kN_ -kP_:1,2 -k -f -
+for name in 'game' 'commands'
+do
+	if eval "\${${name}}"
```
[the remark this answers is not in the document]
illwieckz: This line turns the `game` string contained into the variable `name` into the variable `${game}` then `eval` executes it to get `true` or `false`.

tools/translation/generate_pot.sh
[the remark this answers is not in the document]
illwieckz: Yes, this is the standard syntax for including a file. I an use `source` indeed since there is no plan to make it compatible with another language than bash.

tools/translation/update_po.sh
illwieckz: I'm not sure Weblate does that automatically. Some months ago I noticed some strings were not in Weblate but in source, after having run that script and pushed, Weblate went in sync. Maybe that was a bug in Weblate, but I now always run `./generate_pot.sh` then `./update_po.sh` after merging translation changes from Weblate. In case it's not needed anymore it can't do harm anyway, I always run it to make sure everything is fine.

tools/translation/generate_pot.sh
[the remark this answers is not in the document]
slipher: TIL, `.` is OK I guess

tools/translation/generate_pot.sh
```diff
@@ -24,49 +24,64 @@
...
+	then
+		temp_pot_file="$(mktemp)"
+
+		eval "generate_${name}_pot"
```
slipher, line 79: The `eval` seems to be superfluous. Like in external_deps/build.sh we do just `"build_${pkg}"`

tools/translation/update_po.sh
slipher: It would be nice to have a comment about your experience with that.

tools/translation/generate_pot.sh
illwieckz: Done.

tools/translation/update_po.sh
illwieckz: I updated the translation tool readme and placed the comment there.

illwieckz:
I added the `generate.h` that is now the entry-point, it call `generat_pot.sh` then `update_po.sh` in a row, the readme has been updated.

PR #3003, opened 2024-05-19
votes: Display a proper error on `callvote map` without map

### LINKED ISSUE #2511: `callvote nextmap` tries to load the map ""
Game prints: `callvote: 'maps/.bsp' could not be found on the server ` which is obviously wrong: no map was given after all.

DolceTriade:
If we run `callvote map` without a map name, we should error out with a better error than: `callvote: 'maps/.bsp' could not be found on the server`.

Fixes #2511


slipher (commented):
LGTM

cu-kai (approved):

PR #3006, opened 2024-05-22
Fix crash upon restarting map with kt/ktl/etc. option

slipher:


DolceTriade (approved):

DolceTriade:
I think you messed up your push

slipher:
I found more bot session stuff to clean up, so I moved that commit to #3007.



> I think you messed up your push

Thanks to the CodeQL configuration, pushing any branch now counts as "creating a workflow" according to Github's permission system and was not allowed by the token I was using. I tried to rebase to remove the stuff, but then some commits were somehow not detected as already included in master.

PR #3008, opened 2024-05-24
Fix /devteam angles; clean up spawning code

slipher:


illwieckz (approved):
LGTM and it works for me.

PR #3010, opened 2024-05-27
Add option for the current screen size (via r_mode -2) to resolution list

### LINKED ISSUE #1892: It's impossible to switch back to native screen resolution once you tested another screen size
If your screen has a resolution that isn't in the list, you can't select it, and you thus can't have the correct resolution for your screen.

Example test procedure:
* have a 1366×768 screen or anything else not in the resolution list
* launch a new profile
* see your resolution marked as "custom"
* switch to another resolution
* restart the game (maybe?)
* try to switch back to your normal screen resolution.

slipher:
Fixes #1892

### DIFF
--- a/src/cgame/cg_rocket_dataformatter.cpp
+++ b/src/cgame/cg_rocket_dataformatter.cpp
@@ -76,6 +76,12 @@ static void CG_Rocket_DFResolution( int handle, const char *data )
 		std::string displayString = Str::Format( _( "Custom: %dx%d" ), -w, -h );
 		Rocket_DataFormatterFormattedData( handle, displayString.c_str(), false);
 	}
+	else if ( w == 0 )
+	{
+		// TODO(0.55): show dimensions of screen
+		std::string displayString = _( "Same as screen" );
+		Rocket_DataFormatterFormattedData( handle, displayString.c_str(), false);
+	}
 	else
 	{
 		std::string aspectRatio = DisplayAspectString( w, h );
--- a/src/cgame/cg_rocket_dataformatter.cpp
+++ b/src/cgame/cg_rocket_dataformatter.cpp
@@ -48,7 +48,7 @@ static int GCD( int a, int b )
 	return a;
 }
 
-static const char *DisplayAspectString( int w, int h )
+static std::string DisplayAspectString( int w, int h )
 {
 	int gcd = GCD( w, h );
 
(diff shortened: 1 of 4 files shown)


src/cgame/cg_rocket_dataformatter.cpp
```diff
@@ -62,22 +62,31 @@ static const char *DisplayAspectString( int w, int h )
...
 	{
-		Rocket_DataFormatterFormattedData( handle, "Custom", false );
-		return;
+		std::string displayString = Str::Format( _( "Custom: %dx%d" ), -w, -h );
+		Rocket_DataFormatterFormattedData( handle, displayString.c_str(), false);
+	}
+	else if ( w == 0 )
+	{
+		std::string displayString = Str::Format( _( "Same as screen (%dx%d)" ),
+			cgs.glconfig.vidWidth, cgs.glconfig.vidHeight );
```
illwieckz, line 82: On my end this displays the current resolution of the game, not the native resolution of the screen.

DolceTriade (approved):

src/cgame/cg_rocket_dataformatter.cpp
DolceTriade: I don't think there is anything we can do about this? We can only get what SDL gives us.

illwieckz: By native resolution on the screen, I mean what SDL gives us.

Right now it displays the current resolution of the game, not the screen one.

If I set my screen to `3840×2160` and Unvanquished to windowed `1920×1080`, this line displays “`Same as screen (1920×1080)`”, which is wrong.

slipher: Whoops, could have sworn that was working before... I'll drop the `%dx%d`.

On the bright side, we can remove two `Cvar::GetValue`s since the info is already there.

Maybe for 0.55 we can add `desktopMode` in the glconfig_t.

illwieckz: > Maybe for 0.55 we can add `desktopMode` in the glconfig_t.

Yes.

slipher:
Removed the wrong resolution with "Same as screen". Also moved "Custom" and "Same as screen" to the top of the list instead of the bottom.

illwieckz (approved):
LGTM.

PR #3011, opened 2024-05-28
fix the "wallwalk-on-back" bug 

illwieckz:
> `2024-05-26 20:48:17 +0200 <freem>` Hi. illwieckz, perturbed, Ishq, I have a fix for the problem which makes player "wallwalk on the back".

The first patch is the said fix.

freem also said there is not only  1 but 4 `acosf()` in the same function that receives invalid values. I guess my extra patch is what is needed.


DolceTriade (approved):

slipher:
LGTM

PR #3015, opened 2024-05-29
codeql: sync submodules branches

illwieckz:
Sync submodule branches if branch name ends with `/sync`, like we do with azure pipelines and appveyor.


illwieckz:
This fixes the CI for some branches.

PR #3017, opened 2024-05-30
Include common/Common.h (PCH) first in every .cpp

slipher:
Always including the precompiled header first in every translation unit will help the build be consistent between PCH enabled and disabled.


illwieckz (approved):
LGTM.

slipher:
Merging this now to fix the CI.

PR #3018, opened 2024-05-31
rocket: silence unused parameter warning

illwieckz:
Silence unused parameter warning. That warning turned into error breaks the CI:

```
FAILED: CMakeFiles/cgame-native-dll.dir/src/cgame/rocket/rocket.cpp.o 
In file included from src/cgame/rocket/rocket.cpp:51:
src/cgame/rocket/rocketFocusManager.h:63:33: error: unused parameter 'evt' [-Werror,-Wunused-parameter]
        void ProcessEvent( Rml::Event &evt )
                                       ^
1 error generated.
```

### DIFF
--- a/src/cgame/rocket/rocketFocusManager.h
+++ b/src/cgame/rocket/rocketFocusManager.h
@@ -60,7 +60,7 @@ public:
 
 	VisibleMenusState GetState() { return state_; }
 
-	void ProcessEvent( Rml::Event &evt )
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

illwieckz:
I merge it now because it fixes the CI, which is useful to avoid false positives in other PRs. Also this doesn't change what the code does so the only thing to review is “_does it fix the CI or not?_”.

PR #3019, opened 2024-05-31
malloc and memset cleanup

slipher:
- `malloc`: check return and call Sys::Error on failure. Cut down on redundant zeroing (memsetting after calloc) a bit.
- `memset`: use it less because it has the same issue with passing nullptr being undefined, line memcpy and memmove.


DolceTriade (commented):
seems ok. I noticed that lgun and chaingun couldn't be heard, but I think that's true on master too.

DolceTriade:
Lasgun/chaingun thing is a personal issue. lgtm

PR #3020, opened 2024-06-02
locdamge: Load locational damage config based on class

DolceTriade:
This practically does not affect anything since this mainly affects aliens, which do not consider locational damage, but its still a bug and affects the human-iqm branch since now the human_male model is the same for naked, light armor, and medium armor classes.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/configs/classes/builderupg.locdamage.cfg
+++ b/pkg/unvanquished_src.dpkdir/configs/classes/builderupg.locdamage.cfg
@@ -1,0 +1,8 @@
+//whole body
+{
+  minHeight 0.0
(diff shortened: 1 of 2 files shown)


slipher:
Having to name the config file after a skin name instead of the class name doesn't seem good.

DolceTriade:
> Having to name the config file after a skin name instead of the class name doesn't seem good.

Quite frankly, I considered just naming the locdamage stuff after the class name instead, but didn't think that larmor and marmour were classes, but it seems they are.

Doing this is way more sensible.

slipher:
LGTM

The multiple build failures appear to stem from basing on an old branch.

PR #3021, opened 2024-06-02
Fix misusing GLM vector like float[3]

### LINKED ISSUE #3013: Compiler-unsafe usage of GLM vectors
Recent issues regarding diabolical UB-exploiting compiler optimizations prompt me to bring this up. GLM vectors are implemented somewhat like this:
```
class vec3 {
	float x, y, z;

	float operator[](size_t i)
	{
		switch(i)
		{
		default:
			case 0:
		return x;
			case 1:
		return y;
			case 2:
		return z;
		}
	}
};
```

For compatibility with non-GLM parts of the code which expect an array of 3 floats, we often do as follows:
```
glm::vec3 vec;
some_function( &vec[ 0 ] );
```

This is undefined behavior because we are really just using the address of `x`, which is not part of an array. Compilers could potentially exploit this and break our code. Currently, compilers don't seem to do anything with this, but who knows what the next version may bring?

slipher:
Fixes #3013.

Also rework the coordinate conversion code used to interoperate with the Recast coordinate system. I found one coordinate conversion bug.

### DIFF
--- a/src/cgame/CombatFeedback.cpp
+++ b/src/cgame/CombatFeedback.cpp
@@ -263,7 +263,7 @@ static bool EvaluateDamageIndicator(DamageIndicator *di, bool *draw)
 	di->velocity += glm::vec3(0, 0, -400) * dt;
 	di->origin += di->velocity * dt;
 
-	if ( !CG_WorldToScreen( &di->origin[0], tmp, tmp + 1 ) )
+	if ( !CG_WorldToScreen( GLM4READ( di->origin ), tmp, tmp + 1 ) )
 	{
 		*draw = false;
 		return true;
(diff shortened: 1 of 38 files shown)


illwieckz:
One thing I'm concerned about is to see the GLM migration being finished, and I'm afraid that some commits to fix the `float[3]` would just break the merge of other commits. bmorel said his work was almost finished, I don't know which branch is the most up-to-date on his side, but I found this:

- https://codeberg.org/bmorel/Unvanquished/commits/branch/glm

I would like to see some effort to merge that so we can move to something else.

slipher:
I've been checking out the most recent version of freem's branch, and there are still many cases where the adapting to `float*` is needed. Meanwhile, I have developed a merging workflow where such minor conflicts will not be an impediment: #3026. I think using this workflow, it will actually help me to have the current PR merged.

DolceTriade (approved):

DolceTriade (approved):

PR #3022, opened 2024-06-03
Fix /bot behavior command bugs; don't use SVF_BOT

slipher:


cu-kai:
Shouldn't the bug fix and `SVF_BOT` refactoring be two different PRs?

slipher:
> Shouldn't the bug fix and `SVF_BOT` refactoring be two different PRs?

Well, in general we don't expect every commit, even notionally independent ones, to be in a separate PR, because the overhead of the Github workflow is too high. It would be nice if we could have each commit be independently reviewable, but we would need better collaboration software.

I uploaded these three commits together because I discovered the bug and the unused functions while doing the SVF_BOT refactoring, and splitting them up independently would cause merge conflicts between them.

If you are interested in reviewing only some of the commits, you may LGTM individual commits and I will merge those commits manually.

cu-kai:
Alright, the `/bot` command fixes LGTM. I can't speak for the refactoring though. Seems like more work than needed and could cause a merge conflict for other contributors.

src/sgame/sg_cmds.cpp
```diff
@@ -3166,15 +3167,15 @@ static void Cmd_Tactic_f( gentity_t * ent )
 		if ( Entities::IsAlive( ent ) )
 		{
 			glm::vec3 position = VEC2GLM( ent->s.origin );
-			g_entities[ id ].botMind->userSpecifiedPosition = position;
+			ent->botMind->userSpecifiedPosition = position;
```
necessarily-equal, line 3170: Here I think this should be `target` instead of `ent`: if I'm not mistaken `ent` represents the user launching the typing command. So they can't be a bot.

src/sgame/sg_cmds.cpp
```diff
@@ -3166,15 +3167,15 @@ static void Cmd_Tactic_f( gentity_t * ent )
...
 		}
 		else
 		{
-			g_entities[ id ].botMind->userSpecifiedPosition = Util::nullopt;
+			ent->botMind->userSpecifiedPosition = Util::nullopt;
```
necessarily-equal, line 3174: ditto

src/sgame/sg_cmds.cpp
```diff
@@ -3166,15 +3167,15 @@ static void Cmd_Tactic_f( gentity_t * ent )
...
 		}
 		// use the commanding player's client number as the bot's user specified client number
 		// not all bot actions use this number, but it cannot hurt to set it
-		g_entities[ id ].botMind->userSpecifiedClientNum = ent->s.number;
+		ent->botMind->userSpecifiedClientNum = ent->s.number;
```
necessarily-equal, line 3178: ditto

src/sgame/sg_session.cpp
```diff
@@ -133,7 +133,7 @@ void G_WriteSessionData()
 	for ( i = 0; i < level.maxclients; i++ )
 	{
 		if ( level.clients[ i ].pers.connected == CON_CONNECTED &&
-		     !( g_entities[ i ].r.svFlags & SVF_BOT ) )
+		     !( level.clients[ i ].pers.isBot ) )
```
necessarily-equal, line 136: nit: could remove parenthesis

necessarily-equal (commented):
I didn't test. Some comments, otherwise looks good to me. Not sure if somebody has a branch that explicitely conflicts with the `SVF_BOT` change.

src/sgame/sg_cmds.cpp
slipher: Nice catch. I tested /tactic and made sure it works now.

slipher:
> Not sure if somebody has a branch that explicitely conflicts with the `SVF_BOT` change.

I added back the `isBot |= SVF_FLAGS` lines so that branches that rely on that will continue to work for now. When it is completely removed later, the SVF_BOT definition will be removed so anyone still using it will get a compile error.

slipher:
Reviewed by freem on IRC. freem approves of the idea and says, I also made a commit to get rid of SVF_BOT but used `botMind != nullptr` instead of adding the `isBot` bool, to avoid adding another way to do things. But either way is OK.

PR #3028, opened 2024-06-09
Replace ProjectPointOntoLineOutwards

slipher:
People were complaining about that function because it divides by 0. Replace with an alternative that avoids that, and uses GLM. ProjectPointOntoLineOutwards can be nuked from the engine after.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/beacons.cfg
+++ b/pkg/unvanquished_src.dpkdir/ui/beacons.cfg
@@ -12,10 +12,11 @@ fadeInScale       1.4
 
 hudSize           1
 hudMinSize        0.025
-hudMaxSize        1
+hudMaxSize        0.7
 hudAlpha          0.35
 hudAlphaImportant 0.7
-hudMargin         0.14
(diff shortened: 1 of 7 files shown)


pkg/unvanquished_src.dpkdir/ui/beacons.cfg
```diff
@@ -12,10 +12,10 @@ fadeInScale       1.4
 
 hudSize           1
 hudMinSize        0.025
-hudMaxSize        1
+hudMaxSize        0.7
 hudAlpha          0.35
 hudAlphaImportant 0.7
-hudMargin         0.14
+hudMargin         0.14 // must obey hudMaxSize + 2 * hudMargin < hudSize
```
DolceTriade, line 18: Why

slipher: Fixed the equation which should have said 1 instead of hudSize. The calculation for positioning the beacon does invalid float operations if the beacon does not fit inside the boundaries for drawing.

DolceTriade (approved):

PR #3031, opened 2024-06-10
lua: Move cgame Cmd, Timer, Cvar to shared so sgame can use too

DolceTriade:


src/shared/lua/Cmd.cpp
```diff
@@ -32,21 +32,32 @@ Maryland 20850 USA.
 ===========================================================================
 */
 
-#include "common/Common.h"
-#include "../../cg_local.h"
 #include "register_lua_extensions.h"
 
-static int Cmd_exec(lua_State* L)
+// Manually declare this because we have to include a different file for the SGame and CGame.
+void trap_SendConsoleCommand(const char *text);
```
slipher, line 38: `shared/CommonProxies.h` works for both

DolceTriade: Done.

slipher:
LGTM

PR #3032, opened 2024-06-11
Fix wrong assertion in G_WriteSessionData

slipher:


DolceTriade:
Why is this assertion wrong? Is there a bug here? Did this cause a crash anywhere?

slipher:
I recently added this assertion but it is not always true (there are 2 different code paths for calling G_WriteSessionData and only one kicks bots first). The code works in release mode, but the assertion may be false. Probably I didn't notice it because I was using RelWithDebInfo mode.

DolceTriade (approved):
Please update commit message

PR #3033, opened 2024-06-15
lua: Use traditional Lua iterators

DolceTriade:
LibRocket overrode the builtin Lua iterator behavior with its custom version that would push an index. This did not play with with third party libraries, so RmlUI removed this behavior and used the traditional pairs() behavior.

Tradtionally, the way pairs works is it looks for a __pairs() metamethod that returns a next() function, a metatable, and a nil, so just create a C++ helper that does that. This was adapted from the RmlUI Pairs.h file.


slipher:
I successfully tested it. LGTM.

PR #3034, opened 2024-06-16
Remove some DirToByte()/ByteToDir() usages

VReaperV:
Remove some of the weird crap that is trying to convert 3 floats into one int that is used to access a table with 162 values.

Unfortunately the other usages require changing a bunch of other functions that only take an int.


slipher:
DirToByte is used as a form of compression to decrease network bandwidth usage.

Are these cases ones where the approximation is not good enough and produces some kind of artifact?

VReaperV:
I believe it affects hit/miss with relation to global orientation of a projectile/trace/melee.

For this pr in particular though, network is not affected.

slipher:
Oh, when I saw code about creating an event I assumed I was reading the sgame...

So I agree what the cgame is doing here is stupid, but in its current form the patch doesn't work. Testing it, I find that bullet marks from rifle and mass driver don't show up. There are other callers to that function besides the shotgun one.

Issue #3002, opened 2024-05-16
Bots: missing check for "upward" navcons?


slipher:
The `BotMaybeAttackWhileClimbing` function does an assertion that the last taken navcon should be "upward" (endpoint z coordinate >= start point z coordinate). However the function may be entered without any check that a navcon is active. BotMaybeAttackWhileClimbing is called only by BotClimbToGoal, which is called only by BotMoveUpward. If the bot is a Mantis, BotMoveUpward checks that a navcon is active before calling BotClimbToGoal. But if it is a Dretch, no such check is performed.

sweet235:
I think I forgot to delete this assertion. In an earlier version, there was no wall climb activation. The mantis would try to leap to the higher place.

I guess we can simply delete it.

Issue #3037, opened 2024-06-19
Unvanquished crash with bots


DolceTriade:
# What I did

1. Start `voxelmini` in devmap
2. /bot fill 16 aliens

it crashed

```
Thread 42 "daemon" received signal SIGTRAP, Trace/breakpoint trap.
[Switching to Thread 0x7fffb4c006c0 (LWP 801592)]
0x00007fffcb911234 in BotMaybeAttackWhileClimbing (self=0x7fffc9fa0bd0) at /mnt/media/code/unv-master/src/sgame/sg_bot_nav.cpp:875
875		ASSERT( self->botMind->lastNavconDistance > 0 );
(gdb) bt
#0  0x00007fffcb911234 in BotMaybeAttackWhileClimbing (self=0x7fffc9fa0bd0) at /mnt/media/code/unv-master/src/sgame/sg_bot_nav.cpp:875
#1  0x00007fffcb9103a7 in BotClimbToGoal (self=0x7fffc9fa0bd0) at /mnt/media/code/unv-master/src/sgame/sg_bot_nav.cpp:925
#2  0x00007fffcb9100d7 in BotMoveUpward (self=0x7fffc9fa0bd0, nextCorner=...) at /mnt/media/code/unv-master/src/sgame/sg_bot_nav.cpp:937
#3  0x00007fffcb910f30 in BotTryMoveUpward (self=0x7fffc9fa0bd0) at /mnt/media/code/unv-master/src/sgame/sg_bot_nav.cpp:1040
#4  0x00007fffcb910563 in BotMoveToGoal (self=0x7fffc9fa0bd0) at /mnt/media/code/unv-master/src/sgame/sg_bot_nav.cpp:1097
#5  0x00007fffcb8feeb7 in BotActionRoamInRadius (self=0x7fffc9fa0bd0, node=0x7fff8d0f2740) at /mnt/media/code/unv-master/src/sgame/sg_bot_ai.cpp:1126
#6  0x00007fffcb8fc739 in BotEvaluateNode (self=0x7fffc9fa0bd0, node=0x7fff8d0f2740) at /mnt/media/code/unv-master/src/sgame/sg_bot_ai.cpp:602
#7  0x00007fffcb8fc698 in BotSelectorNode (self=0x7fffc9fa0bd0, node=0x7fff8d0fc370) at /mnt/media/code/unv-master/src/sgame/sg_bot_ai.cpp:276
[...]

PR #3038, opened 2024-06-19
remove an assertion

sweet235:
The BotMaybeAttackWhileClimbing function does an assertion that the last taken navcon should be "upward" (endpoint z coordinate >= start point z coordinate). However the function may be entered without any check that a navcon is active. BotMaybeAttackWhileClimbing is called only by BotClimbToGoal, which is called only by BotMoveUpward. If the bot is a Mantis, BotMoveUpward checks that a navcon is active before calling BotClimbToGoal. But if it is a Dretch, no such check is performed.

Fix https://github.com/Unvanquished/Unvanquished/issues/3002 and https://github.com/Unvanquished/Unvanquished/issues/3037

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -872,8 +872,6 @@ static int WallclimbStopTime( gentity_t *self )
 // check if a wall climbing bot is running into an attackable human entity, if so: attack it
 static void BotMaybeAttackWhileClimbing( gentity_t *self )
 {
-	ASSERT( self->botMind->lastNavconDistance > 0 );
-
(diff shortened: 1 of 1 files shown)


DolceTriade:
lgtm

DolceTriade:
It would be nice to have a better commit message to explain why its wrong. I think you can include slipher's rationale from the bug.

sweet235:
One day has passed. The checks are still not complete. I have been stripped of my permission to press the merge button in this case.

I give up.

Gireen (approved):
i think this can be merged.
@sweet235 do you want to do it yourself 

if no one says something i will merge it in a few days 🙈

PR #3039, opened 2024-06-19
devmap: if devmap, enable instant building

DolceTriade:


slipher:
For me it would be annoying that it steps on the cvar. Particularly if I set cheats and instant building to opposite values, did `map_restart`, and then instant building got flipped.

I like the idea that turning on cheat mode just enables additional cheat commands, which only affect the game if you use them. While developer settings that affect behavior without explicit cheat usage, such as `g_instantBuilding` and `g_neverEnd`, are controlled separately.

DolceTriade:
I dislike the fact that instant building is not enabled automatically via devmap. This was the behavior in tremulous. Maybe I can check for the archive flag?

illwieckz:
Doing this is a regression if I'm right. A previous behavior was to always enable instant building with devmap, without cvar. It meant loading a map with devmap would not run the standard gameplay, which is not good for debugging.

illwieckz:
The `daemon` submodule looks to have been committed by mistake, you can do `git config diff.ignoreSubmodules all` to avoid this, see: https://wiki.unvanquished.net/wiki/Dealing_with_submodules

DolceTriade:
> Doing this is a regression if I'm right

If you're doing development, you are generally testing stuff, and rarely are you actually testing the build timer. It's annoying to have to manually enable the cvar to enable instant building.

PR #3041, opened 2024-06-20
lua: Add ways to automatically run lua on startup

DolceTriade:
- g_luaScript to run at startup
- luaScript spawn flag


src/sgame/sg_main.cpp
```diff
@@ -651,6 +655,14 @@ void G_InitGame( int levelTime, int randomSeed, bool inClient )
 
 	// Initialize Lua
 	Lua::Initialize();
+
+	if ( !g_luaScript.Get().empty() )
```
slipher, line 659: To make this more useful and less redundant with simply putting a Lua command in a map config, how about delaying the command until the first non-fake frame? There's the server startup code

```
// run a few frames to allow everything to settle
	for ( i = 0; i < GAME_INIT_FRAMES; i++ )
	{
		gvm.GameRunFrame( sv.time );
		svs.time += FRAMETIME;
		sv.time += FRAMETIME;
	}
```

During the first frames some entities are still popping into existence and the floating buildings from Radiant-created layouts have to fall to the ground. It could be nice for the script to wait so that entities will all exist and be in the right place.

src/sgame/sg_spawn.cpp
```diff
@@ -1105,6 +1106,16 @@ static void SP_worldspawn()
 	G_SpawnStringIntoCVar( "BPBudgetPerMiner", g_buildPointBudgetPerMiner );
 	G_SpawnStringIntoCVar( "BPRecoveryRateHalfLife", g_buildPointRecoveryRateHalfLife );
 
+	G_SpawnString( "luaScript", "", &s );
```
slipher, line 1109: I think it's premature to add support for Lua embedded in BSPs... people seem to have an expectation that backward compatibility will be maintained forever for every map. It would suck if we immediately get locked in to any API design flaws

src/sgame/sg_main.cpp
DolceTriade: I have been using the timer to run code on the next frame.

```
Timer.add(1, function() print('next frame') end)
```

slipher: I didn't mean anything about rerunning the script every frame. I'm saying it may be better not to run the script in the very first frame because the level isn't truly done loading at that point. For example buildables may not exist yet. So it might mess up a script that does something with buildables.

PR #3042, opened 2024-06-20
lua: Add the concept of an "id" for an entity

DolceTriade:
Recently developments have lead to people wanting to refer to entities for various actions (lock/unlock, Lua actions, etc). This is a mechanism for doing so. Right now, we can set it via Lua or via map builtins.

### DIFF
--- a/src/sgame/sg_entities.cpp
+++ b/src/sgame/sg_entities.cpp
@@ -190,6 +190,11 @@ void G_FreeEntity( gentity_t *entity )
 
 	unsigned generation = entity->generation;
 
+	if ( entity->id )
+	{
+		BG_Free( entity->id );
+	}
+
 	entity->~gentity_t();
 	new(entity) gentity_t{};
 
--- a/src/sgame/sg_spawn.cpp
+++ b/src/sgame/sg_spawn.cpp
@@ -200,6 +200,7 @@ static const fieldDescriptor_t fields[] =
 	{ "dmg",                 FOFS( mapEntity.config.damage ),        F_INT,        ENT_V_UNCLEAR, nullptr },
 	{ "gravity",             FOFS( mapEntity.config.amount ),        F_INT,        ENT_V_UNCLEAR, "amount" },
 	{ "health",              FOFS( mapEntity.config.health ),        F_INT,        ENT_V_UNCLEAR, nullptr },
+	{ "id",                  FOFS( id ),                             F_STRING,     ENT_V_UNCLEAR, nullptr },
 	{ "message",             FOFS( mapEntity.message ),              F_STRING,     ENT_V_UNCLEAR, nullptr },
 	{ "model",               FOFS( mapEntity.model ),                F_STRING,     ENT_V_UNCLEAR, nullptr },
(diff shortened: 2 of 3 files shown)


slipher:
The PR basically only adds an unused variable so I would say it is erring on the side of too small PRs :laughing: 

There is already a way to give an entity a name in the BSP. `name` / `targetname`

DolceTriade:
I was thinking this would be useful for Lua, and not just the BSP. name/targetname are only for mapEntities. I want to give an arbitrary id to any arbitrary entity.

sweet235:
This fits into a web-browser-like model. `targetname` is already required to be non-unique by the game. I do not know what `name` is, but it does not seem to be unique either. The idea here is: you do something like `getEntityById(id)`, and get either 0 or 1 entities. Compare to DOM `getElementsByName()`, `getElementById()`.

Maybe we should enforce `id` to be unique, but that can be done later.

Gireen:
the g_entities index is already treated, and sometimes mentioned as entity id. that could be a bit confusing 

i think this PR is more like a class or group if its not unique anyway.

DolceTriade:
> the g_entities index is already treated, and sometimes mentioned as entity id. that could be a bit confusing

That is the entitynum everywhere I have seen it used.

I think Sweet's anology of the web DOM was my inspiration here as well. In the DOM, the ID isn't enforced to be unique, but its expected to be.

slipher:
I don't get the point of having this for non-BSP entities. Usually in programming, if we want to have a handle to some object we store a pointer/reference/etc. to it, rather than writing a name inside the object itself. An entity can be "named" in a Lua script by assigning to a variable or storing it in a table.

cu-kai:
> The PR basically only adds an unused variable so I would say it is erring on the side of too small PRs 😆
> 
> There is already a way to give an entity a name in the BSP. `name` / `targetname`

`name` and `targetname` are used for linking entities together (e.g. buttons, triggers or various logic entities). This should not be solely relied on for lua interaction because some entities have other functionalities disabled when they are assigned a `name` or `targetname`. For example, doors will not automatically open but require firing after that point.

Any `id` field added as part of this proposal should be cosmetic only and not interfere with the functionality of the entity.

One such use from within lua could be: to find an entity's index within g_entities by its `id` value (could also allow searching by `name`/`targetname` too) for this to be passed to an `entityFire` or `entityLock` command. Remember that entity indexes are not consistent and this is a problem we have run into in the past when using such commands from mapconfigs.

cu-kai (approved):
I don't really know why this needs so much discussion. This is a cosmetic change only to be used by the lua system.

Let's merge it already.

Gireen:
ironically entityshow calls it entityid 😄 
with the index everywhere called entitynum it sounds ok to have a DOM like id for entitys

Gireen (approved):

slipher:
> `name` and `targetname` are used for linking entities together (e.g. buttons, triggers or various logic entities). This should not be solely relied on for lua interaction because some entities have other functionalities disabled when they are assigned a `name` or `targetname`. For example, doors will not automatically open but require firing after that point.

Thanks, now it makes sense.

PR #3044, opened 2024-06-21
Cleanup: call entityNum entityNum consistently in logs instead of entityId

Gireen:
only renamed in logs that admins wont get confused
not tested

must be squashed, no idea if its possible to edit multiple files in one commit with github 🤷


VReaperV:
> must be squashed, no idea if its possible to edit multiple files in one commit with github 🤷

GitHub web interface is evil.

DolceTriade (approved):
Luckily you can squash through webui

PR #3045, opened 2024-06-21
admin: Don't print admin text twice for local clients

DolceTriade:
Also fix a bug where clients with a "loopback" IP are not registered as local clients.

### DIFF
--- a/src/sgame/sg_admin.cpp
+++ b/src/sgame/sg_admin.cpp
@@ -608,23 +608,30 @@ void G_admin_action( const char *action, const char *translation,
 
 	for ( int i = 0; i < level.maxclients; i++ )
 	{
+		gentity_t *ent = &g_entities[ i ];
+
+		// If the client is local, don't send it to them again, because they see console output.
+		if ( ( level.inClient && i == 0 ) || !ent->inuse )
+		{
+			continue;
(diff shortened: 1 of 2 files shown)


cu-kai (approved):
Sensible change. LGTM.

slipher:
This "localClient" thing is weird. Is it supposed to be possible to have "local clients" that aren't running in the same process as the server? ("loopback" refers that type of client.) If so, that's not what we want here, as only the in-process client can read server console messages.

An alternative way to detect the in-process client is `level.inClient && [client num] == 0`.

DolceTriade:
Yeah I had similar concerns, however I did test with separate clients (ie, connected locally to my daemonded process) and I saw single lines outputted whether I locked or the console locked.

I'll try the inClient thing though. I suppose it's never possible for an `inClient` to be anything other than entity 0?

DolceTriade:
@slipher ya, that works too. Changed it to that since that seems safer. I still kept the loopback change as it does sound like a bug.

slipher:
If I connect to a separate local daemonded using `/connect localhost` then I get 127.0.0.1 as the client's address. So I guess that localClient thing was always false before.

LGTM

PR #3046, opened 2024-06-22
teleport: Fix players walking on ice after teleport.

DolceTriade:
We need to reset the pm_flags after teleporting.

### DIFF
--- a/src/sgame/sg_utils.cpp
+++ b/src/sgame/sg_utils.cpp
@@ -281,6 +281,8 @@ void G_TeleportPlayer( gentity_t *player, glm::vec3 const& origin, glm::vec3 con
 		player->client->ps.pm_time = 160;
 	if ( player->client->ps.pm_time != 0 )
 		player->client->ps.pm_flags |= PMF_TIME_KNOCKBACK;
+	else
+		player->client->ps.pm_flags &= ~PMF_ALL_TIMES;
 
 	// toggle the teleport bit so the client knows to not lerp
 	player->client->ps.eFlags ^= EF_TELEPORT_BIT;


slipher:
How can the bug be produced?

DolceTriade:
I did it by calling this function via Lua right when the player spawns.

```
diff --git a/src/sgame/sg_client.cpp b/src/sgame/sg_client.cpp
index f6a31ac0a..deb76f861 100644
--- a/src/sgame/sg_client.cpp
+++ b/src/sgame/sg_client.cpp
@@ -1784,6 +1784,8 @@ void ClientSpawn( gentity_t *ent, gentity_t *spawn, const vec3_t origin, const v
        // (re)tag the client for its team
        Beacon::DeleteTags( ent );
        Beacon::Tag( ent, (team_t)ent->client->ps.persistant[ PERS_TEAM ], true );
+
+       G_TeleportPlayer( ent, {21, 1843, 6}, {0,0,0}, 0.0f);
 }
 ```

This patch repro's effectively what I was doing in Lua.

File: src/sgame/sg_utils.cpp [hunk 1]
```diff
@@ -296,7 +298,7 @@ void G_TeleportPlayer( gentity_t *player, glm::vec3 const& origin, glm::vec3 con
 	BG_PlayerStateToEntityState( &player->client->ps, &player->s, true );
 
 	// use the precise origin for linking
-	VectorCopy( player->client->ps.origin, player->r.currentOrigin );
+	G_SetOrigin( player, origin );
```
slipher, line 301: I tested and this part wasn't necessary to prevent the bug. G_SetOrigin sets the `s.pos` stuff which I believe is not used by clients.

slipher (commented):
I also ran into this bug when working on a mod and worked around it by setting the speed to a small nonzero value. But this fixes the underlying problem, nice!

We might also consider clearing the flag when the time is 0 in PM_DropTimers to prevent this type of bug.

src/sgame/sg_utils.cpp
DolceTriade: I figured it's more correct to use a proper named helper function than not? The intent is more clear using this function than a rando vector copy.

slipher: Using G_SetOrigin sets a bad example because this function doesn't actually work for clients. The actual load-bearing part of this function is `VectorCopy( origin, player->client->ps.origin );` followed by `BG_PlayerStateToEntityState( &player->client->ps, &player->s, true );`. If you delete the VectorCopy, it doesn't work at all. If you delete the G_SetOrigin, then the position for linking is just wrong for 1 frame until it gets updated by pmove.

DolceTriade: https://github.com/Unvanquished/Unvanquished/blob/master/src/sgame/sg_client.cpp#L1674

We use it for spawning as well?

slipher: It's the same situation - G_SetOrigin doesn't do nearly enough on its own. There are 3 other places in ClientSpawn where some origin is set. And the only line we actually want from it is the one to set currentOrigin.

DolceTriade: done

slipher:
LGTM

PR #3048, opened 2024-06-22
Finish up Redsky's circle menu refactor

DolceTriade:
It was already in a pretty good state, but I fixed conflicts and also added right click to close.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/alien_build.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/alien_build.rml
@@ -83,6 +83,9 @@
 	tabs {
 		display: none;
 	}
+	#alienBuildBackdrop {
+		background-color: rgba(50, 0, 0, 26);
+	}
 </style>
 <script src="lua/util.lua" />
 <script>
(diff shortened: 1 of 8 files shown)


Gireen:
I guess this is the one running on the Lab server?

When opening the menus they appear for a short moment upside down.
(not blocking: only small visual annoyance but should be fixable someday 😄 )

The mouse seems to wobble to much when moving to the selected items. I think that could be removed.
(blocking for me: i think that would be really annoying )

slipher:
I think the Lab server has other changes. It makes the icons change size on hover, which this PR doesn't do.

DolceTriade:
I added the scale changes. This has lead to a bug I need to investigate.

Gireen:
Ah i was already wondering how it was done cause i didn't see anything in the PR.

Looking at the code it seems fine. The planned parts from the original PR ( #2904 ) would be nice additions but they can also be added later.

Gireen (approved):

necessarily-equal:
Gentle bump

DolceTriade:
A slightly modified version of this is live on the Bug Squash server. Feel free to try it out. The only difference is that on the bug squash server, the icons scale.

illwieckz:
Hmm, this felt into limbo. It would be cool to have this merged.

slipher:
I vaguely recall that there might have been some bugs...

Since I'm working on input stuff, I added a quick fix to get the closing on right click: #3425. I almost cherry-picked the first commit here to get closing on left click outside the buttons too. But I realized that's not so great without the sectors implementation, since then the menu can close if you missed the button on accident.

PR #3050, opened 2024-06-23
Fixed uninitialized field in NavmeshGenerator

slipher:
Ishq discovered that all the threads could sometimes be immediately canceled, due to the uninitialized canceled_ starting as true.


DolceTriade:
works on my end LGTM

PR #3052, opened 2024-06-29
Telenode is not an alienBuildable.

cu-kai:
This fixes: `<alienPlayer> was caught in the acid` obituary messages due to death being attributed to the wrong MOD.


VReaperV (approved):
LGTM

DolceTriade:
rofl lgtm

Issue #3055, opened 2024-06-30
lower default difficulty of bots


Gireen:
I think bots are to strong for new players, act to aggressive and can easily overwhelm newbies. 
It would be reasonable to lower the difficulty of bots in the default settings, since they are the first thing new players encounter offline or on many servers. 
Regulars play on servers with modified or tweaked bots so they should not be affected so much by it.

DolceTriade:
I think we should make chaingun a bot skill and only grant it at higher levels?

Gireen:
Sure its more work, but if someone wants to implement this it would be better.
The next release is somewhat in sight so if no one makes a skill until then i think its better to disable the chaingun and enable it again with a better solution. 

But the problem here is the overall difficulty not just the chaingun, which had recently the loudest complaints. 
There are many potential ways to make bots easier and i think we should simply try some and see how well they work. 
Sweet mentioned also disabling the Tyrant, that would be easy.
Or to have a max percentage of strong units of bots per team, that would be more complicated.

We could also lower the default skill level of 5.

slipher:
> We could also lower the default skill level of 5.

Yeah that's what I was thinking. I don't like the idea of removing chaingun.

PR #3056, opened 2024-06-30
Disable Chaingun for bots by default

Gireen:
part for #3055
not tested but should work 

it was mentioned multiple times, and recently again on irc that bots with chaingun are to strong or could be considered a cheat.

until they can be weakened in other ways the chaingun should be disabled for bots by default.

server owners or offline players can enable it on their own anytime if they want to

### DIFF
--- a/src/sgame/sg_main.cpp
+++ b/src/sgame/sg_main.cpp
@@ -260,7 +260,7 @@ Cvar::Cvar<bool> g_bot_painsaw("g_bot_painsaw", "whether bots buy the Painsaw", 
 Cvar::Cvar<bool> g_bot_shotgun("g_bot_shotgun", "whether bots buy the Shotgun", Cvar::NONE, true);
 Cvar::Cvar<bool> g_bot_lasgun("g_bot_lasgun", "whether bots buy the Lasgun", Cvar::NONE, true);
 Cvar::Cvar<bool> g_bot_mdriver("g_bot_mdriver", "whether bots buy the Mass Driver", Cvar::NONE, true);
-Cvar::Cvar<bool> g_bot_chaingun("g_bot_chain", "whether bots buy the Chaingun", Cvar::NONE, true);
+Cvar::Cvar<bool> g_bot_chaingun("g_bot_chain", "whether bots buy the Chaingun", Cvar::NONE, false);
 Cvar::Cvar<bool> g_bot_prifle("g_bot_prifle", "whether bots buy the Pulse Rifle", Cvar::NONE, true);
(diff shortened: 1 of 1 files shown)


slipher:
#3057 is a proposed solution for fixing the "cheating" aspect of not being affected by chaingun view shaking.

illwieckz:
I guess it's good to merge the disablement, even if things like #3057 are coming. People would still be able to re-enable the chaingun and propose a re-enablement later.

illwieckz (approved):
LGTM.

slipher:
Why? I don't see any really compelling reason to disable it, with or without the view shaking fix.

illwieckz:
Players have bad experience when it is enabled, and leave the game.

cu-kai:
I should add that the primarily played server (Map & Bot Testing) permits players on the team to decide the value of `g_bot_chain` so this PR is unlikely to make the slightest difference, as the `botequip` command there isn't going away as far as I know.

DolceTriade:
Yeah I agree that this "intermediate" solution is not really doing anything.

Gireen:
#3057 sounds like it could improve the situation, how effective it is must be tested.

The worst case scenario, that this PR addresses is a single new player being alone with bots.
Human bots pick relative early in the game lightarmor and chaingun which is for a inexperienced alien an near impossible enemy, even worse when humans appear together with this equipment.

Modded servers, PvE or PvP with many players are not the focus of this change. Unmodified servers and most important Offline play are the places where new players try the game alone and are affected by the bots the most.

Since Offline play can only be updated by a release, this should be addressed before the next one or the situation stays the same for months or even years.

necessarily-equal:
I'm ok with merging this. This seems like a good idea.

illwieckz:
#3057 still being a draft is a good reason why I suggested to disable first, think about it later (maybe re-enable it later).

If I'm right both unvanquished.net and my server have it disabled from config file.

VReaperV:
If this is already done by server hosts who want it to be disabled, I see no value in this pr.

illwieckz:
The idea is to have a default gameplay that works.

Gireen:
Yeah, the default values are what new players are most likely to encounter. If the release cycle is really getting shorter now, it's not quite as urgent anymore and the jitter PR could be tested in the next release. 

So I would suggest that everyone now vote with their thumbs on the PR as with the motionblur and if no jitter PR or any other improvement attempt has been merged before the next release, then this PR will be merged if the vote is positive.

Gireen:
jitter was merged, the vote is a draw or not interesting and releases are more frequent. 
So this PR has become less urgent with more chances to try other things.

PR #3057, opened 2024-07-02
Make bots affected by chaingun jitter

slipher:
When bots are aiming at enemies, don't erase the view angle perturbations enforced through `delta_angles`. This makes bots affected by chaingun shake. This also affects bots in trappers, but it's harder to characterize the difference there.

Since low-skill bots were then too bad at aiming with chaingun due to the upward drift while firing, I made bots take into account the determinic upward part of the view angle deflection.

Still draft because I need to fix some vertical navcon-related code that sets an absolute value for the pitch angle. The `Obey usercmd_t protocol to set bot view angles` commit needs to be updated to fix that part.


slipher:
It's now ready. I accounted for the 3 places that set the bot's pitch angle directly and tested them.

DolceTriade:
I'll cherrypick on my server for testing and report back.

illwieckz (approved):
LGTM.

illwieckz:
I followed bots equiped with chaingun attacking buildables (so, a non-moving target), and indeed they're affected by the jitter.

sweet235:
> I'll cherrypick on my server for testing and report back.

What is your conclusion?

Issue #3058, opened 2024-07-02
RmlUI transformations affect the console


DolceTriade:
https://github.com/Unvanquished/Unvanquished/assets/1097524/9f410400-ef70-41e8-bdfb-4745d344aee8

Attached is a video. Likely a bug in our transform code.

An interesting thing is that it doesn't scale with all of the items, just that specific one. It happens with that specific position with the circle menus I tested.

Can repro on the `The Lab Experiments | EU` server.

PR #3059, opened 2024-07-02
rmlui: Reset MatrixTransform after rendering

DolceTriade:
Apparently, RmlUI doesn't reset the final matrix transformation so that the last transformation can apply to whatever we end up rendering next, which may end up being the console.

For some reason, this also fixes the jiggly mouse issue when using RmlUI transformations.

Partially fixes https://github.com/DaemonEngine/Daemon/issues/1210

### DIFF
--- a/src/cgame/rocket/rocket.cpp
+++ b/src/cgame/rocket/rocket.cpp
@@ -466,6 +466,8 @@ void Rocket_Render()
 		menuContext->Render();
 	}
 
+	trap_R_ResetMatrixTransform();
+
 }
 
 void Rocket_Update()


slipher:
LGTM

We could say this fixes #3058 but not https://github.com/DaemonEngine/Daemon/issues/1210. Since I still think the engine should be hardened against cgame misbehavior, in addition to fixing the cgame. Assuming the "jiggly mouse issue" is something to do with the 'custom cursor' used it menus, we will need to fix both in any case, since the cursor is drawn by the cgame outside of RmlUi.

PR #3060, opened 2024-07-02
buildables: make g_indestructibleBuildables change existing buildables

DolceTriade:
Previously, g_indestructibleBuildables only took effect for buildables set after the value was set. This was annoying to use because it meant that this value *must* be set before a map is loaded to make it take effect. Further, if you want to disable it, it means you need to change the map for it to stop making buildables invincible.

The new behavior makes the cvar take effect immediately, which is less surprising and more inline with other cvars.

### DIFF
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -34,10 +34,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "CBSE.h"
 #include "sg_cm_world.h"
 
-static Cvar::Cvar<bool> g_indestructibleBuildables(
-		"g_indestructibleBuildables",
-		"make buildables impossible to destroy (Note: this only applies only to buildings built after the variable is set, This also means it must be set before map load for the default buildables to be protected)", Cvar::NONE, false);
-
 /**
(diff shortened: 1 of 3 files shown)


src/sgame/sg_buildable.cpp
```diff
@@ -1523,8 +1519,7 @@ static gentity_t *SpawnBuildable( gentity_t *builder, buildable_t buildable, con
 	built->enabled = false;
 	built->spawned = false;
 
-	built->flags = g_indestructibleBuildables.Get() ?
-		FL_GODMODE | FL_NOTARGET : 0;
+	built->flags = 0;
```
slipher, line 1522: This line should be reverted, or it will only work for buildables constructed before toggling the cvar.

DolceTriade: I don't understand?

Toggling the cvar will override whatever value was originally there?

DolceTriade: Oh I see. I'm dumb

DolceTriade: Done.

slipher:
LGTM modulo removing junk submodule changes

DolceTriade:
undid submodule changes

PR #3061, opened 2024-07-04
bots trace for the floor to dance carefully

sweet235:
A crucial part of the difficulty of higher level bots is their dancing. When they are fighting an enemy, they strafe to be less easy to hit.

I think the strafe code is reasonably difficult to predict users. It is pretty good. But it does have a big deficit: bots will fall down to lower floor levels (or to death zones) while strafing. Human bots with ranged weapons will also try to keep a minimum distance to their enemies, with the same effect.

Fix this by doing a trace for the floor. Only strafe (or go backward) when there is floor to move to.

### DIFF
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -368,6 +368,47 @@ Local Bot Navigation
 ========================
 */
 
+bool BotTraceForFloor( gentity_t *self, uint32_t dir )
+{
+	// this is using a daemon trace to a solid surface
+	// another feasible method might be to do a recast navmesh trace
+	trace_t trace;
+	glm::vec3 dirVec;
+	glm::vec3 viewangles = VEC2GLM( self->client->ps.viewangles );
+	viewangles.x = 0;
+	AngleVectors( viewangles, &dirVec, nullptr, nullptr );
+	// we only handle a subset of the possible values of the bitfield `dir`
+	if ( dir == MOVE_BACKWARD )
+	{
+		dirVec.x = -dirVec.x;
+		dirVec.y = -dirVec.y;
+	}
+	else if ( dir == MOVE_LEFT )
+	{
+		std::swap( dirVec.x, dirVec.y );
+		dirVec.x = -dirVec.x;
+	}
+	else if ( dir == MOVE_RIGHT )
+	{
+		std::swap( dirVec.x, dirVec.y );
+		dirVec.y = -dirVec.y;
+	}
+	else
+	{
+		return false;
+	}
+	glm::vec3 playerMins, playerMaxs;
+	class_t pClass = static_cast<class_t>( self->client->ps.stats[STAT_CLASS] );
+	BG_BoundingBox( pClass, &playerMins, &playerMaxs, nullptr, nullptr, nullptr );
+	float radius = playerMaxs.x - playerMins.x;
+	float height = playerMaxs.z - playerMins.z;
+	glm::vec3 start = VEC2GLM( self->s.origin ) + dirVec * ( 2.f * radius );
+	glm::vec3 end = start;
+	end.z -= height;
+	trap_Trace( &trace, &start[ 0 ], nullptr, nullptr, &end[ 0 ], self->num(), MASK_SOLID, 0 );
+	return trace.fraction < 1.f;
+}
+
 static signed char BotGetMaxMoveSpeed( gentity_t *self )
 {
 	if ( usercmdButtonPressed( self->botMind->cmdBuffer.buttons, BTN_WALKING ) )
@@ -382,19 +423,28 @@ void BotStrafeDodge( gentity_t *self )
 {
 	usercmd_t *botCmdBuffer = &self->botMind->cmdBuffer;
 	signed char speed = BotGetMaxMoveSpeed( self );
+	bool floorRight = BotTraceForFloor( self, MOVE_RIGHT );
+	bool floorLeft = BotTraceForFloor( self, MOVE_LEFT );
 
-	if ( self->client->time1000 >= 500 )
+	if ( self->client->time1000 >= 500 && floorRight )
 	{
 		botCmdBuffer->rightmove = speed;
 	}
-	else
+	else if ( floorLeft )
 	{
 		botCmdBuffer->rightmove = -speed;
+	}
+	else if ( floorRight )
+	{
+		botCmdBuffer->rightmove = speed;
 	}
 
 	if ( ( self->client->time10000 % 2000 ) < 1000 )
 	{
-		botCmdBuffer->rightmove *= -1;
+		if ( ( botCmdBuffer->rightmove < 0.f && floorRight ) || ( botCmdBuffer->rightmove > 0.f && floorLeft ) )
+		{
+			botCmdBuffer->rightmove *= -1;
+		}
 	}
 
 	if ( ( self->client->time1000 % 300 ) >= 100 && ( self->client->time10000 % 3000 ) > 2000 )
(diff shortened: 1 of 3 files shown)


src/sgame/sg_bot_nav.cpp
```diff
@@ -368,6 +368,51 @@ Local Bot Navigation
...
+	float radius = playerMaxs.x - playerMins.x;
+	float height = playerMaxs.z - playerMins.z;
+	// use dirVec * ( 2.f * radius ) as direction offset
+	// in theory, sqrtf(2.f) instead of 2.f would be enough
+	// the slight increase is to safeguard against unexpected map geometry
+	// a bot not dancing is less problematic than a bot falling down
+	glm::vec3 start = VEC2GLM( self->s.origin ) + dirVec * ( 2.f * radius );
+	glm::vec3 end = start;
+	end.z -= height;
+	trap_Trace( &trace, &start[ 0 ], nullptr, nullptr, &end[ 0 ], self->num(), MASK_SOLID, 0 );
```
slipher, line 412: Use the GLM variant: replace `&start[0], nullptr, nullptr, &end[0]` with `start, {}, {}, end`

src/sgame/sg_bot_nav.cpp
```diff
@@ -368,6 +368,51 @@ Local Bot Navigation
 ========================
 */
 
+bool BotTraceForFloor( gentity_t *self, uint32_t dir )
```
slipher, line 371: `botMoveDir_t dir`

src/sgame/sg_bot_nav.cpp
```diff
@@ -368,6 +368,51 @@ Local Bot Navigation
...
+	BG_BoundingBox( pClass, &playerMins, &playerMaxs, nullptr, nullptr, nullptr );
+	float radius = playerMaxs.x - playerMins.x;
+	float height = playerMaxs.z - playerMins.z;
+	// use dirVec * ( 2.f * radius ) as direction offset
+	// in theory, sqrtf(2.f) instead of 2.f would be enough
+	// the slight increase is to safeguard against unexpected map geometry
+	// a bot not dancing is less problematic than a bot falling down
+	glm::vec3 start = VEC2GLM( self->s.origin ) + dirVec * ( 2.f * radius );
+	glm::vec3 end = start;
+	end.z -= height;
```
slipher, line 411: By doing a trace of length of `playerMaxs.z` - `playerMins.z` that starts at a height of `-playerMins.z` above the player's feet, you end up tracing a distance of `playerMaxs.z` below the feet. It's not good to write code that depends on `playerMaxs.z` because the height of the origin within the bbox is arbitrary. The max z could be anywhere from 0 to the entire height of the bbox without affecting collisions.

To me it would make more sense to trace a constant distance below the player's feet.

src/sgame/sg_bot_nav.cpp
```diff
@@ -368,6 +368,51 @@ Local Bot Navigation
...
 */
 
+bool BotTraceForFloor( gentity_t *self, uint32_t dir )
+{
+	// this is using a daemon trace to a solid surface
+	// another feasible method might be to do a recast navmesh trace
+	trace_t trace;
+	glm::vec3 dirVec;
+	glm::vec3 viewangles = VEC2GLM( self->client->ps.viewangles );
+	viewangles.x = 0;
```
slipher, line 378: `viewangles[ PITCH ]`

src/sgame/sg_bot_nav.cpp
sweet235: Do we have any guidelines on how and when and why to use the glm variants? I am lost here. And I bet I'm not the only one.

src/sgame/sg_bot_nav.cpp
sweet235: Do I get the player's feet height by doing

```
glm::vec3 origin = VEC2GLM( self->s.origin );
origin.z += playerMins.z;
```

slipher: That's right

src/sgame/sg_bot_nav.cpp
slipher: The GLM variants just have the first 4 arguments as glm::vec3 instead of vec3_t. So you should definitely use them if all your vectors are glm::vec3.

But come to think of it, you might want to use `G_Trace2` instead of `trap_Trace`. This is because trap_Trace may ignore entities that overlap the starting point of the trace, which is probably not desired here. G_Trace2 does not have a GLM variant.

sweet235: No thx. Your glm migration is beyond me.

sweet235:
Convert to draft, because I do not seem to understand the glm migration guidelines. May some more knowledgeable person continue this.

src/sgame/sg_bot_nav.cpp
sweet235: > To me it would make more sense to trace a constant distance below the player's feet.

That would miss any upward slopes. Better trace from the player's top.

slipher: True, if you are doing a point trace you would need to start higher.

sweet235:
I am now doing this: from the bot's head, go backward for sqrt(2) times its side length, then raytrace down for 2 times its height. This seems to work well.

src/sgame/sg_bot_nav.cpp
```diff
@@ -368,6 +368,50 @@ Local Bot Navigation
...
+	glm::vec3 playerMins, playerMaxs;
+	class_t pClass = static_cast<class_t>( self->client->ps.stats[STAT_CLASS] );
+	BG_BoundingBox( pClass, &playerMins, &playerMaxs, nullptr, nullptr, nullptr );
+	float radius = playerMaxs.x - playerMins.x;
+	float height = playerMaxs.z - playerMins.z;
+	glm::vec3 start = VEC2GLM( self->s.origin ) + dirVec * ( 1.414214f * radius );
+	start.z += playerMaxs.z;
+	glm::vec3 end = start;
+	end.z -= height * 2.f;
+	trap_Trace( &trace, start, {}, {}, end, self->num(), MASK_SOLID, 0 );
```
slipher, line 411: Would be good to include CONTENTS_PLAYERCLIP

src/sgame/sg_bot_nav.cpp
```diff
@@ -368,6 +368,50 @@ Local Bot Navigation
...
+		return false;
+	}
+	dirVec.z = 0.f;
+	glm::normalize( dirVec );
+	glm::vec3 playerMins, playerMaxs;
+	class_t pClass = static_cast<class_t>( self->client->ps.stats[STAT_CLASS] );
+	BG_BoundingBox( pClass, &playerMins, &playerMaxs, nullptr, nullptr, nullptr );
+	float radius = playerMaxs.x - playerMins.x;
+	float height = playerMaxs.z - playerMins.z;
+	glm::vec3 start = VEC2GLM( self->s.origin ) + dirVec * ( 1.414214f * radius );
```
slipher, line 407: Could you comment the geometrical motivation behind square root of 2?

src/sgame/sg_bot_nav.cpp
```diff
@@ -368,6 +368,50 @@ Local Bot Navigation
...
+	else
+	{
+		return false;
+	}
+	dirVec.z = 0.f;
+	glm::normalize( dirVec );
+	glm::vec3 playerMins, playerMaxs;
+	class_t pClass = static_cast<class_t>( self->client->ps.stats[STAT_CLASS] );
+	BG_BoundingBox( pClass, &playerMins, &playerMaxs, nullptr, nullptr, nullptr );
+	float radius = playerMaxs.x - playerMins.x;
```
slipher, line 405: The name `radius` is a bit misleading since it suggests a half-extent, not the full extent. I'd call it `width`

src/sgame/sg_bot_nav.cpp
illwieckz: Also, we have `M_SQRT2`.

PR #3066, opened 2024-07-07
shared: Set count after printing warning in BG_PackEntityNumbers

DolceTriade:
Otherwise the count wrong in the warning making it less helpful.


slipher:
LGTM

PR #3068, opened 2024-07-07
Fix bug when player stands in a disabled teleporter

### LINKED ISSUE #2447: Walking into a "disabled" `trigger_teleport` causes rendering issues
Unsure whether this is a Daemon bug or can be fixed in the gamelogic.

When walking into a trigger_teleport which has been disabled, the "Hall Of Mirrors" effect is present until you exit the trigger volume. You are not teleported to the destination, so the toggle itself works correctly.

I have tried using `env_afx_teleport` but it seems it cannot be toggled (that seems worthy of its own new issue?)

You can test this with map-mission-control_0.4.dpk by entering the alien base, noclipping just past the invisible wall at the start, then standing above the first three lines of eggs. Make sure to trigger the invisible trigger just inside the door to the alien base too.

Example of what I see:
unvanquished_2023-02-12_224958_000

cu-kai:
Fixes #2447

With thanks to @VReaperV for the suggestion on where to look.

### DIFF
--- a/src/cgame/cg_predict.cpp
+++ b/src/cgame/cg_predict.cpp
@@ -398,7 +398,7 @@ static void CG_TouchTriggerPrediction()
 			continue;
 		}
 
-		if ( ent->eType == entityType_t::ET_TELEPORTER )
+		if ( ent->eType == entityType_t::ET_TELEPORTER && !( ent->eFlags & EF_NODRAW ) )
 		{
 			cg.hyperspace = true;
 		}


DolceTriade (approved):

src/cgame/cg_predict.cpp
```diff
@@ -398,7 +398,7 @@ static void CG_TouchTriggerPrediction()
 			continue;
 		}
 
-		if ( ent->eType == entityType_t::ET_TELEPORTER )
+		if ( ent->eType == entityType_t::ET_TELEPORTER && ent->eFlags & ~EF_NODRAW )
```
slipher, line 401: Checking for any flag except EF_NODRAW? Surely that can't be right. Maybe you meant `!( ent->eFlags & EF_NODRAW )`

cu-kai: `EF_NODRAW` is the flag set when the teleporter is disabled, so I think this is correct?

slipher: The condition as written ignores whether EF_NODRAW is set and checks whether any other flag is set.

cu-kai: Turns out I cannot read code properly. I've fixed this in the latest force-push.

DolceTriade (approved):

PR #3074, opened 2024-07-26
Kill ghost buildings to ensure decrementing the spawn counters

sweet235:
You can encounter ghost eggs frequently in the wild. They are very annoying. The egg does not exist, but the spawn counter will never go down to zero. They might be in a map's builtin layout or in a layout someone saved two years ago. One example is map moonbase.

Why are there ghost eggs at all? I do not know. Probably because of changes to the trace code.

Policy dicatates to modify the spawn counters only in the CBSE components. I am not a CBSE expert, so I cannot promise this is correct. Are there any CBSE experts?

### DIFF
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -1838,7 +1838,7 @@ static gentity_t *FinishSpawningBuildable( gentity_t *ent, bool force )
 	{
 		Log::Debug( "^3G_FinishSpawningBuildable: %s startsolid at %s",
 		          built->classname, vtos( built->s.origin ) );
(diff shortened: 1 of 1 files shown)


cu-kai (changes requested):

cu-kai (approved):
LGTM

slipher:
> Why are there ghost eggs at all?

Could be because the buildable has a different bbox size.

LGTM

sweet235:
Someone said bisecting leads to 117e11bcc71417ae14150d635c98082defdfa12a.

sweet235:
Thinking about this, fixing this might actually require a CBSE expert.

PR #3075, opened 2024-08-01
translation: add `completion.txt` and add `write_completion.sh` to write it

illwieckz:
The idea is that, when someone merges translation updates from Weblate, he runs `./generate.sh` which syncs the strings with current source and write a file named `completion.txt`.

The format is like this:

```
3 ar
8 br
91 de
0 eo
54 es
100 fi
100 fr
14 gl
1 hu
0 ia
90 it
28 pl
91 pt
92 pt_BR
81 ru
17 uk
99 zh_Hans
99 zh_Hant
99 zh_Hant_HK
```

First field is completion percentage, second field is language code name.

The purpose is that one day in the future the game may only enable a translation (and list it in menu) once some percentage of completion is reached (like 80%?), it would not prevent translators to set the language by hand, or add a cvar to enable/list all languages even if poorly translated.

I believe that under some level of completion an incomplete translation is counter-productive. Like if you read a full page written in a foreign language then meet some words from your native language, your brain may try to interpret the native word in the foreign language and be very confused.


slipher:
I doubt that making language enabled/disabled state completely automatic is good. Suppose someone translates 80% of strings to Arabic. We should still not enable it yet because we don't implement RTL text rendering, so it's unusable.

Also languages need to have names to be displayed in the menu. Because of that it seems that we might have to maintain a language list by hand anyway?

illwieckz:
> Suppose someone translates 80% of strings to Arabic. We should still not enable it yet because we don't implement RTL text rendering, so it's unusable.

We can write a file to tell urcheon to ignore the arabic files for now.

> Also languages need to have names to be displayed in the menu. Because of that it seems that we might have to maintain a language list by hand anyway?

Maybe there exist ready to use translated list of languages names with code names?

slipher:
> > Suppose someone translates 80% of strings to Arabic. We should still not enable it yet because we don't implement RTL text rendering, so it's unusable.
> 
> We can write a file to tell urcheon to ignore the arabic files for now.

A disadvantage of this is that not all packagers use Urcheon.

> > Also languages need to have names to be displayed in the menu. Because of that it seems that we might have to maintain a language list by hand anyway?
> 
> Maybe there exist ready-to-use translated list of languages names with code names?

Maybe. But I doubt there will be something that has all languages the way we want them without needing any customization. It would suck if we are loading 3 different config files just to make a list of languages.

illwieckz:
I'll do something else.

PR #3081, opened 2024-08-15
do not try to make alien bots crouch at obstacles

sweet235:
There is a mistake in the code deciding whether a human bot should crouch. The crouch activation is done for alien classes too, resulting in wall climbers randomly activating their climb and sometimes getting stuck on walls or ceilings. Fix this.


slipher:
LGTM

PR #3082, opened 2024-08-20
Add /testshader command

slipher:
To see what a texture looks like. It draws a square with the given q3shader, positioned in front of where the player was when the command was executed, similar to `testmodel`.

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -1224,6 +1224,8 @@ struct cg_t
 	char                    testModelName[ MAX_QPATH ];
 	char                    testModelBarrelName[ MAX_QPATH ];
 	bool                testGun;
+	qhandle_t hTestShader;
+	polyVert_t testShaderPoly[ 4 ];
 
 	int                     spawnTime; // fovwarp
 	int                     weapon1Time; // time when BTN_ATTACK went t->f f->t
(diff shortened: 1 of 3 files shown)


src/cgame/cg_view.cpp
```diff
@@ -222,6 +222,48 @@ static void CG_AddTestModel()
...
+class TestShaderCmd : public Cmd::StaticCmd
+{
+public:
+	TestShaderCmd() : StaticCmd("testshader", "draw a polygon with specified q3shader") {}
+
+	void Run( const Cmd::Args &args ) const override
+	{
+		if ( args.Argc() < 2 )
+		{
+			cg.hTestShader = 0;
```
DolceTriade, line 236: Print a message that it's disabling so people running this command without any arguments and expecting usage are less confused?

src/cgame/cg_local.h
```diff
@@ -1224,6 +1224,8 @@ struct cg_t
 	char                    testModelName[ MAX_QPATH ];
 	char                    testModelBarrelName[ MAX_QPATH ];
 	bool                testGun;
+	int hTestShader;
```
DolceTriade, line 1227: qhandle_t?

DolceTriade (approved):
two nits

PR #3084, opened 2024-08-27
NUKE sensor_stage entities after firing

sweet235:
The only use case I have seen in the wild is opening some door once a team reaches a stage. These doors must not be fired a second time, because some door-like entities will respond by closing the door. See map-pierogi for example.


slipher:
LGTM

cu-kai (approved):
Sensible change seeing as unlike here, a team could not stage down in Tremulous. 

While `trigger_stage` could legitimately be fired more than once by not specifying a team (thus, it could be fired twice - once for each team), I have not seen a map where this functionality is used.

The TODO comment can be addressed in a later PR.

PR #3086, opened 2024-09-03
firing a func_rotating entity stops/resumes its rotation

sweet235:
As recently requested. Old code would use `BinaryMover_act` on it, which got very confused, believing this is a door-like thing with start and end positions. This resulted in the brushes ~~vanishing from the map~~ teleporting to the origin.

TODO: test with groups of entities

### DIFF
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -2688,12 +2688,6 @@ static void FuncRotating_act( gentity_t *ent, gentity_t *other, gentity_t *activ
 {
 	gentity_t *master;
 
-	// if this is a non-client-usable thing, return
-	if ( ent->mapEntity.names[ 0 ] && other && other->client )
-	{
-		return;
-	}
-
 	// only the master should be used
 	if ( ent->flags & FL_GROUPSLAVE )
 	{
--- a/src/sgame/sg_spawn_mover.cpp
+++ b/src/sgame/sg_spawn_mover.cpp
@@ -2711,16 +2711,9 @@ static void FuncRotating_act( gentity_t *ent, gentity_t *other, gentity_t *activ
 		if ( ent->s.apos.trDelta[ 0 ] == 0.f && ent->s.apos.trDelta[ 1 ] == 0.f && ent->s.apos.trDelta[ 2 ] == 0.f )
 		{
 			// we are out of phase with a rotation that started with euler angle = 0 at
-			// level.time = 0 now, since we had stopped the rotator for some time.
-			// so we calculate the phase shift of the new rotation, and correct the new
-			// value of the euler angle accordingly.
-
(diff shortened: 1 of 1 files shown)


sweet235:
TODO: think about groups containing `func_rotating` and other entity types.

sweet235:
Doing this is relatively straightforward. We can freeze the func_rotating by setting its angular velocity to zero, and using its current euler angle its new base euler angle. When we resume the rotation, we have to consider the phase shift against a rotation that had never stopped, and adjust the new euler angle accordingly. Thus no change to the code doing the actual rotation is needed.

As far as I can see, only the group problem remains: the group must consist of `func_rotating` entities exclusively.

cu-kai (approved):
I have tested this against the `test-portals-moving` map which has a toggle-able rotator. Seems to work fine.

LGTM.

src/sgame/sg_spawn_mover.cpp
```diff
@@ -2665,30 +2665,94 @@ check either the X_AXIS or Y_AXIS box to change that.
...
+	for ( ent = master; ent; ent = ent->mapEntity.groupChain )
+	{
+		if ( ent->s.apos.trDelta[ 0 ] == 0.f && ent->s.apos.trDelta[ 1 ] == 0.f && ent->s.apos.trDelta[ 2 ] == 0.f )
+		{
+			// we are out of phase with a rotation that started with euler angle = 0 at
+			// level.time = 0 now, since we had stopped the rotator for some time.
+			// so we calculate the phase shift of the new rotation, and correct the new
+			// value of the euler angle accordingly.
+
+			int periodLength = 360000 / ent->mapEntity.speed;  // milliseconds
```
slipher, line 2718: Instead of doing this calculation you can set the `trTime` field to the current time and leave `trBase` unchanged.

sweet235: lol

sweet235:
Having read other act functions, I am now convinced that the group problem is not restricted to this PR. I will leave it to be fixed in a future PR.

src/sgame/sg_spawn_mover.cpp
```diff
@@ -2665,31 +2665,87 @@ check either the X_AXIS or Y_AXIS box to change that.
...
+		}
+		else
+		{
+			for ( int i = 0; i < 3; i++ )
+			{
+				// set the rotation speed to zero
+				ent->s.apos.trDelta[ i ] = 0.f;
+
+				// set the current euler angle as new base euler angle
+				// restricting the angle to the interval 0..360 is not
```
slipher, line 2727: Nit: it's actually -360..360 (I assume speed can be negative to reverse the rotation direction).

slipher (commented):
LGTM

src/sgame/sg_spawn_mover.cpp
sweet235: One day

PR #3090, opened 2024-09-15
Fix CI (Disable missing-braces warning on NaCl)

slipher:
Recent changes in Daemon adding -Wall -Wextra enabled this warning, which is badly implemented on old versions of Clang.

PR #3091, opened 2024-09-15
Build system improvements

slipher:
Builds on #3090.

- Fix a regression of CBSE Python path not propagating to NaCl build (by not using it there at all)
- Propagate Recast configuration options to the NaCl build
- Avoid building/finding some libraries when they are not needed


DolceTriade (approved):
nice clean up

PR #3092, opened 2024-09-19
Add '/r' command for replying to the last private message sender

Grise3:
Hello, 

Here is the `/r` command witch is used to reply to the last private message sender just with `/r`

Like this : `/r [Message]`
and not `/m [Target Name] [Message]`

Thanks to sweet for advice

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -4224,6 +4224,7 @@ static const commands_t cmds[] =
 	{ "listmaps",        CMD_MESSAGE | CMD_INTERMISSION,      Cmd_ListMaps_f         },
 	{ "listrotation",    CMD_MESSAGE | CMD_INTERMISSION,      G_PrintCurrentRotation },
 	{ "m",               CMD_MESSAGE | CMD_INTERMISSION,      Cmd_PrivateMessage_f   },
+	{ "r",               CMD_MESSAGE | CMD_INTERMISSION,      Cmd_ReplyPrivateMessage_f   },
 	{ "maplog",          CMD_MESSAGE | CMD_INTERMISSION,      Cmd_MapLog_f           },
 	{ "me",              CMD_MESSAGE | CMD_INTERMISSION,      Cmd_Me_f               },
 	{ "me_team",         CMD_MESSAGE | CMD_INTERMISSION,      Cmd_Me_f               },
@@ -4441,6 +4442,53 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
 		             name, color.c_str(), msg );
 	}
 }
+
+
+
+/*
+=================
+Cmd_ReplyPrivateMessage_f
+
+Reply by private message to the last private message sender
+=================
+*/
+
+void Cmd_ReplyPrivateMessage_f(gentity_t *ent)
+{
+    char *msg;
+    int target;
+    bool teamonly = false;
+
+    if (!g_privateMessages.Get() && ent)
+    {
+        ADMP( "\"" N_("Sorry, but private messages have been disabled") "\"" );
+        return;
+    }
+
+    if (trap_Argc() < 2)
+    {
+        ADMP( QQ( N_("usage: /r [message]") ) );
+        return;
+    }
+
+    target = ent->client->pers.lastPrivateMessageSender;
+    if (target == -1 || !g_entities[target].client)
+    {
+        ADMP( "\"" N_("No one to reply to.") "\"" );
+        return;
+    }
+
+    msg = ConcatArgs(1);
+
+    if (G_SayTo(ent, &g_entities[target], teamonly ? SAY_TPRIVMSG : SAY_PRIVMSG, msg))
+    {
+        ADMP(va("%s %s", QQ(N_("Private message: ^7$1$")), Quote(msg)));
+        G_LogPrintf("PrivMsg: %d \"%s^*\" \"%s\": %s",
+                     ent->num(), ent->client->pers.netname,
+                     g_entities[target].client->pers.netname, msg);
+    }
+}
+
 
 /*
 =================
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -4444,7 +4444,6 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
 }
 
 
-
 /*
 =================
 Cmd_ReplyPrivateMessage_f
--- a/src/sgame/sg_struct.h
+++ b/src/sgame/sg_struct.h
@@ -346,6 +346,8 @@ struct clientPersistant_t
 	weapon_t          humanItemSelection; // humans have a starting item
 
 	int               teamChangeTime; // level.time of last team change
+	int		  lastPrivateMessageSender; // last private message sender
+
 	namelog_t         *namelog;
 	g_admin_admin_t   *admin;
 


sweet235:
I have thought a while about the obvious drawback: you are typing a reply to user A using `/r`, and a fraction of a second before you send it, you receive a private message from user B. You hit the return key before could possibly notice it. Your message to user B instead of A.

I could not find a simple solution.

slipher:
If a different person sends you a private message at the last moment, they can intercept the message. So I don't think this is a good idea as a console command. If you want to make a convenient way to reply to the last PM sender, I would suggest making it a part of the client chat interface. There are various binds for chatting to everyone (`message_public`), admins (`message_admin`), etc. You could add another one like for the last private message sender. And that way you can determine who will be the recipient at the time the chat is opened, instead of when the message is sent, so you won't have the problem of sending it to the wrong person.

Also, the `lastPrivateMessageSender` variable is never assigned to, so maybe you forgot to upload some of the code.

slipher:
> If a different person sends you a private message at the last moment, they can intercept the message. So I don't think this is a good idea as a console command. If you want to make a convenient way to reply to the last PM sender, I would suggest making it a part of the client chat interface. There are various binds for chatting to everyone (`message_public`), admins (`message_admin`), etc. You could add another command like that for the last private message sender. And that way you can determine who will be the recipient at the time the chat is opened, instead of when the message is sent, so you won't have the problem of sending it to the wrong person.
> 
> Also, the `lastPrivateMessageSender` variable is never assigned to, so maybe you forgot to upload some of the code.

illwieckz:
> I could not find a simple solution. It is probably ok to simply ignore this and live dangerously.

YOLO! 🤸‍♀️️

sweet235:
> And that way you can determine who will be the recipient at the time the chat is opened, instead of when the message is sent, so you won't have the problem of sending it to the wrong person.

Can we really? If we do that, a user would have to close and open the client console each time they want to reply to a different recipient. Not to mention the confusion arising when that command is entered in the games standard input. That is probably not what grise has in mind.

slipher:
> Can we really? If we do that, a user would have to close and open the client console each time they want to reply to a different recipient. Not to mention the confusion arising when that command is entered in the game's standard input. That is probably not what grise has in mind.

This does not involve the console. Try binding a key to `message_public` and pressing the key.

sweet235:
> This does not involve the console. Try binding a key to `message_public` and pressing the key.

How would you prevent users from entering `message_replay` themselves?

Grise3:
> so maybe you forgot to upload some of the code.

Yep. I've added the missing code

sweet235:
In the end, our chat system is on the rudimentary side. We cannot expect @Grise3 to fix it in this PR.

Grise3:
> In the end, our chat system is on the rudimentary side. We cannot expect @Grise3 to fix it in this PR.
 Yep ;)

I can try to make it better (will be hard for me) but not now :D  
and it is not the subject of this pr ^^

DolceTriade:
i haven't reviewed the code yet or tested it, but I like the concept.

I think the edge case of receiving a bunch of PMs is not that relevant and is also fixable. WDYT?

cu-kai:
What about refusing to send the response if a new message was received X seconds before you use `/r`?  You could also allow the client to set how many seconds that might be (`0` to disable).

slipher:
> But how do i get `bind y r yes` without a console command?

Hmm, maybe the message commands can accept an argument, which would pre-fill the text field contents? :stuck_out_tongue: 

> What about refusing to send the response if a new message was received X seconds before you use `/r`? You could also allow the client to set how many seconds that might be (`0` to disable).

It would be annoying if additional messages from the same interlocutor caused the reply to be rejected. So I came up with this algorithm: allow the message if the *last message sender has been the same for the last 3 seconds*. So in addition to `lastPrivateMessageSender`, also track a `lastPrivateMessageSenderTime`. This time would be updated not every time you receive a message, but every time `lastPrivateMessageSender` *changes*.

sweet235:
I'm helping @Grise3 a little.

> I came up with this algorithm: allow the message if the last message sender has been the same for the last 3 seconds

I did that.

One remaining problem is that only the `/m` command sets `lastPrivateMessageSender` (and the time). We should fix this.

Before we merge this, we should replace some spaces by tabs.

Grise3:
> Before we merge this, we should replace some spaces by tabs.

Done !

src/sgame/sg_cmds.cpp
```diff
@@ -4442,6 +4451,58 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
...
+
+	if (!g_privateMessages.Get() && ent)
+	{
+		ADMP( "\"" N_("Sorry, but private messages have been disabled") "\"" );
+		return;
+	}
+
+	if (trap_Argc() < 2)
+	{
+	ADMP( QQ( N_("usage: /r [message]") ) );
```
DolceTriade, line 4477: Maybe it would be a good idea to print who we are replying to here?

src/sgame/sg_cmds.cpp
```diff
@@ -4442,6 +4451,58 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
...
+	return;
+	}
+
+	if ( level.time - ent->client->pers.lastPrivateMessageSenderTime < 3000 )
+	{
+		ADMP( "\"" N_("More than one possible recipient, refusing to send.") "\"" );
```
DolceTriade, line 4483: Maybe we should also print who we would have sent the PM to?

src/sgame/sg_cmds.cpp
```diff
@@ -4442,6 +4451,58 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
...
+	target = ent->client->pers.lastPrivateMessageSender;
+	if (target == -1 || !g_entities[target].client)
+	{
+		ADMP( "\"" N_("No one to reply to.") "\"" );
+		return;
+	}
+
+	msg = ConcatArgs(1);
+
+	if (G_SayTo(ent, &g_entities[target], teamonly ? SAY_TPRIVMSG : SAY_PRIVMSG, msg))
```
DolceTriade, line 4496: Let's also print who we are replying to

DolceTriade (approved):
I think this is fine % some extra UX considerations. Since the user doesn't explicitly say who they are sending the message to, I think it would be good if there were a way for the game to tell you (and also explicitly tell who the message was sent to.

This would allow you to run `/r` with no arguments and check who you would send the message to. Also after sending a message, we confirm who the message was sent to.

src/sgame/sg_cmds.cpp
DolceTriade: To be clear, I mean send a message to the person who sent the command:

Something like (didn't test):
```
ADMP( va( "%s %s", QQ( N_("Sent message to %s") ), g_entities[target].client->pers.netname ) );
```

src/sgame/sg_cmds.cpp
```diff
@@ -4442,6 +4451,58 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
...
+	}
+
+	if ( level.time - ent->client->pers.lastPrivateMessageSenderTime < 3000 )
+	{
+		ADMP( "\"" N_("More than one possible recipient, refusing to send.") "\"" );
+        return;
+	}
+
+	target = ent->client->pers.lastPrivateMessageSender;
+	if (target == -1 || !g_entities[target].client)
```
slipher, line 4488: IIRC `client` will always be set so null-checking it is useless. G_SayTo will check whether there is a connected client or not so you could make this test like `if ( target == -1 || !G_SayTo( ... ) )`

src/sgame/sg_cmds.cpp
slipher: Yeah. Name needs to be quoted though

src/sgame/sg_cmds.cpp
```diff
@@ -4412,6 +4415,12 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
 			count++;
 			Q_strcat( recipients, sizeof( recipients ), va( "%s^*, ",
 			          level.clients[ pids[ i ] ].pers.netname ) );
+			int playerNum = ent - g_entities;
```
slipher, line 4418: Use `ent->num()`

Grise3:
> I think this is fine % some extra UX considerations. Since the user doesn't explicitly say who they are sending the message to, I think it would be good if there were a way for the game to tell you (and also explicitly tell who the message was sent to.
> 
> This would allow you to run `/r` with no arguments and check who you would send the message to. Also after sending a message, we confirm who the message was sent to.

Yeah, we are doing that

It will be like with /m :   
[Sender] => [Receiver] : [message]

src/sgame/sg_cmds.cpp
Grise3: We are doing that. Well are doing it like it do with `/m`

[Sender] => [Target] : [message] 
Like this :

slipher: No, that message is generated client-side for the recipient of the message. What we're saying is that the sender of the message should see something.

Grise3: I don't say that I will do exactly that.
I will do something like this (It will display something like this)

src/sgame/sg_cmds.cpp
Grise3: done

src/sgame/sg_cmds.cpp
Grise3: done

src/sgame/sg_cmds.cpp
Grise3: that's to say ?

src/sgame/sg_cmds.cpp
Grise3: Is it really necessary? It works very well like this, no need to add more code on this side

src/sgame/sg_cmds.cpp
slipher: The suggestion being made here is to show who lastPrivateMessageSender is. In my opinion you don't have to do it

src/sgame/sg_cmds.cpp
slipher: The benefit of this change would be that if the last sender has disconnected from the server, you would get the "No one to reply to" message, instead of nothing happening.

src/sgame/sg_cmds.cpp
Grise3: Done

src/sgame/sg_cmds.cpp
```diff
@@ -4441,6 +4450,59 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
...
+		ADMP( "\"" N_("No one to reply to.") "\"" );
+		ADMP( va( "%s %s", QQ(N_("You would have replied to $1$")), g_entities[target].client->pers.netname));
+		return;
+	}
+
+	msg = ConcatArgs(1);
+
+	if (G_SayTo(ent, &g_entities[target], teamonly ? SAY_TPRIVMSG : SAY_PRIVMSG, msg))
+	{
+		ADMP( va( "%s %s %s", QQ(N_("You have responded to $1$^* : ^2$2$ ")), g_entities[target].client->pers.netname, Quote(msg)));
```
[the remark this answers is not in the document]
Grise3: I think it is better to repeat the message back to the sender :)
like with /m in fact :  
`Private message: [the message]`

src/sgame/sg_cmds.cpp
Grise3: oh. I don't think that it is a good idea. Well, this is the usage. It just say "how to use the command"

src/sgame/sg_cmds.cpp
```diff
@@ -4441,6 +4450,59 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
...
+	{
+		ADMP( "\"" N_("More than one possible recipient, refusing to send.") "\"" );
+        return;
+	}
+
+	target = ent->client->pers.lastPrivateMessageSender;
+	if (target == -1 || !g_entities[target].client)
+	{
+		ADMP( "\"" N_("No one to reply to.") "\"" );
+		ADMP( va( "%s %s", QQ(N_("You would have replied to $1$")), g_entities[target].client->pers.netname));
```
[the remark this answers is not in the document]
Grise3: done

src/sgame/sg_cmds.cpp
```diff
@@ -4441,6 +4450,59 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
...
+
+Reply by private message to the last private message sender
+=================
+*/
+
+void Cmd_ReplyPrivateMessage_f(gentity_t *ent)
+{
+	char *msg;
+	int target;
+	bool teamonly = false;
```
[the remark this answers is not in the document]
Grise3: removed

src/sgame/sg_cmds.cpp
Grise3: done

src/sgame/sg_cmds.cpp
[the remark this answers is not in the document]
slipher: The "You would have replied to" message is still there.

src/sgame/sg_cmds.cpp
```diff
@@ -4441,6 +4450,56 @@ void Cmd_PrivateMessage_f( gentity_t *ent )
...
+	}
+
+	msg = ConcatArgs(1);
+	target = ent->client->pers.lastPrivateMessageSender;
+	if (target == -1 || !G_SayTo(ent, &g_entities[target], SAY_PRIVMSG, msg))
+    {
+        ADMP("\"" N_("No one to reply to.") "\"");
+        return;
+    }
+	if (G_SayTo(ent, &g_entities[target], SAY_PRIVMSG, msg))
```
slipher, line 4493: G_SayTo was already done inside the `if` expression 5 lines up so it doesn't need to be done again.

Also the preceding lines are indented with spaces. They should use tab to match the rest of the file

slipher (commented):
It's getting close to being ready!

Grise3:
> It's getting close to being ready!

I hope so ^^

I'll try to correct the code today.

src/sgame/sg_cmds.cpp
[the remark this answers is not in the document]
Grise3: Eeeh ?? I  will take a look at that today

src/sgame/sg_cmds.cpp
Grise3: it is now out of the "if" and, I replaced spaces by tabs

sweet235:
I will merge this in a day or two if nobody objects.

Grise3:
Ooouuuuééééé!

illwieckz:
> I will merge this in a day or two if nobody objects.

My name is nobody

My name is nobody and I object! So you can merge it!

Grise3:
> > I will merge this in a day or two if nobody objects.
> 
> My name is nobody
> 
> My name is nobody and I object! So you can merge it!

XDD ho no ! Nobody object then we can merge it XD

Grise3:
Ooouuuuééééé ! Thanks!

PR #3093, opened 2024-09-22
motionblur: Disable by default

DolceTriade:
Subjective opinion: It is a net-negative gameplay experience, even on higher settings

### DIFF
--- a/pkg/unvanquished_src.dpkdir/presets/graphics/high.cfg
+++ b/pkg/unvanquished_src.dpkdir/presets/graphics/high.cfg
@@ -21,5 +21,5 @@ seta cg_playerShadows 1
 seta cg_buildableShadows 0
 seta cg_bounceParticles 1
 seta cg_marks on
-seta cg_motionblur 0.05
+seta cg_motionblur 0
 vid_restart
--- a/pkg/unvanquished_src.dpkdir/presets/graphics/ultra.cfg
+++ b/pkg/unvanquished_src.dpkdir/presets/graphics/ultra.cfg
@@ -21,5 +21,5 @@ seta cg_playerShadows 1
 seta cg_buildableShadows 1
 seta cg_bounceParticles 1
 seta cg_marks on
-seta cg_motionblur 0.05
+seta cg_motionblur 0
 vid_restart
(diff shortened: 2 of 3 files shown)


VReaperV (approved):
Vote is there, code/cfg-wise LGTM.

slipher:
References to the cvar should be removed from the preset files. It doesn't make sense to set a variable to the same value in all presets; it should just be controlled independently.

illwieckz:
But we need to provide the user an easy way to reset it to the new default values, re-applying a preset is a good way to do that for all cvars in one go.

I was going to suggest to use `reset` instead, but then I remembered we may do that for many other cvars and then this could be postponed to a later PR.

illwieckz:
Also these days I thought about a kind of new `set` function variant that would reset the cvar _if_ the requested cvar value is the default one, that may even be better.

slipher:
> But we need to provide the user an easy way to reset it to the new default values, re-applying a preset is a good way to do that for all cvars in one go.

If you guys think motion blur is really so bad that we need to make sure it is not turned on due to any pre-existing config files, you could change the cvar name to `cg_motionBlurStrength` or something.

DolceTriade:
> If you guys think motion blur is really so bad that we need to make sure it is not turned on due to any pre-existing config files, you could change the cvar name to cg_motionBlurStrength or something.

This is for new users. If people like it and they haven't disabled it, I assume they are fine with it?

sweet235:
I wonder if we are really the proper target audience for such a vote. I disable anything I can for competetive games.

DolceTriade:
To an extent, we are a proxy for regular users too. If we discover that regular users like this, we should say so here. I'll ask around too.

illwieckz:
Motion blur is currently broken:

unvanquished_2024-10-16_052628_000

necessarily-equal:
I am rather in favor of keeping the motion blur by default for computers that aren't extremely slow. I'd say that the current motion blur effect is slight and only triggers when going very fast, for example with the dragoon pounces.

This is a graphical effect I'd say is quite natural and I hope somewhat helps to make those fast movements look less jumpy for people that don't have very high framerate screens. I imagine that'd be why if you look at a movie, they also appear to have blurry images when looking on paused frames.

necessarily-equal:
Of course, this is something where asking in-game is also a good idea :)

Gireen:
Motionblur doesn't seem to be that popular among gamers in general.
https://www.pcgamer.com/why-people-hate-motion-blur-in-videogames/

It's a shame to deactivate a visually nice effect, but it's always the first thing I turn off cause its annoying for me.

PR #3094, opened 2024-09-23
Update lighting cvars in menus, presets

slipher:


illwieckz:
There is also this (that I should update, the engine code changed a lot since):

- https://github.com/Unvanquished/Unvanquished/pull/2894

Issue #2124, opened 2022-08-30
Inconsistent behaviour between layout callvotes and map callvotes/map/devmap, resulting in map crashes


VReaperV:
If a layout has a large number of buildings (~250), trying to do `/callvote map map layout`, `/map map layout` or `/devmap map layout` results in the game going back to the main menu. The log shows a warning "Warn: CL_ParseGamestate: bad command byte", used to be a warning about there being too many entities for the VM in one of the previous versions.

However, using `/callvote layout layout` after the map was already loaded results in the same layout loading normally.

An example layout showing this behaviour attached for map spacetracks.

[humanpve1.zip](https://github.com/Unvanquished/Unvanquished/files/9454510/humanpve1.zip)

DolceTriade:
We'll take a look. Thanks for reporting. Which map is this layout for?

VReaperV:
`spacetracks`, one of the default maps (`map-spacetracks_1.1.2.dpk`).

bmorel:
Directly related to #573

sweet235:
I suspect the following happens when we try this on a public server: the server continues to run, but every connected clients crashes. This can be tested on bot testing sk7 right now, but may put the server into an unusable state.

sweet235:
> the server continues to run, but every connected clients crashes

This is indeed the case.

cu-kai:
[spacetracks.zip](https://github.com/Unvanquished/Unvanquished/files/11435266/spacetracks.zip)

I have attached a zip containing layouts for spacetracks guaranteed to cause the crash as of 0.54 (use alienpve1-{1|2|3} when testing this)

PR #3098, opened 2024-09-25
Don't transmit some fake entities to clients

slipher:
Don't waste bandwidth on entities that don't do anything for clients.

Hearing about layouts which cause a buffer overflow when they are the built-in layout but not when loaded from a layout file (or vice versa, don't remember which way it was), it occurs to me that this might be related to the fake entities used to spawn buildables. They are known to spawn at different times when a layout file is used, resulting in different entity numbering. So you guys can give this a shot and see if it allows bigger layouts to load

### DIFF
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -1867,6 +1867,7 @@ static void SpawnBuildableThink( gentity_t *ent )
 void G_SpawnBuildable( gentity_t *ent, buildable_t buildable )
 {
 	ent->s.modelindex = buildable;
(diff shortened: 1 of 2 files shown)


DolceTriade:
It does not allow the layout in question to load.

slipher:
I found the issue describing the discrepancy: https://github.com/Unvanquished/Unvanquished/issues/2124. I had it wrong in the OP, the difference caused by whether the layout is loaded following a map_restart or not, not whether it uses a layout file. This is because a new entity baseline is not created after map_restart.

Anyway it's still a good idea not to send useless stuff, regardless of whether it fixes a bug.

PR #3100, opened 2024-09-29
Clearer instructions on how to install dependencies for the repository.

jasonguf:
Hello, in this pull request, I added clearer instructions regarding the dependencies for the repository, and how they can be installed if a user has not already had them installed. 

The commands needed to install each specific dependency are also given.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,26 @@ and a C++14 compiler. The following are actively supported:
 `clang` ≥ 11,
 Visual Studio/MSVC (at least Visual Studio 2019).
 
+For Windows:
+
+Git: Download and install from git-scm.com.
+
+CMake: Download and install from cmake.org.
+
+Python: Download and install from python.org.
+
+python-yaml and python-jinja: Open the command prompt terminal and run the following:
+`pip install pyyaml jinja2`
+
+C++14 Compiler:
+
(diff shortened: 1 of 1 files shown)


slipher:
On Windows, ncurses is not used. We use the submodule at `daemon/libs/pdcursesmod` instead.

Regarding zlib, GMP, Nettle, libcurl, SDL2, PNG, JPEG, Webp, Freetype, Lua, OpenAL, Ogg, Vorbis, and Opus: the common way of getting these for Windows is to use the packages which are automatically downloaded by the build script to the location `daemon/external_deps`. The README is not meant to be encyclopedic, so instructions on alternative methods of obtaining the dependencies would be better suited for the wiki, for example at https://wiki.unvanquished.net/wiki/Compiling_the_source

illwieckz:
We added in the readme a link to that wiki page with the instructions, which has the samepurpose of giving the instructions:

- https://github.com/Unvanquished/Unvanquished/pull/3372

So this PR has been obsoleted by another, let's close this.

illwieckz:
Humm, I confused this PR with:

- https://github.com/Unvanquished/Unvanquished/pull/3368

Some instructions from this can still be useful, but is better written in the wiki.

I created a special entry asking to report this to the wiki instead:

- https://github.com/Unvanquished/Unvanquished/issues/3420

illwieckz:
@jasonguf is you need a wiki account please tell me.

PR #3103, opened 2024-10-05
Fix muzzle flash dlight without flash model

slipher:
Muzzle flash dynamic lights (for weapon firing effects) didn't work if there was no muzzle flash model configured. Before now, asset designers have apparently worked around this situation by putting a flash model with an invisible texture when they want a dlight but not a model.


illwieckz:
I haven't seen any issue when I tested this.

illwieckz (approved):
LGTM.

PR #3106, opened 2024-10-12
trace: partially revert using halfDiagonal in widetrace

DolceTriade:
90202c2 seems to have introduced the bug where you can't hit the medi below you or psaw above your head


slipher:
So the original code has a bug where the CONTENTS_SOLID line-of-sight trace is too long. This could make it think you hit a map surface slightly behind your target, instead of the target you should have hit. This indeed has the potential to cause problems with hitting the medistation. (But it wouldn't change anything with the case of a player standing on your head.)

The proposed change makes the line-of-sight trace too short instead. This could let you hit things through walls. I'll propose a more correct fix in a minute.

slipher:
See #3107

PR #3107, opened 2024-10-12
Fix a G_WideTrace bug (tyrant can't hit medi)

slipher:
It had a bug that the line trace against CONTENTS_SOLID was too long. This could result in choosing the world as the entity when there was actually a closer CONTENTS_BODY entity in front of it.

Thanks to Ishq for finding where in the code the bug is.


DolceTriade:
Can confirm that this fixes the medistation bug.

DolceTriade:
LGTM

PR #3110, opened 2024-10-14
Update graphics settings

VReaperV:
Add settings for static reflections and material system.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
@@ -308,6 +308,29 @@
 				</buttonbar>
 			</panel>
 
+			<tab><translate>Renderer</translate></tab>
+			<panel>
+				<h2><translate>Material system</translate></h2>
+				<row>
+					<h3><translate>Material system</translate></h3>
+					<input cvar="r_materialSystem" type="checkbox" />
+					<p><translate>Use a new system that significantly improves performance. Some drivers may fail with this on (e. g. Mesa and AMD proprietary drivers).</translate>&nbsp;<translate>Experimental.</translate></p>
+				</row>
+				<row>
+					<h3><translate>GPU frustum culling</translate></h3>
+					<input cvar="r_gpuFrustumCulling" type="checkbox" />
+					<p><translate>Cull surfaces that are out of view.</translate>&nbsp;<translate>Experimental.</translate></p>
+				</row>
+				<row>
+					<h3><translate>GPU occlusion culling</translate></h3>
+					<input cvar="r_gpuOcclusionCulling" type="checkbox" />
+					<p><translate>Cull hidden surfaces.</translate>&nbsp;<translate>Experimental.</translate></p>
+				</row>
+				<buttonbar>
+					<button onclick='Cmd.exec("vid_restart")'><translate>Apply now</translate></button>
+				</buttonbar>
+			</panel>
+
 			<tab><translate>Gameplay</translate></tab>
 			<panel>
 				<h2><translate>Building markers</translate></h2>
--- a/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
@@ -16,7 +16,7 @@
 			}
 		</style>
 	</head>
-	<body id="options_graphics" template="window" class="window-options" style="margin-top: 30%; width: 46em;" onload='Events.pushcmd("buildDS resolutions")'>
+	<body id="options_graphics" template="window" class="window-options" style="margin-top: 25%; width: 46em;" onload='Events.pushcmd("buildDS resolutions")'>
 		<h1><translate>Graphics</translate></h1>
 
 		<tabset>
--- a/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
@@ -162,6 +162,13 @@
 					<h3><translate>High-precision rendering</translate></h3>
 					<p><translate>Use high-precision pixel buffers, which reduce color banding.</translate></p>
 				</row>
+
+				<row>
+					<h3><translate>Use bindless textures if possible</translate></h3>
+					<input cvar="r_preferBindlessTextures" type="checkbox" />
+					<p><translate>Use bindless textures if supported by hardware/driver. This might improve performance.</translate>&nbsp;<translate>Experimental.</translate></p>
+				</row>
+
 				<h2><translate>Mip-mapping</translate></h2>
 				<p><translate>Mip-mapping makes distant surfaces look better by using lower-resolution textures on them. This generally does not affect performance.</translate></p>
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
```diff
@@ -216,6 +216,42 @@
 					<input cvar="r_physicalMapping" type="checkbox" />
 					<p><translate>Material appearance based on physical approximations.</translate></p>
 				</row>
+				
+				<h2><translate>Reflections</translate></h2>
+				<row>
+					<h3><translate>Static reflections</translate></h3>
+					<input cvar="r_reflectionMapping" type="checkbox" />
+					<p><translate>Adds static cubemap reflections on supported surfaces. Requires normal, specular and deluxe mapping. Currently unsupported with physical mapping.</translate></p>
```
illwieckz, line 224: Suggestion:

```
<p><translate>Adds static cubemap reflections on supported surfaces. Requires normal, specular and deluxe mapping. Currently unsupported with physical mapping.</translate>&nbsp;<translate>Experimental.</translate></p>
```

This is to ad an (commonly translated) “Experimental” mention, like we do with SSAO.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
```diff
@@ -260,6 +296,29 @@
 				</buttonbar>
 			</panel>
 
+			<tab><translate>Experimental</translate></tab>
```
illwieckz, line 299: Maybe rename the tab to “Renderer”? Experimental looks too generic, there are other experimental options not being listed there (and it's better this way), and those options will be kept once the material system is not considered experimental anymore (we can't require bindless texture so the alternate code will remain).

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
```diff
@@ -260,6 +296,29 @@
...
+			<panel>
+				<h2><translate>Material system</translate></h2>
+				<row>
+					<h3><translate>Material system</translate></h3>
+					<input cvar="r_materialSystem" type="checkbox" />
+					<p><translate>Use a new system that significantly improves performance. Some drivers may fail with this on (e. g. Mesa radeonsi and AMD proprietary drivers).</translate></p>
```
illwieckz, line 305: I would append some `&nbsp;<translate>Experimental.</translate>` string on all those options instead.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
```diff
@@ -260,6 +296,29 @@
...
+				</row>
+				<row>
+					<h3><translate>GPU frustum culling</translate></h3>
+					<input cvar="r_gpuFrustumCulling" type="checkbox" />
+					<p><translate>Cull surfaces that are out of view.</translate></p>
```
illwieckz, line 310: Maybe `Use the GPU to cull surfaces that are out of view.` and `Use the GPU to cull hidden surfaces.`?

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
VReaperV: Done now.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
VReaperV: Hmm, I guess that makes sense. Changed to Renderer now.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
VReaperV: I added it to the first one, adding to all seems redundant though?

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
VReaperV: I  could do that, but the options do both say "GPU ... culling".

slipher:
https://github.com/DaemonEngine/Daemon/pull/1356 will be needed for the material checkbox to actually do anything, since bindless textures are disabled by default.

VReaperV:
I've added an option for `r_preferBindlessTextures` as well.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
illwieckz: Ah yes, that's fine then.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
illwieckz: I prefer for everything experimental be marked as is.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
```diff
@@ -260,6 +302,29 @@
 				</buttonbar>
 			</panel>
 
+			<tab><translate>Renderer</translate></tab>
+			<panel>
+				<h2><translate>Material system</translate></h2>
+				<row>
+					<h3><translate>Material system</translate></h3>
+					<input cvar="r_materialSystem" type="checkbox" />
+					<p><translate>Use a new system that significantly improves performance. Some drivers may fail with this on (e. g. Mesa radeonsi and AMD proprietary drivers).</translate>&nbsp;<translate>Experimental.</translate></p>
```
illwieckz, line 311: You can just say “Mesa” instead of “Mesa radeonsi”. I was able to test the code on Mesa zink too (OpenGL implementation over Vulkan), and the culling code fails like the radeonsi one, and the Mesa code where the breakage occurs is in some shared gallium code.

So material system works on Mesa with the latest fixes after our previous reports, but culling hits some other bugs in the common Mesa code as well.

VReaperV: Ah, I see. Is it only the occlusion culling that crashes on Mesa?

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
VReaperV: Sure, I updated it now.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
VReaperV: Alright, I changed this to just say "Mesa".

illwieckz:
- rebased and fixed merge conflict
- fixed a typo (`Hardwire` instead of `Hardware`)
- moved reflection options to their own table to avoid having the window going out of screen
- simplified `Use the bindless texture extension if supported by hardware/driver and not known to crash on the detected hardware/driver.` string into `Use bindless textures if supported by hardware/driver. This might improve performance.` to not make the window too long, anyway a hardware/driver known to crash is like a hardware/driver not supporting the feature, being for good or wrong reasons.
- moved the option window a bit up to better center it.

PR #3118, opened 2024-10-16
Update language selection

illwieckz:
- options: also enable non-brasilian portuguese
- translation: don't package not enough complete translations that can't be selected in options


slipher:
IIRC the guy who translated Brazilian Portuguese said that the other one was in bad enough shape that it would be preferable to use Brazilian for everyone. To do that we'd need to exclude the un-suffixed `pt` version since that would would always be selected by default otherwise.

illwieckz:
There is only one author for Brasilian portuguese, and 5 for the Portuguese, so maybe real portuguese people actually contributed to the non-brasilian one.

PR #3119, opened 2024-10-17
Add overbright clamp option to gfx menu

slipher:
Needs https://github.com/DaemonEngine/Daemon/pull/1364 or https://github.com/DaemonEngine/Daemon/pull/1350.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
@@ -100,6 +100,11 @@
 					<h3><translate>Lighting system</translate></h3>
 				</row>
 				<row>
+					<h3><translate>Brightness clamping</translate></h3>
+					<input cvar="r_overbrightDefaultClamp" type="checkbox" />
+					<p><translate>Limit max brightness of map lighting of world surfaces and models (better compat with majority of legacy maps)</translate></p>
+				</row>
+				<row>
 					<h3><translate>Light styles</translate></h3>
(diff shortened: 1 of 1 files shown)


pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
```diff
@@ -99,6 +99,11 @@
 					</select>
 					<h3><translate>Lighting system</translate></h3>
 				</row>
+				<row>
+					<h3><translate>Brightness clamping</translate></h3>
+					<input cvar="r_overbrightDefaultClamp" type="checkbox" />
+					<p><translate>Limit max brightness of map lighting of world surfaces and models (better compat with majority of legacy maps)</translate></p>
```
illwieckz, line 105: It's better to not use abbreviations to not make the translation harder to do. Anyway I recommend to use instead something like “`Old brightness limit of map lighting of world surfaces and models (the way legacy maps looked before)`” as it is not really a matter of data compatibility but more a matter of human habituation.

illwieckz:
LGTM % the comment.

illwieckz:
Actually I can commit the suggested easier-to-translate string myself. Let's merge this and start a build of RC3

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
VReaperV: Something like "LDR lighting" might be more concise.

PR #3127, opened 2024-10-17
Reduce reflection cubemap size options

VReaperV:
The previous options were resulting in ~1-2Gb of image data saved for even smaller maps like plat23, per map, when using moderate preset and default spacing.

Prevent crashes and using up tons of disk space.


slipher:
LGTM

PR #3128, opened 2024-10-18
Fix first dropdown selection ignored with custom value

### LINKED ISSUE #3115: [0.55RC2] Switching language in UI doesn't work
As a first step, I tried switching to French in the UI.

image

While I could select the option in the list, clicking on the validation button didn't do anything (no feedback at all). Even restarting the game didn't register the change.

slipher:
Fix #3115


src/cgame/rocket/rocketFormControlSelect.h
```diff
@@ -119,22 +117,24 @@ class CvarElementFormControlSelect : public Rml::ElementFormControlSelect, publi
...
 		{
 			Rml::Element* e = GetOption(i);
 			Rml::String value = e->GetAttribute<Rml::String>( "value", Rml::String() );
 			if ( value == cvarValue )
 			{
-				// If the selection is already correct, do not force set the cvar.
-				ignore_value_change = GetSelection() != i;
+				ignore_value_change = true;
 				SetSelection( i );
+				ignore_value_change = false;
```
DolceTriade, line 129: why we need to do this? Does setselection trigger an event? maybe comment this?

src/cgame/rocket/rocketFormControlSelect.h
```diff
@@ -119,22 +117,24 @@ class CvarElementFormControlSelect : public Rml::ElementFormControlSelect, publi
...
 				return;
 			}
 		}
 		// No matches...Let's add an option for our current value.
 		Add( Str::Format( "Custom Value (%s)", cvarValue ), cvarValue, 0, false );
-		SetSelection( 0 );
 		ignore_value_change = true;
+		SetSelection( 0 );
+		ignore_value_change = false;
```
DolceTriade, line 137: Sounds like this should be a helper function if we always have to do this

src/cgame/rocket/rocketFormControlSelect.h
slipher: Added comment on ignore_value_change.

src/cgame/rocket/rocketFormControlSelect.h
slipher: Eh, we only do it twice, I like the rule of three :)

DolceTriade (approved):

PR #2960, opened 2024-03-30
admin: Add changedevmap command

DolceTriade:
This allows admins to start devmap. useful for group testing when someone doesn't have rcon/console access.

PR #3134, opened 2024-10-20
fix: `changemap` command not found

cu-kai:
In some cases, the `changemap` command may not be found due to incorrect ordering of the array in sg_admin.cpp

This bug was found today on Bunker after the update.

I assume this wasn't found while testing #2960 as the server used to test it probably didn't have custom admin commands.

Thanks @sweet235 for spotting the error.


sweet235:
Sorting stuff is my hobby. LGTM

DolceTriade:
Sorry thanks for fixing. LGTM

PR #3136, opened 2024-10-20
fix: dead/powered down barricade is not predicted

cu-kai:
take 2

### DIFF
--- a/src/cgame/cg_predict.cpp
+++ b/src/cgame/cg_predict.cpp
@@ -147,7 +147,8 @@ static void CG_ClipMoveToEntities( const vec3_t start, const vec3_t mins,
 			if ( ent->eType == entityType_t::ET_BUILDABLE )
 			{
 				BG_BuildableBoundingBox( ent->modelindex, bmins, bmaxs );
-				if ( ent->modelindex == BA_A_BARRICADE && ent->torsoAnim == BANIM_IDLE_UNPOWERED )
+				if ( ent->modelindex == BA_A_BARRICADE && ( ent->torsoAnim != BANIM_IDLE1 || !( ent->eFlags & EF_B_SPAWNED ) ) )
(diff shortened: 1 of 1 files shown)


src/cgame/cg_predict.cpp
```diff
@@ -147,7 +147,7 @@ static void CG_ClipMoveToEntities( const vec3_t start, const vec3_t mins,
 			if ( ent->eType == entityType_t::ET_BUILDABLE )
 			{
 				BG_BuildableBoundingBox( ent->modelindex, bmins, bmaxs );
-				if ( ent->modelindex == BA_A_BARRICADE && ent->torsoAnim == BANIM_IDLE_UNPOWERED )
+				if ( ent->modelindex == BA_A_BARRICADE && ( ent->torsoAnim != BANIM_IDLE1 || !( ent->eFlags & EF_B_SPAWNED ) ) )
```
DolceTriade, line 150: Looking closer at my original fix, I wonder if we could just do `ent->legsAnim == BANIM_POWERDOWN` as the condition and catch everything...

DolceTriade: Seems to work ok in my tests?

DolceTriade:
Ok, before doing some extensive redesign, I think it's fair to say this improves the situation. Perhaps it is fine to submit as is to improve the situation and then we can implement the full design after?

slipher:
I'm not demanding the use of any specific approach. But if anyone is interested in trying the old version with bboxes calculated by the server, they can revert 7f00ec84a911c8b0607025aeefe53be60b95f3d9.

src/cgame/cg_predict.cpp
DolceTriade: eh, it's meh. I think add a note here saying `// TODO: This is not perfect and can be improved.`

DolceTriade:
LGTM

PR #3145, opened 2024-10-22
Some changes in the script to generate a benchmark demo

illwieckz:
- move scripts to `tools/`;
- add some option to make it easy to deal with when running `daemon` instead of `daemon-tty` (no full screen, no mouse capure, etc.),
- disable bot build (make it more comparable to older ones, prevent the bots to build buildables that will stuck bots);
- also rename the default name and write the name right from the start.


illwieckz:
I also disabled bot build in the demo, not only it makes it more comparable with other demos, but when I tried to record it for 0.55 at first a bot was building a telenode right at the top of the ramp that goes to human base and then all the bot team was stuck, making the demo only benchmarking what a bot could be seen while being stuck, instead of recording some walks around and fights…

tools/record-benchmark-demo/record-benchmark-demo
```diff
@@ -41,12 +41,16 @@ cp -a "${script_dir}/record-benchmark-demo.cfg" "${config_dir}"
 if "${daemon_path}" \
 	-homepath "${temp_home_path}" \
 	-set name 'benchmark' \
+	-set r_fullscreen 0 \
+	-set r_customWidth 640 \
```
slipher, line 45: Why 640x480? That's not resolution someone would usually play with

illwieckz: This is not a script to play a game, it is just meant to record a demo, and optionally watch the demo while being recorded. The only reason why it may be useful to look at the demo being recorded is to notice the followed bot is stuck and then not much actions are recorded, saving time by not requiring a replay of the demo to spot the wrongness. As the one using that script it I can say it's more convenient to have this being a tiny window somewhere on the screen. The default behaviour is to not display anything at all anyway. Rendering the game is mostly used for when modifying the demo script and testing it: recording the released benchmark demo file better be done without any risk to input anything.

illwieckz: Especially, when debugging the demo script and looking at the demo being recorded, it is recommended to NOT have the game windows focused (to behave exactly like `daemon-tty`), so the most tiny the windows is, the easier it is.

slipher: Well I assume the benchmark is supposed to give an idea of the performance you would get while actually playing the game. So setting to a tiny resolution fails to reproduce those conditions.

Since you are interested in not seeing the window, I opened a PR for a commit I was sitting on: https://github.com/DaemonEngine/Daemon/pull/1399.

slipher: Sorry total reading comprehension failure on my part. I thought it is for taking the benchmark, not recording a demo.

slipher:
LGTM

tools/record-benchmark-demo/record-benchmark-demo
illwieckz: Yes it's for recording the demo. Running the benchmark just requires to `set demo.timedemo on` and then `demo_play <benchmark_name>` once the demo is donwloaded into `demos/`.

PR #3169, opened 2024-10-30
Add a tool to mark incomplete translation files to be ignored

illwieckz:
The added `ignore_incomplete.sh` script updates the urcheon action list with ignored commands for incomplete translation files.

The current threshold is that less than 50% translation completion is not considered complete enough and then related translation files will not be packaged.

While updating the action list, the tool prints statistics to the user.

This is a new implementation of #3075:

- https://github.com/Unvanquished/Unvanquished/pull/3075

But instead of writing a config file that should be read later by the game (and requires a special implementation in game code), this tells urcheon what to do and relies on already existing features of urcheon.

The downside compared to an in-game implementation is that it's not possible to implement a cvar to enable access to incomplete languages, as incomplete files are not shipped at all to begin with, but this is good enough.

The PR also runs `generate.sh` and then updates the various translation things including that action list.


illwieckz:
The commit updating the strings and the list was merged separately as this is just fully generated and strings have been updated in the mean time so we better push that to weblate, it can be found there: https://github.com/Unvanquished/Unvanquished/commit/337a074f85a826af944525948eb9df5ecfcd61c5

tools/translation/ignore_incomplete.sh
```diff
@@ -0,0 +1,69 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+# ===========================================================================
+
+set -u -e -o pipefail
```
slipher, line 27: Does the script only add things to the ignore list, or can it also un-ignore them?

slipher (commented):
Would it be ideal to have the ignored languages list be the same as our manually curated language menu? Or is there any reason to want them different?

illwieckz:
> Would it be ideal to have the ignored languages list be the same as our manually curated language menu? Or is there any reason to want them different?

It would be ideal to have them exactly the same (well, it would be ideal that the curated list to be the exact inverse of the ignore list).

This focus on making sure incomplete enough translations are not selected by the engine when the user starts the game, it doesn't mean complete enough translations can all be selected in the menus.

To solve the second need, we may migrate from a manually curated language menu to a manually curated list of language names, and automatically build up the language menu from the list of available files, picking the language name from the curated list of language names, and falling back on the code name if language name isn't listed yet.

tools/translation/ignore_incomplete.sh
illwieckz: It also un-ignore them:

1. Read every lines from the action file that do not start with `ignore translation/`:

```sh
action_list="$(grep -v -E '^ignore translation/' "${action_file}")"
```

2. Rewrite the action file without any `ignore translation/*` lines:

```sh
printf '%s\n' "${action_list}" > "${action_file}"
```

3. Add (or readd) to the action file a new `ignore` command for a translation file:

```sh
printf 'ignore translation/%s\n' "${po_file}" >> "${action_file}"
```

slipher: We concluded that some languages such as Arabic may not be ready even at 100%. Maybe it shouldn't remove things?

illwieckz: Right! With luck urcheon has to way to ignore things, an `ignore.txt` file that acts a bit like `.gitignore` (it supports some globbing), and the action file. The `ignore.txt` is purposed more for things like ignoring a whole directory, that's why the script updates the action file instead, which is only purposed to have a single command listed per single file. We can add the arabic file to the `ignore.txt` and then keep the script as simple as it is today.

illwieckz: Now done. I also forgot that the `ignore.txt` file doesn't support complete paths (that's why the action list is better for ignoring a precise file), so I marked any path of `ar.po` to not be packaged.

slipher:
LGTM

PR #3177, opened 2024-11-02
Enable SSAO in presets, don't mark it as experimental anymore

illwieckz:
Once Fog is not behind SSAO:

- https://github.com/DaemonEngine/Daemon/pull/1408

We can mark this feature as non-experimental and enable it in presets.


slipher:
LGTM

Drive-by suggestion for a better description: how about "Darken recesses"?

illwieckz:
> Drive-by suggestion for a better description: how about "Darken recesses"?

I didn't know the “recesses” word yet. 😁️ But I suggest to postpone such reword after 0.55.1 to not break translated strings at the time we advertise the feature.

illwieckz:
I merge it sooner so I can work on building 0.55.1 sooner.

PR #3178, opened 2024-11-03
Reword SSAO description in menu

slipher:


illwieckz:
I merged the extra `Add missing header to sources list` commit.

PR #3184, opened 2024-11-09
Fix dynamic lights intensity, cleanup

VReaperV:
Engine-side pr: https://github.com/DaemonEngine/Daemon/pull/1425

NUKE `trap_R_AddAdditiveLightToScene()`, use `trap_R_AddLightToScene()` instead of it.

### DIFF
--- a/src/cgame/cg_ents.cpp
+++ b/src/cgame/cg_ents.cpp
@@ -305,7 +305,7 @@ static void CG_EntityEffects( centity_t *cent )
 		b = ( cl >> 16 ) & 255;
 		i = ( ( cl >> 24 ) & 255 ) * 4;
 
-		trap_R_AddAdditiveLightToScene( cent->lerpOrigin, i, r, g, b );
+		trap_R_AddLightToScene( cent->lerpOrigin, i, 1.0, r, g, b, 0, 0 );
 	}
 
 	if ( cg.time > cent->muzzleTSDeathTime && CG_IsTrailSystemValid( &cent->muzzleTS ) )
--- a/src/cgame/cg_ents.cpp
+++ b/src/cgame/cg_ents.cpp
@@ -296,16 +296,13 @@ static void CG_EntityEffects( centity_t *cent )
 	// constant light glow
 	if ( cent->currentState.constantLight && ( cent->currentState.eType == entityType_t::ET_MOVER || cent->currentState.eType == entityType_t::ET_MODELDOOR ) )
 	{
-		int cl;
-		int i, r, g, b;
+		const int cl = cent->currentState.constantLight;
+		const float r = ( cl & 255 ) / 255.0;
+		const float g = ( ( cl >> 8 ) & 255 ) / 255.0;
+		const float b = ( ( cl >> 16 ) & 255 ) / 255.0;
+		const float radius = ( ( cl >> 24 ) & 255 ) * 4.0;
 
-		cl = cent->currentState.constantLight;
-		r = cl & 255;
-		g = ( cl >> 8 ) & 255;
-		b = ( cl >> 16 ) & 255;
-		i = ( ( cl >> 24 ) & 255 ) * 4;
-
-		trap_R_AddLightToScene( cent->lerpOrigin, i, 1.0, r, g, b, 0 );
+		trap_R_AddLightToScene( cent->lerpOrigin, radius, 1, r, g, b, 0 );
 	}
 
 	if ( cg.time > cent->muzzleTSDeathTime && CG_IsTrailSystemValid( &cent->muzzleTS ) )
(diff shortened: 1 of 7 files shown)


slipher:
The intensity fix doesn't really have anything to do with the API changes, right? The CG_EntityEffects dividing by 255 thing could be done just as easily on master if we wanted to. Not that it really matters since it's a rarely used feature.

VReaperV:
Yeah, I suppose it could be moved into a different pr, though to really fix it the engine-side nuking of random light scaling (8x in this case IIRC) is also required.

slipher:
I loaded hq-beta28 in Tremfusion to try to get a reference for how bright it is supposed to be, but it doesn't render a light at all.

slipher:
If entity lights never worked in Tremulous, the few maps we have with it are probably just the result of mappers trying larger and larger values hoping something will appear until they give up. So you might as well go ahead and divide by 255 since there is no compatibility concern. (Or rather, the smaller the intensity the more compatible it is.)

I think the working dynamic map lights in Tremulous were made using particle systems.

VReaperV:
Looks like RMLUI cmake is breaking CI for some reason.

VReaperV:
It works fine on my end...

src/cgame/cg_ents.cpp
```diff
@@ -296,16 +296,15 @@ static void CG_EntityEffects( centity_t *cent )
...
-		trap_R_AddLightToScene( cent->lerpOrigin, i, 1.0, r, g, b, 0 );
+		const int cl = cent->currentState.constantLight;
+		const float r = ( cl & 255 ) / 255.0;
+		const float g = ( ( cl >> 8 ) & 255 ) / 255.0;
+		const float b = ( ( cl >> 16 ) & 255 ) / 255.0;
+		const float radius = ( ( cl >> 24 ) & 255 ) * 4.0;
+
+		/* Set intensity to 255.0 here because maps were built with old erroneous code in mind,
+		that wasn't dividing the values after quantising them */
+		trap_R_AddLightToScene( cent->lerpOrigin, radius, 255.0, r, g, b, 0 );
```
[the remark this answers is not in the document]
VReaperV: Originally I thought of settign the intensity somewhere in the middle to have normal looking light brightness, but yeah, apparently it doesn't show up in tremulous, so I've removed this multiplication part now.

VReaperV:
Now it's working again. Magical things happen with CI sometimes.

DolceTriade (approved):
tested ingame and feels good

src/cgame/cg_ents.cpp
```diff
@@ -296,16 +296,13 @@ static void CG_EntityEffects( centity_t *cent )
 	// constant light glow
 	if ( cent->currentState.constantLight && ( cent->currentState.eType == entityType_t::ET_MOVER || cent->currentState.eType == entityType_t::ET_MODELDOOR ) )
 	{
-		int cl;
-		int i, r, g, b;
+		const int cl = cent->currentState.constantLight;
+		const float r = ( cl & 255 ) / 255.0;
+		const float g = ( ( cl >> 8 ) & 255 ) / 255.0;
+		const float b = ( ( cl >> 16 ) & 255 ) / 255.0;
+		const float radius = ( ( cl >> 24 ) & 255 ) * 4.0;
```
slipher, line 303: Why was this changed from intensity to radius? The sgame designates it as intensity; see SP_ConstantLightField.

Probably related: on hqbeta28 I don't see any light being rendered.

slipher (commented):
All of the lights that currently work (particles, trails, muzzle flash) now look correct with the combination of the Unvanquished + Daemon + res-weapons PRs. Only the constantLight thing seems to have an issue.

src/cgame/cg_ents.cpp
VReaperV: Because the original code would call `RE_AddDynamicLightToSceneQ3A()`, which would then call `RE_AddDynamicLightToSceneET()`with `intensity` as the `radius` parameter.

VReaperV:
> Only the constantLight thing seems to have an issue.
> Probably related: on hqbeta28 I don't see any light being rendered.

Because I did this: 
> If entity lights never worked in Tremulous, the few maps we have with it are probably just the result of mappers trying larger and larger values hoping something will appear until they give up. So you might as well go ahead and divide by 255 since there is no compatibility concern. (Or rather, the smaller the intensity the more compatible it is.)

And that light never showed up in Trem, so it's working correctly now.

src/cgame/cg_ents.cpp
slipher: I see. The gamelogic-side call declared it as `trap_R_AddAdditiveLightToScene( const vec3_t org, float intensity, float r, float g, float b )` but then the engine side was different.

I think it would be nice to implement this in some way that the feature is actually usable; the realtime light attached to an entity seems like a useful feature that we ought to have. If the max intensity as 1 we may as well nuke it. That can be a TODO for later though, to scale the intensity + radius in a reasonable way.

slipher:
LGTM for the 3 parts

PR #3185, opened 2024-11-14
bot behavior tree tracing: show name of running action

sweet235:
This is not really neccessary, we already show the line number in the .bt file. But I find myself doing this frequently when I wonder what a bot is up to. Keeping the line numbers in my mind is an impediment to quick debug sessions.

One might object that this PR increases the debug overhead in `AIActionNode_t` from (roughly) a quarter to a third of its size.


slipher:
If you wanted to avoid the extra memory usage, you could search for the function pointer in the action list `AIActions`.

slipher:
LGTM

PR #3190, opened 2024-11-17
human bots fight back when attacked while attacking an alien buildable

sweet235:
Bots will attack an enemy buildable until it is destroyed or they are killed (or sometimes until they want to retreat to heal/buy/build/repair).

Alien bots will dance like crazy while they are attacking a human building. They are not that easy to kill while doing that.

Human bots, on the other hand, stand still and shoot when they are attacking an alien building. That makes them very easy to kill, as they do not fight back.

This PR makes a human bot abort its attack on an alien buildable as soon as it is damaged by an alien player (or bot). It will target the attacker instead.

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -2569,7 +2569,17 @@ void BotPain( gentity_t *self, gentity_t *attacker, int )
 		&& attacker->s.eType == entityType_t::ET_PLAYER
 		&& self->botMind->skillSet[BOT_B_PAIN] )
 	{
-
+		if ( G_Team( self ) == TEAM_HUMANS
+			 && self->botMind->goal.targetsValidEntity()
+			 && G_Team( self->botMind->goal.getTargetedEntity() ) == TEAM_ALIENS
+		     && self->botMind->goal.getTargetedEntity()->s.eType != entityType_t::ET_PLAYER
+		     && attacker->s.eType == entityType_t::ET_PLAYER )
+		{
+			// human bot is attacked by enemy player while busy attacking an
+			// enemy buildable: abort attacking the buildable, fight back instead
+			self->botMind->goal.clear();
+			BotResetEnemyQueue( &self->botMind->enemyQueue );
+		}
 		BotPushEnemy( &self->botMind->enemyQueue, attacker );
 	}
 }


src/sgame/sg_bot_util.cpp
```diff
@@ -2569,7 +2569,17 @@ void BotPain( gentity_t *self, gentity_t *attacker, int )
 		&& attacker->s.eType == entityType_t::ET_PLAYER
 		&& self->botMind->skillSet[BOT_B_PAIN] )
 	{
-
+		if ( G_Team( self ) == TEAM_HUMANS
+			 && self->botMind->goal.targetsValidEntity()
+			 && G_Team( self->botMind->goal.getTargetedEntity() ) == TEAM_ALIENS
+		     && self->botMind->goal.getTargetedEntity()->s.eType != entityType_t::ET_PLAYER
+		     && attacker->s.eType == entityType_t::ET_PLAYER )
+		{
```
slipher, line 2577: I suppose the underlying problem here is that `action fight` doesn't do any prioritization of targets once it has latched onto something. Could we check that the running node is `action fight` to make sure it does not randomly abort other actions? (could use the function pointer or the name string you just added)

sweet235: I'm not sure if looking for `action fight` is the right thing to do. I wrote this code long ago, and i can vaguely remember thinking about looking for `action fight` too. But it feels like an unneccessary complication. If a bot has an enemy player as target, it is guaranteed to be fighting it. What else could it do, using our current actions?

slipher: Technically `action rush` also targets enemy buildables. But I'm not really worried about the standard BTs. This could be annoying if you are trying to script behaviors for a tutorial, a tower defense mode, something like that. If you wanted to program the bots to head straight for an enemy building, it would be impossible because this code keeps deleting the goal.

sweet235: Fair enough. Another method would be to provide more control over how enemies are chosen in the bt files. Including what to do in case of pain.

sweet235:
Another problem with this is `g_bot_reactionTime`, which is 0.5 seconds by default. The bots will not target anything for this time, when using this PR.

PR #3191, opened 2024-11-17
make bots strafe on attack: dragoon, adv dragoon, tyrant

sweet235:
After we merge https://github.com/Unvanquished/Unvanquished/pull/3061, we can make bots using the big alien classes strafe while attacking. The chance of them falling down and losing their invested funds will be greatly reduced.


slipher:
LGTM, didn't test

PR #3192, opened 2024-11-17
CI: bump image macOS-12 -> macOS-14 

slipher:
zlib was claimed to be a dependency of Freetype, but it seems not to be the case. I also tested with a Homebrew-provided Freetype on Mac and it still builds (so it's not like there's special configuration flags in our external_deps flags that sever the dependency).


slipher:
Whatever, this is now a MacOS version bump PR.

sweet235:
LGTM

PR #3194, opened 2024-11-18
human bots buy jetpacks

sweet235:
Make human bots buy jetpacks, but only if they do not want radars instead (as specified by `g_bot_radarRatio`).

This PR does not provide different navcons for jetpack bots, a future PR will address this. The jetpack bots will only use their jetpacks in two cases for now:

- They are fighting a visible enemy
- They are falling too fast

Once the battlesuit is unlocked, bots will prefer it. So the jetpack bots will be rather rare with this PR.

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -873,6 +873,16 @@ static bool TargetInOffmeshAttackRange( gentity_t *self )
 	}
 }
 
+static void BotActivateJetpack( gentity_t *self )
+{
+	if ( BG_InventoryContainsUpgrade( UP_JETPACK, self->client->ps.stats )
+		 && self->client->ps.stats[ STAT_FUEL ] > JETPACK_FUEL_MAX / 4
+		 )
+	{
+		self->botMind->cmdBuffer.upmove = 127;
+	}
+}
+
 // TODO: Move decision making out of these actions and into the rest of the behavior tree
 AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 {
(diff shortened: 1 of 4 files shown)


slipher:
I don't think we have any code for using jetpack in master?

DolceTriade (approved):
seems ok to me.

slipher:
> > I don't think we have any code for using jetpack in master?
> 
> We have to start somewhere, don't we.

The PR title/commit message only described buying jetpacks, not using them, so I didn't realize there is code to use the jetpack as well.

src/sgame/sg_bot_ai.cpp
```diff
@@ -1033,6 +1043,14 @@ AINodeStatus_t BotActionFight( gentity_t *self, AIGenericNode_t *node )
 	// We are human and we either are at fire range, or have
 	// a direct path to goal
 
+	glm::vec3 ownPos = VEC2GLM( self->s.origin );
+	glm::vec3 targetPos = mind->goal.getPos();
+	if ( ownPos.z < targetPos.z + 400 )
+	{
+		// activate the jetpack if we have it, but do not fly too high above the enemy
+		BotActivateJetpack( self );
```
slipher, line 1051: Does this mean the bot will start flying almost every time there is an enemy near?

sweet235: No. It means the bot will start flying if it can see the enemy.

slipher:
OK, so the bots do start flying every time they are trying to fight against a player. That's a funny strategy. Might not combine with some weapons such as painsaw. Seems fine as a novelty only used by a few bots.

Is it intentional that `g_bot_jetpack` is still disabled by default?

LGTM

sweet235:
> Is it intentional that g_bot_jetpack is still disabled by default?

No.

PR #3196, opened 2024-11-19
Don't use a C++ exception when parsing numbers

slipher:
Encountered by illwieckz while testing the Saigo toolchain. It's a bad idea anyway since exceptions can be slow and this case was not even an error.


illwieckz:
It works! ❤️

illwieckz (approved):
LGTM!

DolceTriade (approved):

PR #3198, opened 2024-11-20
Fix download bar being invisible

### LINKED ISSUE #2164: Download bar disappeared on download page
image

On the bottom, we used to have a download bar

VReaperV:
Fixes #2164.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/transition_downloading.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/transition_downloading.rml
@@ -5,7 +5,7 @@
 		<link type="text/rcss" href="loading.rcss" />
 		<style>
 			body {
-				background-color: black;
+				background-color: #00000000;
 			}
 
 			barbox {


VReaperV:
This reveals another bug: the main menu, connecting, downloading, and possibly loading screens are all rendered on top of each other when connecting to a server. I'll add a fix for that later.

pkg/unvanquished_src.dpkdir/ui/transition_downloading.rml
```diff
@@ -5,7 +5,7 @@
 		<link type="text/rcss" href="loading.rcss" />
 		<style>
 			body {
-				background-color: black;
+				background-color: #00000000;
```
slipher, line 8: I think the number of digits is supposed to be divisible by 3? Like `#000` or `#000000`.

slipher: Oh apparently 8 digits means there is an alpha component. So that should make a completely invisible color with alpha = 0!

VReaperV: Yep, it's a hex rgba.

Yes, making it invisible is what I went for. For whatever reason, the stuff under `body id="download"` gets rendered before the `head` part. I tried changing the `z-index` there, but it did not seem to work.

slipher: What do you mean "the head part"?

I don't understand how making the background invisible is supposed to work. Would we see whatever window is drawn below the download one?

VReaperV: Everything between `<head></head>`.

Right now we'd see the connection screen under it. That can be changed to just draw a black background though.

DolceTriade (approved):

pkg/unvanquished_src.dpkdir/ui/transition_downloading.rml
slipher: > Everything between `<head></head>`.

That's just style directives. There isn't any content.

slipher:
I can get it to show up by setting the `z-index` of the `progress` element. It needs to be more than 20 because that's what the `body` element's is set to in `loading.rcss`.

VReaperV:
Closing in favour of #3243.

PR #3199, opened 2024-11-20
Use unity builds for RmlUi

slipher:
I.e. instead of building each source file one at a time, they are concatenated into groups of 8, so that the same headers don't have to be processed so many times. This is about 3x faster on my machine.

Companion: https://github.com/Unvanquished/RmlUi/pull/8

The unity build feature requires CMake 3.16 to work. I think older versions should just ignore the new properties and be fine, but I need to double-check that.

### DIFF
--- a/rmlui.cmake
+++ b/rmlui.cmake
@@ -1,5 +1,8 @@
 # Create libRocket.
 
+option(RMLUI_UNITY_BUILD "Merge RmlUi compilation units for faster build" 1)
+mark_as_advanced(RMLUI_UNITY_BUILD)
+
 set(RMLUI_DIR ${CMAKE_CURRENT_SOURCE_DIR}/libs/RmlUi)
 
 set(Core_HDR_FILES
@@ -559,6 +562,10 @@ set(Lua_SRC_FILES
     ${RMLUI_DIR}/Source/Lua/Vector2i.cpp
 )
 
+set_source_files_properties(
+    ${RMLUI_DIR}/Source/Core/SystemInterface.cpp # includes windows.h
+    PROPERTIES SKIP_UNITY_BUILD_INCLUSION 1)
+
 if (NOT FREETYPE_INCLUDE_DIRS)
     find_package(Freetype REQUIRED)
 endif()
@@ -586,3 +593,15 @@ set_property(TARGET RMLUI_LIB PROPERTY
 set_target_properties(RMLUI_LIB PROPERTIES
     POSITION_INDEPENDENT_CODE ${GAME_PIE}
 )
+
+if (RMLUI_UNITY_BUILD)
+    set_target_properties(RMLUI_LIB PROPERTIES
+        UNITY_BUILD 1
+        # Suppress Visual Studio's "warning MSB8027: Two or more files with the name of XXX.cpp will
+        # produce outputs to the same location. This can lead to an incorrect build result.  The
+        # files involved are C:\YYY\XXX.cpp, C:\ZZZ\XXX.cpp." which seems to be based on an
+        # incorrect heuristic that a .cpp file always has a corresponding object
+        # file with the same name
+        VS_GLOBAL_MSBuildWarningsAsMessages MSB8027
+    )
+endif()


slipher:
I verified that unknown properties don't cause any issue. If you have an older cmake it only means that the build stays the same as it is now.

illwieckz:
I have some questions:

1. Is it some RmlUi feature backported from upstream or some new thing?
2. Will we have to rebase the feature over upstream RmlUi the day we upgrade RmlUi?
3. It it a CMake feature?
4. Does it make debugging harder by intefering with the source layout?
5. Is it possible to not use the feature?
6. Who decided for the “unity” name?

I see the benefit in making the build faster in CI, otherwise I don't really see what this brings, as RmlUi is rarely rebuilt in iterative build.

slipher:
1. It is not a backport.
2. Probably. It's like a 10 minute task.
3. Yes
4. No
5. Sort of. You can set `-DCMAKE_UNITY_BUILD_BATCH_SIZE=1` to make it so that no files are combined, although technically the feature is still being used.
6. I don't know. I believe this term has been used for decades. 



> I see the benefit in making the build faster in CI, otherwise I don't really see what this brings, as RmlUi is rarely rebuilt in iterative build.

It's completely absurd to say making the non-iterative build faster has no benefit.

Remember that if Saigo is adopted, we have to build the whole thing three times when building cgame for distribution!

DolceTriade:
As long as we can disable it easily (maybe make it a top level option) it's fine. If we end up having compiler errors, I assume unity builds will make that terrible.

illwieckz:
> It's completely absurd to say making the non-iterative build faster has no benefit.

I haven't said that it has no benefit, it's just that such situation outside CI is rare so I don't care enough about it to mention it. Doing a full build usually happens when upgrading the distro to the next distro version because that updates the compiler.

slipher:
Getting some annoying spam in Visual Studio now like `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(1091,5): warning MSB8027: Two or more files with the name of ElementForm.cpp will produce outputs to the same location. This can lead to an incorrect build result.  The files involved are C:\unv\Unvanquished\libs\RmlUi\Source\Core\Elements\ElementForm.cpp, C:\unv\Unvanquished\libs\RmlUi\Source\Lua\Elements\ElementForm.cpp.`. I should see if there is a way to prevent that. Or maybe it is just caused by switching from non-unity to unity builds.

slipher:
> As long as we can disable it easily (maybe make it a top level option) it's fine.

Added `RMLUI_UNITY_BUILD` option.



> If we end up having compiler errors, I assume unity builds will make that terrible.

I probably should have clarified that the "concatenation" of files is done via the C preprocessor, with a generated file that `#include`s each .cpp file. So you still get the original source location in the error trace and it's not that bad.



> Getting some annoying spam in Visual Studio now like `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(1091,5): warning MSB8027: Two or more files with the name of ElementForm.cpp will produce outputs to the same location. This can lead to an incorrect build result. The files involved are C:\unv\Unvanquished\libs\RmlUi\Source\Core\Elements\ElementForm.cpp, C:\unv\Unvanquished\libs\RmlUi\Source\Lua\Elements\ElementForm.cpp.`. I should see if there is a way to prevent that.

I managed to suppress this warning.

slipher:
I'll merge this in 2 days if there are no further objections.

illwieckz:
> I probably should have clarified that the "concatenation" of files is done via the C preprocessor, with a generated file that `#include`s each .cpp file. So you still get the original source location in the error trace and it's not that bad.

So I guess it is kind of what HarfBuzz [calls](https://github.com/harfbuzz/harfbuzz/blob/main/BUILD.md) an “_amalgamated_” source.

I'm a bit annoyed by the very confusing “unity” wording. For example the “_Unity crunch build_” or the “_Unity qstat build_” means building binaries from repositories hosted on the GitHub of Unity Software Inc. but now the “_Unity RmlUi build_” means using some CMake feature to build RmlUi…. That looks confusing.

But if that is the reserved name for such implementation in CMake, then LGTM.

illwieckz (approved):

slipher:
"Unity build" is the name commonly used in industry. Probably since before any Unity Software existed.

PR #3200, opened 2024-11-20
Use better math constants

illwieckz:
See:

- https://github.com/DaemonEngine/Daemon/pull/1447


VReaperV:
`angle * π / 180.0` is a degree->radian conversion, and seems to be the case here (although I haven't looked through all usages). So maybe it's better to just replace those with `Math::DegToRad()`.

PR #3202, opened 2024-11-21
Deduplicate bot behavior files

sweet235:
Deduplicate some shared code in bot behavior files. Add a file `heal_humans.bt`, move shared code there. There is no change in function, except for one or two cases where I had forgotten to add a trigger for the medkit.

Add a file `throw_grenade.bt` for the same reason.

Rename three very basic behavior files, for instance `camper.bt` to `example_camper.bt`. Behaviors evolve and are increasingly distributed over several files. There is no way around this, but new users might be overwhelmed by the increasing complexity of behaviors distributed over several files. These users should study the example behaviors first, as these are short and (almost) self contained.


slipher:
LGTM

Maybe behavior trees that are not intended to be used as a top-level tree should also be renamed to indicate this? Like `unstick.bt` -> `subroutine_unstick.bt`.

sweet235:
> Maybe behavior trees that are not intended to be used as a top-level tree should also be renamed to indicate this? Like `unstick.bt` -> `subroutine_unstick.bt`.

I even had a commit for that, but did not push it here, as I could not decide on the prefix yet. `subroutine` seems a little long.

PR #3205, opened 2024-11-23
Deduplicate bot behavior files

sweet235:
Move more shared code into subroutines, no functional change except fixing minor buggy ommisions.

There is almost no duplicated code left.


VReaperV (approved):
LGTM

PR #3207, opened 2024-11-26
ABI update: Use trap_GetPings not playerstate ping

slipher:
Companion to https://github.com/DaemonEngine/Daemon/pull/1450

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -367,6 +367,7 @@ void ScoreboardMessage( gentity_t *ent )
 	string[ 0 ] = 0;
 	stringlength = 0;
 
+	std::vector<int> pings = trap_GetPings();
 	numSorted = level.numConnectedClients;
 
 	for ( i = 0; i < numSorted; i++ )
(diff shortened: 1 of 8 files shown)


illwieckz (approved):
LGTM.

src/sgame/sg_cmds.cpp
```diff
@@ -381,7 +382,7 @@ void ScoreboardMessage( gentity_t *ent )
 		}
 		else
 		{
-			ping = cl->ps.ping < 999 ? cl->ps.ping : 999;
+			ping = pings[ level.sortedClients[ i ] ];
```
VReaperV, line 385: Why is the check for 999 ping dropped here?

src/sgame/sg_main.cpp
```diff
@@ -1827,7 +1829,7 @@ void LogExit( const char *string )
 			continue;
 		}
 
-		ping = cl->ps.ping < 999 ? cl->ps.ping : 999;
+		ping = pings[ level.sortedClients[ i ] ];
```
VReaperV, line 1832: Same here for the 999 ping check.

src/sgame/sg_cmds.cpp
slipher: It's already enforced in SV_CalcPings.

VReaperV (approved):
LGTM

Issue #3208, opened 2024-11-26
The #include directive used in a BT file can crash the game


sweet235:
To reproduce, create a file `bots/foo.bt`:

```
#include "bots/non_existing_file.bt"
```

Then enter the command `/bot add * h 5 foo`.

The game crashes soon after printing this warning:

```
file bots/foo.bt, line 1: file bots/non_existing_file.bt not found
```

There are a few other errors leading to the same outcome, like removing `"bots/non_existing_file.bt"` or just the quotes.

Reading the directives is done in `shared/parse.cpp` by `Parse_ReadToken`, which returns a `bool`. No distinction is made between a syntax error in a directive or, say, the end of the text to parse. Both yield a false result.

In `sgame/sg_bot_parse.cpp`, `CreateTokenList` receives these boolean results, and seems to assume that a false result always indicates the end of the text to parse. This leads to a corrupted behavior tree pointer.

The only fix I can think of right now would be to use a richer set of result values than just true and false.

By the way, notice the comment `TODO: don't crash if there is an unexpected EOF anywhere in the tree` at the function `ReadBehaviorTree`.

sweet235:
One might argue that we could simply NUKE this directive, or at least prohibit its use in BT files. But that would not be a good idea. Currently, our only method to define constants or custom functions (macros with arguments) is `#define`. If we want to share these among several BT files, we need to `#include` them from a shared file.

sweet235:
> The only fix I can think of right now would be to use a richer set of result values than just true and false.

Thinking again, it should suffice to fix the parser to handle EOF correctly. Keeping track of the nesting level and excluding empty input should be enough.

PR #3209, opened 2024-11-26
fix some crashes in the BT parser

sweet235:
Fix https://github.com/Unvanquished/Unvanquished/issues/3208.

Apart from that, fix crashes for the following behavior files:

```
```

```
selector {
```

```
selector { action fight
```

And also fix crashes in some nestings of these constructs.

### DIFF
--- a/src/sgame/sg_bot_parse.cpp
+++ b/src/sgame/sg_bot_parse.cpp
@@ -1169,7 +1169,7 @@ static AIGenericNode_t *ReadActionNode( pc_token_list **tokenlist )
 	BotInitNode( ACTION_NODE, action->run, &node );
 
 	// allow dropping of parenthesis if we don't require any parameters
-	if ( action->minparams == 0 && parenBegin->token.string[0] != '(' )
+	if ( action->minparams == 0 && parenBegin != nullptr && parenBegin->token.string[0] != '(' )
 	{
 		ret = allocNode( AIActionNode_t );
 		*ret = node;
@@ -1350,6 +1350,11 @@ static AIGenericNode_t *ReadNodeList( pc_token_list **tokenlist )
 		return nullptr;
 	}
 
+	if ( current == nullptr )
+	{
+		return nullptr;
+	}
+
 	list = allocNode( AINodeList_t );
 
 	while ( Q_stricmp( current->token.string, "}" ) )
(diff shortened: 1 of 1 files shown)


VReaperV:
> Apart from that, fix crashes for the following behavior files:

Looks like your formatting got broken.

sweet235:
> Looks like your formatting got broken.

Partly a bad joke. The first one is an empty file, or a file containing only whitespace.

src/sgame/sg_bot_parse.cpp
```diff
@@ -1169,7 +1169,7 @@ static AIGenericNode_t *ReadActionNode( pc_token_list **tokenlist )
 	BotInitNode( ACTION_NODE, action->run, &node );
 
 	// allow dropping of parenthesis if we don't require any parameters
-	if ( action->minparams == 0 && parenBegin->token.string[0] != '(' )
+	if ( action->minparams == 0 && parenBegin != nullptr && parenBegin->token.string[0] != '(' )
```
slipher, line 1172: I guess this would prevent a crash but it would still improperly reject a behavior tree consisting of a single action with no arguments.

sweet235: Of course it does. I am just changing from "crash" to "reject".

src/sgame/sg_bot_parse.cpp
```diff
@@ -1632,6 +1638,11 @@ AIBehaviorTree_t *ReadBehaviorTree( const char *name, AITreeList_t *list )
 	}
 
 	tokenlist = CreateTokenList( handle );
+	if ( tokenlist == nullptr )
+	{
+		return nullptr;
```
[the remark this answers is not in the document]
sweet235: I am not aware of this. Existing code does return `nullptr` for some inputs, like

```
foo
```

src/sgame/sg_bot_parse.cpp
sweet235: If you prefer the crash, you can still have it with
```
condition alertedToEnemy
```

src/sgame/sg_bot_parse.cpp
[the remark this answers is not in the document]
slipher: My bad, remembered wrong. It's only added during parsing to prevent recursion.

slipher:
LGTM

PR #3212, opened 2024-11-27
add BT decorator mapResult, the universal decorator

sweet235:
I am trying to tame unconventional control constructs like `fallback` and `concurrent`. Their semantics frequently cause confusion. A lot of what they do can be done by using appropriate decorators. This PR brings the universal decorator, which can implement any decorator.

It is written as `decorator mapStatus( <running>, <success>, <failure> )`. Each of `<running>, <success>, <failure>` can be one of `STATUS_RUNNING, STATUS_SUCCESS, STATUS_FAILURE`. The child node's status is mapped to the status specified that way.

For instance, `decorator return ( STATUS_SUCCESS )` is equivalent to `decorator mapStatus( STATUS_SUCCESS, STATUS_SUCCESS, STATUS_SUCCESS )`.

`decorator invert` is equivalent to `decorator mapStatus( STATUS_RUNNING, STATUS_FAILURE, STATUS_SUCCESS )`.

Any of the 27 possible decorators is equivalent to some invocation of `mapStatus`.

Control constructs like `fallback` can be implemented too.

```
fallback
{
	action fight
	action roam
}
```

Is equivalent to

```
decorator mapStatus( STATUS_RUNNING, STATUS_FAILURE, STATUS_SUCCESS )
{
	sequence
	{
		decorator mapStatus( STATUS_RUNNING, STATUS_FAILURE, STATUS_SUCCESS )
		{
			action fight
		}
		decorator mapStatus( STATUS_RUNNING, STATUS_FAILURE, STATUS_SUCCESS )
		{
			action roam
		}
	}
}
```

(The outermost decorator is not needed if this is the tree's root)

In this case, `decorator invert` would have been enough, but you get the idea.

Using this PR, we could choose to implement the majority of conceivable control structures as syntactic sugar for `sequence, selector, mapStatus`.

More importantly, we do not have to consider adding new and subtly different control structures anymore. A BT programmer only has to understand how `sequence` and `selector` work, as everything else can be constructed when needed.

### DIFF
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -425,8 +425,8 @@ AINodeStatus_t BotDecoratorMapStatus( gentity_t *self, AIGenericNode_t *node )
 {
 	AIDecoratorNode_t *dec = ( AIDecoratorNode_t * ) node;
 
-	AINodeStatus_t result;
-	switch ( BotEvaluateNode( self, dec->child ) )
+	AINodeStatus_t result = BotEvaluateNode( self, dec->child );
+	switch ( result )
 	{
 	case STATUS_FAILURE:
 		result = ( AINodeStatus_t ) AIUnBoxInt( dec->params[ 0 ] );
@@ -436,6 +436,9 @@ AINodeStatus_t BotDecoratorMapStatus( gentity_t *self, AIGenericNode_t *node )
 		break;
 	case STATUS_RUNNING:
 		result = ( AINodeStatus_t ) AIUnBoxInt( dec->params[ 2 ] );
+		break;
+	default:
+		// the child node's result is none of the three, return it unchanged
 		break;
 	}
 
--- a/src/sgame/sg_bot_ai.cpp
+++ b/src/sgame/sg_bot_ai.cpp
@@ -428,13 +428,13 @@ AINodeStatus_t BotDecoratorMapStatus( gentity_t *self, AIGenericNode_t *node )
 	AINodeStatus_t result;
 	switch ( BotEvaluateNode( self, dec->child ) )
 	{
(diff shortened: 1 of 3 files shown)


slipher:
> It is written as `decorator mapStatus( <running>, <success>, <failure> )`

I think it should be `failure, success, running` to match the order of the status enum (and boolean values).

VReaperV:
Shouldn't the `params` indexes be swapped too in this case?

src/sgame/sg_bot_ai.cpp
```diff
@@ -421,6 +421,27 @@ AINodeStatus_t BotDecoratorReturn( gentity_t *self, AIGenericNode_t *node )
...
+		break;
+	case STATUS_SUCCESS:
+		result = ( AINodeStatus_t ) AIUnBoxInt( dec->params[ 1 ] );
+		break;
+	case STATUS_RUNNING:
+		result = ( AINodeStatus_t ) AIUnBoxInt( dec->params[ 2 ] );
+		break;
+	}
+
+	return result;
```
slipher, line 442: This should probably default to returning the original result if it is not one of the 3 valid values. Unless we are going to tighten up `decorator return` (and this one) to forbid random statuses.

sweet235:
> Shouldn't the params indexes be swapped too in this case?

I did exactly that, but by shuffling the case constants around. In this regard, the OP is not accurate anymore.

src/sgame/sg_bot_ai.cpp
sweet235: Did that.

If we really wanted to tighten up, we should probably do it when parsing. That would be rather easy, as we only allow constants as arguments to decorators and actions.

slipher:
LGTM

PR #3213, opened 2024-11-28
Fix reverb bug; simplify locational color grade

slipher:


illwieckz (approved):
If it works, LGTM.

src/cgame/cg_view.cpp
```diff
@@ -1330,94 +1330,65 @@ Sets cg.refdef.gradingWeights
...
-		weight = 1.0f - dist / cgs.gameGradingDistances[i];
-		weight = Math::Clamp( weight, 0.0f, 1.0f ); // Maths::clampFraction( weight )
-
-		// search 3 greatest weights
-		if( weight <= selectedWeight[2] )
-		{
-			continue;
-		}
+		float dist = CM_DistanceToModel( loc, cgs.gameGradingModels[i] );
+		float weight = 1.0f - dist / std::max( 0.01f, cgs.gameGradingDistances[i] );
```
VReaperV, line 1345: Why is the clamp no longer there?

src/cgame/cg_view.cpp
```diff
@@ -1330,94 +1330,65 @@ Sets cg.refdef.gradingWeights
...
+		for ( int j = 1; j < 4 && weight > selectedWeight[ j ]; j++ )
 		{
-			if( weight <= selectedWeight[j] )
-			{
-				break;
-			}
-
-			selectedIdx[j+1] = selectedIdx[j];
-			selectedWeight[j+1] = selectedWeight[j];
+			selectedTex[ j - 1 ] = selectedTex[ j ];
```
VReaperV, line 1349: These seem to always be 0 at this point.

slipher: Do you mean based on inspecting the code or looking at it in a debugger? If the latter, it's because the only known map build using the feature is https://cdn.unvanquished.net/unvanquished_0.24.1_universal/pkg/map-plat23_b11.pk3

VReaperV: By looking at the code. I see `selectedTex` having its values set only where it is declared.

slipher: I see. I have the screenshots from when I (attempted to) test this patch with that map and they are correct... must have made some mistake with building or test commands. I'll take a look later.

slipher: Added the missing line `selectedTex[ j ] = cgs.gameGradingTextures[ i ];`. Now it's working.

illwieckz (approved):

PR #3214, opened 2024-11-29
When levelshot loading fails fall back to the Tremulous path

slipher:


illwieckz:
Hmm, at the time I said:

> In that specific case I'm fine with the warning since I already want to print a warning that the default path is not found anyway.

So that's fine I guess.

Is the `Delete unused RMLUI 'data formatters'` commit required for this?

illwieckz:
What were the other formatters for?

slipher:
> What were the other formatters for?

I recall checking some at the time and some had been used in circle menus, while others were unused for 10+ years.

illwieckz (approved):
I find the `$levelshot` a bit hacky but if that works that's good enough.

I would prefer `${levelshot}` or something like that with an ending markup to remove any ambiguity on the length of the name of the pseudo-variable.

LGTM.

slipher:
> I would prefer `${levelshot}` or something like that with an ending markup to remove any ambiguity on the length of the name of the pseudo-variable.

The feature can't be simulated by replacing `${levelshot}` with any fixed string, so that seems misleading. I think of it as a magic directory.

PR #3218, opened 2024-12-01
add BT files build.bt, subroutine_become_builder_*.bt

sweet235:
The behavior `build` is similar to `default`, but makes a bot build at all costs if possible. This is useful to players alone in a team, who do not have the time to build a base by themselves.

Move some of the shared build code to the files `subroutine_become_builder_aliens.bt`, `subroutine_become_builder_humans.bt`.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/bots/build.bt
+++ b/pkg/unvanquished_src.dpkdir/bots/build.bt
@@ -1,0 +1,118 @@
+selectClass
+{
+	selector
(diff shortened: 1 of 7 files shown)


sweet235:
I added another behavior `build_here.bt`. It makes the bot build at a user specified postition, which is set automatically by `/tactic build_here 1`.

DolceTriade (approved):

PR #3219, opened 2024-12-01
gitignore build only at top level

### LINKED ISSUE #3217: all file names starting with "build" are globally ignored in .gitignore
I tried to add a file `pkg/unvanquished_src.dpkdir/bots/build.bt`. But that file would be ignored by git, as would `build_now_please.bt`. Yet I believe `build.bt` is a reasonable file name, especially since this filename would be visible to non-programmers.

slipher:
Fixes #3217.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ cmake_install.cmake
 vm
 Release
 Debug
-build*
+/build*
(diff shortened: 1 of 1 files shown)


sweet235:
LGTM

PR #3221, opened 2024-12-03
fix g_bot_traceClient not showing the name of the outermost behavior tree

sweet235:
To reproduce the bug, enter these commands:

```
bot add GigaGrise h 5 defend
g_bot_traceClient GigaGrise
```

The output is:
```
[bot] Bot#1 running at <unknown>.bt:28, action roamInRadius
```

The reason is that the outermost behavior tree is never put in the list of running nodes. That makes sense, as it is always running by definition.

Fix this, now the output is:

```
[bot] Bot#1 running at defend.bt:28, action roamInRadius
```


DolceTriade (approved):

slipher:
Thanks!

PR #3222, opened 2024-12-03
NUKE default_aliens.bt, default_humans.bt

sweet235:
That is a good thing to do. They are not that long anymore. Beside that, there is code duplication between them and default.bt.


slipher:
LGTM

PR #3226, opened 2024-12-05
human bots: abort healing when attacked

sweet235:
This fixes a common problem: once human bots decide to go back to a medistation, they will not react to being attacked by aliens players. Once the alien players notice this problem, they exploit it.

Change this: abort the healing when attacked. This makes the bot feel the pain and target any alien damaging it.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/bots/subroutine_heal_humans.bt
+++ b/pkg/unvanquished_src.dpkdir/bots/subroutine_heal_humans.bt
@@ -1,12 +1,19 @@
-selector
+// only do this when not in a fight (not being damaged by enemies for a few seconds)
+// check this condition at every execution, to abort the long running actions below
+// notice that `action heal` and `action roamInRadius` can make a bot run
+// all the way back to the base
+condition timeSinceLastCombat > 2000
 {
-	decorator return( STATUS_FAILURE )
+	selector
 	{
-		action activateUpgrade( UP_MEDKIT )
+		decorator return( STATUS_FAILURE )
+		{
+			action activateUpgrade( UP_MEDKIT )
+		}
+		decorator timer( 3000 )
+		{
+			action heal
+		}
+		action roamInRadius( E_H_MEDISTAT, 250 )
 	}
-	decorator timer( 3000 )
-	{
-		action heal
-	}
-	action roamInRadius( E_H_MEDISTAT, 250 )
 }


VReaperV:
I assume poison damage does not affect `timeSinceLastCombat`?

slipher:
> I assume poison damage does not affect `timeSinceLastCombat`?

Seems that it does.
```
// Update combat timers if this is a client being damaged by another client.
	if (client && source->client && client != source->client) {
		client->lastCombatTime = level.time;
	}
```

slipher:
#3190 seems to be an attempt a more general solution for this.

sweet235:
> https://github.com/Unvanquished/Unvanquished/pull/3190 seems to be an attempt a more general solution for this.

No, that PR is concerned solely with which enemies are preferred as targets. It makes a bot switch its enemy target at certain occasions. This is nothing we have control over in the behavior tree right now.

>> I assume poison damage does not affect timeSinceLastCombat?

> Seems that it does.

That is not optimal indeed. But I would argue it is better than what we are doing now: making bots running home free food.

sweet235:
I encourage everyone to try this to see how bad the current situation is. There are other cases where we should add similar conditions:

- `action equip`, which makes a bot go an armoury
- `behavior subroutine_repair`, which does the same

Building is so imortant that suicide by alien is usually acceptable, as it teleports the bot back to base.

We are already using `timeSinceLastCombat` for `subroutine_reload`. 

`timeSinceLastCombat` is just the BT name for the C++ number `client->lastCombatTime`, which seems to be ancient. Its original purpose is not related to bots at all.

sweet235:
> I assume poison damage does not affect timeSinceLastCombat?

Another problem is damage from ground fires, but this does not affect human bots' last combat time.

VReaperV:
> I encourage everyone to try this to see how bad the current situation is. There are other cases where we should add similar conditions:
> 
> * `action equip`, which makes a bot go an armoury
> * `behavior subroutine_repair`, which does the same
> 
> Building is so important that suicide by alien is usually acceptable, as it teleports the bot back to base.
> 
> We are already using `timeSinceLastCombat` for `subroutine_reload`.
> 
> `timeSinceLastCombat` is just a wrapper around the C++ number `client->lastCombatTime`, which seems to be ancient. Its original purpose is not related to bots at all. I added this wrapper because the information was already there.

Sounds reasonable.

PR #3228, opened 2024-12-06
fix a div#0 in sgame

illwieckz:
A patch by bmorel.


illwieckz:
He also sent the one attached, but we can't apply it for now (the function is not yet migrated to glm):

- [pm_stepevent.patch.txt](https://github.com/user-attachments/files/18038129/pm_stepevent.patch.txt)

slipher:
LGTM

Issue #2974, opened 2024-04-16
`CG_CustomSound` is broken


cu-kai:
When testing the map `mxl-school` which appears to make use of these, the client is dropped by this line when falling off the ledge in the secret room by the computer room as a custom sound is triggered:
https://github.com/Unvanquished/Unvanquished/blob/befc9f4e8e406be4c48f69b3df35660fa91856f5/src/cgame/cg_players.cpp#L92

I have worked around this on Bunker by returning `sounds/null.wav` but that is a hack so I'm hesitant to PR such code until I understand why the existing code is broken.

cu-kai:
Worth noting that the custom sound that causes the crash on `mxl-school` is one listed in this code: `*falling1.wav`

sweet235:
What is a sensible method to handle the case of a missing sound file? I would say: do the same as with missing textures. Print a warning, play nothing.

cu-kai:
I have this exact fix running on bunker but the sound that causes the crash is listed in `cg_customSoundNames` so I'd imagine we should solve the actual bug rather than work around it?
https://github.com/Unvanquished/Unvanquished/blob/befc9f4e8e406be4c48f69b3df35660fa91856f5/src/cgame/cg_players.cpp#L44-L59

cu-kai:
In the Tremulous source code, the array of death sounds have extensions. Here, we do not have these.
See: https://github.com/darklegion/tremulous/blob/master/src/cgame/cg_players.c#L29-L44

PR #3229, opened 2024-12-06
do not crash if a sound is unknown

illwieckz:
Some patch by bmorel

### DIFF
--- a/src/cgame/cg_players.cpp
+++ b/src/cgame/cg_players.cpp
@@ -91,7 +91,8 @@ sfxHandle_t CG_CustomSound( int clientNum, const char *soundName )
 		}
 	}
 
-	Sys::Drop( "Unknown custom sound: %s", soundName );
+	Log::Warn( "Unknown custom sound: %s", soundName );
+	return ci->sounds[ 0 ];
 }
 
 /*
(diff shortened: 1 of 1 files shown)


src/cgame/cg_players.cpp
```diff
@@ -91,7 +91,8 @@ sfxHandle_t CG_CustomSound( int clientNum, const char *soundName )
 		}
 	}
 
-	Sys::Drop( "Unknown custom sound: %s", soundName );
+	Log::Warn( "Unknown custom sound: %s", soundName );
+	return ci->sounds[ 0 ];
```
slipher, line 95: What is `ci->sounds[ 0 ]` supposed to be? I would expect it to simply return 0 like for a shader that is not found

sweet235: Apparently, `ci->sounds[ 0 ]` is supposed to be "*death1".

cu-kai:
Looks like this would fix #2974 

I also had a fix for this though currently I do not use this on my server: https://github.com/cu-kai/Unvanquished/commit/e4e1f8b0883e911364d2b6b4b09596c55738d3f8 (maybe it was lost during a rebase?)

DolceTriade (approved):

illwieckz:
I made it return `sound/null` instead like @cu-kai suggested.

In fact we already to that in another part of the same file:

https://github.com/Unvanquished/Unvanquished/blob/6ee25182ae4faf9e51dd582f13a49bfa741ceac9/src/cgame/cg_players.cpp#L1229-L1233

cu-kai (approved):

PR #3231, opened 2024-12-06
bmorel v0.55.1 glm 1

illwieckz:
Chunk of GLM changes by bmorel.


illwieckz:
I've noticed some wonky spaces but that can be fixed later.

The names of `BG_EvaluateTrajectory` and `BG_EvaluateTrajectoryDelta` in some `math.cpp` looks weird now that they are moved there but we can rename that later.

illwieckz:
The `math.cpp` and `math.hpp` files can get a header.

slipher:
I suggest removing `normalize_warn` and using SafeNormalize instead in all cases. The pattern used here `normalize_warn(foo); bar = glm::normalize(foo)` just warns, but does not actually attempt to prevent the division by zero. All 4 of the cases here are based on the assumption that entities can't overlap each other, but in fact there are some rare exceptions where they can overlap. For example, when someone is noclipping. If you noclip through a rocketpod, the distance vector in the targeting code could become zero. This rare case is not a bug and does not merit special code to handle it, but we should try to avoid division by zero. So SafeNormalize is what we want: don't warn, do prevent undefined behavior. The `normalize_warn` pattern is the opposite: do warn but allow undefined behavior!

There a lot of `&some_glm_vector[0]`s in here which invoke undefined behavior when used as an array (see #3013).

I wish functions wouldn't be moved to different files as they are rewritten, which prevents me from properly seeing the diffs.

illwieckz:
> I suggest removing `normalize_warn` and using SafeNormalize instead in all cases. The pattern used here `normalize_warn(foo); bar = glm::normalize(foo)` just warns, but does not actually attempt to prevent the division by zero. All 4 of the cases here are based on the assumption that entities can't overlap each other, but in fact there are some rare exceptions where they can overlap. For example, when someone is noclipping. If you noclip through a rocketpod, the distance vector in the targeting code could become zero. This rare case is not a bug and does not merit special code to handle it, but we should try to avoid division by zero. So SafeNormalize is what we want: don't warn, do prevent undefined behavior. The `normalize_warn` pattern is the opposite: do warn but allow undefined behavior!

Is it introducing a bug? If it does the job, we can improve that later. My main concern is to limit merge conflict with next commits so we don't get add extra burden to our future self every time we merge a chunk of commits.

> There a lot of `&some_glm_vector[0]`s in here which invoke undefined behavior when used as an array (see #3013).

But is it buggy in practice? That looks temporary to me until the migration is complete. To be honest I would prefer a some GLM2VEC we can easily look for.

> I wish functions wouldn't be moved to different files as they are rewritten, which prevents me from properly seeing the diffs.

Yes, it would have been better to avoid that, but I wonder if reverting this move may increase the merge burden.

illwieckz:
I guess we have to replace every new `&dir[0]` with `GLM4RW(dir)`?

src/sgame/Entities.cpp
```diff
@@ -151,6 +152,7 @@ bool Entities::KnockbackRadiusDamage(Entity& entity, float amount, float range,
 		if (!G_IsVisible(entity.oldEnt, other.oldEnt, MASK_SOLID)) return;
 
 		glm::vec3 knockbackDir = VEC2GLM( other.oldEnt->s.origin ) - VEC2GLM( entity.oldEnt->s.origin );
+		normalize_warn( knockbackDir );
```
slipher, line 155: This isn't actually normalized here so I guess calling this is just an accident.

illwieckz:
I replaced `&v[0]` with `GLM4READ(v)` and `GLM4RW(v)`, but I'm afraid this adds some merge conflict for something that is only temporary and is not known to be buggy with current compilers.

necessarily-equal:
Not quite sure what the status is on this PR, but GLM4RW was removed: https://github.com/Unvanquished/Unvanquished/pull/3374

PR #3232, opened 2024-12-07
Some BT parser fixes

sweet235:
Change from _crash_ to _reject_ for inputs like these:

```
condition alertedToEnemy
```
```
condition alertedToEnemy()
```
```
selector { condition alertedToEnemy
```

Change from _accept_ to _reject_ for inputs like this:
```
selector { action fight()
```

Change from _reject silently_ to _reject with warning_ for inputs like this:
```
selector {
```

### DIFF
--- a/src/sgame/sg_bot_parse.cpp
+++ b/src/sgame/sg_bot_parse.cpp
@@ -53,6 +53,18 @@ static bool expectToken( const char *s, pc_token_list **list, bool next )
 	return true;
 }
 
+static bool prematureEof( pc_token_list **list )
+{
+	const pc_token_list *current = *list;
+
+	if ( !current )
+	{
+		Log::Warn( "Unexpected end of file" );
+		return true;
+	}
+	return false;
+}
+
 static AIValue_t AIBoxToken( const pc_token_stripped_t *token, bool negative = false )
 {
 	if ( token->type == tokenType_t::TT_STRING )
@@ -549,7 +561,7 @@ static const struct AIOpMap_s
 
 static AIOpType_t opTypeFromToken( pc_token_stripped_t *token )
 {
-	if ( token->type != tokenType_t::TT_PUNCTUATION )
+	if ( token == nullptr || token->type != tokenType_t::TT_PUNCTUATION )
 	{
 		return OP_NONE;
 	}
@@ -768,7 +780,7 @@ static AIValueFunc_t *newValueFunc( pc_token_list **list )
 	parenBegin = current->next;
 
 	// if the function has no parameters, allow it to be used without parenthesis
-	if ( v.nparams == 0 && parenBegin->token.string[ 0 ] != '(' )
+	if ( v.nparams == 0 && parenBegin != nullptr && parenBegin->token.string[ 0 ] != '(' )
 	{
 		ret = ( AIValueFunc_t * ) BG_Alloc( sizeof( *ret ) );
 		*ret = v;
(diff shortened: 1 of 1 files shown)


slipher:
We should change the token interface to something sane like
```
class TokenStream {
    token_t GetNext();
    token_t PeekNext();
};
```

and have it just return a token that says `<EOF>`, instead of giving a null pointer.

sweet235:
> We should change the token interface to something sane like [...]

Maybe that would prevent some crashes. But what would `GetNext` do when already at EOF? Return the EOF token again? That might just replace the crash by an endless loop. The only reliable solution is to fix the parser itself.

Your proposed `TokenStream` seems to be stateful. Notice that our current `pc_token_list` is a linked list (it is doubly linked but only the forward links are ever used). That is a stateless construct, which would allow for a clean and generic parser function type of `token list -> tuple(parsed node, token list)`. That is what the bulk of the contemporary literature on parser combinators is using. If we ever need to backtrack for several tokens, a stateful token stream might make things more complicated.

Our parse functions have type `token list reference -> parsed node`, and mutate the caller's token list. That is a little ugly, but still rather managable, as the caller can always save a copy before calling.

sweet235:
> That is a stateless construct, which would allow for a clean and generic parser function type of token list -> tuple(parsed node, token list). That is what the bulk of the contemporary literature on parser combinators is using. 

Well, to be precise, the literature is using a type that is either `SUCCESS` or `FAIL`. The success result contains the tuple, the fail result can be empty of data. I hope you know what I mean.

slipher:
> Your proposed `TokenStream` seems to be stateful. Notice that our current `pc_token_list` is a linked list (it is doubly linked but only the forward links are ever used). That is a stateless construct, which would allow for a clean and generic parser function type of `token list -> tuple(parsed node, token list)`.

Actually we could do my idea with less changes by just getting rid of the reverse links, and making the end of the list an EOF token that points to itself.

> Maybe that would prevent some crashes. But what would `GetNext` do when already at EOF? Return the EOF token again? That might just replace a crash by an endless loop.

Yes. `<eof>` should be an acceptable token in few, if any, contexts, so the parser should quickly run into some error. But if that's really a concern, we could put a finite amount of EOF tokens, say 10, and then allow a crash (which seems preferable to an infinite loop).

sweet235:
> Actually we could do my idea with less changes by just getting rid of the reverse links, and making the end of the list an EOF token that points to itself.

> Yes. <eof> should be an acceptable token in few, if any, contexts, so the parser should quickly run into some error. But if that's really a concern, we could put a finite amount of EOF tokens, say 10, and then allow a crash (which seems preferable to an infinite loop).

I agree that a crash is preferrable over a diverging parser. Looking at the parser code again, chances of divergence are rather low. Recursion will hit the stack limit unless the compiler does some tail call optimization on yet unwritten parser code (the chance of this happening is nonzero). At a quick glance, I did not find an actual loop in the parser. That is due to the language's very simple grammar, which described excusively right-recursive. In general, left recursion can only be done by a loop in a recursive descent parser, as naive left recursion would diverge immediately. But this is not what we are dealing with currently.

I think we could do the self-referencing EOF token at the end of the list without much danger.

sweet235:
Anyway, while this PR does add two checks for null pointers, its main purpose is to deal with premature EOF in the input correctly. I see no reason not merge this as it is.

DolceTriade (approved):

PR #3215, opened 2024-11-29
add server command showBehavior

sweet235:
When working on the BT parser, I frequently find myself doing ad hoc conversions of parts of behavior trees to some text format I can read. I need to do this in order to check if the parsed tree looks like what I expect.

This PR does it properly. It provides a server command `showBehavior <bot>`. This command looks at the bot's behavior tree, converts it to text, and prints the text to the console. The text is such that it could be read by the BT parser, yielding the same parse tree. This operation is an inverse of the parser.

Let me give an example. Assume a behavior file `human.bt`:

```
selectClass
{
	sequence
	{
		condition numOurBuildings( E_H_SPAWN ) < 2 || numOurBuildings ( E_H_REACTOR ) < 1
		{
			spawnAs WP_HBUILD
		}
		spawnAs WP_MACHINEGUN
	}
}

selector
{
	behavior fight_or_flight
	action roam
}
```

And also a behavior file `fight_or_flight.bt`:

```
selector
{
	sequence
	{
		condition alertedToEnemy
		action fight
	}

	sequence
	{
		condition healScore > 0.25
		action heal
	}
}
```

Then we can do the following:

```
/bot add GigaGrise h 5 human
/showBehavior GigaGrise
```

The output will be:
```
selectClass
{
    sequence
    {
        condition ( ( numOurBuildings( 10 ) < 2 ) || ( numOurBuildings( 16 ) < 1 ) )
        {
            spawnAs 22
        }
        spawnAs 9
    }
}
selector
{
    behavior fight_or_flight
    action roam
}
```

This is pretty much the text in `human.bt`, but using number literals as constants. Notice that condition expressions are fully parenthesized. This is required to check priority and associativity of the binary operators. Spaces are used for indentation, as our ingame console does not like tabs.

We can also (recursively) expand subbehaviors with
```
/showBehavior GigaGrise expand
```

The output will be:
```
selectClass
{
    sequence
    {
        condition ( ( numOurBuildings( 10 ) < 2 ) || ( numOurBuildings( 16 ) < 1 ) )
        {
            spawnAs 22
        }
        spawnAs 9
    }
}
selector
{
    selector
    {
        sequence
        {
            condition alertedToEnemy
            action fight
        }
        sequence
        {
            condition ( healScore > 0.25 )
            action heal
        }
    }
    action roam
}
```

A few cases where this can be useful:
- debugging the parser, as explained above
- helping in debugging complex behavior files distributed over several files
- we could write unit tests for the parser, in plain config files
[...]

PR #3233, opened 2024-12-07
Fix another bug in the BT parser

sweet235:
The old parser accepts inputs like

```
selector
{
    action fight
    action roam
}
action danceInTheSun
```

It silently discards everything after the first tree construct is closed, in this case it discards `action danceInTheSun`.

Change this, now the input is rejected. Instead, this warning is printed:

```
Warn: expected end of file, but found `action` in line 6
```


sweet235:
If you want to verify that this, you can use https://github.com/Unvanquished/Unvanquished/pull/3215.

slipher:
LGTM

PR #3236, opened 2024-12-10
Fix a crash related to the /r command

sweet235:
It is possible for users to crash the server using the `/r` command iff they were sent private messages by the server console earlier. To reproduce, start a dedicated server, connect a client. Enter this command at the server console:

```
/m <client> foo
```
To crash the server, enter this command on the client:

```
/r foo
```

The reason for the crash is that we do `ent->num()` with `ent` being the server console. On a dedicated server, the server console is a null pointer. That makes `ent->num()` a huge negative number, which we will save as the message sender. Later we use this number as index in the entity array.

With assertions enabled, the first command should already crash the server.

Fix this by excluding null pointers as saved message senders.

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -4241,6 +4241,7 @@ static void Cmd_ReplyPrivateMessage_f( gentity_t *ent )
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


VReaperV (approved):
LGTM

PR #3238, opened 2024-12-10
humans: refill while holding the blaster

sweet235:
This is an alternative to https://github.com/Unvanquished/Unvanquished/pull/3237. It fixes the same bug, but is a (very subtle) gameplay change.


illwieckz:
I don't think it's bad to always refill the weapon at the armoury whatever is carried at the time.

slipher:
LGTM

Issue #3223, opened 2024-12-04
Getting right-to-left languages rendered correctly


illwieckz:
We have some people contributing Arab and Farsi translations, but we can't display them properly.

Things to do:

- [ ] Migrate to an NaCl supporting a newer C++ (aka Saigo):
  https://github.com/Unvanquished/Unvanquished/issues/3197
- [ ] Upgrade RmlUi (requires a newer C++)
- [ ] Integrate HarfBuzz with FreeType

DolceTriade:
This also would help with full emoji support!! https://github.com/Unvanquished/Unvanquished/pull/2222

PR #3239, opened 2024-12-11
Mark strings as translatable

illwieckz:
- ui: mark strings as translatable
- ui: strip leading and trailing spaces
- translation: do not ship Farsi language files until we can render right-to-left glyphs properly
  * We already had to do it for Arab language for the same reason:
    https://github.com/Unvanquished/Unvanquished/pull/3169
  * We need to fix this to be able to render those languages:
    https://github.com/Unvanquished/Unvanquished/issues/3223

### DIFF
--- a/pkg/unvanquished_src.dpkdir/.urcheon/ignore.txt
+++ b/pkg/unvanquished_src.dpkdir/.urcheon/ignore.txt
@@ -1,1 +1,2 @@
 ar.po
+fa.po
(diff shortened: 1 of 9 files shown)


sweet235:
LGTM, did not test.

illwieckz:
I tested in game and `generate.sh` did not report any error (we may add this test to CI).

PR #3240, opened 2024-12-12
human bots: do not fight with the ckit

sweet235:
Rather continue building/repairing or go to an armoury to buy a real weapon.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/bots/subroutine_fight_or_flight_humans.bt
+++ b/pkg/unvanquished_src.dpkdir/bots/subroutine_fight_or_flight_humans.bt
@@ -15,14 +15,6 @@ selector
 						action buyPrimary( WP_PULSE_RIFLE )
 					}
 				}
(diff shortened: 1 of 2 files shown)


pkg/unvanquished_src.dpkdir/bots/subroutine_fight_or_flight_humans.bt
```diff
@@ -3,6 +3,7 @@ selector
 	sequence
 	{
 		condition alertedToEnemy
+		condition !haveWeapon( WP_HBUILD )
```
slipher, line 6: There is `condition haveWeapon( WP_HBUILD )` in the following block so something doesn't make sense here.

sweet235: yes

slipher:
LGTM

necessarily-equal:
Without testing, LGTM

PR #3242, opened 2024-12-13
add commands viewcon, delcon

sweet235:
Add two commands to maintain navcons. Both require navedit to be enabled, and the user looking at either the start or the end point.

Command `viewcon` displays a line like

```
navcon start, from ( 449, -3979, -918 ) to ( 596, -3995, -750 ), oneway, radius 15
```

Or for jetpack navcons:
```
navcon start, from ( 449, -3979, -918 ) to ( 596, -3995, -750 ), oneway, radius 15, jetpack
```

This is currently the only way to identify jetpack navcons.

Command `delcon` deletes a navcon.

This is required for 0.55.2.

### DIFF
--- a/src/sgame/botlib/bot_nav_edit.cpp
+++ b/src/sgame/botlib/bot_nav_edit.cpp
@@ -427,10 +427,13 @@ public:
 	}
 };
 
-static int findClosestNavcon( OffMeshConnections &cons, rVec &targetPoint )
+// return the navcon number and true if the targetPoint is at the navcon start,
+// false otherwise
+static std::tuple< int, bool > findClosestNavcon( OffMeshConnections &cons, rVec &targetPoint )
 {
 	int resultIndex = -1;
 	float resultDistance = std::numeric_limits<float>::max();
+	bool resultIsStart = true;
 	for ( int i = 0; i < cons.offMeshConCount; i++ )
 	{
 		int n = i * 6;
@@ -445,10 +448,11 @@ static int findClosestNavcon( OffMeshConnections &cons, rVec &targetPoint )
 			{
 				resultIndex = i;
 				resultDistance = distance;
+				resultIsStart = count == 0;
 			}
 		}
 	}
-	return resultIndex;
+	return std::tie( resultIndex, resultIsStart );
 }
 
 class ViewconCmd: public Cmd::StaticCmd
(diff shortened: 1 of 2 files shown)


sweet235:
I propose we merge this as soon as possible, so that it ends up in the upcoming release. What can go wrong, it is just two commands enabled only in local games.

src/sgame/botlib/bot_nav_edit.cpp
```diff
@@ -427,6 +427,132 @@ class AddconCmd: public Cmd::StaticCmd
...
+	for ( int i = 0; i < cons.offMeshConCount; i++ )
+	{
+		int n = i * 6;
+		// look at start and end points
+		for ( int count = 0; count < 2; count++ )
+		{
+			float distanceX = targetPoint[ 0 ] - cons.verts[ n + count * 3 ];
+			float distanceZ = targetPoint[ 1 ] - cons.verts[ n + count * 3 + 1 ];
+			float distanceY = targetPoint[ 2 ] - cons.verts[ n + count * 3 + 2 ];
+			float distance = sqrt( Square( distanceX ) + Square( distanceY ) + Square( distanceZ ) );
```
slipher, line 447: You should be able to use `Distance`/`DistanceSquared` on rVec (it has an `operator float *`)

src/sgame/botlib/bot_nav_edit.cpp
```diff
@@ -427,6 +427,132 @@ class AddconCmd: public Cmd::StaticCmd
 	}
 };
 
+static int findClosestNavcon( OffMeshConnections &cons, rVec &targetPoint )
+{
+	int resultIndex = -1;
+	float resultDistance = std::numeric_limits<float>::max();
```
slipher, line 436: Shouldn't there be some distance limit instead of returning one arbitrarily far away?

sweet235: There is a distance limit, he navcon's radius. The function's name is a bit misleading. It only searches for the closest if two or more navcon circles overlap (this happens).

illwieckz:
Another hiding:

```
src\sgame\botlib\bot_nav_edit.cpp(545): error C2220: the following warning is treated as an error
src\sgame\botlib\bot_nav_edit.cpp(545): warning C4456: declaration of 'i' hides previous local declaration
src\sgame\botlib\bot_nav_edit.cpp(522): note: see declaration of 'i'
```

src/sgame/botlib/bot_nav_edit.cpp
sweet235: yes

slipher:
LGTM

PR #3243, opened 2024-12-13
Fix invisible download bar

### LINKED ISSUE #2164: Download bar disappeared on download page

slipher:
Alternative to #3198 that doesn't use a 100% transparent elements hack.

Fixes #2164.

See the bottom left:
unvanquished_2024-12-13_173318_000


illwieckz (approved):
That looks less hackish indeed, lgtm.

PR #3248, opened 2024-12-15
Bots: rebuild reactor/overmind

sweet235:
Bots do not rebuild these once destroyed. Oops, fix this.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/bots/build.bt
+++ b/pkg/unvanquished_src.dpkdir/bots/build.bt
@@ -29,7 +29,7 @@ selector
 				{
 					behavior subroutine_become_builder_humans
 
-					condition distanceTo( E_H_REACTOR ) <= 700
+					condition distanceTo( E_H_REACTOR ) <= 700 || distanceTo( E_H_REACTOR ) > 999999
 					{
 						decorator return( STATUS_FAILURE )
 						{
@@ -86,7 +86,7 @@ selector
 				{
 					behavior subroutine_become_builder_aliens
 
-					condition distanceTo( E_A_OVERMIND) <= 700
+					condition distanceTo( E_A_OVERMIND) <= 700 || distanceTo( E_A_OVERMIND ) > 999999
 					{
 						decorator return( STATUS_FAILURE )
 						{
--- a/pkg/unvanquished_src.dpkdir/bots/build.bt
+++ b/pkg/unvanquished_src.dpkdir/bots/build.bt
@@ -29,6 +29,7 @@ selector
 				{
 					behavior subroutine_become_builder_humans
 
+					// if reactor is close or does not exist
 					condition distanceTo( E_H_REACTOR ) <= 700 || distanceTo( E_H_REACTOR ) > 999999
 					{
 						decorator return( STATUS_FAILURE )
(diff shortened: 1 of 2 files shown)


slipher:
Is that really a good idea? The consequences of building in a random location are a lot worse for these buildables.

sweet235:
> Is that really a good idea?

That is what they did since 0.54. I broke this in 0.55.2, because I added `condition distanceTo( E_H_REACTOR ) <= 700`. The reason I added this is that a bot might have to traverse a large part of the map to get close to the reactor, and would build while on its way.

To see this, load map terminus and add a few navcons to make the human bots fall down to the rail tracks. A bot will initially build near the reactor, while running `roamInRadius( E_H_REACTOR, 700 )` . Soon it will inevitably fall on one of the rail tracks. Still running `roamInRadius( E_H_REACTOR, 700 )`, it will run to the middle of the map and back until it comes close to the reactor again. It will build this whole time. This navcon file `terminus-human_naked.navcon` is enough to see the problem:

```
1266 2 4069 1126 -50 4064 30 1 63 0
1866 2 3941 1995 -62 3934 30 1 63 0
```

But now I caused a problem much worse: when there is no reactor, `distanceTo` will return `HUGE_VAL`, which prevents the bot from building the reactor forever. It will roam with a ckit until its team has lost.

>  The consequences of building in a random location are a lot worse for these buildables.

In at least 95 out of 100 cases, the location will not be random. Once the reactor/overmind is no more, the only way to become a builder is to suicide by some means. That makes the builder appear next to an existing spawn, where it will build the reactor/overmind. It had been like this for quite some time now, until I broke it.

Even if there is a builder while the reactor/overmind is destroyed, that builder will likely be close to the old reactor/overmind position, and build there.

Nobody seems to have noticed the possibility of a truly random location in the past year.

pkg/unvanquished_src.dpkdir/bots/build.bt
```diff
@@ -29,7 +29,8 @@ selector
 				{
 					behavior subroutine_become_builder_humans
 
-					condition distanceTo( E_H_REACTOR ) <= 700
+					// if reactor is close or does not exist
+					condition distanceTo( E_H_REACTOR ) <= 700 || distanceTo( E_H_REACTOR ) > 999999
```
necessarily-equal, line 33: I would suggest:

 ```suggestion
					condition numOurBuildings( E_H_REACTOR ) == 0 || distanceTo( E_H_REACTOR ) <= 700
```

And likewise for the other 3 conditions.

For the rationale, sounds like a very good idea, thanks !

sweet235: `numOurBuildings` is slower than calculating the distance, as it loops over entities. I added a fast track for spawns, but that is the only fast case for `numOurBuildings`.

necessarily-equal: hmm, ok, fair enough. I think since it has a comment at least the intent is documented. LGTM without test from me

VReaperV (approved):
LGTM

pkg/unvanquished_src.dpkdir/bots/build.bt
slipher: numOurBuildings does need to be used so that under-construction buildables are correctly detected.

PR #3250, opened 2024-12-17
lower g_bot_defaultSkill

Gireen:
recently there where again reports from new players that bots are to strong without knowing how to change the settings. 

The default value should be lowered to accommodate for that. Servers for regulars can still have stronger bots but offline play needs to be focused on new players in its initial settings. 

related to #3055

### DIFF
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -35,7 +35,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include <glm/gtx/norm.hpp>
 #include <glm/gtx/vector_angle.hpp>
 
-static Cvar::Range<Cvar::Cvar<int>> g_bot_defaultSkill( "g_bot_defaultSkill", "Default skill value bots will have when added", Cvar::NONE, 4, 1, 9 );
+static Cvar::Range<Cvar::Cvar<int>> g_bot_defaultSkill( "g_bot_defaultSkill", "Default skill value bots will have when added", Cvar::NONE, 3, 1, 9 );
 static Cvar::Cvar<int> g_bot_alienAimDelay = Cvar::Cvar<int>( "g_bot_alienAimDelay", "make bots of alien team slower to aim", Cvar::NONE, 150 );
 static Cvar::Cvar<int> g_bot_humanAimDelay = Cvar::Cvar<int>( "g_bot_humanAimDelay", "make bots of human team slower to aim", Cvar::NONE, 150 );
 
--- a/src/sgame/sg_bot_util.cpp
+++ b/src/sgame/sg_bot_util.cpp
@@ -35,7 +35,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include <glm/gtx/norm.hpp>
 #include <glm/gtx/vector_angle.hpp>
 
-static Cvar::Range<Cvar::Cvar<int>> g_bot_defaultSkill( "g_bot_defaultSkill", "Default skill value bots will have when added", Cvar::NONE, 5, 1, 9 );
+static Cvar::Range<Cvar::Cvar<int>> g_bot_defaultSkill( "g_bot_defaultSkill", "Default skill value bots will have when added", Cvar::NONE, 4, 1, 9 );
 static Cvar::Cvar<int> g_bot_alienAimDelay = Cvar::Cvar<int>( "g_bot_alienAimDelay", "make bots of alien team slower to aim", Cvar::NONE, 150 );
 static Cvar::Cvar<int> g_bot_humanAimDelay = Cvar::Cvar<int>( "g_bot_humanAimDelay", "make bots of human team slower to aim", Cvar::NONE, 150 );
 


illwieckz:
Why not `3`?

illwieckz:
Actually `unvanquished.net` and `gg.illwieckz.net` use `3` since a very long time now.

Also the value of `3` surfaced again in a recent talk in the chat.

sweet235:
The bot skill is just a number we made up. What we should look at is these cvars:

- `g_bot_humanAimDelay   cvar - "150" - int - make bots of human team slower to aim`
- `g_bot_alienAimDelay`, which is similar
- `g_bot_skillset_baseSkills     cvar - "1:movement 1:fighting 1:feels-pain 1:buy-modern-armor 1:medkit 1:prefer-armor 3:strafe-attack 5:aim-barbs 5:mantis-attack-jump 5:small-attack-jump 5:mara-attack-jump 5:goon-attack-jump 5:tyrant-attack-run 5:a-fast-flee 5:h-fast-flee 5:aim-head 5:safe-barbs 7:predict-aim"`

sweet235:
There has been sudden increase in the bot aim accuracy in https://github.com/Unvanquished/Unvanquished/commit/33fb1572a739bfc4cd40c70adf92bd0d4aa4f5cb. In respone to this, we increased the aim delay for both teams from 100 to 150. Yet I feel like they are still better at aiming than they were before.

sweet235:
> I think bots are to strong for new players, act to aggressive and can easily overwhelm newbies.
It would be reasonable to lower the difficulty of bots in the default settings, since they are the first thing new players encounter offline or on many servers.

We all know the optimal strategy, we have seen scrims. I am not sure what you mean by "aggressive". If it means: they leave their base and attack, I would argue that does actually make them easier to beat. Notice that lowering the skill might make the bots do this more often.

Bots simulate players. An army of 5 players has always been dangerous. Maybe we should also think about the number of bots we add by default. 5 bots can be a lot on small maps.

sweet235:
> Is that a thing?

In theory, we have `g_bot_defaultFill`, which is 5 by default. But I guess the actual number is chosen explicitely by each game host.

slipher:
I see, the `dist/configs` sample server setup thing sets 5. I guess you could consider that a "default" for a server in some sense.

Gireen:
> Why not `3`?

No particular reason. Its more about the direction than the exact value. If no one is against it `3` is also good but not lower than that, to have the option for even weaker bots if newbies find the setting for it. 

> The bot skill is just a number we made up. 

I mean its all made up. 😄 The influence of the skill level is not that big anymore in most calculations but still affects the skilltree.
When used, the skillpoints depend on the level if i see it right. Also the skills from g_bot_skillset_baseSkills are picked by level. 
Of cause we could just change all lvl 5 skills to 6 or higher but that raises the question how the strength distribution over the levels should be. I think lvl 5 should represent an average player which is to strong for a newbie. 


> We all know the optimal strategy, we have seen scrims. I am not sure what you mean by "aggressive". If it means: they leave their base and attack, I would argue that does actually make them easier to beat. Notice that lowering the skill might make the bots do this more often.

Yes thats about constant rushing, but it would not change by lower skills that needs probably a different solution. I don't say this PR fixes #3055, it makes at least some small progress in that direction and other changes, like some of the ones you mentioned, will still be required.

sweet235:
> I mean its all made up. 😄 

Not really. `g_bot_humanAimDelay` and `g_bot_alienAimDelay` appear naturally in the aim calculation. These are of course important parameters, we should probably increase them to 200 or so. How this number is scaled with skill level is hardcoded, but this is open to debate as well.

For reference, in `sg_bot_skilltree.cpp`, we currently have:

```
	static Cvar::Callback<Cvar::Cvar<std::string>> g_skillset_baseSkills(
		"g_bot_skillset_baseSkills",
		"Preferred skills for bots depending on levels, this is a level:skillName key:value list, example: " QQ("6:mantis-attack-jump, 3:prefer-armor"),
		Cvar::NONE,
		"1:movement 1:fighting 1:feels-pain 1:buy-modern-armor 1:medkit "
		"3:strafe-attack "
		"5:aim-barbs 5:mantis-attack-jump 5:small-attack-jump 5:mara-attack-jump 5:goon-attack-jump 5:tyrant-attack-run 5:a-fast-flee 5:h-fast-flee 5:prefer-armor 5:lcannon-tricks "
		"7:aim-head 7:predict-aim 7:safe-barbs "
		, // unused: fast-aim
		G_SetBaseSkillset
		);
```

We can see that even skill level 3 has `strafe-attack`, which we might want to disable for new players.

necessarily-equal:
I'm definitely in favor of making bots easier by default, especially in LAN games. Although I'm thinking maybe we should make the interface for bots in local games much more explicit, so that people at least *know* which bot level they add (otherwise they will always get confused, no matter if we make the default easy or not)

I think a default of 3 is good

necessarily-equal:
> The bot skill number is shown in the score board. But, let me say it again, this is just a made up number.

I mean yes, but if we have a slider that goes from 1 to 9 I think that would be fairly clear (or at least clearer)

Gireen:
maybe there is some misunderstanding. @sweet235 what do you mean with made up number and how is it related to this? Does it prevent us from changing g_bot_defaultSkill?

sweet235:
> what do you mean with made up number and how is it related to this?

The number itself does not have a lot of influence on the actual skill. Adjusting the aim speed can be done with the correspondig cvars, this provides finer grained control. If we want bots to be less intimidating, they should probably not have the skillset skills `buy-modern-amor, strafe-attack`. This is best done in `g_bot_skillset_baseSkills`, even lowering the skill to 1 does not remove `buy-modern-amor` otherwise.

sweet235:
Another cvar with influence on the bot difficulty is `g_bot_reactiontime`.

sweet235:
I propose we set it to 3, and accompany it with https://github.com/Unvanquished/Unvanquished/pull/3254.

Gireen:
The only downside is that `g_bot_reactiontime` will affect all skill levels and make all bots weaker but if lvl9 bots are not used that could be ok too.

PR #3254, opened 2024-12-20
tweak g_skillset_baseSkills in order to make skill level 3 easier

sweet235:
See https://github.com/Unvanquished/Unvanquished/pull/3250 for a discussion.

### DIFF
--- a/src/sgame/sg_bot_skilltree.cpp
+++ b/src/sgame/sg_bot_skilltree.cpp
@@ -153,8 +153,8 @@ void G_InitSkilltreeCvars()
 		"g_bot_skillset_baseSkills",
 		"Preferred skills for bots depending on levels, this is a level:skillName key:value list, example: " QQ("6:mantis-attack-jump, 3:prefer-armor"),
 		Cvar::NONE,
-		"1:movement 1:fighting 1:feels-pain 1:buy-modern-armor 1:medkit "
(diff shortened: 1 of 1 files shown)


sweet235:
Notice that, if we do this, we will receive numerous "bug reports" about the bots being broken. That is how we ended up where we are currently. I will wait a little before merging this.

sweet235:
Maybe this is not a good idea. When left alone, the human bots win every game easily. That is no surprise, as they are firing at aliens who are now approaching in straight lines without any dancing.

While this PR does make the bots on both teams easier to kill, the price is a shift in the overall balance. This would become noticable once there is a user on both teams. The alien player would complain about this being unfair, and rightly so.

slipher:
> When left alone, the human bots win every game easily.

Isn't it already like that now, at level 5?

Gireen:
On the lower levels the actions of players should become more influential than the balance of bots, so the practical consequences might not be that big. 

medikit could also be moved to lvl4 for mitigation.

sweet235:
> Isn't it already like that now, at level 5?

Who knows, both unv.net and illwieckz' server went down to level 3. These are ones I usually look at for statistics.

> On the lower levels the actions of players should become more influential than the balance of bots

That is the case. Notice that lowering the bot level will put the weaker players at a bigger disadvantage in online games.

Issue #3256, opened 2024-12-22
Bug with jetpack or navcon state machine


slipher:
A bot can get stuck in a bad state after going off-course while using a jetpack navcon. In the case described below, the bot continually runs against the wall at the bottom of the navcon and does not activate the jetpack. The symptoms look the same as bug #3253 where a bot without jetpack is trying to use the jetpack navcon.

This bug is not high-priority since it can't be observed while using the default behavior tree. Something breaks it out of the loop right away there.

### Repro instructions

`bots/jetpack-test.bt`:
```
sequence {
	action buy(WP_MACHINEGUN, UP_JETPACK)
	action teleport(1652, 2761, 223)
	action changeGoal(2036, 3759, 458)
	action moveToGoal
	action suicide
}
```

`maps/forlorn-human_naked.navcon`:
```
navcon 3
1822 130 3342 1818 450 3454 30 2 63 1
```

Commands
```
devmap forlorn
g_bot_infiniteFunds 1
g_bot_infiniteMomentum 1
bot add * h 5 jetpack-test
```

Then while the bot is ascending, `slap` it a couple times so the bot does not reach the top.

sweet235:
I figured that there is something that could go wrong at every state. So I added a (maybe temporary) cvar `g_bot_jetpackTimeout`, and set it to 10 seconds by default. The state machine is supposed to be reset to `botJetpackState_t::BOT_JETPACK_NONE` if the navcon end has not been reached after this time.

Now I wonder if this really works in all cases.

sweet235:
This does not seem to be a bug with the jetpack state machine itself, but a more generic problem. I looked at `self->botMind->jetpackState`, it is `BOT_JETPACK_NONE` while the bot is stuck. That is as it should be.

Then I looked at `level.time - self->botMind->lastNavconTime`. This number was ever increasing, and not close to zero as expected when the bot is supposed to start the flight. The reason is that this simple BT file does not seem to trigger the unstick routine, which somehow seems to do a new path search after some time has passed. Without a new path search, the last navcon time is not updated.

To confirm this, try the same with a dretch bot. Use the same navcon file as `forlorn-level0.navcon`, and this behavior `dretch-test.bt`

```
sequence {
	action teleport(1652, 2761, 223)
	action changeGoal(2036, 3759, 458)
	action moveToGoal
	action suicide
}
```

Slap the dretch while it is on the wall. Observe that the dretch has the same problem.

sweet235:
[...]

PR #3257, opened 2024-12-23
command slap: allow deterministic acceleration vector as argument

sweet235:
I am currently using this to debug https://github.com/Unvanquished/Unvanquished/issues/3256.

The slap command can be invoked like:

- `slap Grise` - no damage, random acceleration with magnitude of (roughly) 1000
- `slap Grise 10` - 10 damage, random acceleration with magnitude of (roughly) 1000
- `slap Grise 10 2000` - 10 damage, random acceleration with magnitude of (roughly) 2000

All these forms use an upward pointing random vector as acceleration. The randomness can impede the effective use of this command for debugging.

Add a 4th form:

- `slap Grise 10 100 200 300` - 10 damage, acceleration is the deterministic vector (100, 200, 300)


slipher:
LGTM

PR #3258, opened 2024-12-24
bot: Add E_USERPOS and refactor code to make it more generic

DolceTriade:
Our code only accepts entities, which isn't generic enough to accept either an entity or a position. This allows for writing bts that allow the bots to interact with a user specified position.


sweet235:
> For example, one could write a build-here behavior tree that allows the bot to build at the user specified location.

We do have a behavior `build_here.bt`.

DolceTriade:
Yes, where "here" is where the caller is instead of the bot

sweet235:
I noticed this restriction too. One problem with `E_USERPOS` is that is `E` most probably stands for "entity".

Would this work for other types than vectors too? Maybe a scalar?

DolceTriade:
Hm I didn't think of what the E meant...but that makes sense given that all the other things are entities. I could rename it to something else?

This specific constant depends on something setting the "userSpecifiedPosition" in botMind. Currently, the tactic command will do this, but in the future it could also be Lua.

sweet235:
> Hm I didn't think of what the E meant...but that makes sense given that all the other things are entities.

It took me about a year to figure this out.

> I could rename it to something else?

I have no idea, once we go beyond entities, we are in uncharted territory. I am currently trying to compile a list of what other stuff we might want to name.

> This specific constant depends on something setting the "userSpecifiedPosition" in botMind. Currently, the tactic command will do this, but in the future it could also be Lua.

I also made a `userSpecifiedClientNumber` which is used for the `follow` action.l

sweet235:
After a short discussion elsewhere, I proposed to add arguments to BT files. Someone suggested

```
I think syntax like [behavior] [arg type] [arg]   
```

If we choose to do this, we could NUKE both `userSpecifiedPosition` and `userSpecifiedClientNum` and place this information in arguments instead. These arguments would be set by the commands `tactic command` and `bot behavior`, just as the user specified things are set right now.

I think we could get away with only two types: `number` and `3D vector`. We could even check the types when setting the behavior, eliminating any need for runtime type checks.

PR #3259, opened 2024-12-24
Add a circle menu for bot commands

sweet235:
This is work in progress. I do not have any screenshots, but there is a video at https://users.unvanquished.net/~sweet/videos/bot-command-menu.mp4.

The icons were taken from https://feathericons.com/.


sweet235:
What would be a good key to bind this to?

sweet235:
TODO:

* [ ] Translatable strings
* [ ] Update `g_tacticBehaviors`
* [ ] Bind to a free key on the keyboard
* [ ] Human readable names, do not show something like `stay_here`

Grise3:
Nice!!!

It is a great thing!
maybe "v" for azerty keyboards? Near the "c" for the beacons menus ^^

src/sgame/sg_spawn.cpp
```diff
@@ -1024,15 +1024,17 @@ static void InitDisabledItemCvars()
...
 {
 	static Cvar::Callback<Cvar::Cvar<std::string>> g_tacticBehaviors(
 	   "g_tacticBehaviors",
 		"Allowed behaviors for the tactic command, example: " QQ("default, camper, reckless"),
 		Cvar::NONE,
-		"",
+		defaultTacticBehaviors.c_str(),
 		BG_SetTacticBehaviors
 		);
+	BG_SetTacticBehaviors( defaultTacticBehaviors.c_str() );
```
slipher, line 1037: Why force it back to the default value at the beginning of every match? Seems wrong.

src/cgame/cg_consolecmds.cpp
```diff
@@ -432,6 +432,17 @@ static void CG_BeaconMenu_f()
 	Rocket_DocumentAction( rocketInfo.menu[ ROCKETMENU_BEACONS ].id, "show" );
 }
 
+static void CG_BotCommandMenu_f()
+{
+	if ( !cg.snap
+		|| cg.clientNum != cg.snap->ps.clientNum /* following */
+		|| cg.snap->ps.persistant[ PERS_SPECSTATE ] != SPECTATOR_NOT /* spectating */ )
```
slipher, line 439: Is it supposed to be disabled when you are dead? If so why?

slipher (commented):
I think a suitable substitute for most of the icons can be found among the ones we already have. For example: https://fontawesome.com/v4/icon/bullseye

Why not call it a "bot tactic menu" so that it matches the naming used in sgame?

Any plans to gray out tactics that are not allowed by the server? A config string would be a good way to transmit g_tacticbehaviors

src/sgame/sg_spawn.cpp
sweet235: Hmm, is it really? My first try was to simply give the default value as 4th argument, like in

```
	static Cvar::Callback<Cvar::Cvar<std::string>> g_tacticBehaviors(
		"g_tacticBehaviors",
		"Allowed behaviors for the tactic command, example: " QQ("default, camper, reckless"),
		Cvar::NONE,
		"default, defend, attack, stay_here, follow",
		BG_SetTacticBehaviors
		);
```

But it turns out that the callback function is not called automatically when doing this. The result is an inconsistency between the cvar's value string and the data structure holding the values for actual use.

src/cgame/cg_consolecmds.cpp
sweet235: > Is it supposed to be disabled when you are dead?

No.

sweet235:
> I think a suitable substitute for most of the icons can be found among the ones we already have. For example: https://fontawesome.com/v4/icon/bullseye

I have no particular preference about the icons. The ones I am currently using have the advantage of being MIT licensed. Fontawesome has a custom license I am not qualified to interpret.

Grise3:
azerty keyboards : "n" it is near the "b" key witch is currently the key for the beacon menu

Gireen:
Since we already use FontAwesome it should be fine. But to be sure

> The OFL is listed and recognized by the Free Software Foundation (FSF) as a valid Free Software license. They have added it to their License List. It complies with the Free Software Definition and its four foundational freedoms as defined by the FSF for the GNU project
https://openfontlicense.org/why-use-the-ofl/

The benefit of using icons in text form is also that they scale better.

Grise3:
It is now binded on "n" by default

Grise3:
I will try to add in the tutorial in 2 days

Grise3:
@slipher said : 
> I think a suitable substitute for most of the icons can be found among the ones we already have. For example: https://fontawesome.com/v4/icon/bullseye

There is a vote on the forum for the icons, you can find it [here](https://forums.unvanquished.net/viewtopic.php?p=19502#p19502)

src/sgame/sg_spawn.cpp
sweet235: Nevermind. I made sure this can only be done once.

sweet235:
> Any plans to gray out tactics that are not allowed by the server? A config string would be a good way to transmit g_tacticbehaviors

Maybe in a later PR. Doing too much at once is the best way to stall PRs.

illwieckz:
I was also going to suggest to bind it on `n` too.

my former idea when I redid the binds was:

- `c`: taunt
- `v`: voices (but not implemented yet)
- `b`: beacons

So bot actions on `n` right next to beacons is consistent.

illwieckz:
In fact, many bot actions are similar to the beacons. So maybe we can turn beacons like `attack` or `defend` to also command bots. I wonder if a bot menu would only be needed for things only meant for bots.

sweet235:
> In fact, many bot actions are similar to the beacons. So maybe we can turn beacons like attack or defend to also command bots. I wonder if a bot menu would only be needed for things only meant for bots.

This has been suggested before, and I have thought about it regularly. We could marry the bot tactic command to the beacon command, but to me, this is more of a long term goal right now. There are several differences:

- Beacons can only be placed for the whole team, tactic commands can be given to any number of bots.
- Beacons are all about target positions in space, while the bot tactics make the bots choose targets autonomously. The only exception is `stay_here.bt`, which we might refactor soon, see https://github.com/Unvanquished/Unvanquished/pull/3258.
- In big, open areas, I find it difficult to place beacons accurately. They might end up far from where I want them to be, and I might not even notice that from where I am.
- Beacons are very temporary, bot commands have a lasting effect.

Grise3:
as decided [here](https://forums.unvanquished.net/viewtopic.php?p=19512#p19512), there is a new attack icon 
attack

sweet235:
I find myself in an somewhat unfortunate position.

This PR has been overdue for some time, so I had to make it eventually. None of the icons were selected by myself, but by people who wanted to help me. I am grateful for that help.

However, there seems to be a policy in place to reject any icons not in fontawesome. It is my mistake to have suggested the possibility of icons not in fontawesome. I apolgise for having made that suggestion. Had I known the policy, this would not have happened.

In particular, I would like to apologise to @cu-kai and especially @Grise3, who already made a new icon. Your suggested icons are all in a proper vector graphics format. I do not believe that "icons in text form" scale better.

Gireen:
i think the misunderstanding about adding icons was resolved from my observation, my explanation for how it works:

> the process, as I have seen it here over the years, is to default to what is available in the game. If there is nothing matching available or we can create/add something better then something new is added.

My suggestion for a quick progress would be to merge the menu with the available icon in game, be it font icon or image. And add replacements one by one, that allows room for discussion where its needed and doesn't waists time everywhere else.

On that matter there is also already a defend icon available  with `pkg/unvanquished_src.dpkdir/gfx/feedback/beacons/defend.png`

> I do not believe that "icons in text form" scale better.

Its not about the source files but the use by the engine. The font icons act like vector graphics ingame and look better in different scales.

VReaperV:
? We don't have vector graphics, so they scale the same whether it's a font or not (in fact, font might scale worse due to lower resolution).

Gireen:
Just that i understand this correct, if i have an raster graphic and the same as font icon both created from a vector source.
Display them ingame side by side with 1 em height and increase the text size they will always look the same? Or the text one would look even worse?

sweet235:
> Just that i understand this correct, if i have an raster graphic and the same as font icon both created from a vector source.
Display them ingame side by side with 1 em height and increase the text size they will always look the same? Or the text one would look even worse?

Converting to a font can at best deliver the same scaling behavior as an SVG file.

sweet235:
Could anyone tell me about the current status of the icon bikeshed? I lost track.

cu-kai (approved):
I don't think we should be blocked on the icons bikeshed. That can be a separate PR in my opinion.

LGTM on the condition that we add the license for the current icons from Feather Icons: https://github.com/feathericons/feather/blob/main/LICENSE

We could, for now, place it in the `pkg/unvanquished_src.dpkdir/gfx/feedback/bottactic` folder.

cu-kai:
(Merged at request of @Grise3 and @sweet235)

Grise3:
Thanks Kai !

PR #3260, opened 2024-12-26
Fix env_afx_gravity

sweet235:
It seems our gravity entities always end up using 0 as gravity. This seems to fix that. I found the problem when looking at the function `env_afx_gravity_touch`.

I do not really understand what I am doing here. But I do provide a test map with several gravity zones (rename it from `.zip` to `.dpk`).

[map-blackhole_0.zip](https://github.com/user-attachments/files/18251555/map-blackhole_0.zip)

### DIFF
--- a/src/sgame/sg_spawn_afx.cpp
+++ b/src/sgame/sg_spawn_afx.cpp
@@ -282,6 +282,7 @@ SP_trigger_gravity
 void SP_env_afx_gravity( gentity_t *self )
 {
 	int* amount = &self->mapEntity.config.amount;
+	self->mapEntity.amount = *amount;
 	if ( !( G_SpawnInt( "amount", "0", amount ) || G_SpawnInt( "gravity", "0", amount ) ) )
 	{
 		//TODO: it is highly possible that other situations are broken.


sweet235:
The current state of my investigation:

### Which maps are using gravity entities?

- watah
- mudkip
- zittrig-arena
- transparency
- covid
- scorpius
- paint7
- flap
- cyberspace
- lite
- gamma-core
- combat-t2
- dryout
- obsession
- analmango
- goof
- atcs-upsidedown
- trojan-gates
- mancannon

And some others.

### What is the allowed range of gravity values?

For `g_gravity`, we allow a range from 200 to 2000.

### Are there any maps using negative gravity values?

Yes. `zittrig-arena` is using -250, but that can probably be changed to 0. `covid` is using negative gravity too.

### Are there any maps using a higher gravity than 2000?

Yes, `watah`.

slipher:
That works I guess. It would also be good to get rid of the `env_afx_gravity_reset` function which is useless since the entity does not have any mutable state. If the keyword is not present probably the entity should just warn and delete itself.

sweet235:
> It would also be good to get rid of the env_afx_gravity_reset function which is useless since the entity does not have any mutable state. If the keyword is not present probably the entity should just warn and delete itself.

I am not sure. This probably depends on where we want to go with the lua scripting branches. Maybe it would make sense to keep a gravity entity that does nothing, and set the gravity later during the game.

sweet235:
We should allow for mutable state here.

slipher:
The reset function is not even created unless the number is missing, so I don't see how you could do something useful with it. If a Lua interface were added, it could set the amount directly and wouldn't need a `reset` either.

LGTM regardless

PR #3261, opened 2024-12-26
NUKE entityClass_t

slipher:
The only thing accomplished by this struct was to maintain a counter that appears in one `g_debugEntities` log message informing the number of entities with the same class name. I don't think anyone will miss that.


illwieckz:
It looks like this has been overlooked. @slipher Can this be rebased?

slipher:
Rebased

PR #3263, opened 2024-12-29
Debug draw: use types safe for IPC

slipher:
I think glm::vec3 can be 16 bytes instead of 12 if you turn on SSE optimizations.


VReaperV (approved):
LGTM

PR #3264, opened 2024-12-29
Update FontAwesome to 6.7.2

slipher:
Replay of https://github.com/UnvanquishedAssets/unvanquished_src.dpkdir/pull/177 from the defunct submodule.

The benefit is that there are more icons.

TODO: double-check that everything still looks the same since I replaced the version 6.3 font in the earlier PR with 6.7.


DolceTriade (approved):

illwieckz:
I guess we can merge that?

PR #3265, opened 2024-12-30
Add `bounds_t` with pre-aligned `mins` and `maxs`

illwieckz:
See:

- https://github.com/DaemonEngine/Daemon/pull/1482


illwieckz:
Hmmm, it looks like we need explicit padding:

```
src\engine/qcommon/q_shared.h(302): error C2220: the following warning is treated as an error
src\engine/qcommon/q_shared.h(302): warning C4324: 'bounds_t': structure was padded due to alignment specifier
src\engine/qcommon/q_shared.h(1721): warning C4324: 'cplane_t': structure was padded due to alignment specifier
```

PR #3271, opened 2025-01-02
Cancel rant trample when using +attack

VReaperV:
This was a really annoying behaviour that required momentarily moving backwards in order to attack an enemy or enemy building while trampling, to cancel trample. That never made any sense, so now the trample will just be cancelled when +attack is used. Also removes the hack that was used to make bots do the above, since this fix also makes them cancel trample when attacking.


DolceTriade:
Didn't test code, but the rationale sounds good.

sweet235:
LGTM looking at everything but whitespace

slipher:
LGTM, the rant could use help

PR #3274, opened 2025-01-03
Fix incorrect MDV animation frames

### LINKED ISSUE #1312: “R_AddMDVSurfaces: no such frame” on some map and weapon md3 model.
Simlilar to the bug affecting md3 bullets (DaemonEngine/Daemon#369 #1212), this time with a map md3 model (the chasm sandals).

[map md3 bad frame](https://dl.illwieckz.net/b/unvanquished/bugs/map-md3-bad-frame/unvanquished_2021-02-03_233242_000.jpg)

[map md3 bad frame](https://dl.illwieckz.net/b/unvanquished/bugs/map-md3-bad-frame/unvanquished_2021-02-03_233247_000.jpg)

VReaperV:
Fixes #1312. I'm not entirely sure that the fix is correct, but it does fix the problem on `chasm`. Also note that we no longer have md3 models in assets, so we can't really test those.

### DIFF
--- a/src/cgame/cg_animmapobj.cpp
+++ b/src/cgame/cg_animmapobj.cpp
@@ -198,7 +198,7 @@ void CG_AnimMapObj( centity_t *cent )
 
 	cent->lerpFrame.animation = &anim;
 
-	if ( !anim.loopFrames )
+	if ( !anim.loopFrames && anim.numFrames > 1 )
 	{
 		// add one frame to allow the animation to play the last frame
 		// add another to get it to stop playing at the first frame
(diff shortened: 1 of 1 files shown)


src/cgame/cg_animmapobj.cpp
```diff
@@ -223,8 +223,13 @@ void CG_AnimMapObj( centity_t *cent )
 		cent->animLastState = !( cent->currentState.eFlags & EF_MOVER_STOP );
 	}
 
-	//run animation
-	CG_AMOAnimation( cent, &ent.oldframe, &ent.frame, &ent.backlerp );
+	if ( anim.numFrames > 1 ) {
```
slipher, line 226: Braces should go on separate line

LGTM apart from formatting

VReaperV: Fixed.

PR #3282, opened 2025-01-07
Vampire mode: revert a commit that should not be here

sweet235:
This is an ugly hack, and does not even do anything, as our bots currently ignore markings on buildings.

I want to keep the quality of life bot commits out of this branch, can do that later.


VReaperV:
I did notice buildings becoming marked when testing locally, though I didn't know bots ignored those.

VReaperV (approved):
LGTM

PR #3287, opened 2025-01-09
Vampire mode: do not abuse the g_BPInitialBudget* cvars

sweet235:
Fix the ugly hack I had started with. Do not modify the cvars `g_BPInitialBudgetAliens` and `g_BPInitialBudgetHumans` anymore. Doing this is probably dangerous, as these are serverinfo cvars.

This also reverts a hacky commit I had added to temporarily allow negative values for said cvars.

Directly modify `level.team[ team ].totalBudget` instead.

### DIFF
--- a/src/sgame/sg_buildpoints.cpp
+++ b/src/sgame/sg_buildpoints.cpp
@@ -83,7 +83,6 @@ void G_UpdateBPVampire( int client ) // -1 to update everyone
 void G_UpdateBuildPointBudgets() {
 	int abp = g_BPInitialBudgetAliens.Get();
 	int hbp = g_BPInitialBudgetHumans.Get();
-	// TODO: maybe make vampire mode consider the current BP budgets
 	for (team_t team = TEAM_NONE; (team = G_IterateTeams(team)); ) {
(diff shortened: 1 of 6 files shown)


src/sgame/sg_buildpoints.cpp
```diff
@@ -77,18 +77,19 @@ void G_UpdateBPVampire( int client ) // -1 to update everyone
 		return;
 	}
 
-	trap_SendServerCommand( client, va( "bpvampire %d %d", g_BPInitialBudgetHumans.Get(), g_BPInitialBudgetAliens.Get() ) );
+	trap_SendServerCommand( client, va( "bpvampire %d %d", static_cast<int>( level.team[ TEAM_HUMANS ].totalBudget ), static_cast<int>( level.team[ TEAM_ALIENS ].totalBudget ) ) );
```
DolceTriade, line 80: Maybe make a helper for this and call it in the callbacks for the g_InitialBudget* cvars? Then we don't need the updatevampirebp command at all

sweet235: The `updateBPVampire` command is not needed anymore once we merge this. Or so I believe.

Issue #3275, opened 2025-01-05
Prifle/blaster can't hit medistation underneath them


VReaperV:
It's back, this time plaguing the humans' ability to hamper their own team: https://users.unvanquished.net/~reaper/test.mp4.

PR #3291, opened 2025-01-10
Fix missiles passing through medistation

slipher:
See #3275

### DIFF
--- a/src/sgame/components/MissileComponent.cpp
+++ b/src/sgame/components/MissileComponent.cpp
@@ -83,12 +83,14 @@ static trace2_t MissileTrace(gentity_t* ent)
 
 		if (trWorld.startsolid)
 		{
-			return trWorld;
+			result = trWorld;
 		}
-
-		trace2_t trBody = G_Trace2(ent->r.currentOrigin, ent->r.mins, ent->r.maxs, trWorld.endpos,
-			passent, CONTENTS_BODY, 0);
-		result = trBody.fraction < 1.0f ? trBody : trWorld;
+		else
+		{
+			trace2_t trBody = G_Trace2(ent->r.currentOrigin, ent->r.mins, ent->r.maxs, trWorld.endpos,
+				passent, CONTENTS_BODY, 0);
+			result = trBody.fraction < 1.0f ? trBody : trWorld;
+		}
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


VReaperV:
This doesn't fix the issue on my end (tested with blaster and prifle, dll sgame).

slipher:
> This doesn't fix the issue on my end (tested with blaster and prifle, dll sgame).

Really? I tested the branch again and can't reproduce the issue with either weapon.

VReaperV:
Yeah, I get this result: https://users.unvanquished.net/~reaper/test2.mp4

src/sgame/components/MissileComponent.cpp
```diff
@@ -83,13 +83,14 @@ static trace2_t MissileTrace(gentity_t* ent)
 
 		if (trWorld.startsolid)
 		{
-			return trWorld;
+			result = trWorld;
+		}
+		else
+		{
+			trace2_t trBody = G_Trace2(ent->r.currentOrigin, ent->r.mins, ent->r.maxs, trWorld.endpos,
```
enneract, line 90: The first trace goes the entire length from `ent->r.currentOrigin` to `origin`. The second trace goes only up to `trWorld.endpos`, making it shorter in the bugged case. Thus the two trace fractions can't be directly compared and the ternary below is broken.

Changing the endpoint of this trace to `origin` fixes the issue.

What is the purpose of `pointAgainstWorld` in the first place?

slipher: > Changing the endpoint of this trace to `origin` fixes the issue.

Sure, making both traces use the same endpoint would be another way to fix it.

> What is the purpose of `pointAgainstWorld` in the first place?

It makes the missile use a point trace against map geometry (CONTENTS_SOLID) but a finite sized box trace against CONTENTS_BODY entities.

enneract: > It makes the missile use a point trace against map geometry (CONTENTS_SOLID) but a finite sized box trace against CONTENTS_BODY entities.

Oh, that makes sense, thanks.

> Sure, making both traces use the same endpoint would be another way to fix it.

It's _the_ fix. Comparing fractions from completely different traces doesn't make any sense. Your code is still bugged.

slipher: With the proposed change there is no longer a comparison of two fractions. Only the comparison of a fraction with 1.0.

enneract: You're right, my bad.

I tested your branch (rebased on top of `v0.55.3`) and everything works fine. I can't reproduce Reaper's issue.

Issue #3284, opened 2025-01-08
Vampire mode thread


sweet235:
We are working on a branch `vampire-mode`. It provides a different build point system than the traditional mining system. The main idea is that there is a fixed amount of build points, and no miners. To gain more build points, a team is required to destroy enemy buildables. This steals build points from the opponent team.

I will try to put all information needed to test and host this branch in this issue.

To enbable the vampire mode, use

```
set g_BPTransfer on
```

This is enough to test it.

To host a server, also use

```
set g_BPTransferFactor 0.5
set g_maxMiners 0
```

VReaperV:
Mining rate should also be set to 0, so miners that exist already in layouts don't mess up the BP. The total BP for teams can be adjusted by setting `g_BPInitialBudget{Humans|Aliens}` and running the `updateBPVampire` command.

cu-kai:
> Mining rate should also be set to 0, so miners that exist already in layouts don't mess up the BP. The total BP for teams can be adjusted by setting `g_BPInitialBudget{Humans|Aliens}` and running the `updateBPVampire` command.

Assuming the `g_BPInitialBudget{Humans|Aliens}` was set by the mapconfig, a `updateBPVampire` won't be required, that's only for if we change it after players entered the game.

PR #3292, opened 2025-01-10
Vampire mode

sweet235:
As explained in https://github.com/Unvanquished/Unvanquished/issues/3284 and seen on the map testing and bug squash central servers.

This is the unchanged development branch from which I removed a few temporary commits.

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -2023,9 +2023,12 @@ public:
 
 		if ( cg.bpVampireTime + 3000 + flashDuration > cg.time )
 		{
-			SetInnerRML( Rocket_QuakeToRML( Str::Format( "^%s%s^%s%s",
+			SetInnerRML( Rocket_QuakeToRML( Str::Format( "^7%d ^%s%s^%s%s ^7%d",
+			                                             cg.bpVampire[ TEAM_HUMANS ],
 			                                             hCol, cg.bpVampireBarH,
-			                                             aCol, cg.bpVampireBarA ).c_str(), RP_EMOTICONS ) );
+			                                             aCol, cg.bpVampireBarA,
+			                                             cg.bpVampire[ TEAM_ALIENS ]
+			                                             ).c_str(), RP_EMOTICONS ) );
 		}
 	}
 
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -2006,6 +2006,64 @@ private:
 	int showTime_ = -1;
 };
 
+class BPVampireElement : public HudElement
+{
+public:
+	BPVampireElement( const Rml::String& tag ) :
+			HudElement( tag, ELEMENT_GAME ) {}
+
+	void DoOnUpdate() override
+	{
+		if ( !cg.bpVampire[ TEAM_ALIENS ] || !cg.bpVampire[ TEAM_HUMANS ] )
+		{
+			return;
+		}
+
+		blink();
+
+		if ( cg.bpVampireTime + 3000 + flashDuration > cg.time )
+		{
+			SetInnerRML( Rocket_QuakeToRML( Str::Format( "^%s%s^%s%s",
+			                                             hCol, cg.bpVampireBarH,
+			                                             aCol, cg.bpVampireBarA ).c_str(), RP_EMOTICONS ) );
+		}
+	}
+
+private:
+	bool flashState = false;
+	int flashDuration = 250;
+	int lastFlashTime = cg.time;
+	std::string hCol = "d";
+	std::string aCol = "i";
+
+	void blink()
+	{
+		if ( ( cg.bpVampireTime + 3000 > cg.time )
+		  && ( lastFlashTime + flashDuration < cg.time ) )
+		{
+			if ( cg.bpVampireOld[ TEAM_HUMANS ] > cg.bpVampire[ TEAM_HUMANS ] )
+			{
+				flashState ? hCol = "d" : hCol = "3";
+			}
+
+			if ( cg.bpVampireOld[ TEAM_ALIENS ] > cg.bpVampire[ TEAM_ALIENS ] )
+			{
+				flashState ? aCol = "i" : aCol = "3";
+			}
+			flashState = !flashState;
+			lastFlashTime = cg.time;
+		}
+
+		if ( cg.bpVampireTime + 3000 < cg.time )
+		{
+			// reset the colours so the bar doesn't get stuck
+			hCol = "d";
+			aCol = "i";
+			flashState = false;
+		}
+	}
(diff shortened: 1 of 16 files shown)


DolceTriade (commented):
looks good. some minor cleanup

src/sgame/sg_combat.cpp
```diff
@@ -198,6 +198,182 @@ static const gentity_t *G_FindKillAssist( const gentity_t *self, const gentity_t
...
+		{
+			initialBP = g_buildPointInitialBudget.Get();
+		}
+		level.team[ team ].vampireBudgetSurplus = level.team[ team ].totalBudget - initialBP;
+	}
+	// vampire mode disables the call to G_FreeBudget in the component's
+	// HandleDie method, and does it here instead
+	// reason: HandleDie is called when the building dies, but we are here
+	// when the building explodes (or vanishes)
+	// the explosion happens several seconds after the death
```
slipher, line 373: Why is it necessary to do the accounting at a different time?

VReaperV: This can be exploited otherwise to build something with BP you should not have between the building being destroyed and the actual explosion.

slipher: I mean why don't we do everything upon buildable death like we do now? Doing some stuff later makes `revert` not work correctly. If you use `revert` right after a buildable dies, its death is reverted but its BP is still stolen.

cu-kai (approved):
I think this is a good start, based on the small number of games I've played so far. Like @DolceTriade said, we should iterate on this as required. I believe it may require more testing and fine-tuning, but with more servers able to access this mode, gathering feedback will be easier.

I do have one request though, @sweet235 can you provide easy-to-understand instructions for server hosts to enable this? For example, are there any other ideal cvar values to be set?

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2006,6 +2006,102 @@ class CenterPrintElement : public HudElement
...
+	std::string hCol = "d";
+	std::string aCol = "i";
+	bool hBlink = false;
+	bool aBlink = false;
+
+	const Color::Color humanBPColor = Color::Color( 0.0f, 0.0f, 1.0f, 0.5f );
+	const Color::Color alienBPColor = Color::Color( 1.0f, 0.0f, 0.0f, 0.5f );
+	const Color::Color blinkBPColor = Color::Color( 1.0f, 1.0f, 0.0f, 0.5f );
+
+	void blink()
```
slipher, line 2072: Use capital letter for the member functions `Blink`

slipher:
> I do have one request though, @sweet235 can you provide easy-to-understand instructions for server hosts to enable this? For example, are there any other ideal cvar values to be set?

Maybe we could add scripts `presets/game/bpvampire-on.cfg` and `presets/game/bpvampire-off.cfg` to set the necessary cvars. Then it could be configured by running the command `preset presets/game/bpvampire-on.cfg`. Later we might add a UI button for that.

src/sgame/sg_combat.cpp
sweet235: We do not do everything upon buildable death, and never have. Some things are done when the health reaches zero, others are done when the buildable explodes. Momentum increase is done at explosion time.

It seems needlessly complicated to do momentum at one time and build points at a different time.

> If you use revert right after a buildable dies, its death is reverted but its BP is still stolen.

Yes. And neither the momentum nor the earned funds are corrected either.

sweet235:
> I do have one request though, @sweet235 can you provide easy-to-understand instructions for server hosts to enable this? For example, are there any other ideal cvar values to be set?

This is explained in https://github.com/Unvanquished/Unvanquished/issues/3284. 

> Maybe we could add scripts presets/game/bpvampire-on.cfg and presets/game/bpvampire-off.cfg to set the necessary cvars.

Maybe later. But I doubt this will be neccessary.

sweet235:
> But I doubt this will be neccessary.

Why do I think that? The only required adjustment is that miners have to be disabled. In case a map layout contains miners, it is also good idea to set the BP per miner to zero.

One solution would be to disable miners and remove existing miners automatically. But this PR is already long enough, so let's save that for a later discussion.

src/sgame/sg_combat.cpp
```diff
@@ -198,6 +198,182 @@ static const gentity_t *G_FindKillAssist( const gentity_t *self, const gentity_t
...
+static Cvar::Cvar<bool> g_BPVampireNotifyTeam("g_BPVampireNotifyTeam", "BP transfer experiment team notifications", Cvar::NONE, true);
+static Cvar::Cvar<float> g_BPVampireFactor("g_BPVampireFactor", "BP transfer factor", Cvar::NONE, 0.5f);
+
+static int bpStolenAtThisFrame[ NUM_TEAMS ];
+static int buildablesDestroyedAtThisFrame[ BA_NUM_BUILDABLES ];
+
+static std::vector<buildable_t> alienBuildables = { BA_A_SPAWN, BA_A_BOOSTER, BA_A_BARRICADE, BA_A_ACIDTUBE, BA_A_TRAPPER, BA_A_SPIKER, BA_A_HIVE, BA_A_OVERMIND };
+static std::vector<buildable_t> humanBuildables = { BA_H_SPAWN, BA_H_MGTURRET, BA_H_ROCKETPOD, BA_H_ARMOURY, BA_H_MEDISTAT, BA_H_REACTOR };
+
+static void ResetDestroyedBuildables( team_t team )
```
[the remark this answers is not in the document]
sweet235: That part of the code shows my uncertainty about what how to do this. I thought that sending such messages to clients at each frame could be too much in some situations. So I tried to keep the code flexible enough to accumulate the destructions over several frames in later PRs.

src/cgame/cg_rocket_draw.cpp
```diff
@@ -2006,6 +2006,102 @@ class CenterPrintElement : public HudElement
...
+private:
+	bool flashState = false;
+	int flashDuration = 250;
+	int lastFlashTime = cg.time;
+	std::string hCol = "d";
+	std::string aCol = "i";
+	bool hBlink = false;
+	bool aBlink = false;
+
+	const Color::Color humanBPColor = Color::Color( 0.0f, 0.0f, 1.0f, 0.5f );
```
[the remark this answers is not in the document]
sweet235: Does this work with something not known at compile time like `Color::Color( 0.0f, 0.0f, 1.0f, 0.5f )`?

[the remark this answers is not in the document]
slipher: How is that not known at compile time?

The thing with constexpr is that the class has to specifically support it, so that's why I suggested static const in case it was not implemented.

slipher:
LGTM. Don't forget to squash! (you can add the Co-Authored-By lines for multiple authors)

PR #3293, opened 2025-01-12
Fix Freetype mess, MinGW native gamelogic build

slipher:
06453a8c96118b09b300c402e0d86d62e5d32b16 does a regression of finding Freetype (and populating the variables) only if "NOT BUILD_CLIENT". We always need to find and set it regardless of what happened in Daemon's CMakeLists! It still worked sometimes because rmlui.cmake had a duplicate call to find_package(Freetype). But rmlui.cmake does not link zlib which is supposed to be a dependency of the external_deps Freetype builds used on Windows and Mac.

Note: with some toolchains the build can succeed even if the zlib dependency is not provided. This happens with Visual Studio and I've seen it on Mac too. I think it must depend on whether the linker removes unused symbols or resolves dependencies first. For MinGW you actually have to link zlib to build.


illwieckz (approved):
LGTM.

PR #3294, opened 2025-01-12
Reduce DEFAULT_MOMENTUM_BUILD_MOD to zero

sweet235:
This is just the default value, servers can choose any value by adjusting `g_momentumBuildMod`.

Getting momentum for building is a huge mistake. There are several reasons:

- It seduces new players into building as much as they can, which is usually a bad strategy.
- When you destroy some buidlings in the enemy base, the enemy team gets almost as much momentum from rebuilding than your team gets from the destruction.
  This leads to very strange situations. If you know the enemy team can rebuild, and you attack for the momentum advantage, you should only destroy enemy miners and the main buildable. These are the only buildings that do not yield momentum when built.
- Knowing that rushing the enemy base will feed them momentum encourages camping, especially early in the game.


cu-kai (approved):

DolceTriade:
I think we should come up with an alternate way to gain momentum before we set this to zero.

sweet235:
This turns out to be controversial. It might be a better idea to redesign the whole momentum system instead of doing incremental changes.

cu-kai:
> I think we should come up with an alternate way to gain momentum before we set this to zero.

Such as attacking the enemy players/base?

slipher:
> It seduces new players into building as much as they can, which is usually a bad strategy.

Do they really build because of momentum? I imagine that many new players build because it seems fun or easier than fighting. If they are going to build poorly for any other reason, that only makes them get punished more.

There used to be a notion that forward bases should be encouraged; measures such as momentum for building and BP for miners were implemented to this end. With BP vampire mode this is hardly going to be a viable strategy, but maybe we are making too many changes in the direction of punishing teams that lose buildables: recent changes include giving credits to players who kill buildables, the vampire mode option, and now this proposal.

Gireen:
How much does the momentum affect most players decision making? I think its not that much beside the vague feeling that attacking and building good is, simply because of its complexity.

In optimal condition, with all players knowing how it works, and in its own logic having build momentum zero is also not desired.
It would change from feeding with destroying of buildings to feeding with building. 

Building is an action that is like destroying, something that is potentially benefiting for the team and gives them literal momentum, so it should also affect the momentum score.

Of cause there is the problem of good and bad building, which could be address with something like momentum over time for buildings that last longer, but that will add again another layer of complexity. 

Instead of completely disabling it i think it would be worth a try to reduce DEFAULT_MOMENTUM_BUILD_MOD to 0.5 or 0.4 and revisit it again later if there is more changes or reverting needed.

cu-kai:
> Instead of completely disabling it i think it would be worth a try to reduce DEFAULT_MOMENTUM_BUILD_MOD to 0.5 or 0.4 and revisit it again later if there is more changes or reverting needed.

On Bunker, both the build and decon modifier are 0.5 which has been the case for a long time.

VReaperV:
I think momentum for building is weird, I don't really see a positive effect from it (other than that one time when I worked around a broken zero-g area on mudkip by building a shitton of things).

With the default values the momentum gain is also *really* noticeable, you'd have to be blind to not put 2 and 2 together when building and take advantage of it.

> Building is an action that is like destroying, something that is potentially benefiting for the team and gives them literal momentum, so it should also affect the momentum score.

The difference is that you have to actually put some effort into destroying a building.

> Of cause there is the problem of good and bad building, which could be address with something like momentum over time for buildings that last longer, but that will add again another layer of complexity.

Good and bad building will keep being a thing as long as building is a core mechanic. You can't have it both be important and impossible to do wrong.

DolceTriade:
I'm not going to block. It is a cvar so we can definitely override if we want. It has been tested and players are like or are ambivalent to the change, which is probably sufficient.

Issue #3273, opened 2025-01-03
Dragoon bots' aim is sometimes wrong


sweet235:
The circumstances I observed so far are:
- Dragoon bot wants to pounce at a human
- Dragoon bot wants ot barb a human

In these cases, it can happen that the bot will suddenly stop aiming correctly, and aim upwards almost vertically. It typically continues to do this for a while.

One symptom of this is reading "_(alien bot)_ was barbed" more often than usual in the log. That happens often when this bug occurs while the bot is below a low ceiling.

My current theory is that this bug happens around the function `CalcAimPitch`. It might be wrong.

sweet235:
I am currently trying to figure out whether this goes away when not using ffast-math.

VReaperV:
> That happens often when this bug occurs while the bot is below a low ceiling.

I thought that was just because the ceiling happens to be in the barb's path for hitting a human? Since the trajectory is parabolic.

sweet235:
I found a way to reproduce this, or something very similar. Load map parpax, and then enter these commands:

```
team a
give momentum
g_bot_level4 off
g_bot_infiniteFunds on
bot add * a 5 attack
```

This will add an alien bot. The bot will evolve to advanced dragoon, and move the human base. It will try to barb at a turret when it arrives. But it will not hit the turret, because it will aim too low. Here is a screenshot:

![unvanquished_2025-01-14_082558_000](https://github.com/user-attachments/assets/8888b381-1bc1-431e-8b31-65f4a3cd3993)

The pitch to use for the barb aim is calculated by a function `CalcBarbAimPitch`. This requested pitch is about 2.7 degrees above the horizontal plane. That seems to be reasonable. But somehow, the bot does not use this pitch.

sweet235:
> I am currently trying to figure out whether this goes away when not using ffast-math.

I do not think removing ffast-math changes anything here.

VReaperV:
There's this commented out part, could this be related somehow?
> //in usrcmd angles, a positive angle is down, so multiply angle by -1
> //botCmdBuffer->angles[PITCH] = ANGLE2SHORT(-angle);

sweet235:
> There's this commented out part, could this be related somehow?

I do not think so. This sign inversion is done by the caller, in

```
botCmdBuffer->angles[PITCH] = ANGLE2SHORT( -CalcBarbAimPitch( self, target ) )
```

We would probably have noticed a wrong sign long ago. This function is only called at one place.

[...]

PR #3295, opened 2025-01-14
Revert some commits leading to strange bot aim bugs

sweet235:
See https://github.com/Unvanquished/Unvanquished/issues/3273


slipher:
Thanks for locating this bug.

I tried to bisect with this but it turns out the commits are reverted in backwards order... the most recently added commit needs to be reverted first, not last.

slipher:
Alternative: #3296

sweet235:
https://github.com/Unvanquished/Unvanquished/issues/3273 has become a little verbose. I think the main problem is that the zero on the pitch axis as seen by the bot's command buffer can be offset for an extended period of time.

The reproduction described there might not work reliably without an unrelated bug I caused. While the dragoon bot falls down the elevator shaft, it mistakenly tries to do an upward move. This makes it aim very high and then horizontally again, it short time. I suppose these sudden pitch changes make the problem more visible.

Code setting the pitch for some bot classes has been around for a long time. We should not allow such history dependent offsets.

slipher:
> I think the main problem is that the zero on the pitch axis as seen by the bot's command buffer can be offset for an extended period of time.

The `delta_angles` value in itself is meaningless. It's the changes over time that matter. The value itself equals the sum of all angular disturbances experienced by a client over its lifetime.

cu-kai (approved):
While I have no technical input on this code, as a server host, this does indeed fix the bug with the dragoons.

DolceTriade:
I think we should merge this and then try fixes after.

PR #3296, opened 2025-01-15
Dragoon pounce & barb aiming fix

### LINKED ISSUE #3273: Dragoon bots' aim is sometimes wrong

slipher:
To fix #3273. The 2nd commit is the one to fix it; 1st is just unimportant cleanup.

### DIFF
--- a/src/sgame/sg_bot.cpp
+++ b/src/sgame/sg_bot.cpp
@@ -542,6 +542,12 @@ void G_BotThink( gentity_t *self )
 	// if we were nudged...
 	VectorAdd( self->client->ps.velocity, nudge, self->client->ps.velocity );
 
+	// for real clients this is read off the network as a 16-bit value
+	for ( int &angle : self->botMind->cmdBuffer.angles )
+	{
+		angle &= 65535;
+	}
+
 	// ensure we really want to sprint or not
 	self->client->pers.cmd = self->botMind->cmdBuffer;
 	self->botMind->doSprint(
--- a/src/sgame/sg_bot_nav.cpp
+++ b/src/sgame/sg_bot_nav.cpp
@@ -1015,8 +1015,7 @@ void BotMoveUpward( gentity_t *self, glm::vec3 nextCorner )
 		{
 			self->botMind->cmdBuffer.forwardmove = 127;
 			self->botMind->cmdBuffer.rightmove = 0;
-			float pitch = AngleSubtract( -60.f, SHORT2ANGLE( self->client->ps.delta_angles[ PITCH ] ) );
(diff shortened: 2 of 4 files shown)


sweet235:
I am still trying to understand the consequences of https://github.com/Unvanquished/Unvanquished/pull/3057, even though I spent a considerable amount of time debugging them. Is it correct that https://github.com/Unvanquished/Unvanquished/pull/3057 changed the interface between the bot logic and the core game logic, for all bot classes?

My current theory is that the angles in the bot's command buffer do not correspond to the viewangles anymore. They used to be the same, except that the command buffer uses 65536 instead of 360.0. It seems that now, an additional linear transformation has to be applied to the command buffer angles in order to work as they used to. This transformation is required for all classes, and is specified by the delta angles. Is that correct?

slipher:
> Is it correct that #3057 changed the interface between the bot logic and the core game logic, for all bot classes?

Sort of? In all cases, the equation `bot's output angle[i] = botMind->cmdBuffer.angles[i] + ps.delta_angles[i]` should be enforced. Before #3057, bot code that wanted to set an angle either did (a) set `delta_angles[i]` to 0 and write the desired angle to `cmdBuffer`, or (b) only write the desired angle to `cmdBuffer`. Method (a) works correctly; method (b) is of course wrong but the method (a) code tended to be run frequently enough that the delta angles were almost always 0. The bug in #3057 is I did not notice some of the places using the wrong method (b), so they stopped working.

To make the bots be (sometimes) affected by chaingun jitter and trapper grab, I had to stop setting the delta angles to 0 all the time because those effects are done by setting delta_angles.

sweet235:
> To make the bots be (sometimes) affected by chaingun jitter and trapper grab

The bots were affected by the trapper grab before https://github.com/Unvanquished/Unvanquished/pull/3057.

I will have dig into this to gain further insights about the chaingun. But I wonder if causing all this trouble just for the chaingun bots is really worth it.

In any case, this seems to be a concerted effort by the project heads. Changing such interfaces to several people's work should not be done sneakily. If it is a done at all, a thorough revisit of the affected code places and extensive testing is required. Let me ping @illwieckz and @DolceTriade for clarification.

sweet235:
LGTM. But I did not test this extensively yet.

PR #3300, opened 2025-01-21
Fix acid tube gas being too transparent

slipher:
Regression in c794469d99ba836ca883c66d42013d9e96e6ed00, which changed the denominator of the fraction in CG_CalculateTimeFrac from (cg.time - birth - delay) to (cg.time - birth + delay). The acid cloud was affected because the `alpha` field uses a delay.


sweet235:
LGTM

DolceTriade (approved):

PR #3302, opened 2025-01-22
Vampire mode: fix bug when recently damaged building is marked and replaced

sweet235:
Vampire mode does the call to `G_FreeBudget` when the enemy team is rewarded, and not when the building dies (reaches zero health). However, for historic reasons, this call must not be done when the means of death is one of `MOD_DECONSTRUCT, MOD_REPLACE, MOD_BUILDLOG_REVERT`. Fix this by excluding these cases.

### DIFF
--- a/src/sgame/components/AlienBuildableComponent.cpp
+++ b/src/sgame/components/AlienBuildableComponent.cpp
@@ -64,7 +64,7 @@ void AlienBuildableComponent::Blast(int /*timeDelta*/) {
 	Entities::AntiHumanRadiusDamage(entity, ba->splashDamage, ba->splashRadius, ba->meansOfDeath);
 
 	// Reward attackers.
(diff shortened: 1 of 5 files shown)


slipher:
Is there any good reason that we do any of that team rewarding stuff when the buildable explodes, instead of when it dies? Seems like it would be simpler to do all the stuff on death.

This is a non-blocking comment.

sweet235:
> Is there any good reason that we do any of that team rewarding stuff when the buildable explodes, instead of when it dies?

I have asked myself the same question. Sadly, I have no answer so far. Which is why I have to assume there is some reason I do not know.

sweet235:
Seems like i cannot reopen this. I guess nobody cares anyway.

sweet235:
Ah, now i can.

PR #3303, opened 2025-01-23
Fix realLight keyword for particle coloring

slipher:
A 0-1 color needed to be rescaled to 0-255. This regression may have been caused by Daemon 07251fce, which appears to change the R_LightForPoint results from a 0-255 scale to a 0-1 scale. But I didn't verify.

For now, realLight is only really fixed with overbright clamping enabled, as R_LightForPoint does not scale with the light factor as it should.

Before and after (both with `r_overbrightDefaultClamp 1`):
unvanquished-sectorb17-smoke-particle-reallight
unvanquished-sectorb17-smoke-particle-reallight

### DIFF
--- a/src/cgame/cg_particles.cpp
+++ b/src/cgame/cg_particles.cpp
@@ -2507,7 +2507,16 @@ static void CG_RenderParticle( particle_t *p )
 		if ( bp->realLight )
 		{
 			vec3_t alight, dlight, lightdir;
+
+			// FIXME: at the time of writing, this API is broken with full-range overbright as it
+			// does not include the lightFactor
+			// But clamp it in case that is fixed later (which could lead to values greater than 1.)
+			// This FIXME also applies to trail realLight.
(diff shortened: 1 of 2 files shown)


VReaperV (approved):
I think a proper fix would be to just make particles use lightMapping shader with lightGrid, but LGTM on this as an intermediary solution.

VReaperV:
The same issue also seems to exist with trails.

slipher:
> I think a proper fix would be to just make particles use lightMapping shader with lightGrid, but LGTM on this as an intermediary solution.

Yeah, I'm not sure if we should fix R_LightForPoint, add a shader register flag to force lightmap enablement, or just NUKE the feature and put some constant gray on the vertex color. There are only 4 maps with particle systems that correctly use `realLight`.

> The same issue also seems to exist with trails.

The commit already fixed trails too but I forgot to mention it. Updated the commit message.

PR #3304, opened 2025-01-25
vampire mode: re-enable the BP queue, free BP as in mining

sweet235:
There is a bug in the vampire mode. When a building under construction is destroyed, its BP are not handled correctly. Not even https://github.com/Unvanquished/Unvanquished/pull/3302 seems to fix this. I observed a case where the available budget shown to a builder was 3, but the total budget minus the budget used by existing buildings was 33. So the builder should have had 33 BP available. I do not know the exact cause.

The first vampire commit was too ambitious: we disable the BP queue completely when vampire mode is enabled. Sadly, the vampire branch has been squashed down to a single commit, so we cannot revert the problematic commits anymore.

I propose we re-enable the BP queue for now. This makes the vampire mode playable.  A good value for the BP regen rate is 90 per minute. Any value that does not regen 8 or more BP in 4 seconds will not result in problems. 4 seconds is the time between a building's death and its explosion. 8 BP is the cost of the cheapest buliding.

### DIFF
--- a/src/sgame/components/BuildableComponent.cpp
+++ b/src/sgame/components/BuildableComponent.cpp
@@ -86,9 +86,7 @@ void BuildableComponent::HandleDie(gentity_t* killer, meansOfDeath_t meansOfDeat
 
 	// If destroyed violently, add all build points to queue.
 	if (meansOfDeath != MOD_DECONSTRUCT && meansOfDeath != MOD_REPLACE && meansOfDeath != MOD_BUILDLOG_REVERT) {
-		if (!g_BPVampire.Get()) {
-			G_FreeBudget(team, 0, BG_Buildable(entity.oldEnt->s.modelindex)->buildPoints);
-		}
+		G_FreeBudget(team, 0, BG_Buildable(entity.oldEnt->s.modelindex)->buildPoints);
 	}
 }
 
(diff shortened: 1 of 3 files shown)


sweet235:
Let me elaborate a little. When a building is destroyed, there are two important points of time: the time of death (health reaches zero) and the time of explosion. We do several things when a building is destroyed:

- Reward attackers
- Adjust the used BP budget, put BP in the recovery queue
- When vampire is on: transfer some BP to the other team

Adjusting the BP budget and putting BP in the queue is done at death. This means that BP recovery will start 4 seconds before the explosion. Rewarding attackers is done at explosion time. The current vampire implementation is done when rewarding attackers, as it depends on when the enemy team last damaged the building.

There are several cases where the death and the explosion time are close or equal, like deconstructing, marking and replacing or buildlog reverting.

It has been suggested that we simplify this by doing everything at one of these two points in time. I do not see any reason not to do this.

slipher:
> We do several things when a building is destroyed:
> 
>     * Reward attackers
> 
>     * Adjust the used BP budget, put BP in the recovery queue
> 
>     * When vampire is on: transfer some BP to the other team

And don't forget recording the death in the build log. That's done immediately so it is a source of bugs with reverting effects that only happens a few seconds later.

cu-kai (approved):
Confirmed bug fixed on Bunker.

slipher:
The BP bugs could be caused by failing to call `G_FreeBudget` if a buildable was not damaged by enemies. It's relying on the call stack `G_RewardAttackers -> TransferBPToEnemyTeam -> G_FreeBudget`, but `G_RewardAttackers` returns early if there was no damage by enemies.

DolceTriade (approved):
Incremental fixes are fine imo

PR #3306, opened 2025-01-26
Use RSF_FORCE_LIGHTMAP

VReaperV:
Engine-side pr: https://github.com/DaemonEngine/Daemon/pull/1523

If this flag is used when registering a shader, all `ST_COLORMAP` stages will be changed to `ST_DIFFUSEMAP`. Use it for particles and trails, to make them use the lightGrid in the GLSL code.

Improves performance with particle systems and trails, e. g. with firebombs, and a massive performance improvement on map https://users.unvanquished.net/~sweet/pkg/map-jota_3.0.dpk in the outside area.

There's for the most part no difference in how the particles/trails look, the only differences are due to the GLSL lightGrid being more precise.

Also NUKED the now unused lightGrid code that involved IPC calls.

### DIFF
--- a/src/cgame/cg_consolecmds.cpp
+++ b/src/cgame/cg_consolecmds.cpp
@@ -351,8 +351,7 @@ static void CG_CompleteItem()
 
 static void CG_TestCGrade_f()
 {
-	qhandle_t shader = trap_R_RegisterShader(CG_Argv(1),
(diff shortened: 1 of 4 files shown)


VReaperV:
Added a deprecation warning for `realLight` here.

slipher:
LGTM

PR #3307, opened 2025-01-29
NUKE stupid random delay in weapon reload animation

VReaperV:
This is already accounted for in the reload time, extra 250ms makes it not line up correctly. That code originates from Tremulous, however Tremulous did not have any proper reload animations, so it was lining up properly there. The change itself has a minimal impact on gameplay if noticeable, but removes the annoyance from reload animation.


slipher:
Unfortunately the reload time is a load-bearing part of the gameplay, that gives weaker aliens a chance to attack humans. It's annoying when aesthetic and gameplay things are intertwined like this.

DolceTriade:
Hm, so we were always adding 250ms to the reload time? Regardless of whether we should add this time back explicitly by bumping up all the reload times, I think having this random delay is bad. We can bump reload times if we feel this is unfair.

PR #3309, opened 2025-01-30
vampire mode: fix a bug related to the marked BP budget

sweet235:
no time to type, github will break again soon


DolceTriade:
Please actually describe the bug you're fixing for posterity.

sweet235:
Thank goodness, github is stable once again.

To see the bug, spend all your BP. Then mark a building, let the enemy team damage it, and in the next 5 seconds, replace it by a new building. After the dust has settled, your team is using more BP than you have.

The reason is that we subtract half of the building's BP when it is replaced and was damaged recently. This PR makes the marked BP counting aware of that.

slipher:
Maybe it would make more sense to forbid deconstruction (both direct and via marking) of recently attacked buildables? That way builder would not accidentally take an action that just harms their team.

LGTM I guess, still better than the status quo

Issue #3138, opened 2024-10-21
Stop players from connecting to incompatible servers


DolceTriade:
Currently, we have logic to detect whether or not a player is allowed to connect to the server. However, that logic only triggers after pak downloads. This is problematic because if using 0.55 to connect to a 0.54 server and vice versa, you have to download hundreds of MBs of pkgs before hitting that condition.

One potentially simple change is that moving forward, we report the major version as the protocol so we only see the current compatible servers (or just bump the protocol number on breaking releases).

Otherwise, we need to do something else to detect incompatible servers.

sweet235:
This might be even more problematic. Once you downloaded the 0.55 packages with the 0.54 engine, the 0.55 packages will probably be the new default for local games.

Gireen:
Related issues
#405
#726

illwieckz:
One thing I want is to prevent autodownload of DPK using official names if not served by the server with a prefix directory.

DolceTriade:
I think incompatible servers should not show up at all.

illwieckz:
Then protocol number is the tool for that.

slipher:
Just erasing compatible servers is not that appealing to me... it makes it seem like the master server is down or encourages people to play on servers that weren't updated.

DolceTriade:
I don't follow? If incompatible servers aren't visible how can they be connected to them?

slipher:
If only compatible servers show up and you are using an out of date client, you can only connect to out of date servers.

Unless what you mean is to have the master server ban all servers that aren't on the latest version?

DolceTriade:
That's fine? If you're on 0.54 today, you should be able to connect to 0.54 servers. We should have a real mechanism to convince people to upgrade:
 - news ingame
 - or a place holder server to tell people to update :)

VReaperV:
Fixed by #3310.

Issue #726, opened 2015-05-15
Tell player if their client version is different to server version


ckol:
Even better:
- Server version should be shown in red in server list when it doesn't match /or is older
  than client version.
- Server should check if it's up-to-date and send email to maintainer when it should be updated.

This Issue would then replace #725 and bad user experience.

Amanieu:
Oh boy: http://www.catb.org/jargon/html/Z/Zawinskis-Law.html

I'm personally in favor of just using the protocol version to only show servers that are compatible.

Viech:
Haha, @Amanieu and I had the same discussion recently. I disagree with servers sending mail but I prefer if all servers are visible in the server browser, including incompatible ones, which are marked in some way and can be hidden via a checkbox.

guillaume-uH57J9:
Just encoutered this issue with the followings:

client: unvanquished 0.39
server: unvanquished 0.38

Looks like the client is not forward compatible, nor backward compatible. 
So the warning should be triggered should be triggered for any version difference (not just if the version is newer)

ckol:
Well, there is actually no need to pop-up up a warning. It would be annoying if you know what you are doing and need to get rid of the message each time. It should be sufficient to show version numbers and it could be up to the users to decide if they want to try it or not.

For the novice users, if the versions are the same, they coud show up with an additional green "OK" mark. Otherwise with an red "!" exclamation mark. When you mouse-over, it sould explain what it means.

Zazinskis Law is one thing, but KISS for the users or server admins are another thing. Bring the information where it's needed and present it in the most non-intrusive clear way.

(I've been doing software verification in the industrial area in the past where programmers
tend to hack functionality into the code until the pigs started to fly. But the operators in
low-production-cost countries were people kept on minimum wage which were in charge
to keep the machines up and running night and day without much sleep...)

guillaume-uH57J9:
@ckol +1 

The warning does no need to be a pop-up, it can just be a warning icon, or an exclamation point besides the version number.

Veyrdite:
Nothing has happened yet to fix this and it's a pain for people with old versions of the game.  When they join a newer server they download the latest pk3 (eg unvanquished_0.41.pk3) and then their game crashes because of VM issues between the new package and old engine.  Their game is then broken (will not start, tries to load newest pk3) until they update.

No where along this line are they told what's happening, unless they're wise enough to understand why the game is trying to downloaded the latest unvanquished pk3.

[...]

Issue #405, opened 2014-08-10
Make it easier to know which version of the gamelogic is running


Kangz:
It is difficult to know which version of the gamelogic you are running and it often leads to bugs which are difficult to understand for users (and devs too, I had EU run under 0.29 instead of 0.30 for a week). Some small improvements would make that problem go away:
- Have a ROM cvar in both gamelogics with the latest git tag's name and one with the git commit hash.
- Send the content of these cvars at the gamelogic registration to be printed to the console.
- Have the server latest tag cvar sent as serverinfo to be displayed on the server list.

Then users and server operators won't be able to miss version problems.
- Mentor: Kangz
- Knowledge: Some UI, VM stuff, Cvars
- Difficulty: 4

[Mentored projects on the Unv wiki](http://unvnet.net/wiki/index.php/Mentored_Coding_Project)

Viech:
A few notes on this:

* We have to consider *four* versions: Client, server, client gamelogic, server gamelogic. Both server and its gamelogic's version need to be displayed in the server browser some way (one could be displayed in a column, both would be displayed on mouse-over).
* If a base version (e.g. 0.36) doesn't match, users should be prevented from connecting except if they set a certain cvar (cf. connecting to servers that load pk3dir).

Amanieu:
We have several mechanisms to deal with API versions, all of which we are not making use of...

GAME_API_VERSION
CGAME_IMPORT_API_VERSION
PROTOCOL

VReaperV:
Fixed by #3310. Although it doesn't use specifically git tags/commit hashes for that, it still has the same effect as this proposal.

PR #3310, opened 2025-01-31
Show server versions and a mismatch pop-up

### LINKED ISSUE #3158: Option to launch updater if an update is detected
It seems that some amount of people trying to connect to servers launch daemon directly rather than the updater (I do that as well) and click through to the active server (there's usually just one so it's not like you need to look through them to decide which one to join), without updating.

I propose that if either by getting info from master server or by looking at the versions on the active servers the game can realise there's an update, it should create a popup about it, with a button to close the game and start up the updater (if the updater is found in install directory that is).

VReaperV:
Fixes #3158, #3138, #726, and potentially #405.

Add a `version` field to the server list, which is the engine version used to build the server. If the version is not compatible, show it as `^1!version!` instead, and show a pop-up about version mismatch. Also fixed sort by name for local servers.

Here's how this looks in the server browser when the versions match:
srvb0
If they don't match:
srvb1
And a pop-up if trying to connect to such a server:
srvb2

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/server_mismatch.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/server_mismatch.rml
@@ -6,9 +6,9 @@
 	</head>
 	<body id="server_mismatch" template="window" class="window-options" style="width: 40%; margin-top: 40%; text-align: center; z-index: 30">
 		<h1 style="text-align: center"><translate>Engine version mismatch</translate></h1>
-		<span ><translate>This server uses a different Daemon engine version and may not be compatible.</translate></span>
+		<span ><translate>This server uses a different Daemon engine version, which is not compatible.</translate></span>
 		<br/>
-		<span><translate>The latest Daemon release can be downloaded through the Updater app, at <web dest="download"/>, or in one of the official package repositories.</translate></span>
+		<span><translate>The latest Unvanquished release can be downloaded through the Updater app, at <web dest="download"/>, or in one of the official package repositories.</translate></span>
 		<button style="width:100%; margin-top: 2%;" onclick='Cmd.exec("connectToCurrentSelectedServer")'><translate>Connect anyway</translate></button>
 	</body>
 </rml>
--- a/pkg/unvanquished_src.dpkdir/ui/server_mismatch.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/server_mismatch.rml
@@ -1,0 +1,14 @@
+<rml>
+	<head>
+		<link type="text/rcss" href="/ui/shared/basics.rcss" />
+		<link type="text/template" href="/ui/shared/window.rml" />
+		<link type="text/rcss" href="menu.rcss" />
+	</head>
+	<body id="server_mismatch" template="window" class="window-options" style="width: 40%; margin-top: 40%; text-align: center; z-index: 30">
+		<h1 style="text-align: center"><translate>Engine version mismatch</translate></h1>
+		<span ><translate>This server uses a different Daemon engine version and may not be compatible.</translate></span>
+		<br/>
+		<span><translate>The latest Daemon release can be downloaded through the Updater app, at <web dest="download"/>, or in one of the official package repositories.</translate></span>
+		<button style="width:100%; margin-top: 2%;" onclick='Cmd.exec("connectToCurrentSelectedServer")'><translate>Connect anyway</translate></button>
+	</body>
+</rml>
--- a/src/cgame/cg_rocket_datasource.cpp
+++ b/src/cgame/cg_rocket_datasource.cpp
@@ -36,7 +36,7 @@ Maryland 20850 USA.
 #include "cg_local.h"
 #include "shared/parse.h"
 
-static bool AddToServerList( const char *name, const char *label, const char* version, const char* abiVersion, int clients, int bots,
+static bool AddToServerList( const char *name, const char *label, std::string version, std::string abiVersion, int clients, int bots,
 	int ping, int maxClients, char *mapName, char *addr, int netSrc )
 {
 	server_t *node;
@@ -60,9 +60,9 @@ static bool AddToServerList( const char *name, const char *label, const char* ve
 	node->maxClients = maxClients;
 	node->addr = BG_strdup( addr );
 	node->label = BG_strdup( label );
-	node->version = BG_strdup( version );
+	node->version = version;
 	node->mapName = BG_strdup( mapName );
-	node->abiVersion = BG_strdup( abiVersion );
+	node->abiVersion = abiVersion;
 	rocketInfo.data.serverCount[ netSrc ]++;
 	return true;
 }
@@ -316,11 +316,11 @@ void CG_Rocket_BuildServerList()
 			maxClients = atoi( Info_ValueForKey( info.c_str(), "sv_maxclients" ) );
 			Q_strncpyz( mapname, Info_ValueForKey( info.c_str(), "mapname" ), sizeof( mapname ) );
 
-			const std::string version = Info_ValueForKey( info.c_str(), "daemonver" );
-			const std::string abiVersion = Info_ValueForKey( info.c_str(), "abi" );
+			const std::string version = Info_ValueForKey( info.c_str(), "version" );
+			const std::string abiVersion = Info_ValueForKey( info.c_str(), "abiVersion" );
 			rocketInfo.data.haveServerInfo[ netSrc ][ i ] =
 				AddToServerList( Info_ValueForKey( info.c_str(), "hostname" ), trustedInfo.featuredLabel,
-					version.c_str(), abiVersion.c_str(), clients, bots, ping, maxClients, mapname, trustedInfo.addr, netSrc );
+					version, abiVersion, clients, bots, ping, maxClients, mapname, trustedInfo.addr, netSrc );
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


slipher:
Checking the engine and/or gamelogic version string for an exact match is not quite right, since 0.55.2 clients can happily play on a 0.55.0 server and vice versa. What we should do is compare `version.daemon.abi` (added in 0.55.1) from the server's infostring with the `IPC::SYSCALL_ABI_VERSION` constant. However, `version.daemon.abi` isn't provided now by the `trap_LAN_GetServerInfo` API. So I opened https://github.com/DaemonEngine/Daemon/pull/1541 to make it accessible.

I think it would be good to keep displaying the point versions, as in this PR (and https://unvanquished.net/servers), while coloring it red if `version.daemon.abi` doesn't match. That way we can present information about both kinds of version without the excessive verbosity of two version number columns.

VReaperV:
That sounds reasonable.

slipher:
> However, `version.daemon.abi` isn't provided now by the `trap_LAN_GetServerInfo` API. So I opened [DaemonEngine/Daemon#1541](https://github.com/DaemonEngine/Daemon/pull/1541) to make it accessible.

That bit is now merged.

VReaperV:
Hmm, it looks as though it should be possible to use the server info string to get the version, but I can't get `trap_LAN_ServerStatus()` to work.

VReaperV:
I've rebased these prs on `for-0.56.0/sync`, but so far I still had to put version and ABI version in the smaller info string that we use to build the server list. It could be changed to use the full info string, but then there'd hae to be a longer delay either before the server list is built at all, or before the proper version is displayed.

slipher:
My mistake: there is no ABI version available in the `getinfo` response. The server info response (`SVC_Info()`) only includes a hand-picked list of values. I confused it with the `getstatus` response which includes all SERVERINFO cvars.

But the server info response doesn't include any other `version` field either! I wonder how you made the screenshots in the OP. When I try it I get all empty strings for the versions, which makes sense since SVC_Info doesn't provide anything.

I opened https://github.com/DaemonEngine/Daemon/pull/1547 to actually provide the ABI and Daemon versions.

VReaperV:
Yeah, I still had to modify `SVC_Info()` in https://github.com/DaemonEngine/Daemon/pull/1539/files to get the version.

VReaperV:
Rebased on `for-0.56.0/sync`.

slipher (commented):
Tested successfully. Just some little things to clean up

pkg/unvanquished_src.dpkdir/ui/server_mismatch.rml
```diff
@@ -0,0 +1,14 @@
+<rml>
+	<head>
+		<link type="text/rcss" href="/ui/shared/basics.rcss" />
+		<link type="text/template" href="/ui/shared/window.rml" />
+		<link type="text/rcss" href="menu.rcss" />
+	</head>
+	<body id="server_mismatch" template="window" class="window-options" style="width: 40%; margin-top: 40%; text-align: center; z-index: 30">
+		<h1 style="text-align: center"><translate>Engine version mismatch</translate></h1>
+		<span ><translate>This server uses a different Daemon engine version and may not be compatible.</translate></span>
```
slipher, line 9: Wording could be stronger since we are now using the ABI version that really predicts compatibility. `and may not be compatible` -> `which is not compatible`

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -59,8 +60,9 @@ static bool AddToServerList( const char *name, const char *label, int clients, i
 	node->maxClients = maxClients;
 	node->addr = BG_strdup( addr );
 	node->label = BG_strdup( label );
+	node->version = BG_strdup( version );
```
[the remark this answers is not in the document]
VReaperV: Changed to std::string.

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -336,6 +341,16 @@ void CG_Rocket_BuildServerList()
 		Info_SetValueForKey( data, "maxClients", va( "%d", rocketInfo.data.servers[ netSrc ][ i ].maxClients ), false );
 		Info_SetValueForKey( data, "addr", rocketInfo.data.servers[ netSrc ][ i ].addr, false );
 		Info_SetValueForKey( data, "label", rocketInfo.data.servers[ netSrc ][ i ].label, false );
+
+		std::string version;
+		if ( !Q_stricmp( rocketInfo.data.servers[netSrc][i].abiVersion, IPC::SYSCALL_ABI_VERSION ) ) {
+			version = rocketInfo.data.servers[netSrc][i].version;
+		} else {
+			version = Str::Format( "^1!%s!", rocketInfo.data.servers[netSrc][i].version );
+			Log::Warn( rocketInfo.data.servers[netSrc][i].abiVersion );
```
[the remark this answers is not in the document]
VReaperV: Done.

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -483,16 +499,40 @@ static void CG_Rocket_FilterServerList( const char *table, const char *filter )
...
 			Info_SetValueForKey( data, "addr", rocketInfo.data.servers[ netSrc ][ i ].addr, false );
 			Info_SetValueForKey( data, "label", rocketInfo.data.servers[ netSrc ][ i ].label, false );
+			Info_SetValueForKey( data, "version", rocketInfo.data.servers[netSrc][i].version, false );
 
 			Rocket_DSAddRow( "server_browser", str, data );
 		}
 	}
 }
 
+static Cvar::Cvar<int> cg_currentSelectedServer( "cg_currentSelectedServer",
```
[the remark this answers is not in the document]
VReaperV: Fixed.

src/sgame/sg_main.cpp
```diff
@@ -474,6 +474,7 @@ void G_InitGame( int levelTime, int randomSeed, bool inClient )
 
 	Log::Notice( "------- Game Initialization -------" );
 	Log::Notice( "gamename: %s", GAME_VERSION );
+	Log::Notice( "Engine version: %s", ENGINE_VERSION );
```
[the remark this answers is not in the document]
VReaperV: I see, removed that line.

pkg/unvanquished_src.dpkdir/ui/server_mismatch.rml
VReaperV: Hmm, true, changed it to that.

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -483,16 +498,39 @@ static void CG_Rocket_FilterServerList( const char *table, const char *filter )
...
+class ConnectToCurrentSelectedServerCmd : public Cmd::StaticCmd {
+	public:
+	ConnectToCurrentSelectedServerCmd() : StaticCmd( "connectToCurrentSelectedServer",
+		Cmd::CLIENT, "Connect to cg_currentSelectedServer" ) {}
+
+	void Run( const Cmd::Args& ) const override {
+		const int netSrc = cg_currentSelectedServer;
+		Rocket_DocumentAction( "server_mismatch", "close" );
+		trap_SendConsoleCommand(
+			Str::Format( "connect %s", rocketInfo.data.servers[netSrc][rocketInfo.data.serverIndex[netSrc]].addr ).c_str()
```
slipher, line 519: Storing an index can have bugs with indexing out of bounds if the server list is refreshed.

Optional suggestion: how about using a cvar instead of a command? Since exposing a cvar seems less ugly than an internal use command. Like store the server address in a cvar `ui_serverMismatchAddr`, then in the document put
`onclick=Cmd.exec('connect \"$ui_serverMismatchAddr$\"')`

Or you could just have `cg_currentSelectedServer` be a string instead of an index.

VReaperV: Cvar is what was used there before the latest commit/review, albeit it indexed `servers` rather than holding a string.

slipher: The original version had a command *and* a cvar. My new idea is to get rid of the command and have only a cvar.

VReaperV: I don't think that would work due to `Rocket_DocumentAction()` use in the command.

VReaperV: Now fixed.

slipher: Rocket_DocumentAction is not really needed since `connect` always kills the cgame. But if it were, you can do that from an RML event handler.

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -483,16 +498,38 @@ static void CG_Rocket_FilterServerList( const char *table, const char *filter )
...
 		}
 	}
 }
 
+static std::string cg_currentSelectedServer;
+
+class ConnectToCurrentSelectedServerCmd : public Cmd::StaticCmd {
+	public:
+	ConnectToCurrentSelectedServerCmd() : StaticCmd( "connectToCurrentSelectedServer",
+		Cmd::CLIENT, "Connect to cg_currentSelectedServer" ) {}
```
slipher, line 513: Change the description to "For internal use" or something? It shouldn't sound like the command is usable for something.

LGTM other than that

src/cgame/cg_rocket_datasource.cpp
VReaperV: Without that the error window gets stuck on the connection screen. In fact, the whole main menu is always drawn on it, the stuff on the connection screen just ends up being drawn later.

src/cgame/cg_rocket_datasource.cpp
VReaperV: Good point, changed the description.

PR #3312, opened 2025-02-04
cg_main: reorder the game data loading

illwieckz:
Reorder the game data loading.

- Load map geometry at first so map entities are read and texture options are read before loading textures.
- Load game data explicitly with explicit options before loading configs that may load them implicitly.

Required by:

- https://github.com/DaemonEngine/Daemon/pull/1543

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -425,12 +425,12 @@ static void CG_UpdateMediaFraction( float fraction )
 
 enum cgLoadingStep_t {
 	LOAD_START = 0,
+	LOAD_GEOMETRY,
 	LOAD_TRAILS,
 	LOAD_PARTICLES,
+	LOAD_ASSETS,
 	LOAD_CONFIGS,
 	LOAD_SOUNDS,
-	LOAD_GEOMETRY,
-	LOAD_ASSETS,
 	LOAD_WEAPONS,
 	LOAD_UPGRADES,
 	LOAD_CLASSES,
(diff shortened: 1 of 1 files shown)


DolceTriade (approved):
Didn't test but looks reasonable

src/cgame/cg_main.cpp
```diff
@@ -1313,17 +1308,26 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 
 	CG_SetMapNameFromServerinfo();
 
+	CG_UpdateLoadingStep( LOAD_GEOMETRY );
 	// load the new map
 	CM_LoadMap(cgs.mapname);
 
 	CG_InitMinimap();
 
+	/* Always do this before loading any texture, as texture options may have
```
slipher, line 1317: If you are going to say this please link to the issue/PR, since this code is not merged and may end up working in a different way, or have that issue fixed soon afterward.

slipher (commented):
That's a very hacky way to solve the sRGB code's issues, but it is a good idea for a different reason: see https://github.com/DaemonEngine/Daemon/issues/1079#issuecomment-2641851730. This forces some of the translucent weapon effects to always in front of translucent map surfaces, rather than always behind them.

LGTM modulo comment

VReaperV:
> That's a very hacky way to solve the sRGB code's issues, but it is a good idea for a different reason: see [DaemonEngine/Daemon#1079 (comment)](https://github.com/DaemonEngine/Daemon/issues/1079#issuecomment-2641851730). This forces some of the translucent weapon effects to always in front of translucent map surfaces, rather than always behind them.
> 
> LGTM modulo comment

I assume this just creates the opposite problem: every projectile etc. will be rendered in front of map translucent surfaces that it might actually be behind. This does seem to improve things though, as this problem is comparatively rare.

slipher:
> I assume this just creates the opposite problem: every projectile etc. will be rendered in front of map translucent surfaces that it might actually be behind. This does seem to improve things though, as this problem is comparatively rare.

Correct.

src/cgame/cg_main.cpp
illwieckz: Done.

PR #3316, opened 2025-02-12
presets: disable bloom for now

illwieckz:
Disable bloom because it is not usable with none of our current maps.

It was enabled by mistake because it didn't looked bad while the light was mistakenly clamped. The buggy clamping prevented bloom to happen most of the time. A bad computation was breaking another bad computation and they were cancelling themselves together.

The future of bloom is either:

- We NUKE bloom entirely.
- We enable bloom on maps having lights computed in linear space, and we disable it on maps having lights not computed in non-linear space.
- We enable bloom on maps having lights computed in linear space, and we enable it on maps having lights not computed in non-linear space BUT with a different set of default values specially crafted for such maps.

The current bloom option and values have no future with all maps that have been already published and that will be published before maps use lights computed in linear space.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/presets/graphics/high.cfg
+++ b/pkg/unvanquished_src.dpkdir/presets/graphics/high.cfg
@@ -11,7 +11,7 @@ seta r_specularMapping 1
 seta r_physicalMapping 1
 seta r_reliefMapping 0
 seta r_heatHaze 1
-seta r_bloom 1
+seta r_bloom 0
 seta r_ssao 1
 seta r_subdivisions 4
 seta r_imageFitScreen 1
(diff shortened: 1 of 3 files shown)


illwieckz:
We should do a point release distributing that in a near future.

VReaperV:
Should this be re-evaluated once the recent pr's that change lighting are merged?

slipher:
> Should this be re-evaluated once the recent pr's that change lighting are merged?

Indeed. Previously, using 2 overbright bits, bloom looked very bad on many maps. After cranking that down to 1 overbright bit, most maps are OK. A big exception is Vega: it looked horrible even with 0 overbright bits. But adding tonemap on top makes Vega more or less bearable (though it still looks worse than with no bloom).

illwieckz:
This commit is what makes a new point release mandatory, before any other points like “_the previous release is now old and we better release anew_”.

PR #3318, opened 2025-02-14
cmake: refactor game build configuration to reuse the same code in parent build or subproject build and mutualize code between vm type build

illwieckz:
Refactor game build configuration to reuse the same code in parent build or subproject build and mutualize code between vm type build.

Companion for:

- https://github.com/DaemonEngine/Daemon/pull/1555


illwieckz:
Unneeded by current:

- https://github.com/DaemonEngine/Daemon/pull/1555

PR #3321, opened 2025-02-23
add command botObstacle

sweet235:
Just a simple low level command to enable or disable bot obstacles. Currently, this only works for movers like doors, buttons or platforms. We might extend it to other kinds of entities later.

A few possible use cases:

- Vertically moving doors that sink into the floor, but not completely. If a little bit of the doors sticks out, that will leave an obstacle on the navmesh. We can at least `entityLock` such doors in the open position and remove the obstacle with this command.
- Uncommon causes for obstacles such as the destructable wall on map blackout at `-1870 -444 -185 178 -1`. Destroying it will make a mover appear on the floor, creating an unneccessary obstacle. We want to be able to remove this.
- Debugging bots.

### DIFF
--- a/src/sgame/sg_svcmds.cpp
+++ b/src/sgame/sg_svcmds.cpp
@@ -106,6 +106,54 @@ public:
 };
 static ShowBehaviorCmd showBehaviorRegistration;
 
+class BotObstacleCmd : public Cmd::StaticCmd
+{
+public:
+	BotObstacleCmd() : StaticCmd( "botObstacle", 0, "add or remove a bot obstacle for a given entity" ) {}
+	void Run( const Cmd::Args& args ) const override
+	{
+		if ( args.Argc() != 3 )
+		{
+			PrintUsage( args, "<entity> (on | off)" );
+			return;
+		}
+
+		int entityNum = 0;
+		if ( !Str::ParseInt( entityNum, args.Argv( 1 ) )
+		     || entityNum < MAX_CLIENTS
+		     || entityNum >= level.num_entities )
+		{
+			Print( "invalid entity number" );
+			return;
+		}
+
+		gentity_t *door = &g_entities[ entityNum ];
+		if ( door->s.eType != entityType_t::ET_MOVER )
+		{
+			Log::Notice( "not a ^5MOVER ^*entity " );
+			return;
+		}
+
+		if ( Q_stricmp( args.Argv( 2 ).c_str(), "on" ) == 0 )
+		{
+			glm::vec3 mins = VEC2GLM( door->mapEntity.restingPosition ) + VEC2GLM( door->r.mins );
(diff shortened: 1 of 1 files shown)


VReaperV:
Haven't tested, but code looks good.

slipher:
Isn't there any better way to identify an entity than entity num? Using entity num for `entityLock` caused a lot of grief because entity numbers are not necessarily the same every time. Maybe the target name if it has one, or a position + classname where it selects the closest entity of a given class to the specified coordinates.

DolceTriade:
entityNum is the best way. For Lua, we added a "id": https://github.com/Unvanquished/Unvanquished/pull/3042, but unsure how that can be used here? Maybe with a `.env` file?

src/sgame/sg_svcmds.cpp
```diff
@@ -106,6 +106,54 @@ class ShowBehaviorCmd : public Cmd::StaticCmd
...
+		     || entityNum >= level.num_entities )
+		{
+			Print( "invalid entity number" );
+			return;
+		}
+
+		gentity_t *door = &g_entities[ entityNum ];
+		if ( door->s.eType != entityType_t::ET_MOVER )
+		{
+			Log::Notice( "not a ^5MOVER ^*entity " );
```
slipher, line 133: `Print`

src/sgame/sg_svcmds.cpp
```diff
@@ -106,6 +106,54 @@ class ShowBehaviorCmd : public Cmd::StaticCmd
...
+		int entityNum = 0;
+		if ( !Str::ParseInt( entityNum, args.Argv( 1 ) )
+		     || entityNum < MAX_CLIENTS
+		     || entityNum >= level.num_entities )
+		{
+			Print( "invalid entity number" );
+			return;
+		}
+
+		gentity_t *door = &g_entities[ entityNum ];
```
slipher, line 130: could use an entity `inuse` check here

src/sgame/sg_svcmds.cpp
```diff
@@ -106,6 +106,54 @@ class ShowBehaviorCmd : public Cmd::StaticCmd
...
+		if ( door->s.eType != entityType_t::ET_MOVER )
+		{
+			Log::Notice( "not a ^5MOVER ^*entity " );
+			return;
+		}
+
+		if ( Q_stricmp( args.Argv( 2 ).c_str(), "on" ) == 0 )
```
slipher, line 137: `args.Argv( 2 ) == "on"`. Or better, case insensitively `Str::IsIEqual( args.Argv( 2 ), "on" )`

slipher:
> For Lua, we added a "id": #3042, but unsure how that can be used here? Maybe with a `.env` file?

Yeah the `id` would be basically the same as the `targetname` approach. Using a `.ent` file would be one way to give names to unnamed things. I thought of an idea for naming them automatically: https://github.com/Unvanquished/Unvanquished/issues/3325

Anyway this is a non-merge-blocking discussion; we can retrofit the commands later.

Issue #3078, opened 2024-08-10
Lua_close() segfaults after some Sys::Drop() calls earlier


VReaperV:
I've been getting segfaults at https://github.com/Unvanquished/Unvanquished/blob/2954028438b43714e1dadff8d83ea44c84a55590/src/sgame/lua/Interpreter.cpp#L190 after some earlier Sys::Drop()'s that were called due to unrelated changes. https://github.com/Unvanquished/Unvanquished/blob/2954028438b43714e1dadff8d83ea44c84a55590/src/sgame/sg_spawn.cpp#L956 is one such example.

slipher:
Are you running with gamelogic as a DLL? 

DLL gamelogic is intended for convenient debugging only and we don't try that hard to clean up everything. If the gamelogic is a DLL and Lua is also a DLL, I suppose the latter could retain some state after the gamelogic is unloaded.

VReaperV:
I did run as DLL, yeah.

I think the reason was that `L` was nullptr, so my assumption was that Lua wasn't initialised but was still attempted to be closed.

Issue #1693, opened 2022-01-10
Closing daemon (while some Lua is executing?) may cause a segfault


necessarily-equal:
```gdb
Thread 10 "daemon" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff9ffff640 (LWP 15693)]
0x00007fffc407d093 in luaH_getint () from /nix/store/7n91vqn1kb6vrk11sr30wca9qayis5s1-lua-5.2.4/lib/liblua.so.5.2
(gdb) backtrace
#0  0x00007fffc407d093 in luaH_getint ()
   from /nix/store/7n91vqn1kb6vrk11sr30wca9qayis5s1-lua-5.2.4/lib/liblua.so.5.2
#1  0x00007fffc406ebe1 in lua_rawgeti ()
   from /nix/store/7n91vqn1kb6vrk11sr30wca9qayis5s1-lua-5.2.4/lib/liblua.so.5.2
#2  0x00007fffbc5b38d7 in Rocket::Core::Lua::LuaEventListener::ProcessEvent (
    this=0x7fff9b795050, event=...)
    at /home/afontain/unv/Unvanquished/libs/libRocket/Source/Core/Lua/LuaEventListener.cpp:129
#3  0x00007fffbc50c483 in Rocket::Core::EventDispatcher::TriggerEvents (
    this=this@entry=0x7fff9b7961e0, event=event@entry=0x7fff9b5cd8e0)
    at /home/afontain/unv/Unvanquished/libs/libRocket/Source/Core/EventDispatcher.cpp:182
#4  0x00007fffbc50c694 in Rocket::Core::EventDispatcher::DispatchEvent (
    this=0x7fff9b7961e0, target_element=0x7fff9b792890, name=..., parameters=..., 
    interruptible=<optimized out>)
    at /home/afontain/unv/Unvanquished/libs/libRocket/Source/Core/EventDispatcher.cpp:146
#5  0x00007fffbc4f1d72 in Rocket::Core::Element::DispatchEvent (
    this=<optimized out>, event=..., parameters=..., 
    interruptible=<optimized out>)
    at /home/afontain/unv/Unvanquished/libs/libRocket/Source/Core/Element.cpp:1053
#6  0x00007fffbc4eeb1f in Rocket::Core::RKTEventFunctor::operator() (element=..., 
    this=0x7fff9fffdec0)
[...]

PR #3327, opened 2025-02-26
Implement pedantic shutdown options for gamelogic processes

slipher:
When `g_pedanticShutdown`/`cg_pedanticShutdown` is off (default) and the gamelogic is running in a separate process, skip shutdown code whose only purpose is to deallocate memory. This saves time and avoids spamming anyone with shutdown errors in production for cleanup issues like with Lua (#3078) or RmlUi (#1693).

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1398,8 +1398,24 @@ CG_Shutdown
 Called before every level change or subsystem restart
 =================
 */
+#ifndef BUILD_VM_IN_PROCESS
+#ifdef USING_SANITIZER
+constexpr bool pedanticShutdownDefault = true;
+#else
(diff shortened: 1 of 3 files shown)


illwieckz:
> Personally my reaction is that those syntax highlighters should git gud. But I will do it if you give me LGTM modulo the style comment 😛

Is this blackmail? 😛

LGTM modulo the style comment anyway, that syntax is the only thing I have something to say against.

illwieckz (approved):
LGTM modulo the style.

illwieckz:
Note that my new lines in my examples are not part of my recommendation, it's just to workaroud GitHub not wrapping lines and keep my comments readable.

sweet235:
I guess every second spent on thinking about Github's current code rendering is a waste of human resources.

PR #3328, opened 2025-02-27
Fix cg_drawClock not using local time (in NaCl)

### LINKED ISSUE #1608: "real life" clock shows date in UTC (at least on my computer)

slipher:
Companion: https://github.com/DaemonEngine/Daemon/pull/1569

Update the local time offset once per minute, to avoid making too many syscalls but keep up with daylight savings or time zone changes.

Fixes #1608.

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -3252,10 +3252,29 @@ static void CG_Rocket_DrawConnectText()
 	Rocket_SetInnerRMLRaw( rml );
 }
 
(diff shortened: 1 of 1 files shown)


DolceTriade:
Why bother keeping it in sync every minute? Won't it already be in sync practically every hour when we change map?

slipher:
Because I can keep it really in sync without much code or runtime cost.

illwieckz:
I see it's already merged on engine side, you can merge that as well.

illwieckz (approved):

PR #3331, opened 2025-03-08
ci: use submodules for freetype and lua

illwieckz:
Our submodules may provide symbols not provided by system libraries.


sweet235:
LGTM, did not test.

illwieckz:
Hmmm, this uncover more `-fPIC` madness.

slipher:
> Our submodules may provide symbols not provided by system libraries.

What do you mean by that? Like the OpenAL Soft library which may be built without symbols for the extension functions (and then won't link with Daemon)? If the system libraries don't actually work, the option to use them must be removed entirely.

We might want to avoid building additional code on a platform that runs slowly such as Travis. It would be ideal to have some builds using the option with each of the possible values for better coverage as these options are worth testing. I didn't get around to reporting it but building Freetype as submodule doesn't work if you build both Daemon and Unvanquished.

slipher:
OK, I discovered the [motivation](https://github.com/Unvanquished/Unvanquished/pull/3330#issuecomment-2708200384) for the PR. #3330 uses a Lua 5.3 function while the APT install command requests Lua 5.2.

illwieckz:
It's better to test in CI the build we do in releases anyway.

PR #3337, opened 2025-03-14
luarocket: add player.crosshairEntityNum

sweet235:
This is useful in devmap mode. When inspecting maps, typing entity numbers (or entity IDs) is cumbersome. So I am using binds like this one:

```
/bind o "luarocket \"if player.crosshairEntityNum > 0 then Cmd.exec(string.format('entityShow %d', player.crosshairEntityNum)) end\""
```

This executes command `entityShow N`, where N is the number of the entity I am currently looking at.


slipher:
LGTM

PR #3336, opened 2025-03-13
add fast lookup for entity IDs, command entityShow: accept an entity ID as argument

sweet235:
This is my proposal for supporting the recently introduced entity IDs in game commands and Lua code.

Add a data structure to map entity ID strings to entity numbers efficiently, update it whenever an entity with ID is created or destroyed.

If a user only creates entity IDs that do not parse as integers, an extension to existing commands is straightforward. An integer denotes an entity number as usual, everything else denotes an entity ID. This PR extends the command `entityShow` as an example. Recall that `entityShow` is used like in

```
]/entityShow 227 
⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼ 
#227:            MOVER 
⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼⎼ 
Classname: func_door 
Capabilities: acts thinks resets 
Names:  
{ trigger_bridge_01 } 
State: enabled
```

Assume this entity 227 has id `bridge`. With this PR, this command is equivalent:
```
]/entityShow bridge
```

The ID is now shown by `entityShow`, if it exists.

Lua mechanisms like https://github.com/Unvanquished/Unvanquished/pull/3330 can be extended similarly. In Lua scripts, we can use the types `number` and `string` to distinguish entity number and ID.

To test the example given above, use `map-prometheus_0.6.dpk`, create a file `maps/prometheus.ent` with 

```
q3map2 -exportents prometheus
```

and then edit the .ent file to add the `id` as shown here:

```
{
"classname" "func_door"
"spawnflags" "260"
"noise" "sounds/prometheus_custom/platform_01_working.opus"
"speed" "30"
"targetname" "trigger_bridge_01"
"soundPos1" "sounds/prometheus_custom/platform_01_stop_01.opus"
"soundPos2" "sounds/prometheus_custom/platform_01_stop_01.opus"
"angle" "-2"
"lip" "134"
[...]

PR #3338, opened 2025-03-14
generate automatic entity IDs for map entities without an ID

sweet235:
This is the part I removed from https://github.com/Unvanquished/Unvanquished/pull/3336. I cleaned it up a little.

Please see https://github.com/Unvanquished/Unvanquished/pull/3336#issuecomment-2722680199 for a description.

### DIFF
--- a/src/sgame/sg_spawn.cpp
+++ b/src/sgame/sg_spawn.cpp
@@ -790,6 +790,69 @@ static void G_ParseField( const char *key, const char *rawString, gentity_t *ent
 		G_WarnAboutDeprecatedEntityField(entity, fieldDescriptor->replacement, key, fieldDescriptor->versionState );
 }
 
+static std::unordered_map< std::string, int > autoIdCounters;
+static const std::set< std::string > autoClassnames = {
+	"env_afx_ammo", "env_afx_gravity", "env_afx_heal", "env_afx_hurt",
+	"env_afx_push", "env_afx_teleport",
+
+	"func_bobbing", "func_button", "func_destructable", "func_door", "func_door_model",
+	"func_door_rotating", "func_dynamic", "func_group", "func_pendulum", "func_plat",
+	"func_rotating", "func_spawn", "func_timer", "func_train",
+
+	"sensor_buildable", "sensor_creep", "sensor_end", "sensor_player", "sensor_power",
+	"sensor_stage", "sensor_start", "sensor_support", "sensor_timer",
+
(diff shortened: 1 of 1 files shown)


src/sgame/sg_spawn.cpp
```diff
@@ -790,6 +790,73 @@ static void G_ParseField( const char *key, const char *rawString, gentity_t *ent
 		G_WarnAboutDeprecatedEntityField(entity, fieldDescriptor->replacement, key, fieldDescriptor->versionState );
 }
 
+static std::unordered_map< std::string, int > autoIdCounters;
+static const std::set< std::string > autoClassnames = {
+	"env_afx_ammo", "env_afx_gravity", "env_afx_heal", "env_afx_hurt",
```
slipher, line 795: Why do you need a closed list instead of just doing it for every class?

slipher (commented):
My idea of hashing based on the position (and classname) from the issue thread was intended to make it so that IDs could remain stable even after changes like adding/removing entities elsewhere in the map. Although this (classname_counter) scheme could still be disrupted by such changes, it would only mess up a single entity class at a time. So this PR seems like a good compromise to simplify the implementation by avoiding hashing and get enough stability most of the time.

src/sgame/sg_spawn.cpp
slipher: If you have `std::map<std::string, int> m` you can do `n = ++m[classname]` and the map entry will be automatically created and initialized to 0 if it doesn't exist.

sweet235: Thank you. That was a relic from a more naive method.

src/sgame/sg_spawn.cpp
```diff
@@ -790,6 +790,47 @@ static void G_ParseField( const char *key, const char *rawString, gentity_t *ent
...
+	//   calculate a hash from the entity description text
+	//   convert the hash to something human readable, use this as ID postfix
+	// the more complicated methods are probably not worth it, as we can always assign
+	// IDs of our own choosing in the entity lump / .ent file.
+	std::string classname = spawningEntity->classname;
+	int counter = ++autoIdCounters[ classname ];
+	if ( spawningEntity->id == nullptr )
+	{
+		// subtract 1 from counter because it starts at 1
+		std::string autoId = Str::Format( "%s_%d", classname, counter - 1 );
```
slipher, line 829: Could use postfix ++ if you want starting at 0.

LGTM

PR #3339, opened 2025-03-15
fix a bug in luarocket's player.crosshairEntityNum

sweet235:
It would return the entity you last looked at, indefinitely.


sweet235:
Thank you.

PR #3340, opened 2025-03-15
commands entityFire, entityLock: accept entity IDs

sweet235:
Same as with command `entityShow`.

### DIFF
--- a/src/sgame/sg_svcmds.cpp
+++ b/src/sgame/sg_svcmds.cpp
@@ -43,6 +43,35 @@ Maryland 20850 USA.
 
 #define IS_NON_NULL_VEC3(vec3tor) (vec3tor[0] || vec3tor[1] || vec3tor[2])
 
+static int StringToEntityNum( Str::StringRef str, Str::StringRef commandName )
+{
+	int entityNum;
+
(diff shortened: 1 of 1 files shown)


src/sgame/sg_svcmds.cpp
```diff
@@ -43,6 +43,35 @@ Maryland 20850 USA.
...
+static int StringToEntityNum( Str::StringRef str, Str::StringRef commandName )
+{
+	int entityNum;
+
+	if ( !Str::ParseInt( entityNum, str ) )
+	{
+		entityNum = G_IdToEntityNum( str );
+		if ( entityNum < 0 )
+		{
+			Log::Notice( "%s: no entity with ID %s", commandName, str );
```
slipher, line 55: These could use CommandInteractionMessage

src/sgame/sg_svcmds.cpp
```diff
@@ -43,6 +43,35 @@ Maryland 20850 USA.
...
+			return -1;
+		}
+	}
+
+	if ( entityNum >= level.num_entities || entityNum < MAX_CLIENTS )
```
slipher, line 60: Clients could be valid, for entityShow at least

PR #3341, opened 2025-03-15
fix a bug with auto entity IDs

sweet235:


slipher:
Could be SP_RemoveSelf or other entities that delete themselves if not configured correctly? Checking `inuse` seems better than the classname.

PR #3343, opened 2025-03-16
Skip pointless fraction=0 BlendSkeleton calls

slipher:
I noticed that buildables were doing a lot of trap_R_BlendSkeleton with interpolation fraction 0, which is the same as not blending. With this commit, buildables generally avoid blending if they are not being attacked or used.

### DIFF
--- a/src/cgame/cg_animation.cpp
+++ b/src/cgame/cg_animation.cpp
@@ -157,7 +157,17 @@ void CG_BlendLerpFrame( lerpFrame_t *lf )
 		return;
 	}
 
-	if ( ( lf->blendlerp > 0.0f ) && ( cg.time > lf->blendtime ) )
+	if ( lf->blendlerp < 1.0e-6f )
+	{
+		// Truncate to 0 to avoid wasting CPU/IPC on blending operations for
+		// invisibly small differences.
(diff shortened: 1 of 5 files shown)


src/cgame/cg_animation.cpp
```diff
@@ -209,7 +209,7 @@ void CG_BuildAnimSkeleton( const lerpFrame_t *lf, refSkeleton_t *newSkeleton, co
 	}
 
 	// lerp between old and new animation if possible
-	if ( lf->blendlerp >= 0.0f )
+	if ( lf->blendlerp > 1.0e-6f )
```
DolceTriade, line 212: is there an epsilon constant? Also, for my benefit, what is the difference between lf->blendlerp > 0 and lf->blendlerp > 1.e-6f?

Lastly, if we do keep this number, please include why it was picked, what its implications are, etc.

slipher: I changed this part to `> 0` and moved the code for truncating small values to 0 into the function that sets the `blendlerp` value.

slipher:
Added a new commit that fixes framerate-dependent transition rate between animations.

illwieckz (approved):
LGTM.

PR #3344, opened 2025-03-16
add command entitySetId

sweet235:
Add a command `entitySetId <num> <ID>`. It can be used to give IDs to entities that do not have an ID.

When all else fails, this can be used in cases where people are using raw entity numbers. One example is buildable entities.


sweet235:
Drafted. I changed the custom entity handler PR to support raw entity numbers as well as IDs. So we probably do not need this.

PR #3345, opened 2025-03-16
give IDs to reactor and overmind

sweet235:
Our policy is that the main buildables are unique. So we can simply give them the IDs `reactor` and `overmind` automatically.

### DIFF
--- a/src/sgame/sg_buildable.cpp
+++ b/src/sgame/sg_buildable.cpp
@@ -33,6 +33,7 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 #include "Entities.h"
 #include "CBSE.h"
 #include "sg_cm_world.h"
+#include "sg_spawn.h" // for G_NewString()
 
 /**
  * @return Whether the means of death allow for an under-attack warning.
@@ -1608,6 +1609,7 @@ static gentity_t *SpawnBuildable( gentity_t *builder, buildable_t buildable, con
 			break;
 
 		case BA_A_OVERMIND:
+			built->id = G_NewString( "overmind" );
 			break;
 
 		case BA_H_SPAWN:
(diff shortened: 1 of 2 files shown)


sweet235:
This will be useful for several maps with nonstandard win conditions. One example is `td`, where the game is supposed to end when the reactor is destroyed. Another is the AMP map `statues-clash` (at least I believe that).

slipher:
We might want to make automatic IDs for layout buildables too, like for normal BSP entities. (IIRC there is some nonsense with a temp entity that later spawns a buildable, which is why even BSP-embedded layouts don't get any IDs now.) Since an entity can only have one ID, this PR would interfere with doing that later. But maybe they should just be allowed to have more than one ID?

It might be more robust to have an API for counting or listing buildables of a specific type. Or an API to expose G_MainBuildable. I'm not sure it's a good idea for generated IDs to be reused within the course of a game. Currently there is the bug that if two entities have the same ID and the one currently "owning" it is destroyed, the ID is mapped to nothing afterward, though that should be fixable.

src/sgame/sg_buildable.cpp
```diff
@@ -1608,6 +1609,7 @@ static gentity_t *SpawnBuildable( gentity_t *builder, buildable_t buildable, con
 			break;
 
 		case BA_A_OVERMIND:
+			built->id = G_NewString( "overmind" );
```
DolceTriade, line 1612: WDYT about leaving a bread crumb saying that this will be freed in in the enttiy cleanup function?

DolceTriade (approved):

sweet235:
I think this is a bad idea. Instead, we should provide an event handler for the general case of spawning an entity. It could look like

```lua
function Entities.spawnHandler(entityNum, classname)
    -- do something
end
```

Then we could provide a function to assign IDs to entities in Lua, let us call it `Entities.setId`. Then we could write

```lua
function Entities.spawnHandler(entityNum, classname)
    if classname == "team_human_reactor" then
        Entities.setId(entityNum, "reactor")
    if classname == "team_alien_overmind" then
        Entities.setId(entityNum, "overmind")
    end
end
```

We could also do many other useful things.

Issue #3157, opened 2024-10-25
Inefficient IPC calls


VReaperV:
I plan on making the amount of IPC calls largely static for 0.56, as right now their amount can often cripple performance. This issue is to keep track of such calls.

IPC call|Usage|Solution
-|-|-
`trap_CM_MarkFragments()`|Unused now that fragments are batched|Remove.
-|-|-
`trap_R_RegisterModel()`|Done for each model. Affects loading time?|N/A
-|-|-
`trap_R_RegisterSkin()`|Done for each skin. Affects loading time?|N/A
-|-|-
`trap_R_RegisterShader()`|Done for each shader. Affects loading time?|N/A
-|-|-
`trap_R_AddRefEntityToScene()`|Done for each visible entity, sometimes more than once (e. g. for players/bots, for head, torso, legs, jetpack, radar, weapon etc.)|Add all to shared mem, flush when ready. We don't want stuff appearing/disappearing mid-frame, so don't flush until we have added everything. Alternatively, make `refEntity_t` persistent between frames (this would also help with caching shadowmaps, since engine would then know if an entity hasn't moved within a light source's frustum.
-|-|-
`trap_R_AddPolyToScene()`|Done for all mark fragments, all buildable range spheres, all debug bboxes, once for some test functions, and drawing tracers (?)|Same as above.
-|-|-
`trap_R_AddPolyToScene()`|Done once for each beam (like marauder zap perhaps?)|Same as above.
-|-|-
`trap_R_AddLightToScene()`|Done for all dynamic lights. A few firebombs can produce hundreds of these.|Same as above.
-|-|-
`trap_R_AddAdditiveLightToScene()`|Glowing entities? Don't know how many of those we use|N/A
-|-|-
`trap_R_Add2dPolysIndexedToScene()`|Done for all rml UI elements?|Implement CompiledGeometry?
-|-|-
`trap_R_SetMatrixTransform()`|Used by rml, but how often?|N/A
-|-|-
`trap_R_ResetMatrixTransform()`|Used by rml, but how often?|N/A
-|-|-
`trap_R_ClearColor()`|Done for every beacon, weapon scopes, pain overlay, minimap, cursor, momentum bar, some buildable stuff (health bars?) etc.|
-|-|-
`trap_R_SetColor()`|Similar to the above|Same as above.
-|-|-
`trap_R_SetClipRegion()`|Used mainly for building health bars|Same as above
-|-|-
[...]

PR #3347, opened 2025-03-17
UI: avoid shader handle->name->handle round trips

slipher:
For some UI elements (e.g. weapon and upgrade icons) a round trip of trap_R_RegisterShader( trap_R_ShaderNameFromHandle( handle ) ) was done each time the shader was used, incurring 2 IPC calls. Avoid this by introducing a magic syntax for RML image paths, for example "/$handle/123", which allows a handle to be plumbed through RmlUi.

This gets rid of one of the IPC calls listed in #3157.

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -2313,7 +2313,6 @@ enum {
 // cg_utils.c
 //
 bool   CG_ParseColor( byte *c, const char **text_p );
-const char *CG_GetShaderNameFromHandle( const qhandle_t shader );
 void       CG_ReadableSize( char *buf, int bufsize, int value );
 void       CG_PrintTime( char *buf, int bufsize, int time );
 void CG_SetKeyCatcher( int catcher );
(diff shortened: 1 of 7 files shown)


DolceTriade (commented):
Can we document this somewhere? high level idea seems ok to me, but the magic will be confusing without sufficient bread crumbs

slipher:
> Can we document this somewhere? high level idea seems ok to me, but the magic will be confusing without sufficient bread crumbs

I don't see why it would be difficult to figure out. Easy to grep for `$handle`.

DolceTriade:
> I don't see why it would be difficult to figure out. Easy to grep for $handle.

Yes, but that will result in all usages of $handle too and you need to identify the source. At the very least, I'd like this documented in `GetShaderHandle`. Some of the other magic prefixes were surprises to me as well.

slipher:
> At the very least, I'd like this documented in `GetShaderHandle`.

Done.

illwieckz:
As a side remark, and this also applies to the already existing `$levelshot`, I would prefer if the syntax of such pseudo-variable have an obvious ending token, like `${handle}` or `${levelshot}`.

The idea is good though.

DolceTriade (approved):

slipher:
> As a side remark, and this also applies to the already existing `$levelshot`, I would prefer if the syntax of such pseudo-variable has an obvious ending token, like `${handle}` or `${levelshot}`.

It's not like a variable. More of a magic directory like Unix `/proc`. Refer to https://github.com/Unvanquished/Unvanquished/pull/3214#issuecomment-2513993797.

PR #3353, opened 2025-03-24
cmake: make possible to build both game and engine against the engine Freetype submodule

illwieckz:
Make possible to build both game and engine against the engine Freetype submodule.

Extracted from #3331:

- https://github.com/Unvanquished/Unvanquished/pull/3331

Without the controversial change about default building.

Requires DaemonEngine/Daemon#1617:

- https://github.com/DaemonEngine/Daemon/pull/1617


illwieckz:
Hmm, it's possible that the mac linkage error is specific to the way we craft our `external deps`.

illwieckz:
> Hmm, it's possible that the mac linkage error is specific to the way we craft our `external_deps`.

Fixed.

PR #3359, opened 2025-03-30
move entityType_t in gamelogic

VReaperV:
Engine-side pr: https://github.com/DaemonEngine/Daemon/pull/1631

Patch shared by freem, this moves entityType_t to gamelogic, since the engine doesn't need to know about it.


illwieckz:
I ran the CI in #3362, it went fine.

illwieckz:
Assuming it was tested, LGTM (on both sides).

illwieckz:
I tested and it works.

PR #3360, opened 2025-04-03
Vsay menu

Grise3:
Addition of the Vsay menu.
Screenshot_20250403_155712
The Vsays were chosen by a vote on the forum. The vote is visible [here](https://forums.unvanquished.net/viewtopic.php?t=2512)


sweet235:
LGTM

Gireen:
I think for the vsay menu it'd be better to use text instead of icons. 
The commands are short anyway so it became easier to understand, edit and add new elements. 

Beside that the number of displayed commands can also be increased to seven without becoming too much.

Grise3:
@Gireen wrote :
> I think for the vsay menu it'd be better to use text instead of icons. The commands are short anyway so it became easier to understand, edit and add new elements.
> 
> Beside that the number of displayed commands can also be increased to seven without becoming too much.

Reading your message, I only have one thing that comes to mind: "Are you trolling me?"

Grise3:
**Let’s assume you’re not trolling me.**

> The commands are short anyway so it became easier to understand, edit and add new elements.

I suppose that by "adding new elements," you're referring to the pictures.  
Personally, I prefer creating **each and every vsay picture** rather than settling for replacing the pictures with text (which, let’s be honest, is a **BAD** solution).  

And, pictures make everything look much nicer. Just imagine the result if we used text instead: inconsistent sizes, some smaller, others larger, and overall, a messy and hard-to-read layout.

Gireen:
> Reading your message, I only have one thing that comes to mind: "Are you trolling me?"

🤨 maybe the recent drama by others can make one question every comment. 
but no. that's my honest feedback on this PR, like in any other i comment on. It might have been not oblivious enough but if i don't approve a PR doesn't mean i think it should be rejected, but that i leave the decision to someone else. 

I don't consider it a blocking matter, else i would have said so clearly.
So if you don't want to change anything, and reaper approving it like it is, that's also fine with me. 

With that cleared up let me explain my reasoning. 
The voices are only available in english. Players will also hear the english words and have it easier, even as non native speakers, to learn the meaning. That's why i think the menu items should stay english and only the description translated.

Most commands are short, like the ones you added (attack,defend,yes,no,flank,follow me) with the overall longest one containing three words which is still something that can fit into a circle. 

Some concepts are harder to depict, in the added ones its flank, but there are also sorry, thank you, enemy is near, regroup, enemys spawns down.

Well then now its up to you what you do with the PR. 😄

src/cgame/cg_consolecmds.cpp
```diff
@@ -541,6 +550,7 @@ static const struct cg_cmd_t
 	{ "viewpos",          CG_Viewpos_f,            0                },
 	{ "vote",             0,                       0                },
 	{ "vsay",             0,                       CG_CompleteVsay  },
+	{ "VsayMenu",         CG_VsayMenu_f,           0                },
```
slipher, line 553: The command name should begin with a lowercase letter: `vsayMenu`

src/cgame/cg_rocket_datasource.cpp
```diff
@@ -1531,6 +1566,7 @@ static const dataSourceCmd_t dataSourceCmdList[] =
 	{ "playerList", &CG_Rocket_BuildPlayerList, &CG_Rocket_SortPlayerList, &CG_Rocket_CleanUpPlayerList, &CG_Rocket_SetPlayerListPlayer, &nullFilterFunc, &nullExecFunc, &nullGetFunc },
 	{ "resolutions", &CG_Rocket_BuildResolutionList, &nullSortFunc, &CG_Rocket_CleanUpResolutionList, &CG_Rocket_SetResolutionListResolution, &nullFilterFunc, &nullExecFunc, &CG_Rocket_GetResolutionListIndex},
 	{ "server_browser", &CG_Rocket_BuildServerList, &CG_Rocket_SortServerList, &CG_Rocket_CleanUpServerList, &CG_Rocket_SetServerListServer, &CG_Rocket_FilterServerList, &CG_Rocket_ExecServerList, &nullGetFunc },
+	{ "VsayList", &CG_Rocket_BuildVsayList, &nullSortFunc, &nullCleanFunc, &nullSetFunc, &nullFilterFunc, &nullExecFunc, &nullGetFunc },
```
slipher, line 1569: Also use a lowercase first letter here. Some of these lists may have the requirement to be sorted by ASCII values and might not work if you use uppercase letters since `V` has a different value than `v`.

slipher (commented):
Aren't some of these images the same as ones in the bot tactic menu? If we already have an image, it's better to use the existing one instead of making a copy.

VReaperV:
I urge anyone to not hide any comments in the future, this a public platform. And if it weren't for github's moronic "features", I'd un-hide them, so this will do instead:
image
On top of that, since not so long ago github doesn't allow looking at hidden comments if not logged in, which is anti-thetical to a public platform, and therefore should be doubly avoided.

VReaperV:
Hmm, didn't have the reviews loaded before merging...

VReaperV:
Fixes for the capitalisation are in #3361

PR #3363, opened 2025-04-13
Graphics options update: remove overbright clamping, add tonemapping

slipher:


slipher:
unvanquished_2025-04-13_163718_000

PR #3367, opened 2025-04-17
begin process of moving MODs to bg_mod.h

cu-kai:
This fulfills a TODO in the code, removes duplication and lays the groundwork for other changes we intend to upstream at a later time.

This change is well tested and has been running on Bunker for several months.

### DIFF
--- a/src/shared/bg_mod.h
+++ b/src/shared/bg_mod.h
@@ -1,0 +1,93 @@
+/*
+===========================================================================
+
+Unvanquished GPL Source Code
+Copyright (C) 1999-2005 Id Software, Inc.
+Copyright (C) 2000-2009 Darklegion Development
+
+This file is part of the Unvanquished GPL Source Code (Unvanquished Source Code).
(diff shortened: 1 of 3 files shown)


src/shared/bg_mod.h
```diff
@@ -0,0 +1,93 @@
...
+
+struct meansOfDeathData_t
+{
+	meansOfDeath_t number;
+	const char     *name;
+};
+
+static const meansOfDeathData_t bg_meansOfDeathData[] =
+{
+	{ MOD_UNKNOWN, "MOD_UNKNOWN" },
```
slipher, line 35: The table data should go in a .cpp file. Doing it like this in a header means that every file we build gets its own copy of the table and can cause [ODR](https://en.cppreference.com/w/cpp/language/definition) violations for code that uses it.

cu-kai:
clearly this isn't the right way to do this, so I'll abandon this PR for now

PR #3368, opened 2025-04-25
document how to build on debian trixie

gogo2464:
Hello.

I learnt only a single person has successed to build the game from source.

I did 5 minutes ago then I decided to document all.

I chosen to document the dependencies install from apt and let the user continue with the rest of the build tutorial.

Best regards!

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,23 @@ Required:
 Optional:
 `ncurses`.
 
+## Install dependencies on trixie:
+
+Uninstall `sudo apt purge -y libsdl2-2.0-0 ;` and do:
+
+```bash
+sudo apt install -y libsdl2-dev libopusfile-dev libwebp-dev libglew-dev libopenal-dev ;
+```
+
+sudo apt install -y conda-package-handling
+
+```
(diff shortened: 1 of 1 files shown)


slipher:
Debian package lists can be found [on the wiki](https://wiki.unvanquished.net/wiki/Compiling_the_source#Debian.2FUbuntu). Should work without any uninstall commands or downloading shell scripts. Anyway we don't list the distro-specific package names in the README since that would use a lot of space and too much maintenance burden.

gogo2464:
good idea.? I will move it

illwieckz:
Obsoleted by:

- https://github.com/Unvanquished/Unvanquished/pull/3372

This other PR adds an explicit link to the `Compiling the source` wiki page.

If instructions are missing on the wiki, please tell us which ones are missing.

gogo2464:
I could provide a few commands to help from wiki as well.

illwieckz:
We better improve the wiki than to make the README too large.

@gogo2464 if you need a wiki account please tell me.

PR #3369, opened 2025-04-26
correct the "You have responded to" message in Cmd_ReplyPrivateMessage_f

Grise3:
For example, replying to "foobar" with "hello" showed:
```You have responded to foobar : hello```
but replying to "foo bar" with "hello" showed:
```You have responded to foo : bar```
Corrected

### DIFF
--- a/src/sgame/sg_cmds.cpp
+++ b/src/sgame/sg_cmds.cpp
@@ -4248,7 +4248,7 @@ static void Cmd_ReplyPrivateMessage_f( gentity_t *ent )
 	{
 		setLastPrivateMessageSenderAndTime( target, ent->num() );
 	}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cu-kai** [comment]
This sounds like you have fallen into a common translation pitfall.

Instead of removing the translation, can you try using `Quote` on the netname in the original code? I believe that would fix your problem.

Try testing with something like this and update the PR as required:

```c
ADMP( va( "%s %s %s", QQ( N_("You have responded to $1$^* : ^2$2$ ") ), Quote( g_entities[ target ].client->pers.netname ), Quote( msg ) ) );
```

cu-kai (approved):
LGTM. Also just a word on your original code, be careful to follow the code style (in your original code, you removed the spacing e.g. between the `[`'s

Grise3:
Oh yeah, True.
Thank you!

Grise3:
Thanks Kai :)

PR #3371, opened 2025-04-30
ui: sort graphics presets from highest to lowest and add a message discouraging to use the lowest preset when not absolutely required

illwieckz:
UI: Sort graphics presets from highest to lowest and add a message discouraging to use the lowest preset when not absolutely required.

Previously I made sure to unify our menus to reorder the options in drop-down lists from highest (preferred) to lowest, highest on top. But I forgot to also reorder the graphics presets button (because it is not a drop-down list), and this encourage people to try the “lowest” preset first, which is only meant to be a last resort solution.

So not only I re-ordered the presets from highest to lowest, but I also added a message discouraging to use the lowest preset when not absolutely required.

Already in the past I have seen videos of people playing the game, and tweaking the options with obvious intention to improve performance, but not knowing what options were and doing the opposite. That's part of why I reordered the drop-down menu so even when not understanding what a specific option is, people can blindly select the highest or the lowest options by following conventions.

Recently, I stumbled upon [another video]( https://www.youtube.com/watch?v=nyzYabXS4W4) where someone was streaming the game in 1280×720 resolution with lowest preset and achieving a very solid 70~90fps. The thing is that if such people can get solid a 70~90 fps framerate at this resolution while streaming, his computer is powerful enough to sustain the low preset at the exact same framerate, and the people would not stream visual suffering.

So we better make it clear that the lowest preset is only to be used when the low preset isn't fast enough. This PR does that. Also, inserting the message between all other presets and the lowest one makes it obvious that preset is considered to be apart.

unvanquished_2025-04-30_181733_000

Also, this rework the panel/tab horizontal split. All options window with one-word tab were using split values meant for the gameplay guide, gameplay guide that was using custom values anyway… So this reduces the width of the tab column for anything than the gameplay guide, saving a lot of pixels…


illwieckz:
I also added a mention saying “This is the default” next to the medium preset button.

illwieckz:
We may also reword those preset descriptions, maybe making them less technical, and the bloom mention is actually false, but that's not in the scope of this PR.

illwieckz:
Well, actually, rewording the preset descriptions made easier to write those information messages about the lowest preset, so I did it.

The fact single-line descriptions aren't vertically centered is another issue (fixing it would require a larger revamp of our rmlui stuff).

unvanquished_2025-04-30_194724_000

slipher:
It's not that true that "medium" is the default. Default cvar values don't correspond exactly to any preset. They are kind of a mix of medium and high I think.

Grise3:
> Well, actually, rewording the preset descriptions made easier to write those information messages about the lowest preset, so I did it.
> 
> The fact single-line descriptions aren't vertically centered is another issue (fixing it would require a larger revamp of our rmlui stuff).
> 
> unvanquished_2025-04-30_194724_000

It is better like this.
I am not sure about the position of the texte "When looking for optimisations [...]".
if it is for the lower preset, it should be better to put it as the "same level" than "the faster preset [...]". 

I don't know if what I'm saying is clear. I'll make a picture to explain if it isn't.

illwieckz:
Actually making a clear distinction between the two groups of preset is something I wanted, moving the text to the right aligned with other text may give the illusion it's part of the previous preset description, tricking the brain to skip it, and defeating the clear separation.

The fact that text acts as an horizontal line to split the two parts is on purpose, if there was no text, I would look to add an horizontal line.

Grise3:
> Actually making a clear distinction between the two groups of preset is something I wanted, moving the text to the right aligned with other text may give the illusion it's part of the previous preset description, tricking the brain to skip it, and defeating the clear separation.
> 
> The fact that text acts as an horizontal line to split the two parts is on purpose, if there was no text, I would look to add an horizontal line.

Oh OK I see.
Then the goal is "the lowest preset in a different category"?

illwieckz:
> It's not that true that "medium" is the default. Default cvar values don't correspond exactly to any preset. They are kind of a mix of medium and high I think.

Right, I noticed the engine enables realtime lights by default.
In fact, we may make the `high` preset the default one at some point.

So I removed the “This is the default” string on `medium` preset.

illwieckz:
So, this is how it now looks:

unvanquished_2025-05-24_210939_000

illwieckz:
It's already far better than before, we may improve later if we find better ways.

PR #3374, opened 2025-05-05
Remove GLM4RW in favor of glm-native functions

enneract:
Resolving https://github.com/Unvanquished/Unvanquished/issues/3013 introduced the macro `GLM4RW` (https://github.com/Unvanquished/Unvanquished/commit/d2f8f83ccc131a673589c1e6f9140ff4c5ca5ea2). It technically avoids UB by using `memcpy`, but I'm not a big fan of the hackery.

This PR adds glm versions of a couple of vector functions, making `GLM4RW` obsolete.

It also fixes the related "uninitialized variable" warnings.


src/shared/bg_public.h
```diff
@@ -1968,5 +1956,24 @@ inline float Alignment2D(glm::vec3 u, glm::vec3 v, float z_factor = 0.0f)
...
+
+	if (length2 != 0.0f)
+	{
+		vec *= Q_rsqrt(length2);
+	}
+
+	return length2;
+}
+
+inline void VectorCopy(const glm::vec3 &in, vec3_t out)
```
slipher, line 1971: The VectorCopy from q_shared.h already works for that case.

src/shared/bg_utilities.cpp
```diff
@@ -228,3 +228,127 @@ WARN_UNUSED_RESULT glm::mat3 RotationMatrix( const glm::vec3 &angles ) {
...
 }
+
+glm::vec3 ProjectPointOntoVectorBounded(const glm::vec3 &point, const glm::vec3 &vStart, const glm::vec3 &vEnd)
+{
+	glm::vec3 pVec = point - vStart;
+	glm::vec3 vec = vEnd - vStart;
+	VectorNormalize( vec );
+	// project onto the directional vector for this segment
+	glm::vec3 vProj = vStart + glm::dot(pVec, vec) * vec;
+
```
slipher, line 239: How about `return vStart + clamp(dot(pVec, vec), 0, 1) * vec`?

src/shared/bg_utilities.cpp
```diff
@@ -228,3 +228,127 @@ WARN_UNUSED_RESULT glm::mat3 RotationMatrix( const glm::vec3 &angles ) {
...
+	return vProj;
+}
+
+glm::vec3 ProjectPointOnPlane(const glm::vec3 &point, const glm::vec3 &normal)
+{
+	return point - glm::dot(point, normal) * normal;
+}
+
+glm::vec3 PerpendicularVector(const glm::vec3 &vec)
```
slipher, line 271: Would be good to include the 'assumes vec is normalized' comment

src/shared/bg_public.h
enneract: Thanks, removed.

src/shared/bg_utilities.cpp
enneract: Sounds good, but I'm not sure about corner cases, rounding errors, etc. I'd like to keep the original math for now.

slipher:
LGTM

PR #3388, opened 2025-06-28
Update MAX_PARTICLES

### LINKED ISSUE #2502: MAX_PARTICLES is hit too quickly
MAX_PARTICLES is hit too quickly, on crowded mission map with lot of fire and flamethrower, there is so much particle that firs-person flames of flamethrower may be entirely skipped. Here, I'm firing the flamethrower and actually hit targets (see damage indicators):

[max particles](https://dl.illwieckz.net/b/unvanquished/bugs/max-particles/unvanquished_2023-03-14_180707_001.jpg)

VReaperV:
Combined with https://github.com/DaemonEngine/Daemon/pull/1689 fixes #2502


illwieckz (approved):
LGTM.

slipher:
This is now larger than MAX_REF_ENTITIES (11072 vs. 8192). Would be better to just write the number instead of a nonsense calculation so we could easily see that

PR #3390, opened 2025-07-06
cgame: load map worldspawn in engine before rendering anything

illwieckz:
Load worldspawn as soon as possible before other map data.

I'm afraid we need to load the worldspawnto read the sRGB-related map options  before rendering anything (including the map loading screen)  for enabling/disabling sRGB/linear conversions as soon as possible and getting the UI right.

See on engine side:

- https://github.com/DaemonEngine/Daemon/pull/1692

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1259,10 +1259,7 @@ void CG_Init( int serverMessageNum, int clientNum, const glconfig_t& gl, const G
 	// cg_shadows is latched so we can get it once at the beginning
 	cg_shadows = Util::enum_cast<shadowingMode_t>( trap_Cvar_VariableIntegerValue( "cg_shadows" ) );
 
-	// load a few needed things before we do any screen updates
 	trap_R_SetAltShaderTokens( "unpowered,destroyed,idle,idle2" );
-	cgs.media.whiteShader = trap_R_RegisterShader("gfx/colors/white", RSF_DEFAULT);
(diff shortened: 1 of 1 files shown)


slipher:
I don't believe that would be  necessary. The rendering flow looks like this:
1. Do all 3D rendering to an FBO
2. Run a shader that copies the FBO to the final framebuffer
3. Do all 2D rendering to the final framebuffer

So sRGB settings for 3D rendering need not have any effect on 2D rendering. We should probably stick with naive blending for the 2D rendering since that's what HTML/CSS do.

illwieckz:
Where is the 2D rendering code? I mean the code writing the 2D stuff to a framebuffer.

slipher:
The shader is called "camera effects" I think. The one that does color grades and tone mapping

illwieckz:
I assume the camera one goes before 2D. Well, by 2D I mean UI.

I wonder if what we need is to call `glDisable( GL_FRAMEBUFFER_SRGB )` before rendering UI then to call `glEnable( GL_FRAMEBUFFER_SRGB )` after.

illwieckz:
Well, in `tr_backend.cpp`, a good place would be to disable `GL_FRAMEBUFFER_SRGB` after `backEnd.projection2D = false`  and to re-enable it after `backEnd.projection2D = true`.

illwieckz:
> Well, in `tr_backend.cpp`, a good place would be to disable `GL_FRAMEBUFFER_SRGB` after `backEnd.projection2D = false` and to re-enable it after `backEnd.projection2D = true`.

That does it.

illwieckz:
This seems unneeded for now.

PR #3391, opened 2025-07-11
Fix #3080 (callvote allows invalid map names)

### LINKED ISSUE #3080: Bad Things™ happen if an fs_pakprefix starts with `map-`
We found this out the hard way. I used to have `set fs_pakprefixes map-testing`. This was done to allow me to test higher versions of existing maps without making these versions the default for clients. I placed these maps in a directory `map-testing`, and this worked fine for a while.

Until some genius user found out that you can exploit this to stop the server. In the directory `map-testing`, I had `map-doom_1.dpk`. Users could issue this command: `callvote map testing/map-doom`. This command would not fail. It seems the logic to check if a map exists is broken. The server would try to change to this map, notice it does not exist, and then exit "gracefully". Or so we believe.

I fixed this by renaming the directory. As long as its name does not start with `map-`, I guess I will be fine. But we should fix this. Notice that it seems we can inject path separators into package names in this case. I wonder what other mischief we can cause knowing this.

slipher:
Don't allow voting map names that contain a directory.

See also https://github.com/DaemonEngine/Daemon/pull/1693 for an additional engine-side fix.

### DIFF
--- a/src/sgame/sg_maprotation.cpp
+++ b/src/sgame/sg_maprotation.cpp
@@ -108,7 +108,8 @@ bool G_MapExists( const char *name )
 {
 	// Due to filesystem changes, checking whether "maps/$name.bsp" exists in the
 	// VFS is no longer the correct way to check whether a map exists
-	return trap_FindPak( va( "map-%s", name ) );
+	// Legacy paks not supported here.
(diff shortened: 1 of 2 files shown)


illwieckz (approved):
LGTM.

PR #3399, opened 2025-08-11
Fix classname memory leak

VReaperV:
Fixes leaking allocations from `G_NewString()`.


slipher:
Why is this needed? `classname` always comes from string constants or strings owned by other objects, so I don't see how it could cause a leak.

VReaperV:
Not always: https://github.com/Unvanquished/Unvanquished/blob/7ef97e9db662f9c900cd5d030b381953a44e3d95/src/sgame/sg_spawn.cpp#L722 and a second usage below do `BG_Alloc()` which is then never freed. Lua `id` does something similar but actually frees the allocation later.

slipher:
Oh, the table-based parser. I guess most other F_STRING fields are leaked too. LGTM

PR #3401, opened 2025-08-12
Don't do an engine round-trip to parse entities

VReaperV:


slipher:
LGTM

Good to see more IPC methods getting NUKED

PR #3403, opened 2025-08-13
Fix the remaining memory leaks in gentities

VReaperV:
Follow-up on #3399

Also fixed the atrocious logging functionality that used static strings.

### DIFF
--- a/src/sgame/sg_entities.cpp
+++ b/src/sgame/sg_entities.cpp
@@ -252,20 +252,11 @@ gentity debugging
 =================================================================================
 */
 
-/*
-=============
-EntityToString
-
-Convenience function for printing entities
-=============
-*/
-//assuming MAX_GENTITIES to be 5 digits or less
-#define MAX_ETOS_LENGTH (MAX_NAME_LENGTH + 5 * 2 + 4 + 1 + 5)
 static bool matchesName( mapEntity_t const& ent, std::string const& name )
 {
 	for ( char const* n : ent.names )
 	{
-		if ( !Q_stricmp( name.c_str(), n ) )
+		if ( n && !Q_stricmp( name.c_str(), n ) )
 		{
 			return true;
 		}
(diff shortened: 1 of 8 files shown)


src/sgame/sg_entities.cpp
```diff
@@ -193,15 +194,17 @@ void G_FreeEntity( gentity_t *entity )
...
-	}
-
 	unsigned generation = entity->generation;
 
-	if ( entity->id )
-	{
-		BG_Free( entity->id );
+	for ( const fieldDescriptor_t& field : fields ) {
+		if ( field.type == F_STRING ) {
+			char** string = ( char** ) entity + field.offset;
```
slipher, line 201: Wrong pointer arithmetic. You are multiplying the offset by `sizeof(void*)`

src/sgame/sg_map_entity.h
```diff
@@ -305,24 +305,19 @@ struct mapEntity_t
...
-			{
-				buffer += ", ";
-				buffer += names[i];
+		std::string out;
+		for ( char* name : names ) {
+			if ( !name ) {
+				break;
 			}
+
+			out += ", ";
```
slipher, line 316: Adds unwanted `, ` at the end of the string

src/sgame/sg_entities.cpp
```diff
@@ -252,58 +252,25 @@ gentity debugging
...
-}
-
-const char *etos( const gentity_t *entity )
-{
-	static  int  index;
-	static  char str[ 4 ][ MAX_ETOS_LENGTH ];
-	char         *resultString;
-
-	if(!entity)
-		return "<NULL>";
```
slipher, line 295: It seems this bit is important. I tested g_debugEntities and I'm getting crashes with the entity being null

src/sgame/sg_entities.cpp
```diff
@@ -193,15 +194,17 @@ void G_FreeEntity( gentity_t *entity )
 
 	delete entity->entity;
 
-	if ( entity->classname ) {
-		BG_Free( entity->classname );
-	}
-
 	unsigned generation = entity->generation;
 
-	if ( entity->id )
```
slipher, line 202: I believe `id` is not part of the table-driven parser so this is still needed.

src/sgame/sg_main.cpp
```diff
@@ -395,6 +395,7 @@ static void G_FindEntityGroups()
 				for (int k = 0; comparedEntity->mapEntity.names[k]; k++)
 				{
 					masterEntity->mapEntity.names[k] = comparedEntity->mapEntity.names[k];
+					BG_Free( comparedEntity->mapEntity.names[k] );
```
slipher, line 398: This frees the pointer that we are going to use. If something needs to be freed it's `masterEntity`'s name before assigning to it

src/sgame/sg_entities.cpp
VReaperV: Isn't it this:
```
{ "id",                  FOFS( id ),                             F_STRING,     ENT_V_UNCLEAR, nullptr },
```

src/sgame/sg_main.cpp
VReaperV: Fixed.

src/sgame/sg_entities.cpp
VReaperV: Fixed.

slipher:
LGTM

Issue #579, opened 2015-02-07
Lucifer cannon power warning noise doesn't sound while jetpack on


overvolting:
Flying with jetpack and powering up lucifer cannon doesn't produce the lucifer cannon overload warning.

overvolting:
I would like to add that when you're flying with the jetpack many other sounds also don't play.

For example, the sound of an active saw.

DolceTriade:
Multiple sounds on the same channel?

Kangz:
Yes it is a limitation of the q3 sound model: you can only have one looping sound per entity. I want to move away from this eventually.

Viech:
In that case I won't bother working around this now.

bmorel:
> Flying with jetpack and powering up lucifer cannon doesn't produce the lucifer cannon overload warning

This actually seems fixed now (in 0.52.1) since the jetpack sound is cut to let the luci warn, but not that:

> I would like to add that when you're flying with the jetpack many other sounds also don't play. For example, the sound of an active saw.

although I only have the problem with either luci warning or painsaw, no other weapon seems affected.

bmorel:
> although I only have the problem with either luci warning or painsaw, no other weapon seems affected.

Correction, this also happens with ambient sounds, for example in plat23. This does not, however, happens with buildings sounds like turrets "scan blips" or RC humming.

PR #3406, opened 2025-08-15
Fix looping sounds

VReaperV:
Engine-side pr: https://github.com/DaemonEngine/Daemon/pull/1756

Use the now-implemented persistent option for `trap_S_AddLoopingSound()`. Add `SVF_BROADCAST_ONCE` to `SP_sfx_speaker`.

PR #3407, opened 2025-08-15
Fixes Unvanquished/Unvanquished#579

VReaperV:
Requires #3406 since I didn't test without it/it would create merge conflicts otherwise/both branches target for-0.56.0/sync anyway

Engine-side pr: https://github.com/DaemonEngine/Daemon/pull/1758

Fixes https://github.com/Unvanquished/Unvanquished/issues/579

This requires a cgame change because otherwise it breaks the lcannon warning sound.


sweet235:
LGTM

PR #3417, opened 2025-09-07
update-version-number: also set year

illwieckz:
Also write a year along the version.

Can be used in copyright notices, for example in:

- https://github.com/DaemonEngine/Daemon/pull/1790

### DIFF
--- a/update-version-number.py
+++ b/update-version-number.py
@@ -13,6 +13,11 @@ REPLACEMENTS = [
         r'\1"{version}"'
     ),
     (
+        "daemon/src/common/Defs.h",
+        r'(#define\s+PRODUCT_YEAR\s+)".+"',
+        r'\1"{year}"'
+    ),
+    (
         "macosx/Info.plist",
         r"(<key>CFBundleVersion</key>\s*)<string>[^<]+</string>",
         r"\1<string>{version}</string>"
@@ -39,14 +44,19 @@ MAJOR_REPLACEMENTS = [
 ]
 
 import argparse
+import datetime
 import difflib
 import os
 import re
 
 
-def update_version(version, majorness, dry_run):
+def update_version(version, year, majorness, dry_run):
     changes = {} # filename -> [original content, new content]
     root = os.path.dirname(__file__)
+
+    if not year:
+        year = datetime.date.today().year
+
     replacements = {
         "minor": REPLACEMENTS,
         "major": REPLACEMENTS + MAJOR_REPLACEMENTS,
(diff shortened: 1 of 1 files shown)


slipher:
Why do we need this? It's free software, we don't need to jealously guard the copyright...

illwieckz:
The long term plan is that:

- https://github.com/Unvanquished/Unvanquished/pull/1777

A generic way for Dæmon-based games to run the engine with custom games without having to hardcode the strings.

There are two aspects of this:

1. Properly implementing and defining all the strings a Dæmon-based game can use.
2. Switching from `#define` to some config file (the stalled PR above).

So for both 1. and 2. I look much farther than Unvanquished.

And being exhaustive on 1. helps to better design 2. by properly identifying the needs.

I have other changes to do that falls into 1. that will come later, like properly renaming `PRODUCT_NAME_UPPER` to something like `PRODUCT_MASTER_NAME` to actually describe the usage of it (it's the game name sent to the master server for enlisting), and rename `PRODUCT_NAME_LOWER` to something more meaningful (it's used as a file prefix), and do some clean-up, etc. This way once everything is properly defined as definitions with proper names properly describing the usage, I'll start again working on that config-based de-hardcoding effort, the phase 2.

Actually to better fit 1. I may start writing the whole copyright string instead of the year, even write all the stuff into a dedicated header file before implementing the config file, that would allow us to start plumbing the writing of such config file, with us just needing to change the format (from .h to .conf) the day 2. is implemented.

illwieckz:
> We don't need to jealously guard the copyright...

This is why I used the “Some rights reserved.” string.
Also all licenses we use are based on copyright laws, it's because someone has the copyrights he can give some of them.

illwieckz:
So, yes, this thing is not very important (in fact, I myself find every copyright string stupid to begin with), but I want to properly identify the needs that exist in the industry to design things around those needs. Since we already have to run that script to update versions, it will cost nothing to us to also update that other string, so I can start plumb the things.

illwieckz:
Speaking of year, this also overlaps another topic which is the displayed build date.
We currently use __DATE__ and this is a problem for reproducibility.
We may also want to set some release date that would not be optional, and an optional build date (or we could just drop it).

illwieckz:
I currently put this on hold as it extends to much wider topics I want to take time to design for, and I don't want to make https://github.com/DaemonEngine/Daemon/pull/1790 wait for it.

illwieckz:
Speaking about date, here is what prints the Flatpak build:

```
Unvanquished 0.55.4 Linux amd64 (GCC_15.2.0:c++) Nov 11 2011
```

Yes, they pretend to have built that on `11/11/11`, I wouldn't be surprised if they also pretend to have built that at `11:11:11` too…

That's probably a trick they implemented to improve reproducibility of builds…

So yes, the `__DATE__` thing cannot even be trusted, so us actually writing down the release date would be a good idea.

PR #3422, opened 2025-09-24
Fix bug where the chat field ignores input

slipher:
Also make it possible to bind the mouse wheel keys in the bind menu.


illwieckz (approved):
LGTM.

PR #3424, opened 2025-09-24
ui: reword a sentence to make it easier to understand and translate

illwieckz:
I had no idea what that sentence meant when I tried to translate it in Weblate, especially because it relies on specific English construction (the `<object><dash><adjective>` one), not just English words, and it also makes use of homonyms. For example if we don`t read `map change` as the object, but we read `change` as the object being qualified as persistent, and if we confuse `map` the verb with `map` the name, then “map change-persistent console input” can be understood “to map the input with changes that are persistent” instead of “console input being persistent to map change”. 🙃


illwieckz:
Also it was missing a capital.

slipher:
LGTM

PR #3426, opened 2025-09-25
Fix assertion failure in UI map list

slipher:
There was an unused arg to Str::Format. Behavior of release builds was not affected.

### DIFF
--- a/src/cgame/cg_rocket_draw.cpp
+++ b/src/cgame/cg_rocket_draw.cpp
@@ -1914,7 +1914,7 @@ public:
 			const FS::PakInfo &pak = *FS::FindPak( "map-" + mapName );
 			FS::PakPath::LoadPakPrefix( pak, Str::Format( "meta/%s/", mapName ), ignored );
 			// Unfortunately this spams the console with a 2nd "Loading pak" message
-			FS::PakPath::LoadPakPrefix( pak, Str::Format( "levelshots/", mapName ), ignored );
+			FS::PakPath::LoadPakPrefix( pak, "levelshots/", ignored );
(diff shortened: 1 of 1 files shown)


src/cgame/cg_rocket_draw.cpp
```diff
@@ -1914,7 +1914,7 @@ class LevelshotElement : public HudElement
 			const FS::PakInfo &pak = *FS::FindPak( "map-" + mapName );
 			FS::PakPath::LoadPakPrefix( pak, Str::Format( "meta/%s/", mapName ), ignored );
 			// Unfortunately this spams the console with a 2nd "Loading pak" message
-			FS::PakPath::LoadPakPrefix( pak, Str::Format( "levelshots/", mapName ), ignored );
+			FS::PakPath::LoadPakPrefix( pak, Str::Format( "levelshots/" ), ignored );
```
VReaperV, line 1917: Shouldn't this add `%s` instead?

illwieckz: Yes it looks like the error was the missing `%`.

illwieckz: Unvanquished path: `/meta/mapname/mapname`

Legacy path: `/levelshot/mapname`

This looks to set the directory name, so the mapname shouldn't be set in the legacy case.

The full path concatenation with the dirname and the basename is done later.

slipher: Removed the superfluous `Str::Format` now. `Str::Format("fixed string")` works but is of course pointless.

illwieckz (approved):
LGTM.

PR #3429, opened 2025-09-27
NUKE unused Debug function declaraton

VReaperV:


illwieckz:
Merging it before the usual delay because I need it to properly test the release (and I don't want to delay that).

Issue #1340, opened 2021-03-01
Overmind looks down on tyrant


necessarily-equal:
![Screenshot from 2021-03-01 13-41-39](https://user-images.githubusercontent.com/59283660/109507533-23909a00-7a9f-11eb-8704-22a6fab1f39c.png)

The tyrant isn't supposed to be *that* small.

I suspect the origin of the model is off

Pretuer:
Overmind is a little shy my friend

bmorel:
> The tyrant isn't supposed to be that small.

Overmind simply does not looks at the head of aliens. Size of tyrant:

```
mins              -32 -32 -22
maxs              32 32 70
```

This means if OM looks at origin which is what I would guess, then it looks rants at ~25% of rant's size.
Tyrant is the only alien with that kind of proportions, others being between 50% and 40%.

What would be the other consequences of changing the mins/maxs values of tyrant, here?

necessarily-equal:
Maybe overmind should look at the head of the creature instead. I think this would solve it well?

bmorel:
indeed, using `dir = target.position + target.z` would be the simpler fix.

bmorel:
in cg_buildable:

```
void CG_Buildable( centity_t *cent )
{
  ...
          // Get absolute angles to target.
        VectorSubtract( cg_entities[ es->otherEntityNum ].lerpOrigin, eyeOrigin, dirToTarget );
        VectorNormalize( dirToTarget );
        vectoangles( dirToTarget, angles );
}
```

PR #3438, opened 2025-10-06
Overmind eye to eye

necessarily-equal:
This makes overmind eye aiming go from _its_ eye to _your_ eye instead of from center to center. It also adapts to crouching when in first person view (either playing or spectating).

I think the result fits well and follows well. Brr, creepy.

Screenshot From 2025-10-06 23-10-06

Fixes https://github.com/Unvanquished/Unvanquished/issues/1340.

### DIFF
--- a/src/cgame/cg_buildable.cpp
+++ b/src/cgame/cg_buildable.cpp
@@ -2206,9 +2206,9 @@ void CG_Buildable( centity_t *cent )
 				// HACK: Fixed offset for eye height.
 				// TODO: Retrieve eye origin from skeleton.
 				VectorCopy( es->origin, eyeOrigin );
-				eyeOrigin[ 2 ] += OVERMIND_EYE_Z_OFFSET;
 				vec3_t axis[ 3 ];
 				AnglesToAxis( cent->lerpAngles, axis );
+				VectorMA( eyeOrigin, OVERMIND_EYE_Z_OFFSET, axis[2], eyeOrigin );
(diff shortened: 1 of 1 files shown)


illwieckz:
Excellent!

necessarily-equal:
image
Just added a fix for the Z axis and tried on an overmind with non-aligned axis

illwieckz:
Great!

slipher:
FWIW the Overmind's eye did not cope well with being upside down so I guess the transformations are not 100% correct :) The left/right position of the eye seems to be the inverse of what it should be in that case.

unvanquished_2025-10-08_074829_000

necessarily-equal:
Hmm, the gameplay forbids building an overmind upside down, is there a cvar to unlock that?

slipher:
> Hmm, the gameplay forbids building an overmind upside down, is there a cvar to unlock that?

I did it by editing the configuration files. Add the following to `pkg/unvanquished_src.dpkdir/configs/buildables/overmind.attr.cfg`:
```
+minNormal         0
+allowInvertNormal
```

necessarily-equal:
I agree that the _aiming_ of the eye doesn't work upside down. That was already the case. I only fixed the calculation of the desired angle. I'm tempted to say correcting the aiming when upside-down is out of scope for this PR maybe? Or do you have a suggestion on how to fix that?

slipher:
I don't understand what is supposed to be the distinction between "calculation of the desired angle" and "aiming". But anyway, you can go ahead and merge if you want to.

PR #3440, opened 2025-10-11
Fix spectator 3rd person view YAW limits

necessarily-equal:
This deletes an effectless `AngleNormalize180`, and correct a second one that caused wrap-around by replacing it with a more sensible `Math::Clamp`. You can test in third person view (for example) by following a bot:

* `/bot fill 3`
* `/cg_thirdPerson 1`
* right-click to follow
* look up and down

### DIFF
--- a/src/cgame/cg_view.cpp
+++ b/src/cgame/cg_view.cpp
@@ -319,7 +319,6 @@ void CG_OffsetThirdPersonView()
 	int           cmdNum;
 	usercmd_t     cmd, oldCmd;
 	float         range;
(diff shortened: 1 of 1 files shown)


necessarily-equal:
To clarify a bit: this fixes a bug in 3rd person view following a spectator, where when someone scrolls up ends up wrapping around back to looking from the bottom of the followed character. And this also fixes a warning about the unused result of `AngleNormalize180`

slipher:
LGTM

PR #3441, opened 2025-10-13
README updates (package managers, dependencies)

slipher:


illwieckz:
Yeah, that was just overlooked. It's good.

PR #3448, opened 2025-10-17
Update the Debian build scripts

illwieckz:
I noticed someone tried to follow our dated Debian instructions:

- https://www.youtube.com/watch?v=s-LfjMXlGk8

I did some cleanup and maintenance too.

I also removed the building of nexe because it's useless and the build of them doesn't work on all architectures.

I left the dll building for now, people may still want to run a server using dll for example, that could be discussed later.

### DIFF
--- a/debian/unvanquished-common.install.i386
+++ b/debian/unvanquished-common.install.i386
@@ -2,10 +2,10 @@ debian/paths.conf                          etc/unvanquished/
 dist/scripts/unvanquished                  usr/lib/games/unvanquished/
 dist/scripts/unvanquished-migrate-profile  usr/lib/games/unvanquished/
 
-obj-*-*-*/cgame-native-dll.so    usr/lib/games/unvanquished/
-obj-*-*-*/sgame-native-dll.so    usr/lib/games/unvanquished/
+obj-*-*-*/cgame-native-dll.so     usr/lib/games/unvanquished/
+obj-*-*-*/sgame-native-dll.so     usr/lib/games/unvanquished/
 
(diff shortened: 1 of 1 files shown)


slipher:
I'm confused why we even have this `debian` folder. Who is it supposed to be useful for? The packaging scripts for AUR, Nix, and Flatpak live outside the repository, so it's strange that Debian scripts have pride of place when there isn't even an actual package -- and can't be due to the non-rolling nature of Debian releases. Maybe they could be moved somewhere else?

Regarding `update changelog`, the idea of having a news file in the repo was seemingly rejected [last time](https://github.com/Unvanquished/Unvanquished/pull/1422) it came up.

slipher:
If the goal is to just let people install the game they built on their own machine, we should define a CMake install target (https://github.com/DaemonEngine/Daemon/issues/448).

illwieckz:
> I'm confused why we even have this `debian` folder. Who is it supposed to be useful for? The packaging scripts for AUR, Nix, and Flatpak live outside the repository, so it's strange that Debian scripts have pride of place when there isn't even an actual package -- and can't be due to the non-rolling nature of Debian releases. Maybe they could be moved somewhere else?

I agree with that but right now the files are stored there, so that's where I submit my patches.

We may extract it to another repository in the future.

> Regarding `update changelog`, the idea of having a news file in the repo was seemingly rejected [last time](https://github.com/Unvanquished/Unvanquished/pull/1422) it came up.

This file is only there for satisfying some Debian practices, nothing is written manually there, it's generated. It adds a bit of git noise, but we expect no one to write any news there. Yet again the concern will disappear the day we move it to another repository, but right now it's stored there so I submit my patches there.

PR #3452, opened 2025-11-04
Fix prediction errors if using cg_sprintToggle

slipher:
pm_flags is defined as a 16-bit field by the netcode. PMF_SPRINTHELD was outside this range so it could not be transmitted.


illwieckz (approved):
LGTM.

PR #3455, opened 2025-12-04
Fix keynum_t binds in bind menu

### LINKED ISSUE #3454: Bind menu regression: unfriendly scancodes
There is a regression in 711ea2b662119b05bf1df1c9ef7a64f61d761bbe that causes some keys, which we normally prefer to bind via keycodes, to bind to scancodes in the bind menu. For example when pressing Tab you get `0x2b` instead of `TAB`. Or when pressing Enter you get `0x28` instead of `ENTER`.

slipher:
A regression caused bindable key events not to be sent for any Keyboard::Key that can be translated to an RmlUi key. This is because the bindable key event started being disabled if the normal key event was consumed, but the normal key event was always consumed. This just affected keynum_t Key's since character ones are ignored anyway.

Instead of sending both normal RmlUi key events and bindable key events all the time, send only bindable key events during key binding and only normal key events otherwise. This should be a more reliable design anyway since some keys such as Tab and Enter could be interpreted as ones for navigating the document. So now Escape for canceling a bind will be detected through the bindable key event instead of the normal key event.

Fixes #3454.

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -1469,6 +1469,7 @@ struct rocketInfo_t
 	rocketDataSource_t data;
 	bool renderCursor;
 	qhandle_t cursor;
+	int keyBindingTime = -1; // enable bindable key events and disable normal key events when keyBindingTime = realtime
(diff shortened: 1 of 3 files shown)


illwieckz (approved):
LGTM.

PR #3468, opened 2026-01-08
Enable realtime lighting in medium graphics preset

slipher:
Dynamic lights seem like a pretty basic feature. I was surprised to see them disabled in the medium preset.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/presets/graphics/medium.cfg
+++ b/pkg/unvanquished_src.dpkdir/presets/graphics/medium.cfg
@@ -4,7 +4,7 @@ seta r_accurateSRGB on
 seta r_colorGrading on
 seta r_lightMode 3
 seta r_lightStyles 1
-seta r_realtimeLighting off
+seta r_realtimeLighting on
 seta r_rimLighting 1
 seta r_normalMapping 1
 seta r_deluxeMapping 1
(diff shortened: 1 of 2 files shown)


illwieckz:
I know `medium` and the default options doesn't match yet, but we better make the default match a specific preset and `medium` is a good one.

I do believe that dynamic lights are basic enough and fast enough to be enabled in `medium`, do you think it would be also good to make it the engine default?

The preset description would have to be edited in `pkg/unvanquished_src.dpkdir/ui/options_graphics.rml` accordingly.

slipher:
I feel like the engine's default settings should enable all of the texture types available in shaders. While the engine can disable by default heuristic-based, post-processing effects like anti-aliasing, bloom, and maybe even tonemapping. But relief mapping is an especially expensive texture type that we don't want to have in the medium preset. Maybe it would all work out if "high" is the default though?

If we want to have a specific default preset for new Unvanquished players, another possibility is triggering it with the `cg_welcome` stuff. We'd just need to modify the preset scripts to avoid `vid_restart`ing in that condition.

illwieckz:
We can make “high” the default preset. This doesn't remove the fact that dynamic lights in medium is fine as well. We can do both.

slipher:
> I feel like the engine's default settings should enable all of the texture types available in shaders. While the engine can disable by default heuristic-based, post-processing effects like anti-aliasing, bloom, and maybe even tonemapping. But relief mapping is an especially expensive texture type that we don't want to have in the medium preset.

Actually relief mapping is only in "ultra", and is disabled by default by the engine, so that doesn't quite match my expectation. Anyway "high" is indeed probably closest to the default settings.

Updated the preset descriptions.

illwieckz:
Can you mirror the “high” preset into the default engine configuration?

slipher:
> Can you mirror the “high” preset into the default engine configuration?

https://github.com/DaemonEngine/Daemon/pull/1886

PR #3470, opened 2026-01-16
Don't set sv_killserver upon CA_DISCONNECTED state

slipher:
The engine always shuts down a running cgame upon changing state to CA_DISCONNECTED. So the case of changing to CA_DISCONNECTED should not be reachable in the cgame.


Gireen (approved):
sounds plausible to me

Image

PR #3471, opened 2026-01-17
Pre-build shader variants for spawning buildables

slipher:
Use a skin instead of a refEntity_t.customShader for the human buildable under construction, so that the engine will pre-build the necessary GLSL variants based on loading the skin at load time.


illwieckz (approved):
LGTM.

PR #3475, opened 2026-02-18
Stop requesting no longer existing renderer features

slipher:
Don't try to use renderer features which no longer exist:
- `refEntity_t::lightingOrigin` field (something from GL 1)
- Old shadow stuff e.g. inverse dynamic lights

### DIFF
--- a/src/cgame/cg_view.cpp
+++ b/src/cgame/cg_view.cpp
@@ -141,7 +141,7 @@ void CG_TestGun_f()
 {
 	CG_TestModel_f();
 	cg.testGun = true;
-	cg.testModelEntity.renderfx = RF_MINLIGHT | RF_DEPTHHACK | RF_FIRST_PERSON;
+	cg.testModelEntity.renderfx = RF_DEPTHHACK | RF_FIRST_PERSON;
 }
 
 void CG_TestModelNextFrame_f()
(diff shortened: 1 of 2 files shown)


illwieckz:
Thank you ! Such clean-up is deeply appreciated.

illwieckz:
Was this code actually executed (especially the `CG_StartShadowCaster()` one)? Can the removal of this speed-up the game a bit? Or is it just clean-up of dead code?

VReaperV:
`Don't set refEntity_t's lightingOrigin` looks good, I had a commit somewhere that does the same.

The other commit seems fine too, but haven't tested yet.

slipher:
Removed one more flag (`RF_MINLIGHT`).

> Was this code actually executed (especially the `CG_StartShadowCaster()` one)? Can the removal of this speed-up the game a bit? Or is it just clean-up of dead code?

It should be impossible to set `cg_shadows` greater than 1, so no code was executed beyond the cvar test.

illwieckz:
> It should be impossible to set `cg_shadows` greater than 1, so no code was executed beyond the cvar test.

OK thanks. So dead code removal.

VReaperV:
`Remove no-op RF_MINLIGHT flag` looks good too. Should be removed from engine too though.

slipher:
> `Remove no-op RF_MINLIGHT flag` looks good too. Should be removed from engine too though.

Yes: https://github.com/DaemonEngine/Daemon/pull/1904

illwieckz (approved):
LGTM.

PR #3477, opened 2026-02-23
Add MSAA to ultra preset; remove FXAA option

slipher:
Remove FXAA from options since it doesn't really work.

Add MSAA to options and set it in the "ultra" preset.


illwieckz (approved):
LGTM.

Can merge quick.

PR #3487, opened 2026-03-08
Fix password protection for local servers

slipher:
Fix password protection for servers hosted by a client and fix the `password` cvar registration conflict warning.


slipher:
This should be noted as a server administration-relevant change since the cvar name to password-protect the server has changed.

slipher:
Fixes regression from 3447e9365838fdbc0dd7df0e639b30589e67fec7. Oops!

PR #3488, opened 2026-03-08
Re-add the FXAA option in UI and enable it in high graphics preset, tweak MSAA

illwieckz:
Re-add the FXAA option in UI and enable it in high graphics preset.

Add message saying  MSAA is preferred when both FXAA and MSAA are enabled.

Also reword `r_msaa` and `r_ssao` as `r_MSAA` and `r_SSAO`.

Require:

- https://github.com/DaemonEngine/Daemon/pull/1924

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
@@ -80,12 +80,6 @@
 					<p><translate>Multisample anti-aliasing is preferred when both are enabled.</translate></p>
 				</row>
 
- 				<row>
-					<input cvar="r_FXAA" type="checkbox" />
-					<h3><translate>Fast approximate anti-aliasing (FXAA)</translate></h3>
-					<p><translate>Smooth the edges using a fast but cheap filter, can blur some details.</translate></p>
-				</row>
-
 				<row>
 					<select cvar="r_MSAA">
 						<option value="0"><translate>Off</translate></option>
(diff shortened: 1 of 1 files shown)


illwieckz:
I've set MSAA to 4 in ultra graphics preset because of:

- https://github.com/Unvanquished/Unvanquished/issues/3489

slipher:
I don't think FXAA looks good enough to enable by default, too blurry.

slipher:
LGTM to the following commits:
- `ui: re-add the FXAA option`
- `presets: set MSAA to 4 in ultra graphics preset`
- `ui: list MSAA before FXAA in graphics options`

illwieckz:
> LGTM to the following commits:
> 
>  * `ui: re-add the FXAA option`
>  * `presets: set MSAA to 4 in ultra graphics preset`
>  * `ui: list MSAA before FXAA in graphics options`

Thanks.

> I don't think FXAA looks good enough to enable by default, too blurry.

I tweaked it to be less blurry. People wanting something better can use the ultra preset that enables MSAA, those not having powerful enough hardware for MSAA can still get some antialiasing even if that's not as good as MSAA. The tradeoff looks to be acceptable. So that being said, I _hereby declare_ that we enable it in high preset.

Other remaining commits are just case change, nothing functional. I _hereby declare_ that's the convention we will use.

slipher:
> I tweaked it to be less blurry.

I evaluated that and responded in the FXAA PR thread. It still looked 90% identical with most things being as blurry as before; only one of the sample screenshots became less blurry.

PR #3491, opened 2026-03-14
ui: reword the brightness graphics options

illwieckz:
Reword the brightness panel in the graphics options.

First, our panels usually have a title, add one for consistency.

Then, we have a convention of listing best options before less good options. We usually do this in drop-down lists, but here it applies too.

Also do some rewording changes.

Before:

unvanquished_2026-03-14_234929_000

After:

unvanquished_2026-03-14_235000_000

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
@@ -106,28 +106,31 @@
 
 			<tab><translate>Brightness</translate></tab>
 			<panel>
+				<h2><translate>Brightness</translate></h2>
 				<row>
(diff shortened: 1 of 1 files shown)


slipher:
LGTM

This was not affected by the PR but it should say "Current: 1.00" instead of "Gamma: 1.00" to be consistent with other options.

illwieckz:
Ah yes, will do that as well!

illwieckz:
I also reworded this:

> Consider resetting and using tone-mapping instead.

Into that:

> Consider resetting and using exposure instead.

The advice for using tone mapping as a replacement for gamma was because exposure required tone mapping to be enabled, so the advice was indirect. but now that they are de-correlated we can give the advice for exposure directly

PR #3502, opened 2026-03-22
Remove Tamil from language list

slipher:
It is supposed to be unusable without text shaping.


illwieckz:
Actually I merged outside GitHub.

I also added a commit telling Urcheon to not package it yet (like other languages we cannot render properly yet like Arabic, etc.).

Thanks.

PR #3509, opened 2026-03-27
Force developer mode in server creation UI when NaCl is unavailable

runlevel5:
https://github.com/DaemonEngine/Daemon/pull/1945 needs to be merged first

On platforms without NaCl runtime support the server creation menu now:

- Enable the developer mode checkbox by default

### DIFF
--- a/pkg/unvanquished_src.dpkdir/ui/server_create.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/server_create.rml
@@ -28,10 +28,14 @@
 				end
 				return 'execForm "map $map$ $layout$"'
 			end
+
+			function OnLoad(document)
+				Element.As.ElementFormControlInput(document:GetElementById("devmode")).checked = Cvar.get("vm.nacl.available") == "off"
+			end
 		</script>
 
 	</head>
-	<body template="window" id="createserver" onshow='Events.pushcmd("buildDS mapList")' style="width: 40em; margin: 15% auto auto auto;">
+	<body template="window" id="createserver" onload="OnLoad(document)" onshow='Events.pushcmd("buildDS mapList")' style="width: 40em; margin: 15% auto auto auto;">
 		<h1><translate>Start local/LAN game</translate></h1>
 		<form onsubmit='Events.pushevent(MapMode(document), event)'>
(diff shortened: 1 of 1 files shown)


runlevel5:
Alternatively we could hide the Developer Mode option completely if the goal is to make the game more non-tech users-friendly (I believe not all gamers care or know what NaCl really is)

slipher:
Maybe the simplest possible implementation here can be an `onload` Lua snippet that sets the checkbox to the opposite of `vm.nacl.available`.

slipher:
Well as discussed at https://github.com/DaemonEngine/Daemon/pull/1945, we don't want to have a ton of code for a barely viable use case. So I would really prefer the one-liner to disable the checkbox; I don't think we need warning text or to prevent the box from being unchecked. Having a usable default value should be good enough

pkg/unvanquished_src.dpkdir/ui/server_create.rml
```diff
@@ -21,6 +21,11 @@
 		</style>
 		<script>
 			function MapMode(document)
+				-- If NaCl is not available, always use devmap
+				local nacl = Cvar.get("vm.nacl.available")
+				if nacl == nil or nacl == "0" then
```
slipher: If the cvar is a bool it uses the silly word form `on` or `off` so the parsing is not right.

Gireen:
oh there is a potential problem and reason to prefer the lua method
when the PR is used by a mod or release without a daemon update it will act always as if nacl is not available
then its not possible to start a server without devmap (from the menu)

in lua removing `nacl == nil` should prevent it

but with the `<if>` tag iirc a unset cvar is treated as 0


in every case this PR would also need to have the daemon folder updated after its companion PR is merged

slipher:
The Lua test can be changed to `Cvar.Get(...) == "off"` so the cvar not being present is treated as true.

slipher:
Looking pretty good. It would be a bit better to use the `onload` event instead of `onshow` so that the checkbox doesn't lose its state if you close and reopen the dialog.

Gireen:
And merged. 

thanks for your contribution runlevel5

PR #3513, opened 2026-03-30
Support fixed seeds for random number generators

slipher:
Replace all `rand()` usage with a global RNG and add cvars that can be used to optionally set a fixed seed for the RNG. This is useful for testing various things. For example I have used `cgame.randomSeed` to make scenes involving particle systems that have randomness reproducible for renderer tests. `sgame.randomSeed` might be useful to make tests for bots.


Gireen (approved):
God does not play dice ... when cheats are enabled. 😆 

Code looks good to me. not tested

Image

PR #3514, opened 2026-03-30
Mac CI fix

slipher:


slipher:
OK it works. The job randomly canceled itself this time but it completed the first run: https://dev.azure.com/UnvanquishedDevelopment/Unvanquished/_build/results?buildId=13768&view=results

illwieckz:
Thanks.

VReaperV:
> This commit is written by AI since it was too boring for humans to do.

I was told we don't submit AI-generated code:
> On code side, some people did some experiments but we don't submit code generated by AI.
So why was this merged?

illwieckz:
Ah! You found _the_ exception! 😄️

Anyway it doesn't contribute to the binary.

Just to clarify, in my sentence:

>  but we don't submit code generated by AI.

I meant:

As far as I know, this is not something we do.

It also means that it's not impossible that people submitted code modified by AI but didn't disclose it and there was no way for us to detect it… Who knows, and how to know? If @slipher didn't disclosed it here, no one could have been able to detect it because how short the change is…

I we forbid disclosure, then people will do it without disclosing it. I prefer disclosure.

For reference, my whole comment was:

> AI didn't had a huge impact on us yet.
> On assets side some people may have investigated some upscaling, but there is no real interest in generative AI. It looks like we prefer a game looking a bit dated than looking slop. 😁 
> On code side, some people did some experiments but we don't submit code generated by AI. As an example of such AI usage on code side, I got chatgpt pinpointing a cause of a bug in GCC binutils preventing to build the nacl loader with GCC. I described the problem with a lot of details and chatgpt suggested things I verified, some suggestions were wrong, and one was right. After that initial help we discovered in total 3 ways to fix that bug. Chatgpt didn't wrote anything, it just analysed the thing and suggested causes of bugs I could challenge. 🙂
> Also we didn't got AI generated PR slop to deal with. I know some projects are experiencing some annoyance because of that, but we've dodged that for now.

_Edit:_ Actually it's so small even copyright can't be claimed. 😁

PR #3515, opened 2026-04-01
Bump daemon submodule

runlevel5:
So that the game could support non-x86_64/arm64 Linux architecture by emulating NacL Loader with box64


illwieckz:
Hi! Thanks. We usually use a script to sync the submodules, but after the recent release we haven't fully returned to nominal operability yet. Next time if you need a submodule sync you can just do a comment, so you won't have to go through a complete pull request. 😉️

illwieckz:
No 24h delay because that's just a submodule sync of what is already validated.

PR #3521, opened 2026-04-07
write short X.Y abi version string

illwieckz:
Write short `X.Y` abi version string.

In version `0.56.0` string, only `0.56` says something about the ABI.

Our convention is that any `0.56.Z` version should never break the ABI, so `Z` is worthless mentioning and is confusing.

We may want to print the ABI version, as it is what matters the most about compatibility.

Printing ABI version `0.56.0` when running `0.56.5` game can give false ideas of incompatibility.

Also, `0.56` is shorter than `0.56.0`, which is useful when printing it, especially in tables.

### DIFF
--- a/update-version-number.py
+++ b/update-version-number.py
@@ -34,7 +34,7 @@ MAJOR_REPLACEMENTS = [
     (
         "daemon/src/common/IPC/Common.h",
         r'SYSCALL_ABI_VERSION = "[^"]*"',
-        r'SYSCALL_ABI_VERSION = "{version}"'
+        r'SYSCALL_ABI_VERSION = "{abiVersion}"'
     ),
 ]
 
@@ -52,6 +52,8 @@ def update_version(version, majorness, dry_run):
         "major": REPLACEMENTS + MAJOR_REPLACEMENTS,
     }[majorness]
 
+    abiVersion = ".".join(version.split(".")[0:2])
+
     for filename, pattern, replacement in replacements:
         filename = os.path.join(root, filename)
         if filename not in changes:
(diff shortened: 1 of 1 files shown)


illwieckz:
Also, this is how we wrote the ABI version in 0.55. In 0.56 I wrote it 0.56 by hand to do as before. This script change make sure we continue to do as well as before, but the scripted way.

Gireen (approved):
LGTM 
not tested

Image

slipher:
It doesn't make sense that the script now has two dueling ways of determining which versions are major: the `major`/`minor` argument on the command line, and some sort of string parsing of the version number.

slipher:
Also I could imagine a scenario where we want an ABI bump for a minor version: if we had a case like 0.51.0 where the game was discovered to be unusable after being tagged but before being distributed, and the fix required an ABI change, we might want to have a 0.51.1 ABI.

illwieckz:
> Also I could imagine a scenario where we want an ABI bump for a minor version: if we had a case like 0.51.0 where the game was discovered to be unusable after being tagged but before being distributed, and the fix required an ABI change, we might want to have a 0.51.1 ABI.

If this happens we do a 0.52, even if it's 2 hours before 0.51 is out, even if it's 2 hours after 0.51 is out… Or we can write the 0.51.1 ABI version by hand as an exception, but I would just favor the bumping of the whole.

slipher:
In any case the idea of displaying ABI as version in the UI is not a good one; it will cause just as much confusion. Yes, displaying minor versions could cause some one to think different minors are compatible. But if you display ABI, when 0.56.2 comes out players can think, WTF, why are all the servers still on 0.56 and none of them are updated to the latest version? I think the only way to avoid confusion would to have it be just a compatible yes/no thing, or older/newer/just right.

slipher:
Version list questions aside, getting rid of the `.0` would be OK I guess, as long as we don't leave the script a mess. We can get rid of the `major`/`minor` argument and just always replace the ABI line if we want to be super committed to requiring a major version to change it. Hopefully there would be some assertion that the version string follows the expected format.

illwieckz:
Actually, it works. We can reword the script later if we feel it.

slipher:
This doesn't have any relevance for the 0.56.2 release. It only affects major version releases.

PR #3522, opened 2026-04-12
Move gameplay cvars out from serverinfo to another configstring

slipher:
Remove some gameplay cvars which need to be mirrored to the cgame, but are not especially relevant for deciding which server to join, from the server info. Add a new configstring to use instead of the server info one for transmitting these. This helps to avoid size limitation issues on the server info as in https://github.com/DaemonEngine/Daemon/issues/1215, since the other configstrings' limits are 8k instead of server info's 1k.

Also add a gamelogic-side workaround for https://github.com/Unvanquished/Unvanquished/issues/1102 to ensure that configstrings are reliably mirrored to the client. The server info configstring frequently changes (for example whenever someone changes teams), so I want to ensure that the cvars aren't more exposed to the configstring syncing bug by being moved to a less frequently changed cvar. This is temporary until clients are fixed with https://github.com/DaemonEngine/Daemon/pull/1954. 

I tested the #1102 workaround using a setup like the one described by @DolceTriade in the issue. I tested the gameplay cvar mirroring by changing the BP recovery rate cvar during a game.

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -2188,6 +2188,7 @@ void     CG_ShowScores_f();
 void CG_ExecuteServerCommands( snapshot_t* snap );
 void CG_SetMapNameFromServerinfo();
 void CG_ParseServerinfo();
+void CG_ParseGameplayCvars();
 void CG_SetConfigValues();
 void CG_ShaderStateChanged();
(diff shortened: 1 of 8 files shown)


Gireen (approved):
LGTM 
not tested
Image

PR #3528, opened 2026-04-23
Remove barrel model from /testgun

slipher:
/testgun attempted to add a barrel model formed by adding "_barrel.md3" to the name of the gun model. NUKE this because first, we don't use any barrel models, and second, it won't work any more after the skeleton changes.


Gireen (approved):
LGTM 
not tested

Image

PR #3529, opened 2026-04-23
Configstring-related cleanup

slipher:
- Remove unused stuff
- Fix the bug that you can't see a vote that started before you entered the game.
- Use configstrings for sending pmove params since the desired semantics are the same, and now we have fixed configstrings to be reliable enough.


slipher:
Fixed the compile error with the last commit (pmove params).

Gireen (approved):
LGTM
not tested

Image

PR #3533, opened 2026-04-25
Fix more entity cache-related regressions

### LINKED ISSUE #3516: Gun jiggles while noclipping
While hovering motionless in noclip mode, the 1st-person human weapon appears to jiggle slightly.

Maybe caused by the skeleton refactoring?

slipher:
Fixes #3516, and a bug where you could get disconnected if you lag during a teleport.

The first two commits are picked from https://codeberg.org/unvmods/Unvanquished/commits/branch/unvmods-0.56 and are LGTMed by me.

### DIFF
--- a/src/cgame/cg_ents.cpp
+++ b/src/cgame/cg_ents.cpp
@@ -1250,19 +1250,8 @@ void CG_AddPacketEntities()
 	centity_t* playerEnt = &cg.predictedPlayerEntity;
 
 	BG_PlayerStateToEntityState( ps, &playerEnt->currentState, false );
-	playerEnt->valid = true;
-
-	playerEnt->refEntitiesFrame ^= 1;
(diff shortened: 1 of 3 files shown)


illwieckz:
Thanks you! That's deeply appreciated.

Gireen (approved):
nice finding

Image

illwieckz:
@slipher do you know if one of those patch (and those you already merged) fixes some models being invisible?

I believe I seen @DolceTriade in game reporting that some humans models were invisible as he played alien, and later I spotted invisible human buildables when playing as human, in other games.

slipher:
> @slipher do you know if one of those patch (and those you already merged) fixes some models being invisible?

I doubt it since they only affect one's own player model.

I don't know how to reproduce invisible models and haven't seen the bug myself.

PR #3534, opened 2026-04-25
Clean up APIs for color code and emoticon control

slipher:
- Remove some redundant functions.
- Fix the interface of `Rocket_QuakeToRml`, getting rid of the `RP_QUAKE` flag that didn't do anything.
- For the `<translate>` and `<inlinecvar>` RML elements, instead of having separate `quake` and `emoticons` attribute, just have a single `quake` attribute that enables colors and emoticons together.
- Prevent HTML injection from `<inlinecvar>`


Gireen (approved):
LGTM, not tested

Image

PR #3535, opened 2026-05-02
Enable Vampire mode by default and add vampire and mining gameplay presets

illwieckz:
Enable Vampire mode by default and add `vampire` and `mining` gameplay presets.

Long time overdue.


pkg/unvanquished_src.dpkdir/presets/gameplay/drill.cfg
```diff
@@ -0,0 +1,3 @@
+/seta g_BPVampire off
```
slipher: Usually we don't write a leading `/` in these files.

slipher:
- Usually we don't write a leading `/` on commands for these scripts.
- I think for BP vampire we would also want to set `g_BPBudgetPerMiner` to 0 in case there are miners in the layout.
- There must be a better name than "drill" for the file? E.g. "mining". Also we might want to clarify that the gameplay involved relates to build points. Like `bp_vampire.cfg` and `bp_mining.cfg`
- FWIW the BP is usually a bit higher on the servers where vampire mode is used. Like 120

illwieckz:
> * Usually we don't write a leading `/` on commands for these scripts.

Right

> * I think for BP vampire we would also want to set `g_BPBudgetPerMiner` to 0 in case there are miners in the layout.

Hmmm, we may also prefer preventing the spawn of any miner when BP is enabled.

> * There must be a better name than "drill" for the file? E.g. "mining". Also we might want to clarify that the gameplay involved relates to build points. Like `bp_vampire.cfg` and `bp_mining.cfg`

Well, the idea for those presets were not to be cumulative with other presets. For example in the future we may have a `callvote nextgameplay <name>` command that would set some `g_gameplay` cvar that would be used to make the next game loads `presets/gameplay/<name>.cfg`, and then we would be able to implement a gameplay vote screen at the end of a game, presenting various gameplay names and setting the cvar for the next game. So I expect strings like “deathmatch”, “clanarena”, “capturetheflag”, “kingofthehill”, “juggernaut”, “race”, “soccer” and “curling”.

> * FWIW the BP is usually a bit higher on the servers where vampire mode is used. Like 120

Good to know!

slipher:
> Well, the idea for those presets were not to be cumulative with other presets.

Fair enough. `mining` still sounds better than `drill` though.

illwieckz:
I've set the vampire-mode `g_BPInitialBudget` to `120`, and renamed `drill.cfg` to `mining.cfg`.

illwieckz:
I moved mission stuff to:

- https://github.com/Unvanquished/Unvanquished/pull/3542

Because that commit is still in the writing.

illwieckz:
Something we may need to check is: are buildables loaded before per-map config files are read or before? We want to load buildables after reading per-map config so we can play Vampire mode in maps requiring the placement of a drill for gameplay purpose (like for opening a door).

src/sgame/sg_main.cpp
```diff
@@ -270,7 +270,7 @@ Cvar::Cvar<float>  g_devolveMaxBaseDistance("g_devolveMaxBaseDistance", "Max Ove
 
 Cvar::Cvar<bool>   g_autoPause("g_autoPause", "pause empty server", Cvar::NONE, false);
 
-Cvar::Cvar<int> g_maxMiners("g_maxMiners", "set maximum number of miners per team. -1 = disabled.", Cvar::NONE, -1);
+Cvar::Cvar<int> g_maxMiners("g_maxMiners", "set maximum number of miners per team. -1 = disabled.", Cvar::NONE, 0);
```
Gireen: g_disabledBuildables should be a better way.
then maps that need miners can enable it

alternatively allow maps to set g_maxMiners

Gireen:
A default gamemode should preserve the functionality of self containing mission maps that work without requiring special per map server configs.

src/sgame/sg_main.cpp
slipher: g_disabledBuildables can also be set by the map, so using max miners could be simpler than having to combine disabled buildables from potentially multiple sources.

Gireen: yeah that was the intention. to allow maps that want players to build miners independent of gamemode to overwrite this

slipher:
> We want to load buildables after reading per-map config so we can play Vampire mode in maps requiring the placement of a drill for gameplay purpose (like for opening a door).

Good point. We could set max miners to 1, while setting BP per miner to 0.

illwieckz:
An interesting discussion happened with @cu-kai on chat, reminding that `g_BPBudgetPerMiner 0` is a better choice than `g_maxMiners 0`, because it allows mapper to strategically place drills in maps as weapon reloaders.

With `g_BPBudgetPerMiner 0` we don't have to prevent miners to spawn.

Though we may want a mechanism to prevent useless drills to spawn in vampire mode, like would be all drills in stock Unvanquished maps, because it's weird to have useless things in base, especially when it takes precious places, but that's a cosmetic thing and we have less urge for it.

The `g_BPBudgetPerMiner 0` thing makes the vampire mode easier to deliver as a default gameplay, and we can improve after that.

illwieckz:
> > We want to load buildables after reading per-map config so we can play Vampire mode in maps requiring the placement of a drill for gameplay purpose (like for opening a door).
> 
> Good point. We could set max miners to 1, while setting BP per miner to 0.

This may better be a map option, but as a start that may work.

illwieckz:
So in the current state,

vampire:

```
seta g_BPVampire on
seta g_maxMiners 1
seta g_BPBudgetPerMiner 0
seta g_BPInitialBudget 120
```

mining:

```
seta g_BPVampire off
seta g_maxMiners -1
seta g_BPBudgetPerMiner 50
seta g_BPInitialBudget 80
```

slipher:
LGTM

Gireen (approved):
some useless miners are fine as free barricades or "environment buildings" 

Image

illwieckz:
OK LET'S GO

[OK LET'S GO](https://www.youtube.com/watch?v=AWM5ZNdWlqw)

PR #3537, opened 2026-05-03
Fix muzzlePS particle system corruption

### LINKED ISSUE #1950: Blaster particle in your face (not always)
May be different than:

- https://github.com/DaemonEngine/Daemon/issues/1947

When firing the blaster with some time separated, all particles lit-up on the aimed far surface.

When maintaining the blaster firing, at some point the particles start lighting at blaster position.

First:

Image

Then:

Image

slipher:
Due to CG_AddPlayerWeapon updating muzzlePS before creating and/or destroying it, attachment updates could be issued for an already deleted particle system and then corrupt a different particle system that was created in the same slot. Fix this by removing duplicated update code and doing things in the right order.

Fixes https://github.com/DaemonEngine/Daemon/issues/1950

### DIFF
--- a/src/cgame/cg_weapons.cpp
+++ b/src/cgame/cg_weapons.cpp
@@ -1493,19 +1493,6 @@ void CG_AddPlayerWeapon( refEntity_t* parent, playerState_t* ps, centity_t* cent
 
 	if ( CG_IsParticleSystemValid( &cent->muzzlePS ) )
 	{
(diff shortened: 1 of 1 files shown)


slipher:
CI failure is irrelevant - build bot's network crapping out.

illwieckz (approved):
LGTM. ✅️

PR #3538, opened 2026-05-03
Fix entity cache allocator bugs: fixes invisible models

slipher:


src/cgame/EntityCache.cpp
```diff
@@ -156,21 +156,10 @@ void AddRefEntities( centity_t* cent, std::vector<refEntity_t>& ents ) {
...
-			entityCache.Free( oldOffset, cent->refEntitiesCount, true );
-
-			for ( refEntity_t* ent = entities + oldOffset, *ent2 = entities + cent->refEntitiesOffset;
-				ent < entities + oldOffset + cent->refEntitiesCount;
-				ent++, ent2++ ) {
-				*ent2 = *ent;
-			}
-		}
-
+		uint16_t newOffset = entityCache.Alloc( ents.size() + frameCount );
```
DolceTriade, line 162: Are you worried that before the new Alloc could expand the previous allocation but now you are guaranteed to get a new allocation?

src/cgame/EntityCache.h
```diff
@@ -66,11 +66,16 @@ class EntityCache {
...
-	static constexpr uint32_t blockCount   = PAD( MAX_REF_ENTITIES / 64, 64 );
-	static constexpr uint32_t blockCountL2 = ( blockCount + 63 )   / 64;
+	using block1_t = uint64_t;
+	using block2_t = uint64_t;
+	static constexpr uint32_t blockSize1 = std::numeric_limits<block1_t>::digits; // "digits" = num bits
+	static constexpr uint32_t blockSize2 = std::numeric_limits<block2_t>::digits;
 
-	uint64_t blocksL2[blockCountL2];
-	uint64_t blocks[blockCount];
+	static constexpr uint32_t blockCount   = MAX_REF_ENTITIES / blockSize1;
```
DolceTriade, line 74: why is it safe to remove the padding?

src/cgame/EntityCache.cpp
slipher: There was a bug when the new allocation overlaps the previous one, but starts at a lower address. So yeah I want to guarantee a new, non-overlapping one to avoid tricky cases.

src/cgame/EntityCache.h
slipher: The class assumes that the max number of entities is 4096 * blockCountL2. So if MAX_REF_ENTITIES is not divisible by 4096 but you round up the quotient, it would just go off the end of the array.

DolceTriade (approved):

PR #3539, opened 2026-05-03
Server list: resurrect ping column

### LINKED ISSUE #3531: Ping not shown in server list
In 0.56 the ping column is not visible in the in-game server list. When adding the version column, none of the other column widths were decreased, so they now add up to 111%.

slipher:
Fixes #3531.


Gireen (approved):
LGTM, not tested

Image

PR #3540, opened 2026-05-03
WIP: Do not spawn any miner when no miner is allowed.

illwieckz:
Do not spawn any miner if no miner is allowed.

To be used with the Vampire mode:

- https://github.com/Unvanquished/Unvanquished/pull/3535


slipher:
I think during entity spawning it's too early to assume that the game rules are known. In a local game a user might want to change them using the UI, assuming we add one. And servers that set game rules from a map config (e.g. based on layout name) may not have executed the configs yet. If I'm not mistaken, at this point `G_MapConfigs` has run, but the commands are still waiting in the buffer to be executed. That's why I suggested setting BP per miner to 0 rather than destroying miners.

PR #3542, opened 2026-05-04
Add mission gameplay preset

illwieckz:
Add mission gameplay preset.

Other cvars may be needed, like forbidding the joining of some team, and we may want to split them (human mission, alien mission).

### DIFF
--- a/pkg/unvanquished_src.dpkdir/presets/gameplay/mission.cfg
+++ b/pkg/unvanquished_src.dpkdir/presets/gameplay/mission.cfg
@@ -1,0 +1,3 @@
+seta g_BPVampire off
+seta g_maxMiners 1
+seta g_BPInitialBudget 999


slipher:
I'm not convinced that this makes sense; mission or scenario maps don't have that much in common. Let's check some examples:

- Mission One has obstacle courses for both sides, intended to be followed by normal Tremulous gameplay. Any normal building rules should suffice, but giving 1000 BP would be bad. There is a map entity that destroys everything in the obstacle course parts so there shouldn't be any exploit of leaving drills there.
- mission-mantis has an obstacle course for humans, intended to be followed by PvE against alien bots. A custom config is required to lock the alien team. The alien layout seems to be just eggs so it doesn't really matter if they have the BP to power it.
- tremship is another one designed to work with a default Tremulous config. Humans are locked in their starting area for a fixed time for a tower defense scenario. Then they are given access to the alien base. This one might work better with vampire mode disabled. I saw this map on the Deadbeef server with vampire mode and after playing a while the bots had ended up with 0 BP for humans, but a bunch of buildables left, which were unpowered. Maybe the BP-based unpowering should be disabled in Vampire mode?

Also for maps that need more BP, the map has the ability to control that. operation-dretch uses this to give the aliens 5000 BP.

PR #3543, opened 2026-05-08
Don't unpower buildables due to BP in vampire mode

slipher:
Unpowering buildables when there is not enough BP to support buildables is a mechanic that should belong to mining mode.


illwieckz:
…and since BP is lost by destroying buildables, deactivating a buildable is meant to be done by destruction anyway.

illwieckz (approved):
LGTM.

PR #3544, opened 2026-05-09
Add blend regime option in graphics presets and in graphics option window

illwieckz:
Add blend regime option in graphics presets and in graphics option window.

It forces the naive blend regime in `low` and `lowest` presets. Blending sRGB textures with sRGB lightmaps without conversions is a good enough approximation for low presets, especially when specular and normal mapping is disabled (which is in low preset). This also avoids conversion round trips that lose a lot of precisions when high precision framebuffers aren't used, which aren't used in low presets. So all of that put together means smaller and faster code in low presets, while looking more pleasant.

This requires:

- https://github.com/DaemonEngine/Daemon/pull/1869

I also added the option for the sRGB accuracy in the graphics option window. We missed it for 0.56.

I'm thinking about also disabling the sRGB accuracy in `medium` preset (which is the default one), because the approximation is good enough, seriously.

### DIFF
--- a/pkg/unvanquished_src.dpkdir/presets/graphics/low.cfg
+++ b/pkg/unvanquished_src.dpkdir/presets/graphics/low.cfg
@@ -1,6 +1,7 @@
 reset r_clear
 reset r_picmip
 reset r_rimLighting
+seta r_forceBlendRegime 1
 seta r_accurateSRGB off
 seta r_colorGrading off
 seta r_lightMode 3
--- a/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
@@ -160,6 +160,13 @@
 					<p><translate>Non-physical lighting technique to make models less dark.</translate></p>
 				</row>
 				<row>
+					<select cvar="r_forceBlendRegime" style="width:20em;">
+						<option value="0"><translate>Detect</translate></option>
+						<option value="1"><translate>Forced naive blend (cheap approximation)</translate></option>
+					</select>
+					<h3><translate>Blend regime</translate></h3>
+				</row>
+				<row>
 					<select cvar="cg_shadows" style="width:20em;">
 						<option value="1"><translate>Blobs</translate>&nbsp;(<translate>low</translate>)</option>
 						<option value="0"><translate>None</translate></option>
--- a/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
+++ b/pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
@@ -160,6 +160,11 @@
 					<p><translate>Non-physical lighting technique to make models less dark.</translate></p>
 				</row>
 				<row>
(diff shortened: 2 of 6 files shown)


pkg/unvanquished_src.dpkdir/presets/graphics/low.cfg
```diff
@@ -1,6 +1,7 @@
 reset r_clear
 reset r_picmip
 reset r_rimLighting
+seta r_forceBlendRegime 1
```
slipher, line 4: Could be worth commenting that this goes together with the 8-bit framebuffer.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
```diff
@@ -159,6 +159,19 @@
 					<input cvar="r_halfLambertLighting" type="checkbox" />
 					<p><translate>Non-physical lighting technique to make models less dark.</translate></p>
 				</row>
+				<row>
+					<h3><translate>sRGB conversion accuracy</translate></h3>
+					<input cvar="r_accurateSRGB" type="checkbox" />
+					<p><translate>Disabling accuracy makes color conversions faster but less accurate.</translate></p>
+				</row>
+				<row>
+					<select cvar="r_forceBlendRegime" style="width:20em;">
```
slipher, line 168: Could also be just a 0/1 checkbox since we don't usually add debug options to the menu.

illwieckz:
> I'm thinking about also disabling the sRGB accuracy in `medium` preset (which is the default one), because the approximation is good enough, seriously.

In fact I forgot that our default preset is the `high` one :

- https://github.com/DaemonEngine/Daemon/pull/1886

So our default preset will still use accurate color conversion. We can use the less accurate one on `medium` one, it's fine.

pkg/unvanquished_src.dpkdir/ui/options_graphics.rml
illwieckz: Well, for readability we use “checked” as “better”, and here disabled (`0`) is better. I can hide the debug option though.

illwieckz: I removed the debug option.

slipher:
> I'm thinking about also disabling the sRGB accuracy in `medium` preset (which is the default one), because the approximation is good enough, seriously.

It may not make much difference visually but I can't imagine it will save much time either!

illwieckz:
That's my only argument for not setting it. 😁️

But since our default preset is now `high`, I don't see any reason to not enable in `medium` an optimization we can enable.

illwieckz:
Actually I have seen the impact of the per-surface-shader color conversion in CPU profilers in the past. I only looked at the differences between with and without conversion, but I know that color conversion isn't that cheap (this is also why I want to use naive blending in low presets). The rgbGen stuff is done repeatedly many times on every frame.

For the camera shader, the benefit is likely only a small binary size optimization, but optimization may have a little impact in rgbGen CPU code.

illwieckz:
Merging it for my local release candidates for the release.

PR #3546, opened 2026-05-09
BP vampire UI - turn on/off from the menu

slipher:
Adjust the BP vampire network commands so the BP bar will immediately show or hide based on whether the game mode is enabled.  Add buttons for the gameplay presets in the menu.

I haven't tested what happens if you change the mode in the middle of the game, but the point is it should at least work if you change it at the beginning before anything happens.


illwieckz:
Thanks!

> I haven't tested what happens if you change the mode in the middle of the game, but the point is it should at least work if you change it at the beginning before anything happens.

As a _server setup_ menu it's a good start. In the future we may have a votable `g_gameplay` cvar to set it and things like that, but that server setup menu would already do the job.

slipher:
Forgot the screenshot.

image

illwieckz:
Maybe rename Presets as Gameplay?

illwieckz:
Well, Gameplay may be harder to translate.

illwieckz (approved):
LGTM.

illwieckz:
Merging it for my local release candidates for the release.

PR #3552, opened 2026-05-14
admin: Fix segfault in bot skill

DolceTriade:
MAX_CLIENTS is wrong because only level.maxclients will be populated with a client. We could also optionally harden G_Bot_SetSkill to return early if bot->client is null...

repro:
- load map
- `/bot skill 5`


slipher:
LGTM

PR #3553, opened 2026-05-18
Fix entity cache bug: your human player invisible in mirrors + hall-of-mirrors

slipher:
Thanks to yet another entity cache bug, the predicted player ref entities could get corrupted. This is because when AddRefEntities tries to reallocate the ref entities group, it does not take into account that there may be LerpTag queries or attachments pointing at the old offset. So some of the entity operations get pointed at a random wrong entity.

The bug had the following consequences:
- Your own human player model is invisible in mirrors/portals.
- Partial hall-of-mirrors effect in some circumstances. It is easiest to
  see like this: Stand in front of a mirror or portal with MSAA enabled,
  bloom enabled, and r_clear disabled, and fire the rifle. You will see
  white bloom garbage. With other settings it can manifest as darkening
  or blurring.

unvanquished_2026-05-17_233347_000

The bug became more visible following https://github.com/Unvanquished/Unvanquished/commit/ee3406e7f8c360fd0a59ab17c59fd85ca2f978ce which abandons the
attempt to reallocate the entity range in-place if possible. This makes
sense since the bug occurs only if the entity IDs change.

Fix the bug by preallocating a large ref entities array for the
predicted player entity. That way when ref entities are added for the
predicted player from multiple places, there will be no resizing/moving
of the entity array. Fortunately, it seems that all other centity_t
besides predicted player always add their entities in a single
AddRefEntities call so they don't use resizing.

Add a warning to make sure no one is moving the entity allocations in the future.

### DIFF
--- a/src/cgame/EntityCache.cpp
+++ b/src/cgame/EntityCache.cpp
@@ -159,7 +159,17 @@ void AddRefEntities( centity_t* cent, std::vector<refEntity_t>& ents ) {
 
 	if ( ents.size() + frameCount > cent->refEntitiesCount ) {
 		uint16_t newOffset = entityCache.Alloc( ents.size() + frameCount );
-		std::copy_n( entities + cent->refEntitiesOffset, frameCount, entities + newOffset );
+
+		if ( frameCount > 0 ) {
+			Log::Warn( "EntityCache: entity %d had to move its ref entity array", cent - cg_entities );
+			// This doesn't work in general because stuff referencing the entity id's will end up wrong
+			// The entityAttachment field within the same array could be fixed easily, but LerpTag
+			// queries cannot so just don't do it
+
+			// Halfheartedly attempt to move the array
+			std::copy_n( entities + cent->refEntitiesOffset, frameCount, entities + newOffset );
+		}
+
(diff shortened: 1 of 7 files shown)


illwieckz:
Thanks!

src/cgame/EntityCache.cpp
```diff
@@ -143,24 +143,33 @@ static void UpdateEntity( refEntity_t& ent, const uint16_t id ) {
...
 				ents.size() )
 		);
 	}
 
 	if ( ents.size() + frameCount > cent->refEntitiesCount ) {
 		uint16_t newOffset = entityCache.Alloc( ents.size() + frameCount );
-		std::copy_n( entities + cent->refEntitiesOffset, frameCount, entities + newOffset );
+
+		if ( frameCount > 0 ) {
+			Log::Warn( "EntityCache: entity %d had to move its ref entity array" );
```
DolceTriade: contains format specifier but no arg

DolceTriade (approved):

DolceTriade:
This code is very new to me so I am not super well equipped to review it. However, I didn't notice any regressions.

src/cgame/EntityCache.cpp
slipher: Fixed.

PR #3557, opened 2026-06-06
tools/translate: extract strings from attr.cfg files

illwieckz:
Extract strings from `attr.cfg` files

- open files as read-only and as UTF-8
- make the pot printer reusable
- extract strings from attr.cfg files

### DIFF
--- a/tools/translation/generate_attr_pot.py
+++ b/tools/translation/generate_attr_pot.py
@@ -1,5 +1,29 @@
 #!/usr/bin/env python3
 
+# ===========================================================================
+#
+# Copyright (c) 2026 Unvanquished Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+# ===========================================================================
+
 import re
 import sys
 
--- a/tools/translation/pot_printer.py
+++ b/tools/translation/pot_printer.py
@@ -1,5 +1,29 @@
 #!/usr/bin/env python3
 
+# ===========================================================================
+#
+# Copyright (c) 2026 Unvanquished Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+# ===========================================================================
+
 import argparse
 import collections
 import datetime
--- a/tools/translation/generate_gender_pot.pl
+++ b/tools/translation/generate_gender_pot.pl
@@ -1,5 +1,29 @@
 #! /usr/bin/perl -w
 
+# ===========================================================================
+#
+# Copyright (c) 2012 Unvanquished Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+# ===========================================================================
+
 no locale;
 use strict;
 use warnings;
--- a/tools/translation/generate_rml_pot.py
+++ b/tools/translation/generate_rml_pot.py
@@ -1,5 +1,29 @@
 #!/usr/bin/env python3
 
+# ===========================================================================
+#
+# Copyright (c) 2021-2026 Unvanquished Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+# ===========================================================================
+
 import re
 import sys
 


illwieckz:
Something I wanted to do since 2021 when we worked on extracting strings from RML and things like that…

Finally, I did it for `attr.cfg` files as well! 🎉

illwieckz:
I don't know if in the C++ code we process them, but at least we can start translating them.

illwieckz:
I added a commit to translate `humanName` and `description` from `.attr.cfg` files in circle menus.

Actually I wonder if we should better translate them at parsing time… I don't know if those files are reparsed when switching the language. Also there was some existing code (tutorial) already translating them “just before display”.

slipher:
Why do you need to modify `.po` files? Shouldn't the Weblate service add new strings to the per-language databases automatically?

illwieckz:
> Why do you need to modify `.po` files? Shouldn't the Weblate service add new strings to the per-language databases automatically?

I don't know why, but it doesn't work. Maybe it reads the `.po` files, not the `.pot` one?

illwieckz:
Well, the `game.pot` file is configured in Weblate. But experience shows we better sync strings ourselves.

tools/translation/generate_attr_pot.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python3
+
+import os.path
+
+exec(open(os.path.join(os.path.dirname(os.path.realpath(__file__)), "pot_printer.py")).read())
```
illwieckz: The reason why I did this ugly thing:

```py
exec(open(os.path.join(os.path.dirname(os.path.realpath(__file__)), "pot_printer.py")).read())
```

Instead of this other ugly thing:

```py
sys.path.append(os.path.dirname(os.path.realpath(__file__)))
import pot_printer
```

Is that the `exec` doesn't write any temporary file in the `tools/translation/` directory. And despite not caching precompiled python files, the script with the `exec` is as slow as the script variant with the `import`.

slipher:
> > Why do you need to modify `.po` files? Shouldn't the Weblate service add new strings to the per-language databases automatically?
> 
> I don't know why, but it doesn't work. Maybe it reads the `.po` files, not the `.pot` one?

They say it can; you just need to configure it. https://docs.weblate.org/en/weblate-5.3.1/admin/addons.html#update-po-files-to-match-pot-msgmerge

illwieckz:
> > > Why do you need to modify `.po` files? Shouldn't the Weblate service add new strings to the per-language databases automatically?
> > 
> > 
> > I don't know why, but it doesn't work. Maybe it reads the `.po` files, not the `.pot` one?
> 
> They say it can; you just need to configure it. https://docs.weblate.org/en/weblate-5.3.1/admin/addons.html#update-po-files-to-match-pot-msgmerge

Ah, good to know, that's an addon disabled by default but we can enable it.

Anyway the `ignore_incomplete` script disables languages that are not translated enough, and this requires the po files to be updated first.

So since we need to run the script to update the pot file, when this scripts update the po file, the only difference is that it's us committing the update, not weblate.

In all cases we need to 1. run the script to update the pot file and 2. commit the changes.

When we update the po files and update the list of enabled language and commit those changes at the same time, we 1. run the script to update all the files and 2. commit the changes.

The cost is exactly the same for the human, and both pot, po and language list are never desynchronised.

If we disable po update in the script it means we go from:

- 1 human commit to update everything

to:

- 1 human commit to update the pot file
- 1 weblate commit to update the po files
- 1 human commit to update the enabled language list

In fact it's even worse because it would be in practice:

- 1 human commit to update the pot file from days 0
   and update the enabled language list from day -1
- 1 weblate commit to update the po files from day 0
- 1 human commit to update the pot file from days 1
   and update the enabled language list from day 0

The `msgmerge` Weblate extension looks useful when you also use the `xgettext` Weblate extension, meaning you never run any script and just push your `.cpp` files to the repository. But we require our script as we have our own pot generators that are not `xgettext`.

illwieckz:
I modified the scripts to not write a modified `.po` or `.pot` file when the only modification is the POT date.

I noticed some scripts didn't had any license header, so I copied the one from the others.

I noticed the license I used in the files that already had license headers was MIT (not BSD 3 clause).

Usually we use BSD 3 clause for permissive license in our code base. I strongly dislike the 3rd clause, but if we want to unify we can use BSD 3 clauses on all those files instead.

@slipher and @DolceTriade I not only need your opinion on the license but I also need your acknowledgement for whatever license we use because some of those files that were lacking a license were written or modified by you.

illwieckz:
I added `generate_map_pot.py` and `generate_arena_pot.py` to extract translatable strings from `.map` and `.arena` files.

They are written in `data.pot`.

This requires Esquirel with this patch:

- https://github.com/DaemonEngine/Urcheon/pull/72

This also requires UnvanquishedAssets:

- https://github.com/UnvanquishedAssets/UnvanquishedAssets

The `generate.sh` tool only produces `game.pot` by default, one can produce the `data.pot` this way:

```sh
./generate.sh data --data-dir=~/Mapping/Unvanquished/UnvanquishedAssets
```

I doubt the game code already implement the translation of those strings yet.

illwieckz:
@slipher: the `msgmerge` extension changes the formatting:

- https://github.com/Unvanquished/Unvanquished/pull/3560/changes

Screenshot 2026-06-08 at 23-35-37 Translations update from Hosted Weblate by weblate · Pull Request #3560 · Unvanquished_Unvanquished

And since having Weblate do the update just makes things worse even without that formatting change (we have to run the update script and the updates will be committed anyway), it looks better to just not use the Weblate's msgmerge addon.

slipher:
> added `generate_map_pot.py` and `generate_arena_pot.py` to extract translatable strings from `.map` and `.arena` files.

Please move this to a separate PR. Maybe it's just me, but in my opinion, it is bad form to make major additions to a PR after reviewers have started to look at it.

illwieckz:
> > added `generate_map_pot.py` and `generate_arena_pot.py` to extract translatable strings from `.map` and `.arena` files.
> 
> Please move this to a separate PR. Maybe it's just me, but in my opinion, it is bad form to make major additions to a PR after reviewers have started to look at it.

OK done:

- https://github.com/Unvanquished/Unvanquished/pull/3561

slipher:
Thanks for investigating the msgmerge stuff. Sounds good to continue with manual updates.

However I wonder if it wouldn't be easier for both of us to keep the `.po` updates out of the PR and just push them from the command line. For the author, including those updates in the PR seems laborious and error-prone, since every time you change any other commit, a redo of the last commit may be needed. It would be easy to mistakenly use an outdated version of the strings in the end. (Also they could be outdated because strings were changed on master.) And as a reviewer there wouldn't be so many junk changes to ignore.

illwieckz:
> However I wonder if it wouldn't be easier for both of us to keep the `.po` updates out of the PR and just push them from the command line.

Sure, I wanted to be sure the produced files looked fine, but usually I do the update from command line.
I removed that commit.

slipher:
LGTM to these commits:
- `tools/translation: add license headers to perl file`
- `cg_rocket_datasource: translate humanName and description from .attr.cfg files in circle menus`

illwieckz:
> * cg_rocket_datasource: translate humanName and description from .attr.cfg files in circle menus

I merged this one.

tools/translation/pot_printer.py
```diff
@@ -0,0 +1,61 @@
...
+    parser.add_argument("file", nargs="*")
+    argv = parser.parse_args()
+
+    translation_dict = collections.defaultdict(list)
+    errors = 0
+
+    files = argv.file
+
+    if not files:
+        files = sys.stdin
```
slipher: This won't work. It's used as a file name in the next line

tools/translation/denoise_po_diff.sh
```diff
@@ -0,0 +1,58 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+# ===========================================================================
+
+set -u -e -o pipefail
+
+short_po_diff () {
+	git diff --unified=0 --no-prefix -- "${1}" \
+	| sed '/^diff --git/d;/^index /d;/^--- /d;/^\+\+\+ /d;/^@@ /d' \
```
slipher: Parsing the diff output itself seems unnecessary. How about something like
```
erasedate() {
    sed -e 's/^"POT-Creation-Date:.*/POT-Creation-Date: 0/'
}
diff -q <(git show HEAD:${filename} | erasedate) <(erasedate < ${filename})
```

slipher (commented):
I have some suggestions in the branch `slipher/attrpot-review`. There are 3 commits there:
- Use normal Python importing instead of reading the file and evaling it.
- Fix wrong escaping of strings containing newlines.
-  Use a simpler approach for finding strings in the .cfg files. Trying to parse/tokenize the whole file is a lot of unnecessary code and just makes it less robust because if you get off track with the in-string state or whatever then you lose the rest of the file. Also it fixes the bug with your version that newlines were dropped.

illwieckz:
> Use normal Python importing instead of reading the file and evaling it.

Weird, when I did the same (that's what I did first), it polluted the source directory by writing temporary files there, and also didn't found the `pot_printer.py` file… And now that seems to work and to not pollute things. !!??

Well, if that works, we can do it then.

> Use a simpler approach for finding strings in the .cfg files. Trying to parse/tokenize the whole file is a lot of unnecessary code and just makes it less robust because if you get off track with the in-string state or whatever then you lose the rest of the file. Also it fixes the bug with your version that newlines were dropped.

I'm not sure doing things the lazy way is better than actually parsing the format…

illwieckz:
Well, it indeed wrote a `__pycache__/` folder, here is where I wrote why I didn't went the `import` way:

- https://github.com/Unvanquished/Unvanquished/pull/3557#discussion_r3369483496

tools/translation/generate_attr_pot.py
illwieckz: It looks like the `path.append()` is not needed for the import, I don't know why it didn't worked on my end without that.

Anyway, the `import` modifies the source directory by creating such files:

```
__pycache__/pot_printer.cpython-312.pyc
```

tools/translation/pot_printer.py
illwieckz: Fixed, and this wasn't the only mistake.

tools/translation/denoise_po_diff.sh
illwieckz: Good idea

illwieckz:
I imported:

- Fix wrong escaping of strings containing newlines.

I also made the new line being printed as `\\n` and not a simple ` ` whitespace.

tools/translation/denoise_po_diff.sh
illwieckz: One thing with `git diff` is it it doesn't care about where you are when you run it. Running `git show` and `diff` requires to first move to the git repository root directory.

illwieckz: …and to recompute the path of files relatively to the root.

illwieckz: Ah it looks like we can do `HEAD:./${1}` instead of `HEAD:${1}` to use relative paths…

illwieckz: I implemented it.

illwieckz:
> Use a simpler approach for finding strings in the .cfg files. Trying to parse/tokenize the whole file is a lot of unnecessary code and just makes it less robust because if you get off track with the in-string state or whatever then you lose the rest of the file.

At least this lazy approach can likely be modified easily to extract strings from `.map` files without requiring an external tool like `esquirel`, but I wouldn't call it “less robust” to actually implement the format than lazily capturing stuff that somewhat looks like what we look for. I definitely see the advantage of dealing with less code, but I wouldn't call it “less robust” to actually have the script knows the actual semantics of what's parsed.

> Also it fixes the bug with your version that newlines were dropped.

It was just that I didn't knew that new lines meant actual new lines and not a simple split string with a space.

slipher:
> Well, it indeed wrote a `__pycache__/` folder, here is where I wrote why I didn't went the `import` way:

That just needs to be added to gitignore, like every repository does. Strange that we don't have it already since there are several other Python scripts like CBSE.


> At least this lazy approach can likely be modified easily to extract strings from `.map` files without requiring an external tool like `esquirel`, but I wouldn't call it “less robust” to actually implement the format than lazily capturing stuff that somewhat looks like what we look for. I definitely see the advantage of dealing with less code, but I wouldn't call it “less robust” to actually have the script knows the actual semantics of what's parsed.

It's more robust since if there is a mistake at some point, it will get back on track for the next thing. Like UTF-8 where it's designed so that you can tell which byte in a multi-byte character is the first one, so if one byte is lost or changed, the rest of the file is read correctly. Anyway these files don't even follow any kind of well-defined grammar, so it doesn't make much sense to try to parse the whole thing. The most we can assume is that tokens are usually parsed with `COM_ParseExt`. But then for each attribute the programmer can decide at random e.g. whether newlines are allowed before/between the value(s). And from the standpoint of `COM_Parse`, `humanName "Shotgun"` is identical to `"humanName" Shotgun` so you can't really tell what's a key and what's a value.

illwieckz:
> That just needs to be added to gitignore, like every repository does. Strange that we don't have it already since there are several other Python scripts like CBSE.

Well, I dislike the idea that something writes temporary files in the source repository, but that's not a big deal.

> And from the standpoint of `COM_Parse`, `humanName "Shotgun"` is identical to `"humanName" Shotgun` so you can't really tell what's a key and what's a value.

OK, that's a good argument. If there is no format to parse, we cannot parse it.
So let's do the lazy capture way then.

As a side note, we may later replace those custom formats with something like toml or something like that.

illwieckz:
So I imported your missing improvements.

slipher:
LGTM. You can put whichever MIT/BSD variant you want for my part. I will ping Ishq about that

illwieckz:
So I merged all but the remaining license headers stuff.
I also added `__pycache__/` to `.gitignore`.

illwieckz:
I updated the `pot`/`po` files, so the `attr` strings are being submitted to Weblate as we speak.

DolceTriade:
ya feel free to relicense stuff from me too

illwieckz:
> ya feel free to relicense stuff from me too

Thanks

illwieckz:
Now that I start translating those strings, I wonder if it would not have been better to split lines into different strings.

For example we have:

```
Crucial alien structure.
Allows for upgrades to be purchased and structures to be built.
Destruction slows the alien advance.
```

and:

```
Crucial human structure.
Allows for upgrades to be purchased and structures to be built.
Destruction slows the human advance.
```

Those are 2 strings, not 6. So we translate twice the `Allows for upgrades to be purchased and structures to be built.` string.

On the other hand, doing them as single multiline strings retains some context.

illwieckz:
One counter argument is this string:

```
Alien construction form.
Much like the base granger, but can now walk on walls.
Possesses a weak spit blob for ranged attacks,
which can slow down enemies or extinguish fires.
Extinguishing fires is rewarded by credits.
```

One sentence is actually split into multiple lines, so we better not split the translatable string!

PR #3561, opened 2026-06-08
tools/translate: create data.pot for map translatable strings

illwieckz:
Create `data.pot` for map translatable strings

I added `generate_map_pot.py` and `generate_arena_pot.py` to extract translatable strings from `.map` and `.arena` files.

They are written in `data.pot`.

This requires Esquirel with this patch:

- https://github.com/DaemonEngine/Urcheon/pull/72

This also requires UnvanquishedAssets:

- https://github.com/UnvanquishedAssets/UnvanquishedAssets

The `generate.sh` tool only produces `game.pot` by default, because the data repository isn't known from this repository.

One can produce the `data.pot` this way, asking the `data` translation files to be generated explicitly, and passing the data path explicitly:

```sh
./generate.sh data --data-dir=~/Mapping/Unvanquished/UnvanquishedAssets
```

### DIFF
--- a/tools/translation/generate_arena_pot.py
+++ b/tools/translation/generate_arena_pot.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+
+# ===========================================================================
+#
+# Copyright (c) 2026 Unvanquished Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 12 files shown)


illwieckz:
I implemented the translation of the `location` RML tag (entity positional name) and `levelname` RML tag (map name used on map load screen and other things).

There may be other things to translate, like messages displayed in the center of the screen (I don't know yet how it's implemented) but that's already a good start and it proves that works.

tools/translation/generate_arena_pot.py
```diff
@@ -0,0 +1,153 @@
...
+    in_comment = False
+    in_block = False
+    in_key = False
+    in_value = False
+    in_string = False
+    key = ""
+    value = ""
+    string = ""
+
+    for character in content:
```
illwieckz, line 54: Maybe @slipher will want to adapt his lazy “simplified attr parser” for this.

What we need to capture is this line:

```
longname "Platform 23"
```

tools/translation/generate_map_pot.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def debug(msg):
+        if is_debug:
+            print(f"Debug in file {filename}#strings on line {line}: {msg}", file=sys.stderr)
+
+    lines = str(content).split('\n')
+
+    pattern = re.compile("(?P<num>[0-9]*): (?P<string>.*)")
+
+    for entry in lines:
```
illwieckz, line 55: Maybe @slipher will want to adapt his lazy “simplified attr parser” for this, and this would avoid requiring an external tool.

What we need to capture are those lines:

```
"message" "Platform 23"
```

illwieckz: NetRadiant doesn't do this, but other tools may do this: there can be random amount of whitespace before, between and after tokens.

PR #3564, opened 2026-06-15
Improve translatable strings

illwieckz:
The purpose is to make the strings easier to translate.

"Alien build weapon2" is obviously not meant to be translated, maybe it's not displayed (the description for this is even `null`), but we better have it right. It doesn't have to be different with non-advanced granger.


illwieckz:
Well, no need to wait for a complete day for that.

PR #3566, opened 2026-06-15
ui: make the the evo price string translatable

illwieckz:
Make the the evo price string translatable.

Other build menu files are crafted this way.

The `Price:` string is already translated, just not used there.


slipher:
LGTM

PR #3567, opened 2026-06-23
Convert generate_gender_pot.pl to Python

slipher:
Convert the script from Perl to Python via LLM. LLM conversion can be a nice way do achieve https://github.com/Unvanquished/Unvanquished/issues/2200 where desired, since it is highly accurate for small scripts. For this one it worked on the first try.

If desired, I could add a second commit to integrate the flags and boilerplate output junk that the other .pot generation scripts have.


illwieckz:
Related topic on AI disclosure:

- https://forums.unvanquished.net/viewtopic.php?t=2628

PR #3568, opened 2026-06-27
WIP: Testing the upcoming deps and setting the CI for macOS arm64

illwieckz:
WIP: Testing the upcoming deps and setting the CI for macOS arm64

PR #3571, opened 2026-07-09
Tell daemon cmake to pass CBSE_PYTHON_PATH to the NaCl cmake subinvocation

illwieckz:
Tell daemon cmake to pass `CBSE_PYTHON_PATH` to the NaCl cmake subinvocation.

The branch names ends with `/sync` because there is an engine counterpart:

- https://github.com/DaemonEngine/Daemon/pull/1991

But since the feature is already broken, we can merge this in any order. This will start to work once it's merged on both side, but merging only on one side whatever the side will not add more breakage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,7 @@ endif()
 
 if (BUILD_SGAME)
     include(tools/cbse/CBSE.cmake)
+	list(APPEND GAME_NACL_VM_INHERITED_OPTIONS "CBSE_PYTHON_PATH")
 
     # Avoid generating CBSE in the NaCl subproject.
     if (FORK EQUAL 2)


CMakeLists.txt
```diff
@@ -196,6 +196,7 @@ endif()
 
 if (BUILD_SGAME)
     include(tools/cbse/CBSE.cmake)
+	list(APPEND GAME_NACL_VM_INHERITED_OPTIONS "CBSE_PYTHON_PATH")
```
slipher, line 199: The way we do this in the existing case is:
```
set(NACL_VM_INHERITED_OPTIONS ${NACL_VM_INHERITED_OPTIONS} RC_MAX_LAYERS RC_MAX_NEIS)
```

This works because plain `set` does not changed the cached value, only the value within the script execution.

But see the comment below: generating CBSE is not supposed to happen in the nested build.

Also bad whitespace.

CMakeLists.txt
```diff
@@ -196,6 +196,7 @@ endif()
...
 
     # Avoid generating CBSE in the NaCl subproject.
     if (FORK EQUAL 2)
```
slipher, line 202: The nested build for NaCl is not supposed to generate CBSE. Can you give steps to repro the bug of it trying to generate CBSE in there?

illwieckz: It happened with this arm64 macOS CI build:

- https://github.com/Unvanquished/Unvanquished/pull/3568

Our CI script sets a custom python path and later the subinvocation attempts to use that Python and failes.

slipher: It's a bug if the subinvocation is attempting to build CBSE.

